PR #23, opened 2016-09-14
Change generic string type to object path type.

mulkieran:
Modify default values appropriately.


mulkieran:
Do not merge yet. I expect there will be more.

PR #26, opened 2016-09-19
Some style and whitespace fixes

agrover:
Here are some initial whitespace and other fixes to more closely hew to Rust's idiomatic coding style.


agrover:
Actually, I think I understand a little bit more now and the enum changes are not quite correct. Let me work on it some more and maybe do another PR soon.

PR #43, opened 2016-09-29
Master map

mulkieran:
Start using ```map``` instead of ```for``` loop. This is probably the more rustic approach. Also, it would be desirable to automate the transformation of these values to the correct types as much as possible, possibly in a similar way to that done in stratis-cli.


mulkieran:
@agrover Why do I need the ```as_slice().into_iter()``` calls in order to avoid a borrowing error at compile time? Isn't there a better way?

agrover:
@mulkieran yes, see #44

PR #44, opened 2016-09-29
Simplify get_list_items a little

agrover:
Move the Iter into get_list_items, rather than a non mut ref. This lets get_list_items use it, instead of generating another iterator to map over.


agrover:
Just to expand on this, a non-mut iterator is pretty useless, since you can't call next() on it (next() takes `&mut self`). The old code was creating a new iterator from the passed-in iterator and then using that. If the original iterator is moved into get_list_items() or passed as a mutable reference then it will work. This patch chooses to move the iterator rather than passing a mut ref to it, since the calling functions don't make further use of the iterator.

PR #46, opened 2016-09-29
Master misc

mulkieran:


mulkieran:
screenshot from 2016-09-29 17-20-31

Note that there is no checking between the type we declare and that gets introspected and the value that we actually construct. So, it is up to us to get that correct.

mulkieran:
Whoops! This needs a fix, I think.

mulkieran:
@agrover , @trgill rebased! How's it look?

PR #49, opened 2016-10-03
Miscellaneous minor edits

mulkieran:
The second commit takes a position, though.

### DIFF
--- a/src/sim_engine.rs
+++ b/src/sim_engine.rs
@@ -38,7 +38,6 @@ impl Engine for SimEngine {
 
         let deleted_pool = self.pools.remove(name);
 
-        // TODO deal with not found
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


src/sim_engine.rs
```diff
@@ -38,7 +38,6 @@ impl Engine for SimEngine {
 
         let deleted_pool = self.pools.remove(name);
 
-        // TODO deal with not found
```
agrover: Can you explain more? I lean towards returning a 'notfound' error, since I can think of cases where the client might want to know it passed a bad pool name.

trgill: ok - that sounds good

mulkieran: It should be about what the operation accomplishes, not about what actions it finds it has to perform. You have a reasonable chance at idempotency if you think that way, which will probably be helpful in the long run.

I don't see client code getting any benefit from an error being returned, given that what it wanted has been accomplished. The CLI can behave differently, and maybe should. But, it should be able to do that by invoking GetPoolObjectPath().

PR #65, opened 2016-10-06
Refine the SimEngine and dbus error reporting

mulkieran:


mulkieran:
So, in my existing dbus tests for previous implementation, I differentiate a bit between different situations that might cause create_pool() to fail. Here I have one more situation modelled in the engine.
The question is, do we need error codes for these, or should we just have strings.

Here is the rundown:

1. The devices are already in use by something and we dare not touch them. This is an internal Stratis decision. If the client is the CLI, it would be really nice to give them that information, obviously. Right now, they would get a generic stratis error with a generic stratis chunk of text.
2. The pool name is already in use. The argument is the same, and the CLI user would really like to know that that was their mistake, since the action they would take is completely different from that for case (1).
3. Everything was great, and something terrible happened while creating the pool. Again, this is a wholly different situation from the previous 2,  and we want to distinguish it from the others.

So, I think we need to do something better than just return a generic stratis error to the client of the dbus API for all these things. And given that their may be other things that interact with our engine besides the dbus API, I think we should consider their information requirements.

Also, and this is a principle which I think is important, I think Stratis should catch and wrap all its errors, rather than deliberately allowing errors from underlying implementations to propagate. Any exceptions that Stratis raises from underlying libraries that aren't Stratis errors are not part of its API and are a bug to be fixed. The reasons are that any other approach has the following flaws:
1. Lack of information hiding...if you change the implementation then you have to fake up the old exceptions on top of what the new implementation does.
2. Impossible to maintain API. Your clients are unlikely to believe wholeheartedly that the exceptions that they might throw are part of their API, especially if they are propagating them by accident from their underlying libraries.

mulkieran:
This still has lots of problems, for one:

The thing contained by ```EngineError::Stratis``` constructor should probably be more informative than a nullary field is. Right now, these are called EngineErrorEnums, and they are all nullary. Probably, they should be expanded to take arguments, probably of type str, as the best generic way to give further information. So, to the code ```ALREADY_EXISTS``` the question is what already exists? The same for the code ```BUSY```, the question is, what is busy?

agrover:
yeah, having the error include a string seems helpful.

mulkieran:
This still has a lot of problems, for one:

Their should be some way for some of the valuable information in the EngineErrors to be transmitted to the user on the d-bus.

mulkieran:
rebased.

mulkieran:
For example, in ```destroy_pool()``` we need to know what is busy. Maybe the busy thing will be at the filesystem level, maybe at the device level, maybe it will be just that the pool is busy. We get back an error from the engine. We should return the error code, properly translated, and the more specific message, from the engine. (This means we need to add a BUSY to the dbus error enum.)

In ```create_pool()```, we have another busy, an AlreadyExists, and a generic error that can be mapped to generic error.

So, ```code_to_message_items(engine_to_dbus_err(...))``` needs a rethink.

mulkieran:
This PR does two things:

* Makes the SimEngine use randomization in order to have errors that it could not otherwise model. Some errors it can only model in the most general terms, it uses "X" for those.
* Does the necessary work to get those errors and the helpful information associated with them onto the dbus.

There is probably room for refinement.

@agrover, @tgill review?

agrover:
Without looking yet at the content of this PR, please update the PR title, and let's ping our colleague who is not tgill but rather @trgill :smile:

agrover:
Thanks. But also, I'm still confused by asking for a review (and implicitly to merge if it looks good) while the title still says "DO NOT MERGE".

mulkieran:
"DO NOT MERGE" means "do not merge". This PR makes a number of fairly significant changes. It seems foolish to refine to the last degree functions that may have to be completely rewritten or tossed pursuant to a review. So, I didn't do that. Of course, if there are no significant objections to the PR I will turn out to have made the wrong choice.

agrover:
looks good to me so far. So, nix and io errors still are present in EngineError, I think we talked about having engines do some error conversions? Is that coming or did plans change?

What do you think about renaming StratisErrorEnum to DbusErrorEnum or similar? It is actually DBus-specific I believe.

mulkieran:
I wouldn't mind changing _every_ error enum to just be ErrorEnum and just using the module name for disambiguation.

StratisErrorEnum is definitely part of the dbus API. But, I'm thinking that it might be part of some other way we set up of talking to the engine as well, like some other RPC mechanism might also want to send back the identical set of constants. So, I don't think it should be given a DBus specific name.

mulkieran:
About the Nix and IO constructors in EngineError enum:

We don't actually want to use those two constructors in our real code, ever. But, we have to handle code that might return them in our real engine, for sure.

We have just two options:

* In EngineError, define a single constructor, Stratis. In StratisErrorEnum, define a constructor that has the explicit job of wrapping other errors. Always wrap these system/io errors in that constructor. Pursue as our goal, using that constructor as little as possible, but instead handling those errors more intelligently.

* Keep it like it is, but pursue as our goal never using the IO and Nix constructors, but rather returning everything as Stratis.

Either way, explicitly document our intention, and say that that stuff does not constitute part of our public API, it is just a bug.

These two approaches are roughly equivalent. I think the only question is which is easier to implement and more extensible.

mulkieran:
rebased

mulkieran:
rebased.

agrover:
LGTM :+1:

mulkieran:
@trgill review?

PR #75, opened 2016-10-07
Master small fixes

mulkieran:


src/sim_engine.rs
```diff
@@ -48,12 +48,10 @@ impl Engine for SimEngine {
 
     fn list_pools(&self) -> StratisResult<BTreeMap<String, Box<StratisPool>>> {
 
-        let mut stratis_pools = BTreeMap::new();
-
-        for (name, pool) in &self.pools {
-
-            stratis_pools.insert(name.clone(), pool.copy());
-        }
+        let stratis_pools = self.pools.iter().fold(BTreeMap::new(), |mut acc, x| {
```
agrover: BTreemap implements FromIterator so maybe see if a chain of iter(), cloned(), maybe a map()?, and then collect() lets you do this and avoid the fold

src/sim_engine.rs
```diff
@@ -48,12 +48,10 @@ impl Engine for SimEngine {
...
+            acc.insert(x.0.clone(), x.1.copy());
+            acc
+        });
 
         Ok(stratis_pools)
```
agrover: don't need this line/variable if you wrap the previous expession in `Ok()`

src/dbus_api.rs
```diff
@@ -223,12 +223,12 @@ fn create_pool(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
 
     match result {
         Ok(_) => {
-            let dbus_contex_clone = dbus_context.clone();
-            let object_path_message_item = create_dbus_pool(dbus_contex_clone);
+            let dbus_context_clone = dbus_context.clone();
```
agrover: It might be tidier to pass `&dbus_context` to create_dbus_pool and do the clone there, instead of here, outside.

src/dbus_api.rs
```diff
@@ -53,9 +53,9 @@ impl DbusContext {
             engine: engine.clone(),
         }
     }
-    pub fn get_next_id(&mut self) -> String {
+    pub fn get_next_id(&mut self) -> u64 {
         self.next_index += 1;
-        self.next_index.to_string()
+        self.next_index
```
agrover: Looks good. BTW My immediate reaction was "why isn't the callsite in this commit too, surely we need to change that as well?" but we don't because it already had a (previously extraneous) `to_string()` call.

src/sim_engine.rs
mulkieran: This is done.

src/sim_engine.rs
mulkieran: This is done.

src/dbus_api.rs
mulkieran: I didn't touch this, because I didn't introduce that clone, and I don't know the reason for the clone in the first place.

mulkieran:
The final three commits are new, with the last one being more ambitious. The rest do not need further review. The original version of "Handle possibility of failure" was wrong.

AFAICT, there is no good general way to abstract the operations of extracting values form MessageItems or wrapping them up appropriately. The latter is something that it is necessary to do in the Stratis CLI, and I have implemented it using what seems like the only good technique, but that will not work in Stratisd because it would require that Rust have dependent types. So, the best that can be done is to write convenience methods for the commons cases, as far as I can tell. That is the plan, which I will attempt to implement in a subsequent PR.

PR #76, opened 2016-10-12
Changes to dbus api methods

mulkieran:
Uses the recommended approach for extracting data from message.

Distinguishes between missing args and invalid types of args.

If invalid type of arg, gives the _index_ of the argument, not its actual value, which I think is just as good.

Also, swaps the order of devs and raidtype in create_pool.


mulkieran:
This is not ready yet, I will see if the similar can be done with the append calls.

mulkieran:
@agrover review?

It looks like the append*() methods, if used to their fullest extent, would be full of surprises, so I'm just going to rest on this. The last three commits are therefore just bug fixes and minor fixups.

Issue #34, opened 2016-09-26
Translation concerns should be explictly postponed.


mulkieran:
Otherwise, we waste the time of the translators, among other things.

agrover:
It seems to me like Linux is localized for user-facing programs, but admin-oriented tools are generally not. e.g. btrfs-tools is not. lvm2's is incomplete and disabled. Cockpit is localized. Right now I'd think Stratis falls into the category of tools where i18n is not expected.

mulkieran:
Sounds good. Let us consider it postponed until it is either needed or customary, which may be never. I'll close this tomorrow unless there is an objection.

mulkieran:
Ok, closing!

PR #83, opened 2016-10-13
Get rid of an internationalization TODO.

mulkieran:
We explicitly agreed that this could be postponed to a distanct future in #34.


mulkieran:
@trgill review?

PR #84, opened 2016-10-13
create DeferredAction queue to update the tree outside of dbus method.

trgill:

### DIFF
--- a/src/dbus_consts.rs
+++ b/src/dbus_consts.rs
@@ -77,7 +77,6 @@ impl HasCodes for StratisErrorEnum {
 
     fn get_error_string(&self) -> &str {
         match *self {
-            // TODO deal with internationalization/do this better
             StratisErrorEnum::STRATIS_OK => "Ok",
             StratisErrorEnum::STRATIS_ERROR => "A general error happened",
             StratisErrorEnum::STRATIS_NULL => "Null parameter was supplied",


trgill:
@agrover  @mulkieran  review?

agrover:
Please commit cargo.lock and remove it from .gitignore. rationale: we are an app, not a library: http://doc.crates.io/guide.html#cargotoml-vs-cargolock

Please add more of a commit message. This patch actually does 3 things: reformat some unrelated code (accidentally?), add the deferred action stuff, and also change how we're handling the message loop from tree.run() to conn.iter(). It would be nice to mention these things and why they were all necessary and related.

also, now that you've added more changes to this PR, please update the PR title accordingly :smile: 

Make sure to also include signed-off-by on these additional commits.

src/sim_engine.rs
```diff
@@ -6,15 +6,21 @@ use blockdev::BlockDevs;
...
-
+use std::fmt::Debug;
 use std::path::Path;
 use std::collections::BTreeMap;
 use std::iter::FromIterator;
 
 use types::StratisResult;
 
 
+pub trait StratisPool: Debug {
```
agrover: StratisPool is a trait that all Engine implementations will be using. Therefore it would be better to put it in engine.rs, because sim_engine.rs should just have the specific code and data types associated with its implementation of the Engine and StratisPool traits.

agrover: It also might be good (now that struct Pool is gone) to just call this trait Pool. i.e make the generic traits the short, generic words w/o prefix.

This will happen to let StratisEngine call its struct that implements the Pool trait be called StratisPool :smile:

PR #86, opened 2016-10-14
create DeferredAction queue to update the tree outside of dbus method.

trgill:
Note: Some unrelated code was reformatted due to running rustfmt.

The dbus binding doesn't support adding/removing objects while
in a dbus method.  The choice was made to make the tree immutable
while processing a call for an object in that tree.

The author of the binding suggested processing
add/remove operations via a TODO (DeferredAction) queue.

Details can be found at:
https://github.com/diwic/dbus-rs/issues/52

Deferring updates to the tree required the main
message loop to be restructured.  Using the tree.run() method
passed ownership, so the tree became imutable in the loop. So
rather than calling tree.run(), we implement the run code
copied from TreeServer.  This allows us to maintain ownership
of the tree and update it based on the contents of the queue.

### DIFF
--- a/src/dbus_api.rs
+++ b/src/dbus_api.rs
@@ -24,17 +24,26 @@ use dbus::tree::MTFn;
 use dbus::tree::MethodResult;
 use dbus::tree::MethodInfo;
 use dbus::tree::Tree;
+use dbus::tree::ObjectPath;
+use dbus::ConnectionItem;
 
 use dbus_consts::*;
 
 use engine::Engine;
 use types::{StratisResult, StratisError};
 
-#[derive(Clone,Debug)]
+#[derive(Debug)]
+pub enum DeferredAction {
+    Add(ObjectPath<MTFn<TData>, TData>),
+    Remove(ObjectPath<MTFn<TData>, TData>),
+}
+
+#[derive(Debug)]
 pub struct DbusContext {
     pub next_index: u64,
     pub pools: BTreeMap<String, String>,
     pub engine: Rc<RefCell<Engine>>,
+    pub action_list: Vec<DeferredAction>,
 }
 
 
@@ -44,6 +53,7 @@ impl DbusContext {
             next_index: 0,
             pools: BTreeMap::new(),
             engine: engine.clone(),
+            action_list: Vec::new(),
         }
     }
     pub fn get_next_id(&mut self) -> u64 {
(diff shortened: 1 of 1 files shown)


trgill:
@agrover @mulkieran review?

I waill address the remainder of agrovers comments in a subsequent PR.

agrover:
:+1: looks good, thanks!

mulkieran:
AFAICT, from a client's point of view, all this does is to increase the likelihood of getting an object path that doesn't yet exist. Since that could happen anyway, the only important thing for the client to understand is that it is wrong to assume that, if the object path is not there, that's because it is already gone, instead of not yet arrived, which the client could assume before this change.

Assuming that the ```create_pool()``` stuff is good, you still need to do something similar in ```destroy_pool()```, AFAIU.

Couldn't you just as easily register the DeferredAction in ```create_pool()``` rather than ```create_dbus_pool``` in order to make things more consistent with functions that do not need to call a whopping big internal function, but still need to register an action, e.g., ```destroy_pool()```? I think we'll have a bunch of destroy* functions like that.

agrover:
I don't think there is any client-visible difference. Let's look at `run()` in dbus_api.rs. We're calling tree.handle(), then sending back return/error messages (client now receives and now knows new object path), then processing the action_list. We could do the last two things in reverse order but either way there's no race because we will always process the action_list before we process the next possible place the new object path could be used by the client -- the next call to tree.handle().

agrover:
...we could be in trouble if c.register_object_path fails, but that should never fail :smile: :smile: :smile:

diwic:
@agrover  You can call c.register_object_path from inside the method handler, as well as creating the new ObjectPath. All that needs to be deferred is the actual insertion of the object path into the tree.

agrover:
Good point. It's better to put all the things that can fail in the method handler, so we can return an error and keep going.

Also, if we ever have a method that adds multiple object paths, we likely want to do what `Tree::set_registered` does, to clean up 0..n-1 if c.register_object_path(N) fails.

mulkieran:
I am glad that it is the case that if the client gets an object with the object path and a method call fails that means that
1. There is a bug.
2. The object path was there but has gone away.

Not having to add:

3. The object is not there yet.

is actually a very great improvement for the client. Maybe that is worth documenting in the source, since that guarantee is really quite important.

PR #90, opened 2016-10-18
Example of not-working test infrastructure

mulkieran:
Signed-off-by: mulhern amulhern@redhat.com

### DIFF
--- a/tests/engine_test.rs
+++ b/tests/engine_test.rs
@@ -1,0 +1,1 @@
+extern crate stratisd;


mulkieran:
I was attempting to follow the directions at: https://doc.rust-lang.org/book/testing.html#the-tests-directory, but it all fails. I think this is because stratisd is an application, not a library, but I'm not sure. Here's the error:

```
[mulhern@dhcp-25-209 stratisd]$ cargo test
   Compiling stratisd v0.1.0 (file:///home/mulhern/my-contributions/stratisd)
tests/engine_test.rs:1:1: 1:23 error: can't find crate for `stratisd` [E0463]
tests/engine_test.rs:1 extern crate stratisd;
                       ^~~~~~~~~~~~~~~~~~~~~~
error: Could not compile `stratisd`.

To learn more, run the command again with --verbose.
```

I was trying to figure out a way to run some tests against the engine. Putting them in a separate directory seems like the correct thing, because:

* They should be designed to pass both against the Sim Engine and against the real engine.
* They may become quite complicated.

Is there a way we can split stratisd into two parts, the app part and the lib part?

mulkieran:
Oboleted by PR #164.

PR #94, opened 2016-10-18
Add parameters to pool interface / place holders for pool interface pass parameters to engine

trgill:
- initial place holders for pool parameters
- fixed problem with interface name
- added place holders for parsing the parameters to create_volumes
- insert a mapping of the object path name to the pool name in the dbus_context


trgill:
@agrover @mulkieran review?

src/dbus_api.rs
```diff
@@ -121,7 +121,30 @@ fn list_pools(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
 
 
 fn create_volumes(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
+    let message: &Message = m.msg;
+    let mut iter = message.iter_init();
+    let dbus_context = m.path.get_data();
```
mulkieran: We're aiming to get the stuff having to do with actually doing the action to all happen after the stuff that involves parsing the message information. So, line 126 and 127 should be moved down to right above line 134, or thereabouts.

src/dbus_api.rs
```diff
@@ -121,7 +121,30 @@ fn list_pools(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
...
+
+    match dbus_context.borrow_mut().pools.get(&object_path.clone().to_string()) {
+        Some(pool) => println!("{}: {}", object_path, pool),
+        None => println!("{} not  found.", object_path),
+    }
 
+    let _volumes: Array<(&str, &str, &str), _> = try!(iter.read::<Array<(&str, &str, &str), _>>()
+        .map_err(|_| MethodErr::invalid_arg(&2)));
+
+    // let result = {
```
mulkieran: I'ld like to ditch commented out code.

src/dbus_api.rs
```diff
@@ -121,7 +121,30 @@ fn list_pools(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
...
+
+    match dbus_context.borrow_mut().pools.get(&object_path.clone().to_string()) {
+        Some(pool) => println!("{}: {}", object_path, pool),
+        None => println!("{} not  found.", object_path),
+    }
+
+    let _volumes: Array<(&str, &str, &str), _> = try!(iter.read::<Array<(&str, &str, &str), _>>()
+        .map_err(|_| MethodErr::invalid_arg(&2)));
+
+    // let result = {
```
mulkieran: I'ld like to ditch this commented out code.

src/dbus_api.rs
```diff
@@ -121,7 +121,30 @@ fn list_pools(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
 
 
 fn create_volumes(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
+    let message: &Message = m.msg;
+    let mut iter = message.iter_init();
```
mulkieran: You can still follow this up with the protocol for extraction that we have arrived at:

```
if iter.arg_type() == 0 {
    return Err(MethodErr::no_args());
}
let spec: Array<(&str, &str, &str), _> = try!(iter.read::<Array<(&str, &str, &str), _>>().map_err(|_| MethodErr::invalid_arg(&0));
```

then print it out however you want.

src/dbus_api.rs
```diff
@@ -241,6 +297,13 @@ fn create_pool(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
             let dbus_context_clone = dbus_context.clone();
             let object_path: dbus::Path = create_dbus_pool(dbus_context_clone);
             let (rc, rs) = code_to_message_items(StratisErrorEnum::STRATIS_OK);
+
+            // Insert a mapping of the object path name to the pool name to be
+            // used in the pool interface methods to look up the Stratis pool.
```
mulkieran: This cloning really worries me, because I know it means we're making a copy every time. There seem to be issues of efficiency and correctness. Other than that, I definitely think we need this table, somehow, so this is right.

mulkieran: And I think the comment will very soon seem completely redundant, so might as well just leave it out.

src/dbus_api.rs
```diff
@@ -165,31 +188,64 @@ fn create_dbus_pool<'a>(dbus_context: Rc<RefCell<DbusContext>>) -> dbus::Path<'a
 
     let f = Factory::new_fn();
 
-    let create_volumes_method = f.method(CREATE_VOLUMES, (), create_volumes);
+    let create_volumes_method = f.method(CREATE_VOLUMES, (), create_volumes)
+        .in_arg(("volumes", "a(sss)"))
+        .out_arg(("results", "a(sqs)"))
```
mulkieran: I think the signature for this should probably be "a(oqs)", since I believe the first part of the result will be an object path. There may be a few more like that.

src/dbus_api.rs
trgill: Ok, I addressed this issue (if I understand it correctly)

src/dbus_api.rs
trgill: Ok

src/dbus_api.rs
trgill: done

src/dbus_api.rs
trgill: done

trgill:
rebased on top of volume to filesystem rename....

src/dbus_api.rs
```diff
@@ -241,6 +296,11 @@ fn create_pool(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
             let dbus_context_clone = dbus_context.clone();
             let object_path: dbus::Path = create_dbus_pool(dbus_context_clone);
             let (rc, rs) = code_to_message_items(StratisErrorEnum::STRATIS_OK);
+
+            let dbus_context_clone = dbus_context.clone();
```
mulkieran: What is the news on the cloning? I think at one time @agrover suggested that the context should be cloned within the ```create_dbus_pool()``` method.

src/dbus_api.rs
```diff
@@ -120,7 +120,29 @@ fn list_pools(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
...
+    let message: &Message = m.msg;
+    let mut iter = message.iter_init();
+
+    if iter.arg_type() == 0 {
+        return Err(MethodErr::no_arg());
+    }
+
+    let _volumes: Array<(&str, &str, &str), _> = try!(iter.read::<Array<(&str, &str, &str), _>>()
+        .map_err(|_| MethodErr::invalid_arg(&0)));
+
```
trgill: yes.

all the println! calls are going to be removed once I get it working.  this is intended as
a place holder for making the calls to engine.

src/dbus_api.rs
agrover: Yeah a clone just bumps the refcount on the Rc.

The style comment I was making was just that we could pass a ref to the Rc<RefCell<DbusContext>> into create_dbus_pool and then it could clone internally, and that that would be slightly tidier.

PR #96, opened 2016-10-19
Shorten enum names and re-order

mulkieran:
These changes arise out of a discussion on #65.

On the whole they should be non-controversial.

Even the "Shorten..." one shouldn't be troublesome, because if needed the names that go with the values that are published on the d-bus or other mechanism can be extended to include ```STRATIS_```, i.e., they don't need to be in the constructor names.


mulkieran:
@agrover , @trgill review?

agrover:
I'm cool with all these except f99c88e, because I think there's a natural order that differs from alphabetical, also because "single" is kind of bad so I think we may end up with another name (maybe "none"? "linear"? TBD) so it would then move back.

mulkieran:
@grover I think that it is likely to turn out to be the case in the end that our list of raid levels becomes more subtle and looses the property of having any reasonable "natural" order. But I'll drop the last commit, it is not too important.

mulkieran:
@grover, @trgill review?

agrover:
:+1:

PR #101, opened 2016-10-20
Add simulated block devices to SimEngine implementation.

mulkieran:


mulkieran:
@agrover, @trgill  review?

src/sim_engine/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod engine;
```
agrover: I think you can also re-export stuff here, so in main.rs it could still say `use sim_engine::SimEngine` if desired.

agrover:
Also, this PR adds SimDev but also significantly puts sim_engine in its own directory, which I would probably say deserves mention in the PR title.

mulkieran:
@agrover, @trgill review?

PR #102, opened 2016-10-20
framework for adding a filesystem to a pool / wrap internal members of dbus_context in Rc<RefCell<>>

trgill:
added NotFound to ErrorEnum
added code to create track a filesystem associated with a pool in the SimEngine
added initial code for the filesystem interface (parameters)
added BTreeMap to track filesystems in the SimPool
changed internal memebers of dbus_context to be wrapped in a Rc<RefCell<>> to simplify borrowing issues
surpressed compiler warnings by adding _ to variable names in stubbed out prototypes
added stubs for pool interface


trgill:

src/sim_engine/pool.rs
```diff
@@ -59,10 +66,18 @@ impl Pool for SimPool {
...
             raid_level: self.raid_level.clone(),
             online: true,
             checking: false,
         };
         Box::new(pool_copy)
     }
+    fn create_filesystem(&mut self,
+                         _filesystem_name: &str,
+                         _mount_point: &str,
+                         _size: &str)
```
mulkieran: Change ```_size``` to a u64. This is certainly better than a string, although not ideal.

src/sim_engine/engine.rs
```diff
@@ -89,6 +89,15 @@ impl Engine for SimEngine {
 
         Ok(())
     }
+    fn get_pool(&mut self, name: &str) -> EngineResult<&mut Box<Pool>> {
```
mulkieran: Is it possible to drop the ```&mut```?

trgill:
@mulkieran am I good to merge after addressing those two issues?  (I'll wait for agrover too)

src/sim_engine/pool.rs
agrover: We do have a `Sectors` type, how about that?

src/sim_engine/engine.rs
agrover: no, because of the return type. I think we tried this and compiler caught it. It's returning a mutable reference into the Engine, so it needs to start with a mutable reference to the Engine.

agrover:
:+1:

src/sim_engine/pool.rs
mulkieran: We should take the opportunity to define exactly what Sectors represent in that case and why it is suitable or adequate for this method. Right now, our design document does not use the word "sector" at all.

It looks like we can get rid of SumSectors pretty soon, based on progress of https://github.com/rust-lang/rust/issues/34529.

I am actually the world's leading expert in the representation of and computation with allocated address ranges for block storage, see: https://github.com/mulkieran/justbytes, and https://mojo.redhat.com/docs/DOC-1089158. (It's like being on the North Korean Olympic curling team). But, it would be a shame to waste my expertise in this area, if it would be useful.

src/sim_engine/engine.rs
mulkieran: I'm puzzled by the fact that it has to return a mutable reference to a Box.

src/sim_engine/pool.rs
```diff
@@ -2,18 +2,24 @@
...
 
+use std::collections::BTreeMap;
 use std::vec::Vec;
 
 use engine::EngineResult;
 use engine::Pool;
 
 use super::blockdev::SimDev;
 
+#[derive(Debug, Clone)]
```
mulkieran: Maybe you could break this currently tiny definition and implementation out to a separate file. It will get bigger.

src/dbus_api.rs
```diff
@@ -141,10 +143,111 @@ fn list_pools(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
...
+
+    let dbus_context = m.path.get_data();
+    let object_path = m.path.get_name();
+    let return_message = message.method_return();
+
+    println!("Object Path {}", object_path);
+
+    let pool_name = match dbus_context.pools.borrow_mut().get(&object_path.to_string()) {
+        Some(pool) => pool.clone(),
+        None => return Err(MethodErr::invalid_arg(&0)),
```
mulkieran: I think this should be a return message with an Err result on the dbus.

src/dbus_api.rs
```diff
@@ -141,10 +143,111 @@ fn list_pools(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
...
+    };
+
+    let mut b_engine = dbus_context.engine.borrow_mut();
+    let ref mut pool = match b_engine.get_pool(&pool_name) {
+        Ok(result) => result,
+        Err(x) => {
+            let (rc, rs) = engine_to_dbus_err(&x);
+            let (rc, rs) = code_to_message_items(rc, rs);
+            let entry = MessageItem::Struct(vec!["o".into(), "q".into(), "s".into()]);
```
mulkieran: I think it is necessary to use specific correct MessageItem constructors for the components of the message.

src/sim_engine/engine.rs
mulkieran: But, the more I look, the more it looks like it has to be. OK.

mulkieran:
There are more changes I'm requesting. I realize now I could probably have done that as a second review.

src/sim_engine/pool.rs
trgill: @agrover ok if I change _size to u64 and we tackle this topic in another PR?

agrover: yup.

src/dbus_api.rs
trgill: I changed to:
    
        return Err(MethodErr::invalid_arg(&format!("object path not found {}",
                                                       object_path.to_string())))
Ok?

src/dbus_api.rs
```diff
@@ -141,10 +143,111 @@ fn list_pools(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
...
+
+    let mut filesystems: Array<(&str, &str, &str), _> =
+        try!(iter.read::<Array<(&str, &str, &str), _>>()
+            .map_err(|_| MethodErr::invalid_arg(&0)));
+
+    let dbus_context = m.path.get_data();
+    let object_path = m.path.get_name();
+    let return_message = message.method_return();
+
+    println!("Object Path {}", object_path);
```
trgill: done

src/sim_engine/pool.rs
trgill: ok if we do that on next PR?

mulkieran: Sure.

trgill:
PR review comments have been addressed.

PR #105, opened 2016-10-26
Create filesystem in engine and map from object path to pool name, filesystem name pair in dbus

mulkieran:


mulkieran:
@agrover @trgill review?

This is incomplete because:

1. The engine method can't fail, but it should be able to.
2. The engine method should probably be ```create_filesystems()``` with a plural, so that we can realize efficiencies possible with creating multiple filesystems at once.

But, I think it would be better to get this in sooner rather than later, so that we can have more working bits.

src/dbus_api.rs
```diff
@@ -43,19 +43,21 @@ pub enum DeferredAction {
 
 #[derive(Debug, Clone)]
 pub struct DbusContext {
+    pub action_list: Rc<RefCell<Vec<DeferredAction>>>,
```
agrover: My first reaction to alphabetizing struct members is negative. This doesn't help readability. I have a hard time articulating a better policy other than "put them in the order they should be in" or "it doesn't matter" but I do think there are reasons to order members a certain way, and a policy to alphabetize them is counterproductive.

This commit also mixes a substantive change with the ordering change which is stylistic, so it would've been nice for them to be in separate commits.

agrover:
An `EngineResult<()>` return can return `EngineError`. See engine.rs:43.
```
pub type EngineResult<T> = Result<T, EngineError>;
```
it just returns nothing (aka `()`) on success.

mulkieran:
Like ```create_pool()```, ```create_filesystem()``` should have some failure modes that model possible modes of failure that might occur in real life. It doesn't, because I didn't implement them, something which I am deferring until later. That is what "The engine method can't fail, but it shoudl be able to" means.

mulkieran:
DbusContext fields dis-ordered. @trgill , ready to merge.

PR #106, opened 2016-10-27
Initial implementation of list_filesystems

trgill:
Added Filesystem trait to engine.rs
Created filesystem.rs in sim_engine/
Added filesystem to mod.rs
Added list_filesystem to the sim_engine/pool.rs
Added name, mount_point, and size to the SimFilesystem
Loop on the items in filesystems (not filesystems.next())
Remove mut from filesystems (not needed)
Add object_manager() to the base interface
Remove "name" field from Pool and Filesystem

### DIFF
--- a/src/dbus_api.rs
+++ b/src/dbus_api.rs
@@ -33,7 +33,6 @@ use dbus_consts::*;
 use engine;
 use engine::Engine;
 use engine::EngineError;
-
 use types::StratisResult;
 
 #[derive(Debug)]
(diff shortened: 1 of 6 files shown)


trgill:

agrover:
please fix typos in e0ac54c commitmsg, and I am :+1:

src/dbus_api.rs
```diff
@@ -323,6 +322,7 @@ fn list_filesystems(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
 
     let msg = match result {
         Ok(filesystem_tree) => {
+            println!("filesystems size = {}", filesystem_tree.len());
```
mulkieran: Drop println.

src/sim_engine/filesystem.rs
```diff
@@ -6,15 +6,29 @@ use engine::Filesystem;
 
 #[derive(Debug, Clone)]
 pub struct SimFilesystem {
+    pub name: String,
```
mulkieran: Could we leave out name field as it is redundant because we already map name to fileystem object.

src/dbus_api.rs
```diff
@@ -286,7 +284,60 @@ fn destroy_filesystems(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
 }
 
 fn list_filesystems(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
-    Ok(vec![m.msg.method_return().append3("/dbus/cache/path", 0, "Ok")])
+    let message: &Message = m.msg;
+    let mut iter = message.iter_init();
+
+    if iter.arg_type() != 0 {
```
mulkieran: Drop these three lines, there should not be any args.

mulkieran:
I have three requests, two of which are marked as outdated, I don't know why.

src/dbus_api.rs
mulkieran: We could start complaining if there are too many args, but that seems like overkill. In any case, we should do it consistently, if we decide to.

src/sim_engine/filesystem.rs
trgill: we've got a "name" field in the Pool already... we should either remove or keep both, right?

src/dbus_api.rs
mulkieran: Oh, I get it. Well that's an addition to the standard approach. What we're really complaining about is extra args...in that case we should complain about it everywhere we've filled in a method. Could you toss in a few more in the completed methods?

src/sim_engine/filesystem.rs
mulkieran: Yes. I'm all for removing both.

agrover: makes sense

src/dbus_api.rs
trgill: Ok, the only other completed method that takes no arguments is list_pools.  I updated it.

I think the check makes sense. But if folks disagree, I'll take them out.

PR #110, opened 2016-10-31
Abstract pool lookup

mulkieran:
```
   Compiling stratisd v0.1.0 (file:///home/mulhern/my-contributions/stratisd)
src/dbus_api.rs:99:22: 99:54 error: borrowed value does not live long enough
src/dbus_api.rs:99     let pool = match dbus_context.engine.borrow_mut().get_pool(&pool_name) {
                                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/dbus_api.rs:88:75: 108:2 note: reference must be valid for the lifetime 'a as defined on the block at 88:74...
src/dbus_api.rs:88                            -> Result<&'a mut Box<Pool>, Vec<MessageItem>> {
                                                                                             ^
src/dbus_api.rs:99:5: 106:7 note: ...but borrowed value is only valid for the statement at 99:4
src/dbus_api.rs:99     let pool = match dbus_context.engine.borrow_mut().get_pool(&pool_name) {
                       ^
src/dbus_api.rs:99:5: 106:7 help: consider using a `let` binding to increase its lifetime
src/dbus_api.rs:99     let pool = match dbus_context.engine.borrow_mut().get_pool(&pool_name) {
                       ^
error: Could not compile `stratisd`.
```

### DIFF
--- a/src/dbus_api.rs
+++ b/src/dbus_api.rs
@@ -253,14 +253,9 @@ fn create_filesystems(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
         default_return; return_message);
 
     let mut b_engine = dbus_context.engine.borrow_mut();
-    let ref mut pool = match b_engine.get_pool(&pool_name) {
-        Ok(result) => result,
-        Err(x) => {
-            let (rc, rs) = engine_to_dbus_err(&x);
-            let (rc, rs) = code_to_message_items(rc, rs);
-            return Ok(vec![return_message.append3(default_return, rc, rs)]);
-        }
-    };
+    let ref mut pool = engine_try!(b_engine.get_pool(&pool_name);
+                                   default_return;
+                                   return_message);
 
     let ref mut list_rc = ErrorEnum::OK;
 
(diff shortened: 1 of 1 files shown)


mulkieran:
@agrover Any hints on how to make this work?

agrover:
I think the borrowed engine just needs to live longer? Try doing what other places do, with `b_engine` binding?

mulkieran:
I already tried that and I get the same error.

mulkieran:
The gains aren't great, but they will seem like more as we develop more methods.

I want to know why the compiler doesn't complain about the ```default_return``` variable, which is used twice, although in two mutually exclusive contexts.

The big take home is that abstracting getting the pool object from the pool name is impossible because the lifetime of the pool must not extend beyond the lifetime of the engin.

mulkieran:
@agrover @trgill review?

agrover:
There's more to do but this is going in the right direction. :+1:

mulkieran:
Yes. I think it pretty obviously needs a macro, called, ```try_dbus```, so I'll try to add that.

mulkieran:
The last two are minor fixes, but they're fixes on code that was affected by previous commits, so I think they might as well be put here.

I have been wary of using macros, for obvious reasons, but I think these are unambitious macros, with a well-defined purpose, analogous to the ```try!``` macro and fairly easily understood, so I think I'll stick with them.

I think making an uber-macro that did both tasks, got the name from the ojbect path and then the pool from the name would add a degree of complexity and specificyt that would not be helfpul.

So, I'll put this pu for review once more.

@agrover @trgill review?

trgill:
:+1:

agrover:
Can we replace the return_sig vars with consts in dbus_consts?

mulkieran:
I don't think that would make sense since they are specific to the function.

agrover:
:+1:

PR #115, opened 2016-11-01
Replace BTreeMap with BidirMap / Correct types for Size properties

trgill:
Using a bi-directional map will allow looking up the dbus
object path given the Stratis name and looking up the
Stratis name given the dbus object path.

Fixed compile problem 
Corrected type of Size property


trgill:

mulkieran:
:+1:

PR #116, opened 2016-11-01
Fix a compiler error.

mulkieran:
Whoops, that should never happen.


mulkieran:
Already fixed.

PR #118, opened 2016-11-02
Implement some Get*ObjectPath methods

mulkieran:


mulkieran:
@agrover @trgill review?

I believe the ```unwrap()``` 's are justified since our code should always ensure that the object paths extracted from the various maps are valid.

At one point, I have to take a tuple of references to Strings and convert it into a reference to a tuple of Strings, which I have to do by cloning the Strings. I'm not aware of a better way.

I deferred the two other Get*ObjectPath methods until later because they seem to have become a bit ill-defined, in the case of ```GetCacheObjectPath()``` or more tricky, in the case of ```GetDevObjectPath()```.

mulkieran:
Whoops, wait a sec. A failure to unwrap should return INTERNAL_ERROR.

mulkieran:
@agrover @trgill review?

I think it is ready now.

trgill:
:+1:  looks good

src/dbus_api.rs
```diff
@@ -667,12 +715,48 @@ fn destroy_pool(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
...
-    Ok(vec![m.msg.method_return().append3("/dbus/filesystem/path", 0, "Ok")])
+    let message: &Message = m.msg;
+    let mut iter = message.iter_init();
+
+    if iter.arg_type() == 0 {
+        return Err(MethodErr::no_arg());
+    }
+    let pool_name: &str = try!(iter.read::<&str>().map_err(|_| MethodErr::invalid_arg(&0)));
+
+    if iter.arg_type() == 0 {
```
agrover: is this supposed to be here? identical code on line 740

mulkieran: Yes. It's not quite identical. The first time the ```pool_name``` is read, the second time the ```name``` of the filesystem.

mulkieran: ```iter.read``` advances through the args, so that chunk is used to check if there exists a next arg. it is the technique recommended by the dbus-rs developer.

agrover:
Ah yes I recall now. 👍

PR #119, opened 2016-11-02
add ListDevs and ListCacheDevs

trgill:
Put all relevant traits in engine.rs
Remove unused files
Implement list_cache_devs and list_devs
Remove list_cache (duplicate)
Move the Filesystem trait to filesystem.rs (from engine.rs)
Add required methods to SimDev and SimCacheDev
Add working list_blockdevs and list_cachedevs to pool.rs
Correct signature of destroy_filesystems
Use default_return where possible.


trgill:

agrover:
:+1:

src/sim_engine/blockdev.rs
```diff
@@ -26,14 +28,30 @@ pub struct SimDev {
...
+    fn copy(&self) -> Box<Dev> {
+        let simdev_copy = SimDev {
+            name: self.name.clone(),
+            rng: self.rng.clone(),
+            state: self.state.clone(),
+        };
+        Box::new(simdev_copy)
+    }
+    fn get_id(&self) -> String {
+        let id = self.name.to_str();
```
mulkieran: See the one-liner below that should be just as good.

But, I'm quite puzzled by an id that seems to be generated from a PathBuf.

src/sim_engine/blockdev.rs
```diff
@@ -26,14 +28,30 @@ pub struct SimDev {
     pub state: State,
 }
 
-/// Implement Debug for SimDev explicitly as ThreadRng does not derive it.
```
trgill: Sorry - will fix

src/cache.rs
```diff
@@ -1,4 +1,9 @@
 // This Source Code Form is subject to the terms of the Mozilla Public
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
+use std::fmt::Debug;
 
+pub trait Cache: Debug {
+    fn copy(&self) -> Box<Cache>;
+    fn get_id(&self) -> String;
```
trgill: I think we've got work to do in this area, so this is not intended to be a final step.  I'm just trying to replicate what we had in the C version.

In this PR, I'm using it to return an "ID" to the dbus calls ListDevs and ListCacheDevs.  This will be displayed to the user in the CLI.  Similar to how ListPools uses the key of the map.

Does this make sense?


In my next PR, I need a way to uniquely identify a element in the Vec to so I can delete it.

src/sim_engine/cache.rs
```diff
@@ -0,0 +1,65 @@
...
+        Box::new(cache_copy)
+    }
+
+    fn get_id(&self) -> String {
+        let id = self.name.to_str();
+
+        match id {
+            Some(x) => return String::from(x),
+            None => return String::from("Conversion Failure"),
+        }
```
trgill: I find the original easier to read, but maybe your approach is more "rustic"?

src/sim_engine/cache.rs
```diff
@@ -0,0 +1,65 @@
...
+    }
+    pub fn update(&mut self) {}
+
+    /// Checks usability of a SimCacheDev
+    pub fn usable(&self) -> bool {
+        self.state == CacheState::OK
+    }
+}
+
+impl fmt::Debug for SimCacheDev {
```
mulkieran: Yes. I'll look into that. The randomization should definitely be configurable.

src/cache.rs
trgill: I opened issue #122 to track your concern about using paths as IDs.

mulkieran:
:+1:

PR #123, opened 2016-11-03
Remove cache and devs

trgill:
Change block_devs and cache_devs maps to not store the pool. It isn't needed
Added code to create dbus objects for devs when a pool is created
Changed DeferredAction::Remove to take a string
Added remove_dbus_object_path
Implemented remove_cache_devs and remove_devs
Added eq method to Cache and Dev traits to facilitate remove from vec
Added remove_blockdev and remove_cachedev to Pool trait
Added get_id and eq to Filesystem
Implemented remove_blockdev and remove_cachedev in sim_engine
Correct output name of create_filesystems_method to "filesystems"
implement fmt::Display for SimDev
reformat eq code to be more readable for blockdev and cache

### DIFF
--- a/src/dbus_api.rs
+++ b/src/dbus_api.rs
@@ -765,7 +765,7 @@ fn create_dbus_pool<'a>(mut dbus_context: DbusContext) -> dbus::Path<'a> {
 
     let create_filesystems_method = f.method(CREATE_FILESYSTEMS, (), create_filesystems)
         .in_arg(("filesystems", "a(sst)"))
-        .out_arg(("results", "a(oqs)"))
+        .out_arg(("filesystems", "a(oqs)"))
         .out_arg(("return_code", "q"))
         .out_arg(("return_string", "s"));
 
--- a/src/dbus_api.rs
+++ b/src/dbus_api.rs
@@ -51,8 +51,8 @@ pub struct DbusContext {
     pub engine: Rc<RefCell<Engine>>,
     pub action_list: Rc<RefCell<Vec<DeferredAction>>>,
     pub filesystems: Rc<RefCell<BidirMap<String, (String, String)>>>,
-    pub block_devs: Rc<RefCell<BidirMap<String, (String, String)>>>,
-    pub cache_devs: Rc<RefCell<BidirMap<String, (String, String)>>>,
+    pub block_devs: Rc<RefCell<BidirMap<String, String>>>,
+    pub cache_devs: Rc<RefCell<BidirMap<String, String>>>,
 }
 
 impl DbusContext {
@@ -542,9 +542,9 @@ fn add_devs(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
         match result {
             Ok(_) => {
                 let object_path: dbus::Path = create_dbus_blockdev(dbus_context.clone());
-                dbus_context.block_devs.borrow_mut().insert(object_path.to_string(),
-                                                            ((&pool_name).clone(),
-                                                             String::from(dev.to_str().unwrap())));
+                dbus_context.block_devs
+                    .borrow_mut()
+                    .insert(object_path.to_string(), String::from(dev.to_str().unwrap()));
                 let (rc, rs) = ok_message_items();
                 let entry = MessageItem::Struct(vec![MessageItem::ObjectPath(object_path), rc, rs]);
                 vec.push(entry);
@@ -614,9 +614,9 @@ fn add_cache_devs(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
         match result {
             Ok(_) => {
                 let object_path: dbus::Path = create_dbus_cachedev(dbus_context.clone());
-                dbus_context.cache_devs.borrow_mut().insert(object_path.to_string(),
-                                                            ((&pool_name).clone(),
-                                                             String::from(dev.to_str().unwrap())));
+                dbus_context.cache_devs
+                    .borrow_mut()
+                    .insert(object_path.to_string(), String::from(dev.to_str().unwrap()));
                 let (rc, rs) = ok_message_items();
                 let entry = MessageItem::Struct(vec![MessageItem::ObjectPath(object_path), rc, rs]);
                 vec.push(entry);
@@ -737,6 +737,14 @@ fn create_pool(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
             let object_path: dbus::Path = create_dbus_pool(dbus_context.clone());
             let (rc, rs) = ok_message_items();
             dbus_context.pools.borrow_mut().insert(object_path.to_string(), String::from(name));
+            dbus_context.pools.borrow_mut().insert(object_path.to_string(), String::from(name));
+            for dev in blockdevs {
+                let dev_object_path: dbus::Path = create_dbus_blockdev(dbus_context.clone());
+                dbus_context.block_devs
+                    .borrow_mut()
+                    .insert(dev_object_path.to_string(),
+                            String::from(dev.to_str().unwrap()));
+            }
             return_message.append3(MessageItem::ObjectPath(object_path), rc, rs)
         }
         Err(x) => {
--- a/src/dbus_api.rs
+++ b/src/dbus_api.rs
@@ -41,7 +41,7 @@ use types::StratisResult;
 #[derive(Debug)]
 pub enum DeferredAction {
     Add(ObjectPath<MTFn<TData>, TData>),
-    Remove(ObjectPath<MTFn<TData>, TData>),
+    Remove(String),
 }
 
 #[derive(Debug, Clone)]
@@ -223,6 +223,10 @@ fn ok_message_items() -> (MessageItem, MessageItem) {
 
 fn default_object_path<'a>() -> dbus::Path<'a> {
     dbus::Path::new(DEFAULT_OBJECT_PATH).unwrap()
+}
+
+fn remove_dbus_object_path(dbus_context: &DbusContext, path: String) {
+    dbus_context.action_list.borrow_mut().push(DeferredAction::Remove(path));
 }
 
 fn list_pools(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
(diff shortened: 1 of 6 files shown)


trgill:

I fought with the Remove changes a bit.  The 'static lifetime specifier concerns me in this code:

      DeferredAction::Remove(path) => {
                        let object_path: dbus::Path<'static> = dbus::Path::new(path.clone())
                            .unwrap();
                        tree.remove(&object_path);
                        c.unregister_object_path(&path);
     }

Does it look ok?

agrover:
looks fine. I think you can omit the type altogether in that one spot. Also if you call `unregister_object_path` first you don't need the clone(). But either way, :+1: good work.

src/dbus_api.rs
```diff
@@ -478,7 +482,65 @@ fn list_cache_devs(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
...
+    let to_remove_devs = devs.map(|x| Path::new(x)).collect::<Vec<&Path>>();
+
+    for dev in to_remove_devs {
+        let result = pool.remove_cachedev(dev);
+        match result {
+            Ok(_) => {
+                let cache_devs_map = dbus_context.cache_devs.borrow_mut();
+                if let Some(remove_object_path) =
+                       cache_devs_map.get_by_second(&String::from(dev.to_str()
+                    .unwrap())) {
```
mulkieran: This unwrap() should probably be handled as an internal error. Better than just crashing.

src/dbus_api.rs
```diff
@@ -478,7 +482,65 @@ fn list_cache_devs(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
 }
 
 fn remove_cache_devs(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
-    Ok(vec![m.msg.method_return().append3("/dbus/cache/path", 0, "Ok")])
+    let message: &Message = m.msg;
+    let mut iter = message.iter_init();
+
+    let devs: Array<(&str), _> = try!(iter.read::<Array<(&str), _>>()
```
mulkieran: This skips a step, but with contents of PR# #124 it should be able to be boiled down to one line.

src/dbus_api.rs
```diff
@@ -530,9 +592,9 @@ fn add_devs(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
         match result {
             Ok(_) => {
                 let object_path: dbus::Path = create_dbus_blockdev(dbus_context.clone());
-                dbus_context.block_devs.borrow_mut().insert(object_path.to_string(),
-                                                            ((&pool_name).clone(),
-                                                             String::from(dev.to_str().unwrap())));
+                dbus_context.block_devs
+                    .borrow_mut()
+                    .insert(object_path.to_string(), String::from(dev.to_str().unwrap()));
```
mulkieran: I dunno how this unwrap should be handled, but it probably should be.

src/dbus_api.rs
```diff
@@ -625,7 +687,64 @@ fn add_cache_devs(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
     Ok(vec![return_message.append3(MessageItem::Array(vec, return_sig.into()), rc, rs)])
 }
 fn remove_devs(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
-    Ok(vec![m.msg.method_return().append3("/dbus/cache/path", 0, "Ok")])
+    let message: &Message = m.msg;
+    let mut iter = message.iter_init();
+
+    let devs: Array<(&str), _> = try!(iter.read::<Array<(&str), _>>()
```
mulkieran: This should use method from PR #124.

src/dbus_api.rs
```diff
@@ -625,7 +687,64 @@ fn add_cache_devs(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
...
+    let to_remove_devs = devs.map(|x| Path::new(x)).collect::<Vec<&Path>>();
+
+    for dev in to_remove_devs {
+        let result = pool.remove_blockdev(dev);
+        match result {
+            Ok(_) => {
+                let block_devs_map = dbus_context.block_devs.borrow_mut();
+                if let Some(remove_object_path) =
+                       block_devs_map.get_by_second(&String::from(dev.to_str()
+                    .unwrap())) {
```
mulkieran: unwrap should be handled.

src/dbus_api.rs
```diff
@@ -737,6 +856,14 @@ fn create_pool(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
             let object_path: dbus::Path = create_dbus_pool(dbus_context.clone());
             let (rc, rs) = ok_message_items();
             dbus_context.pools.borrow_mut().insert(object_path.to_string(), String::from(name));
+            dbus_context.pools.borrow_mut().insert(object_path.to_string(), String::from(name));
+            for dev in blockdevs {
+                let dev_object_path: dbus::Path = create_dbus_blockdev(dbus_context.clone());
+                dbus_context.block_devs
+                    .borrow_mut()
+                    .insert(dev_object_path.to_string(),
+                            String::from(dev.to_str().unwrap()));
```
mulkieran: unwrap

src/dbus_api.rs
```diff
@@ -964,8 +1091,10 @@ pub fn run(engine: Rc<RefCell<Engine>>) -> StratisResult<()> {
                         tree.insert(path);
                     }
                     DeferredAction::Remove(path) => {
-                        tree.remove(path.get_name());
-                        c.unregister_object_path(path.get_name());
+                        let object_path: dbus::Path<'static> = dbus::Path::new(path.clone())
+                            .unwrap();
```
mulkieran: unwrap

src/engine.rs
```diff
@@ -44,15 +44,19 @@ pub type EngineResult<T> = Result<T, EngineError>;
 pub trait Dev: Debug {
     fn copy(&self) -> Box<Dev>;
     fn get_id(&self) -> String;
+    fn eq(&self, other: &Path) -> bool;
```
mulkieran: This is not the usual definition of eq. Something should give, either the name or the type of other.

src/engine.rs
```diff
@@ -44,15 +44,19 @@ pub type EngineResult<T> = Result<T, EngineError>;
...
 }
 
 pub trait Cache: Debug {
     fn copy(&self) -> Box<Cache>;
     fn get_id(&self) -> String;
+    fn eq(&self, other: &Path) -> bool;
```
mulkieran: Not the meaning of eq.

src/engine.rs
```diff
@@ -44,15 +44,19 @@ pub type EngineResult<T> = Result<T, EngineError>;
...
 }
 
 pub trait Filesystem: Debug {
     fn copy(&self) -> Box<Filesystem>;
+    fn get_id(&self) -> String;
+    fn eq(&self, other: &Filesystem) -> bool;
```
mulkieran: This _is_ eq.

src/engine.rs
```diff
@@ -63,6 +67,8 @@ pub trait Pool: Debug {
                          -> EngineResult<()>;
     fn add_blockdev(&mut self, path: &Path) -> EngineResult<()>;
     fn add_cachedev(&mut self, path: &Path) -> EngineResult<()>;
+    fn remove_blockdev(&mut self, path: &Path) -> EngineResult<()>;
```
mulkieran: I'm going to take these Path arguments as being a temporary hack, like the fact that get_id() is based on the Path.

I think that we will have to figure out something far better to pass to the engine methods, but I don't know what that is.

src/sim_engine/blockdev.rs
```diff
@@ -53,6 +60,10 @@ impl Dev for SimDev {
             None => return String::from("Conversion Failure"),
         }
     }
+
+    fn eq(&self, other: &Path) -> bool {
```
mulkieran: As above, this is not any definition of equality. It includes an ```unwrap()``` which is bad, and conflicts with Rust's own definition of equality. I think it might be possible to have SimDev implement PartialEq and Eq correctly and go from there. Dev cannot implement PartialEq, because it cannot be Sized.

I already tried to have

src/sim_engine/cache.rs
```diff
@@ -75,4 +75,7 @@ impl Cache for SimCacheDev {
             None => return String::from("Conversion Failure"),
         }
     }
+    fn eq(&self, other: &Path) -> bool {
```
mulkieran: Same as above.

src/sim_engine/filesystem.rs
```diff
@@ -25,4 +25,11 @@ impl Filesystem for SimFilesystem {
         };
         Box::new(filesystem_copy)
     }
+    fn get_id(&self) -> String {
+        self.mount_point.clone()
+    }
+
+    fn eq(&self, other: &Filesystem) -> bool {
+        self.get_id() == other.get_id()
```
mulkieran: This one works, because it is really equality. The fact that you can use ```==``` must be because String implements PartialEq. Is it possible to skip the cloning and do ```self.mountpoint == other.mountpoint```? I think it could work.

src/sim_engine/pool.rs
```diff
@@ -94,4 +96,34 @@ impl Pool for SimPool {
...
+    fn remove_blockdev(&mut self, path: &Path) -> EngineResult<()> {
+        let index = self.block_devs.iter().position(|x| x.eq(path));
+        match index {
+            Some(index) => {
+                self.block_devs.remove(index);
+                return Ok(());
+            }
+            None => {
+                return Err(EngineError::Stratis(ErrorEnum::NotFound(String::from(path.to_str()
+                    .unwrap()))))
```
mulkieran: unwrap.

src/sim_engine/pool.rs
```diff
@@ -94,4 +96,34 @@ impl Pool for SimPool {
...
+        let index = self.cache_devs.iter().position(|x| x.eq(path));
+
+        match index {
+            Some(index) => {
+                self.cache_devs.remove(index);
+                return Ok(());
+            }
+            None => {
+                return Err(EngineError::Stratis(ErrorEnum::NotFound(String::from(path.to_str()
+                    .unwrap()))))
```
mulkieran: unwrap

src/engine.rs
trgill: Ok, I'll change the name.  Any suggestions?  How about is_same()?

src/sim_engine/filesystem.rs
trgill: I did this because we agree that mountpoint is a bad ID.  So, when we add a good one, we just update get_id().  Not sure I see why comparing mountpoing is better here.

mulkieran: When you clone a String you make a copy. This one is about efficiency...it seems wrong to make two exact copies and compare them when you can just compare the two things directly.

src/engine.rs
mulkieran: ```is_same``` is probably overstating the case in the presence of multipath. Maybe ```has_same```?

mulkieran:
I just noticed another thing...there are a couple of places in this PR and in existing code where the type of the incoming message is converted to ```Array<(&str), _>```, but it should be ```Array<&str, _>```.

I guess the tuple is automatically converted to a &str by Path::new(), but it would still be better to fix.

src/dbus_api.rs
mulkieran: I misread what was happening there. It looks to me like you don't need to do the conversion to Path and construction of the ```to_remove_devs``` iterator at line 508 at all. I think you could just iterate over ```devs.iter()``` or something at line 510.

trgill: Line 511 takes a Path as a parameter, right?  So we do need the conversion.

mulkieran: Something like:

```
for dev in devs {
    let result = pool.remove_cachedev(Path::new(dev));

    ...
    ... cache_devs_map.get_by_second(&String::from(dev)) ...
```

seems better to me if possible because

* it gets rid of that suspicious looking unwrap() (I agree it ought to be safe)
* it means one conversion instead of two
* it isolates converting to Path to one place instead of making it part of the overall structure of the code
  (I like this, because I think the signature of ```remove_cachedev``` is definitely wrong)

trgill: Ok

src/engine.rs
trgill: sounds good

src/sim_engine/filesystem.rs
trgill: The other parameter in this case is a trait - Filesystem - it doesn't have a mountpoint like SimFilesystem.  I could add a get_mountpoint(), but it would have to do the same thing.

trgill: Ok, I just pushed the updates that I could address.  Good to merge with you?

PR #124, opened 2016-11-03
Write a method to get an argument off the dbus.

mulkieran:
Use it where appropriate.


mulkieran:
@trgill review?

This simply shortens getting messages off the dbus to a single line for each message, which I think is about as good as it can get.

trgill:
:+1:

PR #126, opened 2016-11-04
Make simulator configurable over d-bus

mulkieran:


mulkieran:
```configure_simulator``` cna only theoretically return an error, so I avoided adding a new error that would never be used.

mulkieran:
This changes the default to completely deterministic behavior. @trgill review?

trgill:
:+1:

PR #127, opened 2016-11-04
Use Entry interface to avoid two lookups.

mulkieran:


mulkieran:
This never looked like a benefit, and there is an existing record of having used the Entry interface when it actually helped, so this can be closed.

PR #133, opened 2016-11-08
Unbox the values of the sim engine's pool.

mulkieran:
Adjust other methods accordingly.

### DIFF
--- a/src/dbus_api.rs
+++ b/src/dbus_api.rs
@@ -233,13 +233,11 @@ fn list_pools(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
     let dbus_context = m.path.get_data();
     let ref engine = dbus_context.engine;
 
-    let result = engine.borrow().list_pools();
-
     let return_message = m.msg.method_return();
     let return_sig = "s";
 
-    let msg = match result {
-        Ok(pool_tree) => {
+    let msg = match engine.borrow().list_pools() {
+        Ok(ref pool_tree) => {
             let msg_vec =
                 pool_tree.keys().map(|key| MessageItem::Str(format!("{}", key))).collect();
             let item_array = MessageItem::Array(msg_vec, return_sig.into());
--- a/src/dbus_api.rs
+++ b/src/dbus_api.rs
@@ -236,7 +236,7 @@ fn list_pools(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
     let return_message = m.msg.method_return();
     let return_sig = "s";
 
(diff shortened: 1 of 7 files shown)


mulkieran:
The ```list_``` and ```copy()``` methods are still a puzzle.

agrover:
I think `copy` may be eligible for refactoring. Thoughts on this, as perhaps replacing `get_pool` and `list_pools` and letting us remove `copy`?

```
fn pools(&mut self) -> EngineResult<BTreeMap<String, &mut Pool>> {
        Ok(self.pools
            .iter_mut()
            .map(|x| (x.0.clone(), x.1 as &mut Pool))
            .collect())
    }
```

mulkieran:
@agrover I think you're suggesting ```list_*s``` methods can be renamed to ```*s``` methods. Then ```get_*``` methods can be implemented in terms of the corresponding ```*s``` method. I'll give it a try.

mulkieran:
@agrover @trgill review?

The label is still on there because I want to try to exercise these changes by expanding the sim engine a bit. OTOH, I'ld like feedback if anything seems really wrong.

I'm dubious about commit "Make the various list functions...", because I'm not sure that it would make sense to be able to mutate each object, and obviously it will make the method more difficult to use.

@agrover I don't know if you were actually suggesting this, but for efficiency reasons I'ld rather not get all the pools just to get the one pools I'm interested in. So, I left ```get_pool()``` alone, just changed its name to ```pool()```.

agrover:
Great! Although I'd prefer `get_pool` kept its name, I think it better connotes there's a (possibly-failing) lookup happening. Also, can `*s()` methods fail? Do they still need to return EngineError?

Also great to see `copy` is no longer needed. I'm very positive about this PR.

mulkieran:
@agrover for the sim engine the ```*s``` shoul dnever fail. I'm unsure abou tthe real engine, though.

agrover:
They could return an empty collection but I don't think they could return an error. I think we are safe in assuming each engine is traversing its in-memory data structures to generate the returned items (should never fail) rather than doing I/O to build the list.

mulkieran:
@agrover OK. That may be optimistic, but if it turns out wrong we can reverst it easily. I didn't try to make the dbus methods perfect, as my understanding is that they will go away anyway, wiht the ehlpf og GetMangeedObjects.

agrover:
yup. :+1:

mulkieran:
My hopes of finding anything wrong with this before noon are receding. @trgill Feel free to merge whenever you're ready.

PR #134, opened 2016-11-09
Minor fixes to sim engine behaviors

mulkieran:


mulkieran:
@agrover @trgill review? These are just minor fixes, each makes one more test pass.

src/sim_engine/engine.rs
```diff
@@ -71,12 +72,19 @@ impl Engine for SimEngine {
     }
 
     fn destroy_pool(&mut self, name: &str) -> EngineResult<()> {
+        let entry = match self.pools.entry(name.into()) {
+            Entry::Vacant(_) => return Ok(()),
```
agrover: Trying to destroy a pool that doesn't exist should return an error.

agrover: Any benefits in this particular case of using `entry()` over `get()`?

mulkieran:
@agrover I don't think the method should return an error when pool not found.

One way to think of this method is to think that it has to do some work, namely some sort of removal. If it is not vouchsafed the opportunity to do this work, execute certain instructions, it should fail. I believe that that view causes all sorts of problems.

The second point of view is that the method needs to achieve something and that it should guarantee that it can do it so long as certain conditions are satisfied. This is the notion of pre- and post-conditions, which has a lot of other names like "Design by Contract". Here the post-condition of the ```destroy_pool``` method is that the pool should be destroyed. If the pool is not even there, the post-condition is immediately achieved, so the method returns Ok(()). I believe that this is a more helpful view.

What sort of problems does the first view cause? Problems with reasoning and failures of code. An obvious problem is if for some reason two requests to destroy the pool come in from two entities, for each of which the pool exists at the time the instructions was issued.. The first one will succeed, but with your recommended construction the second will fail, and all this just depends on who gets in first. With mine, both will get what they wanted without any extra noise. This is good. With your suggested approach, the client code has to look out for ```POOL_NOTFOUND``` error, but has no means of interpreting its meaning. It could be that it got in there late and some other instruction had already removed the pool, it could be there is a bug in the stratis code (the most likely first suspicion), it could be a mis-spelling in the name.

I'm not saying that it is wrong to write code that returns an error if the pool turns out not to exist when the command is sent. The CLI can do that somehow, if desired. But the ```destroy_pool``` method is the building block on which all sorts of other code can and will be built. So, it must have the simplest and clearest interface that it can have. I think that the pre-condition/post-condition approach is the best available for achieving this goal.

src/sim_engine/engine.rs
mulkieran: I tried both. Considerations of lifetimes seemed to make the ```get()``` approach just as hard to read. So I went with the entry approach because it was more efficient.

agrover:
Ok thanks. DBC seems like a big topic that deserves further discussion, but we can merge this PR in the meantime :+1:

mulkieran:
Triggering a fully new build by closing and reopening.

trgill:
:+1:

PR #141, opened 2016-11-11
Replace BlockDev::new() with BlockDev::initialize()

agrover:
If we are creating a pool from multiple blockdevs and there's something
wrong with one of the later ones, we will abort but already have written
Stratis headers to the earlier ones. This new function takes multiple
paths to create blockdevs on at once. This allows it to check all of them
first for min size, is it a blockdev, etc before overwriting the initial
data on any of them.

Make BlockDev::new() use batch_new() internally.

### DIFF
--- a/src/strat_engine/blockdev.rs
+++ b/src/strat_engine/blockdev.rs
@@ -38,7 +38,7 @@ pub struct MDA {
 
 #[derive(Debug, Clone, PartialEq)]
 pub struct BlockDev {
-    pub stratis_id: String,
+    pub pool_id: String,
     pub dev: Device,
     pub id: String,
     pub path: PathBuf,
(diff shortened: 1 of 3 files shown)


agrover:
review @trgill and/or @mulkieran ?

src/strat_engine/pool.rs
```diff
@@ -56,7 +56,7 @@ impl Pool for StratPool {
     }
 
     fn add_blockdev(&mut self, path: &Path) -> EngineResult<()> {
-        let bd = try!(BlockDev::new(&self.uuid, Path::new(path), MIN_MDA_SIZE, true));
+        let bd = try!(BlockDev::new(&self.uuid, path, MIN_MDA_SIZE, true));
```
mulkieran: The near future plan is to change this method to ```add_blockdevs``` (plural). In that case, ```BlockDev::new``` would be the wrong method to use. The correct method to use would be ```batch_new```. In that case, there would be no place in the code that actually used ```BlockDev::new``` and therefore it would be nice if ```BlockDev::new``` would go away.

src/strat_engine/blockdev.rs
```diff
@@ -55,87 +55,109 @@ impl BlockDev {
...
+    }
 
-        if !metadata.file_type().is_block_device() {
-            return Err(EngineError::Io(io::Error::new(ErrorKind::InvalidInput,
-                                                      format!("{} is not a block device",
-                                                              path.display()))));
-        }
+    /// Initialize multiple blockdevs at once. This allows all of them
+    /// to be checked for usability before writing to any of them.
+    pub fn batch_new(stratis_id: &str,
```
mulkieran: It would be helpful to add a header comment about the parameters or even rename ```stratis_id``` to ```pool_id``` since, AFAICT, that is what it is. Please do.

src/strat_engine/blockdev.rs
```diff
@@ -55,87 +55,109 @@ impl BlockDev {
...
+                                                      format!("mda size {} is less than \
                                                                minimum ({})",
-                                                              path.display(),
+                                                              *mda_size,
                                                               *MIN_MDA_SIZE))));
         }
 
-        let dev = try!(Device::from_str(&path.to_string_lossy()));
+        // Check stuff for all candidates before writing anything
+        let mut good_paths = Vec::new();
```
mulkieran: It looks like ```good_paths``` is redundant. If you encounter an error for any path, you exit. If you get down to the next block, then all the paths you were passed were good path, so you can just reuse paths. It makes it look like you're filtering, and only using the good paths, but if I read more carefully it seems like that is not what you're doing. Please see if you can remove the use of ```good_paths``` entirely.

src/strat_engine/blockdev.rs
```diff
@@ -55,87 +55,109 @@ impl BlockDev {
...
+        for path in paths {
+            let metadata = try!(fs::metadata(path));
```
mulkieran: It would be nice if the chunk from line 87 to line 126 were a single function that returned an EngineResult. Then use ```find()``` to find the first path that is an error and return that error if so, otherwise proceed. This will make everything easier when we have to abstract dealing with paths that are duplicates or not paths at all and all those other issues. Please see if you can abstract the checking into a single function.

src/strat_engine/blockdev.rs
```diff
@@ -55,87 +55,109 @@ impl BlockDev {
...
 
-        let dev_size = try!(blkdev_size(&f));
-        if dev_size < MIN_DEV_SIZE {
-            return Err(EngineError::Io(io::Error::new(ErrorKind::InvalidInput,
-                                                      format!("{} too small, {} minimum",
-                                                             path.display(),
-                                                             ByteSize::b(MIN_DEV_SIZE as usize)
-                                                                 .to_string(true)))));
+        let mut bds = Vec::new();
+        for (path, dev, dev_size) in good_paths {
```
mulkieran: As pointed out, I think ```paths``` or ```paths.iter()``` is all that is necessary here, ```good_paths``` should be able to be removed entirely.

src/strat_engine/blockdev.rs
```diff
@@ -55,87 +55,109 @@ impl BlockDev {
                mda_size: Sectors,
                force: bool)
                -> EngineResult<BlockDev> {
-        let metadata = try!(fs::metadata(path));
+        Ok(try!(Self::batch_new(stratis_id, &[path], mda_size, force)).pop().unwrap())
+    }
```
mulkieran: As I pointed out below...it would be great if this function above could be eliminated completely once the engine refactoring is completed.

src/strat_engine/blockdev.rs
```diff
@@ -55,87 +55,109 @@ impl BlockDev {
...
+            let mut f = try!(OpenOptions::new()
+                .read(true)
+                .write(true)
+                .open(path)
+                .map_err(|_| {
+                    io::Error::new(ErrorKind::PermissionDenied,
+                                   format!("Could not open {}", path.display()))
+                }));
+
+            if !force {
```
mulkieran: I reopened the issue about this. I'm pretty confident that we agreed not to force.

src/strat_engine/blockdev.rs
mulkieran: ```new``` and ```batch_new``` don't seem to describe what the method does. Something like ```appropriate``` or ```occupy``` would be better. Please consider a name change.

src/strat_engine/blockdev.rs
agrover: The goal was really to hold on to dev and dev_size values that we needed to get (to check if they're valid) but also we need to initialize fields in BlockDev, once all tests are passed. Paths doesn't need to be in there -- we can just zip the `vec<(dev, dev_size)>` with the `vec<paths>` in the following loop, and call the former "path_info" perhaps?

src/strat_engine/blockdev.rs
agrover: great idea, will do!

src/strat_engine/blockdev.rs
```diff
@@ -50,92 +50,112 @@ pub struct BlockDev {
...
-                last_updated: Timespec::new(0, 0),
-                used: 0,
-                length: (*mda_size / 2 * SECTOR_SIZE) as u32,
-                crc: 0,
-                offset: SectorOffset(*mda_size / 2),
-            },
-            mda_sectors: mda_size,
-            reserved_sectors: MDA_RESERVED_SIZE,
-        };
+            let dev = try!(Device::from_str(&path.to_string_lossy()));
```
mulkieran: Please try to move ```dev``` down to its first use, at around line 155. Or, even better, don't calculate it in this method, calculate it in the chunk of code starting at180.

agrover: `Device::from_str` actually checks that path points to a valid block device. (I'll remove the redundant check immediately above in this patch.) I think we want to verify this before trying to open it, before reading the first 4k (if !force), and definitely before calling `blkdev_size`.

Line 180, that's down in `setup()`? That's a separate code path for when we're given a device that is already initialized, and we're gathering the info to construct the corresponding `BlockDev`. We need to obtain a `Device` in either case, to create the `BlockDev`.

agrover:
any remaining issues @trgill @mulkieran ?

trgill:
:+1:   Looks good... 

No need to change now, but while looking at it, I was wondering if it would be better to store the ID as a Uuid rather than covert the Uuid to a string.

agrover:
agree, will do that in the future.

mulkieran:
:+1:  ok, I'm closing and reopening to trigger CI, just in case.

mulkieran:
Closing...

mulkieran:
reopening...

PR #137, opened 2016-11-10
Implement rename filesystem

trgill:
Added code for dbus and sim_engine.
Implemented rename filesystem for sim_engine
Update uuid to version 0.3.1 and enable v4 features (minor changes)
Make uuid key for filesystems table in the Pool
The uuid version change required minor updates to strat_engine

Signed-off-by: Todd Gill <tgill@redhat.com>

PR #143, opened 2016-11-15
Add some macros for when there is no return value.

mulkieran:


mulkieran:
These are good for when you want to use the original macros, but can't because there is no default return value. ```rename_filesystem``` is a good example.

mulkieran:
@trgill @agrover review?

agrover:
suggest updating this PR to use the new macros once #137 is merged.

trgill:
:+1:  I'll merge this and then update my PR to use the macros.

PR #144, opened 2016-11-16
Blockdev improvements

agrover:
* Use `uuid::Uuid` for blockdev uuids instead of `String`
* Add `BlockDev::avail_areas()` so higher layers can get the usable portion of a blockdev
* Try to ensure blockdevs are unused by this or other pools, and are unique
* Plumb `force` all the way up through the DBus API layer

### DIFF
--- a/src/dbus_api.rs
+++ b/src/dbus_api.rs
@@ -917,11 +917,12 @@ fn create_pool(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
     let name: &str = try!(get_next_arg(&mut iter, 0));
     let raid_level: u16 = try!(get_next_arg(&mut iter, 1));
     let devs: Array<&str, _> = try!(get_next_arg(&mut iter, 2));
+    let force: bool = try!(get_next_arg(&mut iter, 3));
 
     let blockdevs = devs.map(|x| Path::new(x)).collect::<Vec<&Path>>();
 
     let dbus_context = m.path.get_data();
-    let result = dbus_context.engine.borrow_mut().create_pool(name, &blockdevs, raid_level, true);
+    let result = dbus_context.engine.borrow_mut().create_pool(name, &blockdevs, raid_level, force);
 
     let return_message = message.method_return();
 
@@ -1084,6 +1085,7 @@ fn get_base_tree<'a>(dbus_context: DbusContext) -> StratisResult<Tree<MTFn<TData
         .in_arg(("pool_name", "s"))
         .in_arg(("raid_type", "q"))
         .in_arg(("dev_list", "as"))
+        .in_arg(("force", "b"))
         .out_arg(("object_path", "o"))
         .out_arg(("return_code", "q"))
         .out_arg(("return_string", "s"));
--- a/src/dbus_api.rs
+++ b/src/dbus_api.rs
@@ -921,7 +921,7 @@ fn create_pool(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
     let blockdevs = devs.map(|x| Path::new(x)).collect::<Vec<&Path>>();
 
     let dbus_context = m.path.get_data();
-    let result = dbus_context.engine.borrow_mut().create_pool(name, &blockdevs, raid_level);
+    let result = dbus_context.engine.borrow_mut().create_pool(name, &blockdevs, raid_level, true);
 
     let return_message = message.method_return();
 
--- a/src/dbus_api.rs
+++ b/src/dbus_api.rs
@@ -679,7 +679,7 @@ fn add_devs(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
     let mut vec = Vec::new();
 
     for dev in devs {
-        let result = pool.add_blockdev(Path::new(dev));
+        let result = pool.add_blockdev(Path::new(dev), true);
         match result {
             Ok(_) => {
                 let object_path: dbus::Path = create_dbus_blockdev(dbus_context.clone());
--- a/src/dbus_api.rs
+++ b/src/dbus_api.rs
@@ -659,6 +659,7 @@ fn add_devs(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
     let mut iter = message.iter_init();
 
     let devs: Array<&str, _> = try!(get_next_arg(&mut iter, 0));
+    let force: bool = try!(get_next_arg(&mut iter, 1));
 
     let dbus_context = m.path.get_data();
     let object_path = m.path.get_name();
@@ -679,7 +680,7 @@ fn add_devs(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
     let mut vec = Vec::new();
 
     for dev in devs {
-        let result = pool.add_blockdev(Path::new(dev), true);
+        let result = pool.add_blockdev(Path::new(dev), force);
         match result {
             Ok(_) => {
                 let object_path: dbus::Path = create_dbus_blockdev(dbus_context.clone());
@@ -878,6 +879,7 @@ fn create_dbus_pool<'a>(mut dbus_context: DbusContext) -> dbus::Path<'a> {
 
     let add_devs_method = f.method(ADD_DEVS, (), add_devs)
         .in_arg(("devs", "as"))
+        .in_arg(("force", "b"))
         .out_arg(("results", "a(oqs)"))
         .out_arg(("return_code", "q"))
         .out_arg(("return_string", "s"));
--- a/src/strat_engine/blockdev.rs
+++ b/src/strat_engine/blockdev.rs
@@ -38,9 +38,7 @@ pub struct MDA {
 
 #[derive(Debug, Clone, PartialEq)]
 pub struct BlockDev {
-    pub pool_uuid: Uuid,
     pub dev: Device,
-    pub dev_uuid: Uuid,
     pub path: PathBuf,
     pub sectors: Sectors,
     pub mdaa: MDA,
@@ -79,8 +77,6 @@ impl BlockDev {
         for (dev, (path, dev_size)) in devices.into_iter().zip(dev_info.into_iter()) {
 
             let mut bd = BlockDev {
-                pool_uuid: pool_uuid.to_owned(),
-                dev_uuid: Uuid::new_v4(),
                 dev: dev,
                 path: path,
                 sectors: Sectors(dev_size / SECTOR_SIZE),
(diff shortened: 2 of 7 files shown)


agrover:
review @mulkieran @trgill ?

src/strat_engine/blockdev.rs
```diff
@@ -320,8 +322,8 @@ impl BlockDev {
         buf[4..20].clone_from_slice(STRAT_MAGIC);
         LittleEndian::write_u64(&mut buf[20..28], *self.sectors);
         // no flags yet
-        buf[32..64].clone_from_slice(self.pool_id.as_bytes());
-        buf[64..96].clone_from_slice(self.id.as_bytes());
+        buf[32..64].clone_from_slice(self.pool_uuid.as_bytes());
+        buf[64..96].clone_from_slice(self.dev_uuid.as_bytes());
```
agrover: hmm. oops. Nope no expansion, we actually want 32 ascii hex bytes. Lemme fix that...

agrover: instead of
```
buf[32..64].clone_from_slice(self.pool_uuid.as_bytes());
```
we want
```
buf[32..64].clone_from_slice(self.pool_uuid.simple().to_string().as_bytes());
```
Fixing and rebasing... (nice catch!)

src/strat_engine/blockdev.rs
```diff
@@ -51,10 +53,10 @@ impl BlockDev {
     /// Initialize multiple blockdevs at once. This allows all of them
     /// to be checked for usability before writing to any of them.
     pub fn initialize(pool_uuid: &Uuid,
-                      paths: &[&Path],
+                      devices: BTreeSet<Device>,
```
mulkieran: This is a step in the right direction, glad to see it.

src/strat_engine/blockdev.rs
```diff
@@ -101,22 +103,24 @@ impl BlockDev {
...
-            bds.push(bd);
+            bds.insert(bd.dev_uuid.clone(), bd);
         }
         Ok(bds)
     }
 
     /// Gets device and device sizes, and returns an error if devices
     /// cannot be used by Stratis.
-    fn dev_info(paths: &[&Path], force: bool) -> EngineResult<Vec<(Device, u64)>> {
+    fn dev_info(devices: &BTreeSet<Device>, force: bool) -> EngineResult<Vec<(PathBuf, u64)>> {
```
mulkieran: This function and its use would be easier to work with as a simple function without an internal loop, and a map and a fold at the call site, I think. This is a suggestion, not a request.

src/strat_engine/engine.rs
```diff
@@ -45,9 +47,30 @@ impl Engine for StratEngine {
...
-        let pool_uuid = Uuid::new_v4();
-        let bds = try!(BlockDev::initialize(&pool_uuid, blockdev_paths, MIN_MDA_SIZE, true));
-        let pool = StratPool::new(name, pool_uuid, &bds, raid_level);
+        let mut devices = BTreeSet::new();
+        for path in blockdev_paths {
+            let dev = try!(Device::from_str(&path.to_string_lossy()));
+            devices.insert(dev);
+        }
+
+        if devices.len() != blockdev_paths.len() {
```
mulkieran: I'm in favor of this not being an error.

src/strat_engine/pool.rs
```diff
@@ -56,10 +65,24 @@ impl Pool for StratPool {
     }
 
     fn add_blockdev(&mut self, path: &Path) -> EngineResult<()> {
-        let bd = try!(BlockDev::initialize(&self.uuid, &[path], MIN_MDA_SIZE, true))
-            .pop()
+        let dev = try!(Device::from_str(&path.to_string_lossy()));
```
mulkieran: Please file an issue about the problem alluded to in the commit message. It seems bad that pools can steal devs from each other.

src/strat_engine/pool.rs
```diff
@@ -56,10 +69,24 @@ impl Pool for StratPool {
...
```
mulkieran: I think ```add_blockdev``` needs a force parameter as well.

agrover:
The only rationale is that we need pool and blockdev uuids when writing out metadata. Well, can't we pass those in to `save_state()`? I think so, let me try that :grin: It seems a little funny that a Blockdev won't actually know its own unique name, but it does get rid of the redundancy, which of course is a potential source of bugs.

src/dbus_api.rs
```diff
@@ -917,11 +917,12 @@ fn create_pool(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
     let name: &str = try!(get_next_arg(&mut iter, 0));
     let raid_level: u16 = try!(get_next_arg(&mut iter, 1));
     let devs: Array<&str, _> = try!(get_next_arg(&mut iter, 2));
+    let force: bool = try!(get_next_arg(&mut iter, 3));
```
mulkieran: One last thought: You will also need to change the ```in_arg``` value for this method. That's the value that controls the information returned by the Introspect method.

src/strat_engine/engine.rs
agrover: I'm afraid I must disagree. If the user asks for a pool comprised of three blockdevs, we either need to give them a pool of three blockdevs, or fail.

src/dbus_api.rs
agrover: ah!! yes!!

src/strat_engine/blockdev.rs
```diff
@@ -381,4 +381,33 @@ impl BlockDev {
     pub fn dstr(&self) -> String {
         format!("{}:{}", self.dev.major, self.dev.minor)
     }
+
+    /// List all used areas in the blockdev.
+    fn used_areas(&self) -> Vec<(SectorOffset, Sectors)> {
```
mulkieran: Maybe ```reserved_areas``` or ```stratis_areas``` is better.

src/strat_engine/blockdev.rs
```diff
@@ -381,4 +381,33 @@ impl BlockDev {
...
+        // Flag end mda zone as used
+        used.push((SectorOffset(*self.sectors - *BDA_STATIC_HDR_SIZE - *self.mda_sectors),
+                   BDA_STATIC_HDR_SIZE + self.mda_sectors));
+        used
+    }
+
+    /// List all unused, available extents in this blockdev.
+    pub fn avail_areas(&self) -> Vec<(SectorOffset, Sectors)> {
+        let used = self.used_areas();
+        assert_eq!(used.len(), 2);
```
mulkieran: Is this correct? The code below seems extremely ornate if the number of these "reserved" areas is certainly just two.

src/strat_engine/blockdev.rs
```diff
@@ -8,6 +8,8 @@ use std::path::{Path, PathBuf};
 use std::io;
 use std::str::{FromStr, from_utf8};
 use std::cmp::Ordering;
+use std::collections::BTreeMap;
```
mulkieran: ok, thanks.

src/strat_engine/blockdev.rs
mulkieran: I mean the function and its one use could probably be split up as a map and a fold and a simple ```dev_info``` function without a loop in it, probably.

src/strat_engine/engine.rs
mulkieran: If I write out a grocery list that says:

* milk
* peanuts
* prunes
* pretzels
* prunes

I would rather that the person with the list got me what I'ld asked for than that they returned empty handed because I had listed prunes twice.

src/strat_engine/blockdev.rs
agrover: Hmm. Well it really depends on where we keep track of allocations in the blockdev. In Froyo, the Blockdev tracked this itself, but maybe Stratis tracks this in the Flex layer, and so all Blockdev has to report is its reserved areas.

Ok, since [YAGNI](https://en.wikipedia.org/wiki/You_aren't_gonna_need_it) let me just have avail_area report the offset/length of its usable area (reduced by metadata areas) and we can wait and see how we end up tracking blockdev allocations.

agrover:
`try!()` and `map()` don't get along, and I think that will bite us here with what you're suggesting for `dev_info()`. I'd love to be wrong, you want to try to get it working?

agrover:
```
touch foo
mkdir tmp
mv foo bar tmp
mv: cannot stat 'bar': No such file or directory
```
I think we need an all-or-nothing approach to our API calls.

mulkieran:
Yes, we need an all or nothing approach. But that's not at issue here. We would give them what they requested, because we are batching it up.

The venerable unix mv's default behavior on my system does not constitute support for either of our viewpoints, it does the move and complains:

```
[mulhern@dhcp-25-209 temp]$ touch foo bar baz
[mulhern@dhcp-25-209 temp]$ mv foo bar foo baz dir
mv: cannot stat 'foo': No such file or directory
[mulhern@dhcp-25-209 temp]$ echo $?
1
[mulhern@dhcp-25-209 temp]$ ls
dir
[mulhern@dhcp-25-209 temp]$ mv x dir
mv: cannot stat 'x': No such file or directory
[mulhern@dhcp-25-209 temp]$ echo $?
1
```
So, mv takes things one at a time, moves what it can, and complains about the things it can't. Bad mv.
You want to complain and not do the move. I want to do the move and not complain. Probably either approach is slightly better than mv's.

I think the problem is that that little loop checks two things at once.
1. Are there duplicate devices?
2. Do some of the devices not actually exist at all?

If some of the requested devices do not exist at all, an error is the correct response. But if some devices were asked for twice, continuing is the correct response.

But, since the loop exits if the device does not actually exist, what we get at the end if we do not return with an error in the middle of the loop is exactly the devices the user asked for, even if they asked for them in two different ways.

Why not change the return type of ```create_pool```? It could contain anything, but right now a sufficiently informative type would be an unsigned int which indicates how many devices were added to the pool when the pool was created. That's a well established technique for indicating helpful information like how many bytes were actually read from the file, or whatnot. Then, the client can check the value if it want to know whether the number of devices in the pools is fewer than the number of strings that were passed at the higher level.

mulkieran:
I'm happy to try the fold of the map approach, but it would be more convenient to work on it from the stratis-storage repo then from your personal repo, so I'm happy to wait until this PR is completed.

agrover:
Ok, let's try it! Returning `()` always seems like a waste.

mulkieran:
:+1: Suffering from review exhaustion. Feel free to merge. Please remember to file issue about BTree* vs. Hash* data structures.

agrover:
Thanks, merging.

PR #156, opened 2016-11-29
Pluralize add_blockdevs and add_cachedevs

mulkieran:


mulkieran:
@agrover @trgill review?

This is preliminary. The ultimate goal is to pluralize a bunch of the functions in engine, pool, etc. I used the function that is currently already pluralized, ```create_pool``` as it takes multiple devices, as a model, and found a bunch of things I thought I had better change before I could really use it as a model.

The significant changes are:

* No longer have object paths for devices that Stratis uses. We do not have any good names for these devices, and therefore, mapping between the object path and some name is going to be a huge problem. Therefore, if possible, we should avoid it.
* Change ```Engine::create_pool``` API so that it returns a list of the device nodes of the devices used. This is a bit richer than just the number of devices added/
* No longer check the engine structures to see if the device already belongs to another pool. Instead, when getting the device info, look for the Stratis magic number and the pool uuid.
* Some refactoring inside BlockDev::initialize.

And finally, an initial attempt to pluralize ```add_blockdevs```.

agrover:
Hi some initial comments. Looking very positive!
* don't need to pass Device to dev_info() by ref since it's `Copy`. (A nit, but just mentioning in case it makes life easier somewhere)
* mapping over dev_info() doesn't let us return err as soon as we see one, like before. How big of a deal do you think that is?

agrover:
oh, `dev_infos` is not `collect()`ed, it's still just an iterator! Never mind 😃

trgill:
If we eliminate object paths for blockdevs, how will we publish alerts for a device failure?  We do it at the pool level and include information in the alert that allows the user to identify the problem blockdev?

mulkieran:
@trgill Yes, we will have to talk about the devices in the context of the stratisd instance/pool that they belong to.

src/sim_engine/pool.rs
```diff
@@ -60,14 +61,12 @@ impl SimPool {
 }
 
 impl Pool for SimPool {
-    fn add_blockdev(&mut self, path: &Path, _force: bool) -> EngineResult<()> {
-        self.block_devs.push(SimDev::new_dev(self.rdm.clone(), path));
-        Ok(())
+    fn add_blockdevs(&mut self, _paths: &[&Path], _force: bool) -> EngineResult<Vec<PathBuf>> {
+        Ok(vec![])
```
trgill: This is just a place holder?

mulkieran: Yes. I should have mentioned that I need to rework the sim engine a little bit.

src/dbus_api.rs
```diff
@@ -701,47 +687,35 @@ fn add_devs(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
     let dbus_context = m.path.get_data();
     let object_path = m.path.get_name();
     let return_message = message.method_return();
-    let return_sig = "(oqs)";
+    let return_sig = "s";
```
mulkieran: This is a tricky one. We pass a list of paths as argument, although, as I've commented elsewhere, I think it should be a list of strings, so we can, for example, use the vpd83 information if it is available, instead of a path. And then, the engine figures out what those paths actually correspond to, and add some set of devices to the pool. There is no longer a 1-to-1 mapping between the values passed in and the actual devices used, as there may be duplicates and all sorts of other issues.

The best alternative return I have been able to come up with is a pair of values.
* a dict from device nodes actually added to sets of paths passed
* a vec of paths that didn't contribute to any of the device nodes because, e.g., a path actually corresponded to a device that was already in the pool

This is possible, because initialize() could return the mapping that could then be passed up to the ```add_*``` methods. In that case, the return value would have the signature ```(a{sas}as)```.

From the point of view of our stratis client, this could be nice. In some sort of verbose mode, it would actually display a little table of devices added and why, as well as of devices ignored.

mulkieran:
@agrover @trgill review?

I believe that the only substantive changes since last time are the final two commits. ```SimPool::add_blockdevs()``` and ```add_cachedevs()``` are incomplete, but I think it's better to go ahead than to let the PR sit too much longer.

mulkieran:
@agrover @trgill final review?

trgill:
:+1:

agrover:
@mulkieran please resolve conflicts

mulkieran:
rebased, there was only one,trivial, conflict.

agrover:
I think `Bad` might be better characterized as `Invalid`, at least for the cases it is used in this PR.

mulkieran:
Changed ```Bad``` constructor to ```Invalid```.

mulkieran:
@agrover I think you're good with this. I'll merge tomorrow if you don't merge tonight.

PR #158, opened 2016-11-29
Initial version of create_snapshot

trgill:
Remove &mut from get_filesystem_id it is not necessary
Add vector of ancestors to the SimFilesystem this is a placeholder
Add get_filesystem_by_name to lookup filesystem by name for a Pool

### DIFF
--- a/src/sim_engine/filesystem.rs
+++ b/src/sim_engine/filesystem.rs
@@ -15,7 +15,9 @@ pub struct SimFilesystem {
     pub name: String,
     pub mount_point: String,
     pub quota_size: u64,
+    pub nearest_ancestor: Option<Uuid>,
 }
+
 impl SimFilesystem {
     pub fn new_filesystem(name: &str, mount_point: &str, quota_size: Option<u64>) -> SimFilesystem {
         SimFilesystem {
@@ -23,6 +25,7 @@ impl SimFilesystem {
             uuid: Uuid::new_v4(),
             mount_point: mount_point.to_owned(),
             quota_size: quota_size.unwrap_or(DEFAULT_FILESYSTEM_QUOTA_SIZE),
+            nearest_ancestor: None,
         }
     }
 }
@@ -46,5 +49,9 @@ impl Filesystem for SimFilesystem {
     fn rename(&mut self, new_name: &str) -> EngineResult<()> {
         self.name = String::from(new_name);
         Ok(())
+    }
+
+    fn add_ancestor(&mut self, parent: Uuid) {
+        self.nearest_ancestor = Some(parent);
     }
 }
--- a/src/sim_engine/pool.rs
+++ b/src/sim_engine/pool.rs
@@ -51,6 +51,12 @@ impl SimPool {
 
         new_pool
     }
+
+    // If the source doesn't exist, return an error - otherwise the UUID of the soruce
+    fn validate_snapshot_source_exists(&mut self, source: &str) -> EngineResult<Uuid> {
+        let filesystem = try!(self.get_filesystem_by_name(&source));
+        Ok(filesystem.get_id())
+    }
 }
 
 impl Pool for SimPool {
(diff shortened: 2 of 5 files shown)


trgill:

agrover:
In sim_engine modifications, can all the `match`s be `try!`s? They appear to be of the form
```
let bar = match foo() {
    Ok(x) =>x,
    Err(e) -> return Err(e),
}
```
which is just `try!()`.

src/sim_engine/pool.rs
```diff
@@ -51,6 +51,16 @@ impl SimPool {
 
         new_pool
     }
+
+    fn check_snapshot_source(&mut self, source: &str) -> EngineResult<Uuid> {
```
mulkieran: Is ```check_snapshot_source``` the best choice of name here? It could use a comment, as well.

src/sim_engine/filesystem.rs
```diff
@@ -47,4 +50,9 @@ impl Filesystem for SimFilesystem {
         self.name = String::from(new_name);
         Ok(())
     }
+
+    fn add_ancestor(&mut self, parent: Uuid) -> EngineResult<()> {
```
mulkieran: Since this is a private function, and ```push()``` can only panic, might as well have it return ```()``` rather than an ```EngineResult``` for now.

src/sim_engine/filesystem.rs
```diff
@@ -15,14 +15,17 @@ pub struct SimFilesystem {
...
 }
+
 impl SimFilesystem {
     pub fn new_filesystem(name: &str, mount_point: &str, quota_size: Option<u64>) -> SimFilesystem {
         SimFilesystem {
             name: name.to_owned(),
             uuid: Uuid::new_v4(),
             mount_point: mount_point.to_owned(),
             quota_size: quota_size.unwrap_or(DEFAULT_FILESYSTEM_QUOTA_SIZE),
+            ancestors: Vec::new(),
```
mulkieran: I don't know why ```ancestors``` is a ```Vec```. Couldn't it just as well be a single UUID called ```parent``` for the uses it is getting, maybe with type ```Option<Uuid>```?

trgill: Yeah, I wasn't sure if we might want to track all ancestors or just the immediate parent.  I'll change to Option<Uuid> for now.

src/sim_engine/pool.rs
trgill: how about validate_snapshot_source_exists?  I'll add a comment.

trgill:
@agrover @mulkieran Ok - I addressed the comments.

src/sim_engine/pool.rs
```diff
@@ -139,6 +162,14 @@ impl Pool for SimPool {
         Err(EngineError::Stratis(ErrorEnum::NotFound(String::from(name))))
     }
 
+    fn get_filesystem_by_name(&mut self, name: &str) -> EngineResult<&mut Filesystem> {
+        let id = match self.get_filesystem_id(name) {
```
mulkieran: This looks like an opportunity for ```try!```.

src/sim_engine/pool.rs
```diff
@@ -51,6 +51,16 @@ impl SimPool {
 
         new_pool
     }
+
+    fn check_snapshot_source(&mut self, source: &str) -> EngineResult<Uuid> {
+        let filesystem = match self.get_filesystem_by_name(&source) {
```
mulkieran: ```try!``` here?

src/sim_engine/pool.rs
```diff
@@ -104,6 +114,19 @@ impl Pool for SimPool {
         Ok(())
     }
 
+    fn create_snapshot(&mut self, snapshot_name: &str, source: &str) -> EngineResult<()> {
+
+        let parent_id = try!(self.check_snapshot_source(source));
+
+        try!(self.create_filesystem(&snapshot_name, &String::from(""), None));
```
mulkieran: What would happen if you changed the signature of ```create_filesystem``` so that it returned the just created filesystem object?

Maybe it would be helpful for ```create_snapshot``` to do something similar.

src/sim_engine/filesystem.rs
```diff
@@ -15,14 +15,17 @@ pub struct SimFilesystem {
...
 }
+
 impl SimFilesystem {
     pub fn new_filesystem(name: &str, mount_point: &str, quota_size: Option<u64>) -> SimFilesystem {
         SimFilesystem {
             name: name.to_owned(),
             uuid: Uuid::new_v4(),
             mount_point: mount_point.to_owned(),
             quota_size: quota_size.unwrap_or(DEFAULT_FILESYSTEM_QUOTA_SIZE),
+            ancestor: None,
```
mulkieran: This and other ancestors should be changed to ```parent``` or possibly ```nearest_ancestor``` since parent's can go away.

agrover:
👍

src/sim_engine/pool.rs
trgill: The pool can't insert it into the BTreeMap and return a mutable reference, right?

Are you suggesting we return the mutable object, then the caller needs to make another call to get it inserted into the BTreeMap once they are done with it?

src/sim_engine/pool.rs
trgill: Ok

mulkieran:
:+1:

PR #164, opened 2016-12-02
Turn stratisd into a library.

mulkieran:


mulkieran:
@agrover @trgill review?

Note that this has the other benefit that now docs are generated properly.

mulkieran:
I sort of exaggerated about the docs being generated properly. They were always being generated properly, but the only thing in stratisd's public interface was ```debug```.

mulkieran:
I'm not really so sure that this change achieves much, since it would be necessary to expose the engines as part of the API in order to test them as part of the integration tests, and it's better to be conservative about exposing stuff as part of the public API.

mulkieran:
Not worth doing at this time. There are some good reasons to split Stratisd into a front-end and a back-end, but possible to get by with some sort of internal testing for a while.

PR #166, opened 2016-12-02
Add some testing to project and CI

mulkieran:


mulkieran:
@agrover @trgill review?

This does not actually do much, but it does show that we can do testing, even quickcheck style, and include it in CI.

trgill:
:+1:

agrover:
👍 

So we're past the issue in #90?

mulkieran:
@agrover: We can't do tests on the whole thing unless it is a library. We can do unit tests at every level of module and submodule, which is almost like doing integration tests. It does give us an incentive to organize modules and submodules carefully, to facilitate those tests.

PR #171, opened 2016-12-06
Include facilities for renaming a pool

mulkieran:


mulkieran:
@agrover @trgill review?

mulkieran:
I'm trying the request a review feature that seems to have appeared.

trgill:
:+1:   I had one question about returning a "internal error" vs "ok" for a case that shouldn't happen.  Your choice to address or leave as is...

mulkieran:
There is a tricky bit that I just realized that I'ld like to discuss in meeting.

mulkieran:
User 1 and User 2 both call ```rename()``` on same pool using Stratis CLI. This a two step process. First they get the object path for the pool, then they invoke the Rename() method on the object path with the new name:

```
User 1                                                    User 2
Gets pool object path for "old name"
                                                              Get pool object path for "old name"
Calls rename("name1"), pool is now "name1"
                                                              Calls rename("name2"), pool is now "name2"
```

User 2 just won, and User 1 was told that they succeeded, and they did, but then they lost to User 2. This maybe isn't so bad, because User 2 could have gone after User 1 completed, getting the object path by "name1" instead and poor User 2 would still have lost.

There are a bunch of variations on this theme, like User 1 successfully adds some devices to a pool named "pool", but that pool is actually named "somepool" when their command is done, because User 1 got busy renaming.

How do we feel about these? If we're not OK with these scenarios, then we need to start calling methods on the Manager, and dbus objects that represent pools and filesystems should be restricted to having only read-only properties, no methods.

agrover:
I think we're ok with the race. This should be communicated to the client with DBus PropertyChanged signals in the future.

agrover:
I think we're all 👍 on this, so I'm going to merge this in an hour if there are no objections (or nobody beats me to it).

PR #173, opened 2016-12-07
Use macros for shared code in *Engine methods

mulkieran:


mulkieran:
@grover @trgill review?

I happened to notice the extra files after doing the restructuring, so I thought they might as well be removed.
The directory restructuring is a good idea anyway, but the motivation right now is to make importing the macros that I define in the new file macros.rs occur in a sensible manner. I exposed the engine API directly, but require people to go through the ```sim_``` or ```strat_engine``` subpackages to access those things directly. This only happens when choosing between engines. The macros are completely straight forward, but sharing the code between the two engines wherever possible seems like a good idea. There is a sense in which testing the sim engine also tests parts of the strat engine by that means, which is nice.

mulkieran:
Whoops! One more thing.

mulkieran:
@agrover @trgill really review?

Now, commits are:
1) Remove two unused...
2) Update D-Bus mapping when destroying (bug fix).

Directory restructuring commits:

3) Push engine...
4) Use macros...
5) Now make destroy_pool()
6) Update D-Bus API to return result...

This is the most urgent of my outstanding PRs, since the others could build on this to a greater or lesser extent.

mulkieran:
Whoops. Need to queue up up object path removal as well.

mulkieran:
@agrover @trgill review please?

agrover:
looks fine to me. 👍

PR #176, opened 2016-12-09
Remove dead code.

mulkieran:
These methods are no longer used, since we do not currently construct dbus
objects for the devs.

This is a follow up to PR #156.


mulkieran:
@agrover @trgill review?

trgill:
:+1:

PR #179, opened 2016-12-10
Simplify Pool, Filesystem implementation and API.

mulkieran:

### DIFF
--- a/src/dbus_api.rs
+++ b/src/dbus_api.rs
@@ -507,10 +507,15 @@ fn destroy_filesystems(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
 
     let filesystems: Array<&str, _> = try!(get_next_arg(&mut iter, 0));
 
+    let mut filesystem_names = Vec::new();
+    for name in filesystems {
+        filesystem_names.push(name);
+    }
+
     let dbus_context = m.path.get_data();
     let object_path = m.path.get_name();
     let return_message = message.method_return();
-    let return_sig = "(qs)";
+    let return_sig = "s";
     let default_return = MessageItem::Array(vec![], return_sig.into());
 
     let pool_name = dbus_try!(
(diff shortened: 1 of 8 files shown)


mulkieran:
@agrover @trgill review?

I would not recommend looking at the commits individually, things are created and then completelye destroyed within this PR. I will see about squashing commits to remove some of that, if changes basically OKd.

Main changes are:
* Go back to mapping in pool from filesystem name to filesystem. The filesystem struct contains its UUID. I know the UUID is important, but, from management perspective right now, the name is the way to do the lookup. So lets stick with that.
* Eliminate chunks of Filesystem and Pool API that are unnecessary. The chunks are deemed unnecessary if they are not currently presented to the D-Bus layer for any reason.
* Implement ```rename_filesystem``` as part of Pool, akin to how ```rename_pool``` is part of Engine. Implementation of the first is much the same as that of the second.
* Make D-Bus ```rename_filesystem``` (called on filesystem) return a boolean, just like D-Bus ```rename_pool```.
* Make D-Bus ```destroy_filesystems``` return an array of strings, containing the names of the filesystems actually destroyed.
* Pluralize Engine's ```destroy_filesystems```.
* Macroize where possible. These macros should be considered more as scaffolding than as a permaenent structure, mostly.
* For StratEngine put StratFilesystem implementation into a separate file.

I plan to finish up this PR by pluralizing ```create_filesystem``` and by squashing some commits.

In the meantime, I'ld like your feedback.

trgill:
why move functions like destroy_filesystem() into a macro?

mulkieran:
@trgill What is special about ```destroy_filesystems```?

trgill:
nothing special about destroy_filesystems - just curious why it is better defined as a macro?  This PR changed it from a function to macro, right?

mulkieran:
The bodies of the functions are defined as macros. The general purpose is to force the two engine implementations to share source code, and thus to share characteristics, like struct fields, that they probably should share.

Soon, probably, it will become apparent for many of the methods, that they won't be doing entirely identical things. I think the only ones that have a chance of being implemented as they are forever are the ```rename*```, ```pools()```, and ```get_pool()``` methods, since they are exclusively about juggling data structures. My expectation is that ```pools()``` is likely to go away with the ```GetManagedObjects``` refactoring, anyway.

For the functions for which the current macros are probably inadequate, we'll have to see what happens next. We might find that much remains the same. Take ```destroy_pool()``` as an example. Most of the current macro is still going to be shared. What will change is the very end, since the real engine will have to mess with actual metadata while the sim engine can, at least now, just mess with a data structure, and will never end up dong the same things as the real engine. But, we may find that that is the thing to abstract that action, leaving most of the macro intact.

Why macros instead of shared functions? Because the shared functions have to know the type of self, which, if they are truly shared, must be Engine. Engine structs are opaque, and Rust does not have inheritance. So it seems like macros are the only choice, and because they are hygienic, they are actually a pretty good choice.

The drawback of macros is that they are slightly harder to debug. I've tried to keep them very formulaic.

mulkieran:
Superceded by #181.

PR #180, opened 2016-12-12
Add rustfmt.toml to set options for rustfmt

trgill:
Initial version turns on ordering of imports.  This is just a proposal.  Seems like automatic ordering 
is a good idea.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,2 @@
+format_strings = false
+reorder_imports = true
--- a/src/dbus_api.rs
+++ b/src/dbus_api.rs
@@ -4,31 +4,23 @@
 
 use bidir_map::BidirMap;
 
-use std::borrow::Cow;
-use std::cell::Cell;
-use std::cell::RefCell;
-use std::fmt::Display;
-use std::path::Path;
-use std::rc::Rc;
-use std::vec::Vec;
-
 use dbus;
-use dbus::Connection;
 use dbus::BusType;
+use dbus::Connection;
+use dbus::ConnectionItem;
 use dbus::Message;
 use dbus::MessageItem;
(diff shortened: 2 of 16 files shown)


trgill:
@agrover @mulkieran this ok with you?

agrover:
Where are all the rustfmt options documented? I'm curious what `format_strings` is supposed to do.

I'm still a little unclear on what it's doing. It appears to sometimes treat a series of `use` lines separated by whitespace differently from consecutive `use` lines.

trgill:
Docs for rustfmt: 
https://github.com/rust-lang-nursery/rustfmt 

This command provides the valid options for rustfmt.toml along with a short description:
rustfmt --config-help 

I think rustfmt attempts to leave any empty lines "as is".  In this case it make looks a bit odd.  I can fix.  Just need to decide on whatever convention we want.

agrover:
OK. Ideally I'd prefer std, external crate, and internal use decls to be grouped separately (and in that order) but if the tool can't do that then I'll live with it 😃 

👍 

Does `format_strings=false` actually change anything in this patch? If not let's omit changing it from the default for now.

mulkieran:
I agree with @agrover  in that I like the less specific imports, like std:: to come before the more specific imports, like super:: but everything to be alphabetical within that. Maybe it can be persuaded somehow? I will look at the options tomorrow.

trgill:
@agrover - no the format_strings=false doesn't change anything.  I'll remove it.

@mulkieran the only other related flag I see is:  

reorder_imported_names = true

It also alphabetizes elements containted within {}, so for example it would change:

-use std::io::{Read, Write, ErrorKind, Seek, SeekFrom};
to
+use std::io::{ErrorKind, Read, Seek, SeekFrom, Write};

PR #182, opened 2016-12-14
Initial work on pluralizing remove* methods

mulkieran:


mulkieran:
@agrover @trgill preliminary review?

The basic changes are:

* Clean up CacheDev and Dev API a little. I won't be able to make more progress until GetManagedObjects refactoring.
* Abstract resovling of paths to devices. It currently doesn't do much, but that is where we whould be doing things like following links, handling vpd83 info, etc.
* Move a bunch of BlockDev code into two separate modules: metadata (handling metadata) and setup (which is about batch processing of blockdevs for setup and initialization).

Things I am postponing to some future PR:
* Testing metadata code. It is easy to test and should be tested, but we know that it will change, so we might as well wait until later. Hopefully, what is there now more or less captures the gestalt.
* Formalizing errors raised by metadata code, for the same reason as above.
* Concerning myself with the return type of ```find_all()``` in setup.rs. No other code depends on it at this time.

Next steps are to consider whether it makes sense to write a purpose-built data structure for Devs and CacheDevs now, or whether it is possible to go direct to pluralizing ```remove_blockdev``` and ```remove_cachedev```.

agrover:
First of all, this is some pretty extensive work, well done. This should make our upcoming metadata changes go much more smoothly.

I like moving stuff to metadata.rs but I'm less favorable to the move of things into setup.rs. I think everything within it could just as easily stay in blockdev.rs, especially since blockdev.rs is now down to 100 lines!

I think adding more explanatory text to commit messages would be worthwhile. Not just what, but why the change is an improvement. Also be careful about unrelated changes slipping in, I think one of the first couple has a function param change name that is not mentioned in the message. Mentioning everything in the message helps us reviewers know it didn't just slip in. The other place I'd like more in the commitlog is around leaving read/write_mdax() unimplemented.

I'm also not quite sure about a couple function names: `determine ownership`? (Maybe just `ownership` or `owner`?) MDAGroup and MDA  `read` and `write` don't actually do I/O, they're more about converting between structs and their in-buffer representation.

mulkieran:
Superceded by #192.

PR #183, opened 2016-12-14
Initial version of integration test framework.

trgill:

### DIFF
--- a/stratis-tests/src/blockdev_tests.rs
+++ b/stratis-tests/src/blockdev_tests.rs
@@ -1,0 +1,62 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use std::collections::BTreeSet;
+use std::path::Path;
+use std::str::FromStr;
+
+use uuid::Uuid;
+
+use common::*;
+
+use devicemapper::Device;
+
+use stratisd::engine::strat_engine::blockdev::BlockDev;
+use stratisd::engine::strat_engine::consts::MIN_MDA_SIZE;
+
+pub fn test_blockdev_force_flag(blockdev_paths: &Vec<&Path>) -> TestResult<()> {
+
+    let mut devices = BTreeSet::new();
+    let mut devices_copy = BTreeSet::new();
+    for path in blockdev_paths {
+        let dev = try!(Device::from_str(&path.to_string_lossy()));
+        devices.insert(dev);
+        devices_copy.insert(dev.clone());
+    }
+
+    // Initialzie devices with force = ture
+    let fake_pool_uuid = Uuid::new_v4();
+    try!(BlockDev::initialize(&fake_pool_uuid, devices, MIN_MDA_SIZE, true));
+
+    // Try to initialzie again with different uuid, force = false - this should fail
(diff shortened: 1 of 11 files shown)


trgill:

I think this is just a step towards what we want.  There is still work that needs to be done around logging and reporting etc.

trgill:
Hold on... there is a warning due to the new way things are compiled.  Will fix.

mulkieran:
I think that the idea of splitting stratisd into a front end (the D-Bus part) and a back end (the engine part) should go back on the table.

stratis-tests/src/test_consts.rs
```diff
@@ -0,0 +1,3 @@
+
+pub const OK_TO_DESTROY_DEV_ARRAY: &'static str = "ok_to_destroy_dev_array";
```
mulkieran: I don't know why something called an array has type string.

stratis-tests/src/blockdev_tests.rs
```diff
@@ -0,0 +1,62 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use std::collections::BTreeSet;
+use std::path::Path;
+use std::str::FromStr;
+
+use uuid::Uuid;
+
+use common::*;
```
trgill: Ok

trgill:
I agree with @mulkieran  that we should put the idea of "splitting stratisd into a front end (the D-Bus part) and a back end (the engine part) " should go back on the table.

stratis-tests/src/blockdev_tests.rs
```diff
@@ -0,0 +1,62 @@
...
+            info!("PASS: initialize of already initialzied blockdevs failed (intended)");
+        }
+    }
+
+    Ok(())
+}
+
+pub fn test_blockdevs(blockdev_paths: &Vec<&Path>) -> TestResult<()> {
+    info!("Starting BlockDev Tests...");
+    let mut devices = BTreeSet::new();
```
mulkieran: You should be able to use ```resolve_devices``` in place of this block of code here.

stratis-tests/src/blockdev_tests.rs
```diff
@@ -0,0 +1,62 @@
...
+use devicemapper::Device;
+
+use stratisd::engine::strat_engine::blockdev::BlockDev;
+use stratisd::engine::strat_engine::consts::MIN_MDA_SIZE;
+
+pub fn test_blockdev_force_flag(blockdev_paths: &Vec<&Path>) -> TestResult<()> {
+
+    let mut devices = BTreeSet::new();
+    let mut devices_copy = BTreeSet::new();
+    for path in blockdev_paths {
```
mulkieran: You can use ```resolve_devices``` here.

stratis-tests/src/blockdev_tests.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use uuid::Uuid;
+
+use common::*;
+
+use devicemapper::Device;
+
+use stratisd::engine::strat_engine::blockdev::BlockDev;
+use stratisd::engine::strat_engine::consts::MIN_MDA_SIZE;
+
```
mulkieran: A comment here, stating the property being tested would be helpful.

trgill:
Replaced by PR #199

PR #192, opened 2016-12-22
Refactor BlockDev implementation contents 

mulkieran:


mulkieran:
Supercedes #182.

@agrover, @trgill review?

Things I'm deferring until a subsequent PR:
* moving all or most of ```read_mdax()```, ```write_mdax()```, ```write_hdr_buf()```,  ```main_bda_size()```, and ```aux_bda_size()``` out of BlockDev impl. They are not in the way, so they are not terribly urgent. It will be useful to abstract disk geometry a bit, however.
* Purpose built data structures to handle BlockDevs.

agrover:
I'm still not sure why blockdev setup and initialize should go someplace other than blockdev.rs. Other layers are going to have setup and initialize too, where will that code go?

mulkieran:
@agrover which other layers?

mulkieran:
@agrover @trgill changes made, review?

agrover:
Yep, I just find the mod nesting unnecessary. Pushed a commit to de-mod. If that's acceptable then I say 👍

mulkieran:
@agrover I'ld prefer that the auxiliary functions were defined within the functions that used them, since it is not clear that they will ever be useful to any other functions. So, I've made that change.

mulkieran:
@agrover @trgll review?

"Have blockdev::initialize..." completely overwrites the changes in "Allow a default value...".

If "Have blockdev::initalize..." is OK, I'll squash it into "Allow a default value..." for final commit.

trgill:
:+1:

mulkieran:
The only change is that "Have blockdev::initialize..." and "Allow a default value for MDA size..." are both dropped.

@agrover Is this good to go?

PR #193, opened 2017-01-04
Master object manager

mulkieran:


mulkieran:
This review introduces two data structures, lookup tables for the DBus API. They are consistent w/ each other. The question is, are they a good idea. What is the effect of the static dbus Paths?

agrover:
Is all of this to better support GetPoolObjectPath? I thought the ObjectManager approach would let us do away with that method, which means that all this related code is also not needed.

agrover:
Also, it may make things easier to use TData::ObjectPath to store the name. This would save having to look it up.

mulkieran:
The answer to the first question is no. We absolutely have to be able to map between object paths and names because the engine does not know about object paths.

agrover:
ok, but you knew the name and object path when you created the dbus object, I was thinking you could just save the name in the TData, and then when executing a method it has access to the name without doing another lookup.

mulkieran:
That is a thought...because if the name changes, well, we have the object path at the time that we change the name or not, so we could update it then. This also means that ```get_pool_name()``` and ```get_pool_uuid()``` could simply get these values from the data attached to the object path. _However_, this requires some tight reasoning to be confident that we never change these values except when we are able to change the values attached to the object path _at the same time_. That would have to be true in all cases in order to make this work and make these table unnecessary (I would be delighted if these tables were made unnecessary).

mulkieran:
#206 is making good progress. If it can just remove the necessity for these tables, then there is no point in putting these improvements in.

mulkieran:
See https://github.com/diwic/dbus-rs/issues/73 for a fuller discussion of the problem we currently face.

We _could_ decorate the object path with data, like the name and the uuid. That would almost work. However, probably the correct approach, given the use of GetManagedObjects(), would be to change ```DestroyPool``` and ```DestroyFilesystems``` so that the pool/filesystms were specified by means of object paths. The problem is that there is no easy way to go from those object paths, which are really just strings, to the data attached to the object paths in the tree. So, such a scheme would still necessitate a look-up table for those situations, making decorating the object path in the tree redundant. This would still be better than the current case, since the table would only have to be one-way, not bidirectional, would save a surprising amount of trouble.

An alternative is to make two Destroy methods, one for Pool interface which destroys Pools and for Filesystem interface which destroys files. We have talked about that before. It would, sort of, solve the problem in this particular case. But the problem, which is simply that passing an ObjectPath as a parameter is currently not that helpful unless you have a table for properties, might resurface in another context anyway. Also, there are drawbacks to this approach which I have discussed before. The only reason to take this approach is if we have decided, once and for all, that we will never have an interface that allows object paths to be passed as arguments. Such an interface could be realized, and could even be made efficient, but at the cost of more complexity in the client and the service.

So, my current plan is to proceed with #206 
* to make our look-up tables for pools and filesystems one-directional
* to make ```DestroyPool()``` and ```DestroyFilesystems()``` take object paths as arguments
* to require the lookup tables to map from an object path as Path to exactly the same data as the properties that we publish
* to implement the properties in terms of these lookup tables that hold their values
* and to hope that we can eventually ditch these lookup tables when the dbus implementation allows us to map from object paths received as arguments to object paths in the tree.

@agrover, @trgill, @tasleson any thoughts?

mulkieran:
Superceded by #206 and subsequent. The only interesting thing left here is the change to ActionQueue's ```push_remove()``` method in last commit.

PR #194, opened 2017-01-05
Some GetManagedObjects refactoring preliminary work

mulkieran:


src/dbus_api.rs
```diff
@@ -39,6 +39,9 @@ use engine::RenameAction;
 
 use types::StratisResult;
 
+const STRATIS_BASE_PATH: &'static str = "/org/storage/stratis1";
+const STRATIS_BASE_SERVICE: &'static str = "org.storage.stratis1";
```
agrover: No consts left in dbus_consts.rs, fold into dbus_api.rs in the future?

trgill:
:+1:  Ok

src/dbus_api.rs
```diff
@@ -266,7 +274,7 @@ fn ok_message_items() -> (MessageItem, MessageItem) {
 }
 
 fn default_object_path<'a>() -> dbus::Path<'a> {
-    dbus::Path::new(DEFAULT_OBJECT_PATH).unwrap()
+    dbus::Path::new("/").unwrap()
```
agrover: Yeah, when to use consts should probably go in the (to-be-written) style guide. Especially in this case where we are in a function that just wraps the const, and there's really no other valid value for the const other than the one, using a const seems unnecessary. Consts are also really best for when more than one place relies on a value, which is not the case for the method names below. There is a trade-off, since it becomes a little harder to see all the defined methods that used to be grouped in a single place, but I still tend to think this is a preferable approach.

src/dbus_api.rs
```diff
@@ -317,9 +325,11 @@ fn create_dbus_filesystem<'a>(mut dbus_context: DbusContext) -> dbus::Path<'a> {
                               STRATIS_BASE_PATH,
                               dbus_context.get_next_id().to_string());
 
+    let interface_name = format!("{}.{}", STRATIS_BASE_SERVICE, "filesystem");
+
```
agrover: this binding strictly isn't needed.. did you consider the alternative and decide to do it this way for clarity or line length reasons?

mulkieran: aesthetics.

src/dbus_api.rs
mulkieran: I'm planning some reorganization.

PR #197, opened 2017-01-06
Master redundancy

mulkieran:


mulkieran:
@agrover @trgill preliminary review?

The main purpose of this PR is to change GetErrorCodes and GetRaidLevels methods to RO properties.

To do that it is necessary to define RAID levels as something that the engine, not just the D-Bus layer knows about. So a significant part of this PR is pushing this concept down into the engine.

Other things are that I changed RaidLevels to RedundancyCodes since the word Redundancy is sufficientgly general, and RAID isn't, and because RaidLevels makes no sense with modern RAID concepts.

Also, the properties are a list of pairs, name and value, instead of triples,name, value, and description, since description is a pain to maintain and isn't really useful to caller. Probably the manual is the best place for descriptiosn.

I'm still writing some tests, and may try to fix up the commits to be slightly tidier, but I'm failrly satisified with this PR. If you are too, it can go in today.

mulkieran:
Note: There is a wierd problem, but I think it may be a problem with a client library. Compare the following:

```
[mulhern@dhcp-25-209 stratisd-client-dbus]$ busctl --user get-property org.storage.stratis1 /org/storage/stratis1 org.storage.stratis1.Manager RedundancyCodes
a(sq) 1 "RAID0" 0
```

From QT5DbusViewer
```
Received reply from :1.6899
  Arguments: [Variant: [Argument: a(sq) {[Argument: (sq) "RAID0", 0]}]]
```

From dbus-python
```
(Pdb) Manager.Properties.RedundancyCodes(self._proxy)
dbus.Array([dbus.Struct((dbus.String('RAID0'), dbus.UInt16(0)), signature=None)], signature=dbus.Signature('(sq)'), variant_level=1)
```

Note how the bottom 2 think that they're dealing with a variant type, while busctl does not seem to.
Note also that it is not the type within the array that is a variant, but the array itself.

agrover:
I think this looks 👍 . My only (perhaps bikeshedding) comment is a slight preference for RedundancyValues (and ErrorValues) over *Codes, as being slightly more clear.

mulkieran:
Regarding the spurious variant type in the signature of the returned value of the property calls...

The bad news is:
1. After some poking around I still have no idea why this is happening. I don't think it is due to dbus-python doing introspection since I can not observe it doing introspection and stratisd's introspection data is correct.
2. I'm not sure that reporting the problem would do any good, since dbus-python maintainer is uninterested.

The good news is that I think we can ignore this problem, as the type of the contained variant is correct, and we can use the value without any difficulty. The main problem is that our tests won't be able to check the types of properties for correctness because they will have the wrong type, so at some point our tests may fail to catch an actually, rather than a spuriously incorrect type. But I think that can be got around by extending into-dbus-python's signature method a bit, so I'm going to look into that.

trgill:
:+1:

mulkieran:
Superceded by #198.

PR #199, opened 2017-01-10
build as a bin and lib.

trgill:
Move main.rs to bin/stratisd.rs to allow for building stratisd as bin.

Added lib.rs so that the remainder of the application is compiled as a
lib crate.

Added two sample integration tests blockdev_tests.rs and pool_tests.rs.

Added [dev-dependencies] to include Json parser, log, and env_logger for
use in the integration tests.

Note: see README.md in the tests directory for instructions on enabling test logging.

Log levels currently supported are debug and info.

Added test_config.json to hold parameters for tests.  Currently it only
has 1 array element "ok_to_destroy_dev_array_key".  This array contains a
a list of devices that can be safely destroyed during testing.

Moved dbgp! macro to stratisd.rs - it is the only place it is used - we
should probably switch to a logging crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "stratisd"
+name = "libstratis"
 version = "0.1.0"
 authors = ["Andy Grover <agrover@redhat.com>", "Mulhern <amulhern@redhat.com>", "Todd Gill <tgill@redhat.com>"]
 build = "build.rs"
@@ -29,3 +29,6 @@ features = ["v4"]
 
 [dev-dependencies]
 quickcheck = "0.4"
+rustc-serialize = "0.3.21"
+log = "0.3"
+env_logger="0.3.5"
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -2,7 +2,8 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-#![allow(dead_code)] // only temporary, until more stuff is filled in
+#[macro_use]
+extern crate libstratis;
 
 extern crate devicemapper;
 #[macro_use]
@@ -29,28 +30,17 @@ extern crate enum_derive;
 #[cfg(test)]
 extern crate quickcheck;
 
-mod types;
-mod consts;
-mod dbus_consts;
-#[macro_use]
-mod stratis;
-mod dbus_api;
-mod engine;
-
 use std::io::Write;
 use std::error::Error;
 use std::process::exit;
 
-use stratis::{DEBUG, VERSION};
-
-use types::{StratisResult, StratisError};
-
 use clap::{App, Arg};
 
-use engine::Engine;
-use engine::sim_engine::SimEngine;
-use engine::strat_engine::StratEngine;
-
+use libstratis::engine::Engine;
+use libstratis::engine::sim_engine::SimEngine;
+use libstratis::engine::strat_engine::StratEngine;
+use libstratis::stratis::VERSION;
+use libstratis::types::{StratisResult, StratisError};
 
 fn write_err(err: StratisError) -> StratisResult<()> {
     let mut out = term::stderr().expect("could not get stderr");
@@ -59,6 +49,17 @@ fn write_err(err: StratisError) -> StratisResult<()> {
     try!(writeln!(out, "{}", err.description()));
     try!(out.reset());
     Ok(())
+}
+
+pub static mut DEBUG: bool = false;
+
+macro_rules! dbgp {
+    ($($arg:tt)*) => (
+        unsafe {
+            if DEBUG {
+                println!($($arg)*)
+            }
+        })
 }
 
 fn main() {
@@ -87,7 +88,7 @@ fn main() {
         }
     };
 
-    let r = dbus_api::run(engine);
+    let r = libstratis::dbus_api::run(engine);
 
     if let Err(r) = r {
         if let Err(e) = write_err(r) {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,0 +1,33 @@
+extern crate devicemapper;
+#[macro_use]
+extern crate clap;
+#[macro_use]
+extern crate nix;
+extern crate crc;
+extern crate bidir_map;
+extern crate byteorder;
+extern crate uuid;
+extern crate time;
+extern crate dbus;
+extern crate term;
+extern crate rand;
+extern crate serde;
+
+#[cfg(test)]
+extern crate quickcheck;
+
+pub mod types;
+pub mod consts;
+#[macro_use]
+pub mod stratis;
+pub mod engine;
+pub mod dbus_api;
+mod dbus_consts;
+
+
+#[macro_use]
+extern crate custom_derive;
+#[macro_use]
+extern crate newtype_derive;
+#[macro_use]
+extern crate enum_derive;
--- a/tests/blockdev_tests.rs
+++ b/tests/blockdev_tests.rs
@@ -14,7 +14,7 @@ use libstratis::engine::strat_engine::metadata::MIN_MDA_SIZE;
 
 use std::path::Path;
 
-use util::blockdev_utils::clean_blockdevs_headers;
+use util::blockdev_utils::clean_blockdev_headers;
 use util::test_config::TestConfig;
 use util::test_consts::DEFAULT_CONFIG_FILE;
 use util::test_result::TestError::Framework;
@@ -87,8 +87,7 @@ pub fn test_blockdev_setup() {
 
     let device_paths = safe_to_destroy_devs.iter().map(|x| Path::new(x)).collect::<Vec<&Path>>();
 
-    try_clean_devs!(&device_paths);
-    info!("devices cleaned for test");
+    clean_blockdev_headers(&device_paths);
 
     assert!(match test_blockdev_force_flag(&device_paths) {
         Ok(_) => true,
--- a/tests/blockdev_tests.rs
+++ b/tests/blockdev_tests.rs
@@ -1,0 +1,101 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#[macro_use]
+extern crate log;
+extern crate uuid;
+extern crate devicemapper;
+extern crate libstratis;
+#[macro_use]
+mod util;
+
+use libstratis::engine::strat_engine::blockdev;
+use libstratis::engine::strat_engine::metadata::MIN_MDA_SIZE;
+
+use std::path::Path;
+
+use util::blockdev_utils::clean_blockdevs_headers;
+use util::test_config::TestConfig;
+use util::test_consts::DEFAULT_CONFIG_FILE;
+use util::test_result::TestError::Framework;
+use util::test_result::TestErrorEnum::Error;
+use util::test_result::TestResult;
+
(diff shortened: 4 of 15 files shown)


trgill:
@mulkieran @agrover review please?

FYI - I had to put a #[ignore] tag on the test_pools().  We can remove when we implement cleanup of the disk label for destroy_pool().

Cargo.toml
```diff
@@ -29,3 +29,6 @@ features = ["v4"]
 
 [dev-dependencies]
 quickcheck = "0.4"
+rustc-serialize = "0.3.21"
+log = "0.3"
+env_logger="0.3.5"
```
trgill: Will add newline here.

agrover:
Wow you've been busy! 😃

- I think some of the stuff about running the tests, log levels etc deserves to have greater visibility than the checkin comment. Put it in the README.md?
- This one commit is pretty big. Are there ways it could separate logical changes into separate commits more? Moving dbgp, adding deps and updating Cargo.lock, making into a lib, test framework, the initial tests themselves -- these all build on each other but seem logically distinct to me.
- in the commitlog for the libification, can you take a stab at discussing the other methods you investigated before settling on this approach? Having a record of why this is better than the Cargo approach will be great if we get questions from contributors down the line.

Much thanks! This is going to be so nice to have this in place.

trgill:
@agrover  Ok - thanks.  Will do.

trgill:
@agrover I broke up the commit into steps that would compile and follow your suggestions as much as possible.

mulkieran:
@trgill Is there some preliminary code for releasing the pools properly in the works?

trgill:
@mulkieran I don't have any code for releasing the pools properly.  I talked with @agrover about the issue briefly.  I think he is considering some alternatives based on how recoverable we want to make deleted pools.

tests/blockdev_tests.rs
```diff
@@ -0,0 +1,89 @@
...
+
+pub fn test_blockdev_force_flag(blockdev_paths: &Vec<&Path>) -> TestResult<()> {
+
+    let mut devices = BTreeSet::new();
+    let mut devices_copy = BTreeSet::new();
+    for path in blockdev_paths {
+        let dev = try!(Device::from_str(&path.to_string_lossy()));
+        devices.insert(dev);
+        devices_copy.insert(dev.clone());
+    }
```
trgill: I need to change this to use blockdev::resolve_devices().  Will do that now.

src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+#[cfg(test)]
+extern crate quickcheck;
+
+pub mod types;
+pub mod consts;
+#[macro_use]
+pub mod stratis;
+pub mod engine;
+pub mod dbus_api;
+#[allow(dead_code)]
```
mulkieran: What dead code does this allow? Should the dead code get removed?

trgill: Fixed

Cargo.toml
mulkieran: Is your comment above totally obsolete?

src/bin/stratisd.rs
```diff
@@ -61,6 +52,17 @@ fn write_err(err: StratisError) -> StratisResult<()> {
     Ok(())
 }
 
+pub static mut DEBUG: bool = false;
```
mulkieran: If ```dbgp!``` was moved, rather than copied, shouldn't there be a corresponding ```-``` entry in this commit?

tests/util/test_result.rs
```diff
@@ -0,0 +1,82 @@
...
+extern crate libstratis;
+
+use libstratis::engine::EngineError;
+use self::rustc_serialize::json::ParserError;
+use std::fmt;
+use std::io::Error;
+
+#[derive(Debug)]
+pub enum TestErrorEnum {
+    _Ok,
```
mulkieran: Might as well get rid of constructors with underscores. Also, it is possible that ```_Ok``` will always be unnecessary, for Ok is by definition not an error.

tests/util/test_result.rs
```diff
@@ -0,0 +1,82 @@
...
+    Error(String),
+    JsonError(String),
+    _NotFound(String),
+}
+
+impl TestErrorEnum {
```
mulkieran: See https://github.com/stratis-storage/stratisd/pull/207/commits/b2ab75b804b7c3cdee9ce0c0107a2d44d8883df0 where some improvements are made to the existing error structure. Probably this commit should get off on the right foot by including those improvements.

tests/util/test_result.rs
```diff
@@ -0,0 +1,82 @@
...
+    }
+}
+
+impl fmt::Display for TestErrorEnum {
+    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        write!(f, "{}", self.get_error_string())
+    }
+}
+
+// Define a common error enum.
```
mulkieran: Please remove this comment.

src/bin/stratisd.rs
```diff
@@ -2,7 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-#![allow(dead_code)] // only temporary, until more stuff is filled in
+// only temporary, until more stuff is filled in
```
mulkieran: Well, rustfmt did us a favor here. You can drop this commit with PR #209, but now we know that you can still remove the ```#![allow(dead_code)]``` directive and accompanying comment in one of these commits.

src/bin/stratisd.rs
mulkieran: Ok, looks good now.

tests/util/blockdev_utils.rs
```diff
@@ -0,0 +1,87 @@
...
+		        Err(e) => {
+		        	error!("Failed to clean header. {:?}", e);
+		        	false
+			    }
+			}
+		)
+    }
+}
+
+#[allow(dead_code)]
```
mulkieran: You should be able to drop this directive when you add the commit that uses this method.

tests/util/blockdev_utils.rs
```diff
@@ -0,0 +1,87 @@
...
+            Err(err) => {
+                panic!("Failed to clean signature on {:?} : {:?}", path, err);
+            }
+        }
+    }
+
+    Ok(())
+}
+
+#[allow(dead_code)]
```
mulkieran: Looks like this method can be private from the start, so  dead code directive never needed.

tests/util/blockdev_utils.rs
```diff
@@ -0,0 +1,87 @@
...
+    let mut f = try!(OpenOptions::new().write(true).open(path));
+    let zeroed = [0u8; (SECTOR_SIZE * 8) as usize];
+
+    try!(f.write_all(&zeroed[..(SECTOR_SIZE * 8) as usize]));
+    try!(f.write_all(&zeroed[..(SECTOR_SIZE * 8) as usize]));
+
+    Ok(())
+}
+
+#[allow(dead_code)]
```
mulkieran: This function isn't used anywhere in the PR, can you remove it?

trgill:
@mulkieran I think I've addressed your comments except for 1.

If I don't tag blockdev_utils::get_ownership() with #[allow(dead_code)], I get a warning on the test compile.  The function is called in a test, I tried tagging with a

tests/util/blockdev_utils.rs
```diff
@@ -0,0 +1,74 @@
...
+        Err(err) => {
+            let error_message = format!("{} for device {:?}", err, path);
+            return Err(TestError::Framework(TestErrorEnum::Error(error_message)));
+        }
+    };
+
+    Ok(ownership)
+}
+
+pub fn clean_blockdevs_headers(blockdev_paths: &Vec<&Path>) -> TestResult<()> {
```
mulkieran: Seems like this has too many plurals in the name, ```clean_blockdev_headers``` might seem more natural. Also, it might be possible to substitute for this method a map and then a find, or something.

tests/util/blockdev_utils.rs
```diff
@@ -0,0 +1,74 @@
...
+		        	error!("Failed to clean header. {:?}", e);
+		        	false
+			    }
+			}
+		)
+    }
+}
+
+#[allow(dead_code)] // Needed to avoid warning on test build
+pub fn get_ownership(path: &PathBuf) -> TestResult<DevOwnership> {
```
mulkieran: I think the warning is caused because this method is only used in a test that is currently ignored. The comment could say that.

tests/util/test_config.rs
```diff
@@ -0,0 +1,127 @@
...
+use std::io::prelude::Read;
+use util::test_consts::OK_TO_DESTROY_DEV_KEY;
+use util::test_result::TestError;
+use util::test_result::TestErrorEnum;
+
+use util::test_result::TestResult;
+
+pub struct TestConfig {
+    filename: String,
+    json: Option<Json>,
```
mulkieran: Please use a name that defines the thing, not its format.

tests/util/test_result.rs
```diff
@@ -0,0 +1,76 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+extern crate rustc_serialize;
```
mulkieran: Everything in this file makes sense to me.

tests/util/test_config.rs
```diff
@@ -0,0 +1,127 @@
...
+                return Err(TestError::Framework(TestErrorEnum::JsonError("Array conversion \
+                                                                          failed"
+                    .into())))
+            }
+        };
+
+        Ok(ok_to_destroy_dev_list)
+    }
+
+    pub fn read_test_config(&mut self) -> TestResult<()> {
```
mulkieran: A comment would be helpful here.

mulkieran: I think you can make this method private.

tests/util/test_config.rs
```diff
@@ -0,0 +1,127 @@
...
+            Err(_) => {
+                return Err(TestError::Framework(TestErrorEnum::Error("Failed to read test config"
+                    .into())))
+            }
+        }
+
+    }
+
+
+    pub fn get_safe_to_destroy_devs(&mut self) -> TestResult<Vec<String>> {
```
mulkieran: Method header comment helpful here.

tests/util/test_config.rs
```diff
@@ -0,0 +1,127 @@
...
+use util::test_result::TestErrorEnum;
+
+use util::test_result::TestResult;
+
+pub struct TestConfig {
+    filename: String,
+    json: Option<Json>,
+}
+
+macro_rules! test_try_get_destroy_devs {
```
mulkieran: Maybe this could be a function, not a macro, as well.

trgill:
RE:  removing OK_TO_DESTROY_DEV_KEY and using string literal

I think we are going to be adding other sections to the JSON.  It would be best to keep the keys in one place.

trgill:
How is PR #207 is going to make the tests obsolete?

One test is ignored because it fails due to a bug  releasing devices that needs to be fixed.  I'll remove the #ignore tag on it.  It won't run when the JSON is empty anyway.  But when we use it on real devices it will fail.

mulkieran:
I don't mean the tests will be obsolete, I mean that the code they rely on will be. So, tests will have to be fixed up immediately in order to compile by PR #207.

trgill:
I think the adjustments for PR #207 will be mostly to the code in blockdev_utils.rs.  I'd suggest we merge this first - then 207.

mulkieran:
Ok! :+1: .

PR #200, opened 2017-01-11
Add Version property to manager interface

mulkieran:


mulkieran:
@agrover @trgill review?

I suspect this is ready to go. I didn't use the ```crate_version``` macro, because the Rust compiler does not allow it in a constant context. I think the ```env``` macro is special because it is not a macro, it is a compiler built-in.

PR #202, opened 2017-01-12
Add appropriate emits changed designation to existing properties.

mulkieran:


mulkieran:
@agrover @trgill review? This is an easy ack. All it does is makes sure the introspection information includes attribute ```<annotation name="org.freedesktop.DBus.Property.EmitsChangedSignal" value="const"/>``` for our existing properties, which are all const.

PR #207, opened 2017-01-17
Update blockdev code for latest metadata format (0.7.5)

agrover:
WIP for the blockdev stuff I've been working on. It delves a little into some error rework, and then gets into the heart of the changes.

### DIFF
--- a/src/engine/strat_engine/blockdev.rs
+++ b/src/engine/strat_engine/blockdev.rs
@@ -3,36 +3,30 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 use std::io;
-
 use std::collections::BTreeSet;
 use std::collections::BTreeMap;
 use std::fs::File;
-use std::io::{ErrorKind, Read, Seek, SeekFrom, Write};
+use std::io::ErrorKind;
 use std::fs::{OpenOptions, read_dir};
-
 use std::os::unix::prelude::AsRawFd;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 
 use time::Timespec;
 use devicemapper::Device;
-use crc::crc32;
 use uuid::Uuid;
 
 use types::{Sectors, SectorOffset};
 use engine::{EngineResult, EngineError, ErrorEnum};
 
 use consts::*;
-
-use super::metadata::{SigBlock, validate_mda_size};
+use super::metadata::{StaticHeader, BDA, validate_mda_size, BDA_STATIC_HDR_SECTORS};
 use super::engine::DevOwnership;
-
 pub use super::BlockDevSave;
 
 type DevUuid = Uuid;
 type PoolUuid = Uuid;
 
-const BDA_STATIC_HDR_SIZE: Sectors = Sectors(8);
 const MIN_DEV_SIZE: u64 = GIGA;
 
 ioctl!(read blkgetsize64 with 0x12, 114; u64);
@@ -70,36 +64,15 @@ pub fn find_all() -> EngineResult<BTreeMap<PoolUuid, BTreeMap<DevUuid, BlockDev>
 
         let mut f = try!(OpenOptions::new()
             .read(true)
-            .open(devnode)
-            .map_err(|_| {
-                io::Error::new(ErrorKind::PermissionDenied,
-                               format!("Could not open {}", devnode.display()))
-            }));
+            .open(devnode));
 
-        let mut buf = [0u8; 4096];
-        try!(f.seek(SeekFrom::Start(0)));
-        try!(f.read(&mut buf));
-
-        match SigBlock::determine_ownership(&buf) {
-            Ok(DevOwnership::Ours(_)) => {}
-            Ok(_) => {
-                return Ok(None);
-            }
-            Err(err) => {
-                let error_message = format!("{} for devnode {}", err, devnode.display());
-                return Err(EngineError::Engine(ErrorEnum::Invalid, error_message));
-            }
-        };
+        let static_header = try!(StaticHeader::setup(&mut f));
+        let bda = try!(BDA::load(&mut f, static_header));
 
         Ok(Some(BlockDev {
             dev: dev,
             devnode: devnode.to_owned(),
-            sigblock: match SigBlock::read(&buf, 0, Sectors(try!(blkdev_size(&f)) / SECTOR_SIZE)) {
-                Ok(sigblock) => sigblock,
-                Err(err) => {
-                    return Err(EngineError::Engine(ErrorEnum::Invalid, err));
-                }
-            },
+            bda: bda,
         }))
     }
 
@@ -112,9 +85,9 @@ pub fn find_all() -> EngineResult<BTreeMap<PoolUuid, BTreeMap<DevUuid, BlockDev>
 
         match setup(&devnode) {
             Ok(Some(blockdev)) => {
-                pool_map.entry(blockdev.sigblock.pool_uuid)
+                pool_map.entry(blockdev.bda.header.pool_uuid)
                     .or_insert_with(BTreeMap::new)
-                    .insert(blockdev.sigblock.dev_uuid, blockdev);
+                    .insert(blockdev.bda.header.dev_uuid, blockdev);
             }
             _ => continue,
         };
@@ -135,7 +108,10 @@ pub fn initialize(pool_uuid: &PoolUuid,
 
     /// Gets device information, returns an error if problem with obtaining
     /// that information.
-    fn dev_info(dev: &Device) -> EngineResult<(PathBuf, u64, DevOwnership)> {
+    /// Returns a tuple with the blockdev's path, its size in bytes,
+    /// its ownership as determined by calling determine_ownership(),
+    /// and an open File handle, all of which are needed later.
+    fn dev_info(dev: &Device) -> EngineResult<(PathBuf, u64, DevOwnership, File)> {
         let devnode = try!(dev.path().ok_or_else(|| {
             io::Error::new(ErrorKind::InvalidInput,
                            format!("could not get device node from dev {}", dev.dstr()))
@@ -151,11 +127,7 @@ pub fn initialize(pool_uuid: &PoolUuid,
 
         let dev_size = try!(blkdev_size(&f));
 
-        let mut buf = [0u8; 4096];
-        try!(f.seek(SeekFrom::Start(0)));
-        try!(f.read(&mut buf));
-
-        let ownership = match SigBlock::determine_ownership(&buf) {
+        let ownership = match StaticHeader::determine_ownership(&mut f) {
             Ok(ownership) => ownership,
             Err(err) => {
                 let error_message = format!("{} for device {}", err, devnode.display());
@@ -163,7 +135,7 @@ pub fn initialize(pool_uuid: &PoolUuid,
             }
         };
 
-        Ok((devnode, dev_size, ownership))
+        Ok((devnode, dev_size, ownership, f))
     }
 
     /// Filter devices for admission to pool based on dev_infos.
@@ -172,15 +144,15 @@ pub fn initialize(pool_uuid: &PoolUuid,
     fn filter_devs<I>(dev_infos: I,
                       pool_uuid: &PoolUuid,
                       force: bool)
-                      -> EngineResult<Vec<(Device, (PathBuf, u64))>>
-        where I: Iterator<Item = (Device, EngineResult<(PathBuf, u64, DevOwnership)>)>
+                      -> EngineResult<Vec<(Device, (PathBuf, u64, File))>>
+        where I: Iterator<Item = (Device, EngineResult<(PathBuf, u64, DevOwnership, File)>)>
     {
         let mut add_devs = Vec::new();
         for (dev, dev_result) in dev_infos {
             if dev_result.is_err() {
                 return Err(dev_result.unwrap_err());
             }
-            let (devnode, dev_size, ownership) = dev_result.unwrap();
+            let (devnode, dev_size, ownership, f) = dev_result.unwrap();
             if dev_size < MIN_DEV_SIZE {
                 let error_message = format!("{} too small, minimum {} bytes",
                                             devnode.display(),
@@ -188,14 +160,13 @@ pub fn initialize(pool_uuid: &PoolUuid,
                 return Err(EngineError::Engine(ErrorEnum::Invalid, error_message));
             };
             match ownership {
-                DevOwnership::Unowned => add_devs.push((dev, (devnode, dev_size))),
+                DevOwnership::Unowned => add_devs.push((dev, (devnode, dev_size, f))),
                 DevOwnership::Theirs => {
                     if !force {
-                        let error_str = format!("First 4K of {} not zeroed", devnode.display());
+                        let error_str = format!("First 8K of {} not zeroed", devnode.display());
                         return Err(EngineError::Engine(ErrorEnum::Invalid, error_str));
-
                     } else {
-                        add_devs.push((dev, (devnode, dev_size)))
+                        add_devs.push((dev, (devnode, dev_size, f)))
                     }
                 }
                 DevOwnership::Ours(uuid) => {
@@ -211,28 +182,26 @@ pub fn initialize(pool_uuid: &PoolUuid,
         Ok(add_devs)
     }
 
-    match validate_mda_size(mda_size) {
-        None => {}
-        Some(err) => {
-            return Err(EngineError::Engine(ErrorEnum::Invalid, err));
-        }
-    };
+    try!(validate_mda_size(mda_size));
 
     let dev_infos = devices.into_iter().map(|d: Device| (d, dev_info(&d)));
 
     let add_devs = try!(filter_devs(dev_infos, pool_uuid, force));
 
     let mut bds = BTreeMap::new();
-    for (dev, (devnode, dev_size)) in add_devs {
+    for (dev, (devnode, dev_size, mut f)) in add_devs {
+
+        let static_header = StaticHeader::new(pool_uuid,
+                                              &Uuid::new_v4(),
+                                              mda_size,
+                                              Sectors(dev_size / SECTOR_SIZE));
+        let bda = try!(BDA::initialize(&mut f, static_header));
+
         let bd = BlockDev {
             dev: dev,
             devnode: devnode.clone(),
-            sigblock: SigBlock::new(pool_uuid,
-                                    &Uuid::new_v4(),
-                                    mda_size,
-                                    Sectors(dev_size / SECTOR_SIZE)),
+            bda: bda,
         };
-        try!(bd.write_sigblock());
         bds.insert(devnode, bd);
     }
     Ok(bds)
@@ -243,106 +212,15 @@ pub fn initialize(pool_uuid: &PoolUuid,
 pub struct BlockDev {
     pub dev: Device,
     pub devnode: PathBuf,
-    pub sigblock: SigBlock,
+    pub bda: BDA,
 }
 
 impl BlockDev {
     pub fn to_save(&self) -> BlockDevSave {
         BlockDevSave {
             devnode: self.devnode.clone(),
-            total_size: self.sigblock.total_size,
+            total_size: self.bda.header.blkdev_size,
         }
-    }
-
-    // Read metadata from newest MDA
-    pub fn read_mdax(&self) -> EngineResult<Vec<u8>> {
-        let younger_mda = self.sigblock.mda.most_recent();
-
-        if younger_mda.last_updated == Timespec::new(0, 0) {
-            let message = "Neither MDA region is in use";
-            return Err(EngineError::Engine(ErrorEnum::Invalid, message.into()));
-        };
-
-        let mut f = try!(OpenOptions::new().read(true).open(&self.devnode));
-        let mut buf = vec![0; younger_mda.used as usize];
-
-        // read metadata from disk
-        try!(f.seek(SeekFrom::Start((*BDA_STATIC_HDR_SIZE + *younger_mda.offset) * SECTOR_SIZE)));
-        try!(f.read_exact(&mut buf));
-
-        if younger_mda.crc != crc32::checksum_ieee(&buf) {
-            return Err(EngineError::Io(io::Error::new(ErrorKind::InvalidInput, "MDA CRC failed")));
-            // TODO: Read end-of-blockdev copy
-        }
-
-        Ok(buf)
-    }
-
-    // Write metadata to least-recently-written MDA
-    fn write_mdax(&mut self, time: &Timespec, metadata: &[u8]) -> EngineResult<()> {
-        let aux_bda_size = (*self.aux_bda_size() * SECTOR_SIZE) as i64;
-
-        if metadata.len() > self.sigblock.mda.mda_length as usize {
-            return Err(EngineError::Io(io::Error::new(io::ErrorKind::InvalidInput,
-                                                      format!("Metadata too large for MDA, {} \
-                                                               bytes",
-                                                              metadata.len()))));
-        }
-
-        let older_mda = self.sigblock.mda.least_recent();
-        older_mda.crc = crc32::checksum_ieee(metadata);
-        older_mda.used = metadata.len() as u32;
-        older_mda.last_updated = *time;
-
-        let mut f = try!(OpenOptions::new().write(true).open(&self.devnode));
-
-        // write metadata to disk
-        try!(f.seek(SeekFrom::Start((*BDA_STATIC_HDR_SIZE + *older_mda.offset) * SECTOR_SIZE)));
-        try!(f.write_all(&metadata));
-        try!(f.seek(SeekFrom::End(-aux_bda_size)));
-        try!(f.seek(SeekFrom::Current((*older_mda.offset * SECTOR_SIZE) as i64)));
-        try!(f.write_all(&metadata));
-        try!(f.flush());
-
-        Ok(())
-    }
-
-    pub fn write_sigblock(&self) -> EngineResult<()> {
-        let mut buf = [0u8; SECTOR_SIZE as usize];
-        self.sigblock.write(&mut buf, 0);
-        try!(self.write_hdr_buf(&self.devnode, &buf));
-        Ok(())
-    }
-
-    pub fn wipe_sigblock(&mut self) -> EngineResult<()> {
-        let buf = [0u8; SECTOR_SIZE as usize];
-        try!(self.write_hdr_buf(&self.devnode, &buf));
-        Ok(())
-    }
-
-    fn write_hdr_buf(&self, devnode: &Path, buf: &[u8; SECTOR_SIZE as usize]) -> EngineResult<()> {
-        let aux_bda_size = (*self.aux_bda_size() * SECTOR_SIZE) as i64;
-        let mut f = try!(OpenOptions::new().write(true).open(devnode));
-        let zeroed = [0u8; (SECTOR_SIZE * 8) as usize];
-
-        // Write 4K header to head & tail. Sigblock goes in sector 1.
-        try!(f.write_all(&zeroed[..SECTOR_SIZE as usize]));
-        try!(f.write_all(buf));
-        try!(f.write_all(&zeroed[(SECTOR_SIZE * 2) as usize..]));
-        try!(f.seek(SeekFrom::End(-(aux_bda_size))));
-        try!(f.write_all(&zeroed[..SECTOR_SIZE as usize]));
-        try!(f.write_all(buf));
-        try!(f.write_all(&zeroed[(SECTOR_SIZE * 2) as usize..]));
-        try!(f.flush());
-
-        Ok(())
-    }
-
-    pub fn save_state(&mut self, time: &Timespec, metadata: &[u8]) -> EngineResult<()> {
-        try!(self.write_mdax(time, metadata));
-        try!(self.write_sigblock());
-
-        Ok(())
     }
 
     /// Get the "x:y" device string for this blockdev
@@ -350,20 +228,18 @@ impl BlockDev {
         self.dev.dstr()
     }
 
-    /// Size of the BDA copy at the beginning of the blockdev
-    fn main_bda_size(&self) -> Sectors {
-        BDA_STATIC_HDR_SIZE + self.sigblock.mda_sectors + self.sigblock.reserved_sectors
-    }
+    pub fn save_state(&mut self, time: &Timespec, metadata: &[u8]) -> EngineResult<()> {
+        let mut f = try!(OpenOptions::new().write(true).open(&self.devnode));
+        try!(self.bda.save_state(time, metadata, &mut f));
 
-    /// Size of the BDA copy at the end of the blockdev
-    fn aux_bda_size(&self) -> Sectors {
-        BDA_STATIC_HDR_SIZE + self.sigblock.mda_sectors
+        Ok(())
     }
 
     /// List the available-for-upper-layer-use range in this blockdev.
     pub fn avail_range(&self) -> (SectorOffset, Sectors) {
-        let start = self.main_bda_size();
-        let length = self.sigblock.total_size - start - self.aux_bda_size();
+        let start = Sectors(BDA_STATIC_HDR_SECTORS) + self.bda.header.mda_size +
+                    self.bda.header.reserved_size;
+        let length = self.bda.header.blkdev_size - start;
         (SectorOffset(*start), length)
     }
 }
--- a/src/engine/strat_engine/blockdev.rs
+++ b/src/engine/strat_engine/blockdev.rs
@@ -87,7 +87,7 @@ pub fn find_all() -> EngineResult<BTreeMap<PoolUuid, BTreeMap<DevUuid, BlockDev>
             }
             Err(err) => {
                 let error_message = format!("{} for devnode {}", err, devnode.display());
-                return Err(EngineError::Stratis(ErrorEnum::Invalid(error_message)));
+                return Err(EngineError::Engine(ErrorEnum::Invalid, error_message));
             }
         };
 
@@ -97,7 +97,7 @@ pub fn find_all() -> EngineResult<BTreeMap<PoolUuid, BTreeMap<DevUuid, BlockDev>
             sigblock: match SigBlock::read(&buf, 0, Sectors(try!(blkdev_size(&f)) / SECTOR_SIZE)) {
                 Ok(sigblock) => sigblock,
                 Err(err) => {
-                    return Err(EngineError::Stratis(ErrorEnum::Invalid(err)));
+                    return Err(EngineError::Engine(ErrorEnum::Invalid, err));
                 }
             },
         }))
@@ -159,7 +159,7 @@ pub fn initialize(pool_uuid: &PoolUuid,
             Ok(ownership) => ownership,
             Err(err) => {
                 let error_message = format!("{} for device {}", err, devnode.display());
-                return Err(EngineError::Stratis(ErrorEnum::Invalid(error_message)));
+                return Err(EngineError::Engine(ErrorEnum::Invalid, error_message));
             }
         };
 
@@ -185,14 +185,15 @@ pub fn initialize(pool_uuid: &PoolUuid,
                 let error_message = format!("{} too small, minimum {} bytes",
                                             devnode.display(),
                                             MIN_DEV_SIZE);
-                return Err(EngineError::Stratis(ErrorEnum::Invalid(error_message)));
+                return Err(EngineError::Engine(ErrorEnum::Invalid, error_message));
             };
             match ownership {
                 DevOwnership::Unowned => add_devs.push((dev, (devnode, dev_size))),
                 DevOwnership::Theirs => {
                     if !force {
                         let error_str = format!("First 4K of {} not zeroed", devnode.display());
-                        return Err(EngineError::Stratis(ErrorEnum::Invalid(error_str)));
+                        return Err(EngineError::Engine(ErrorEnum::Invalid, error_str));
+
                     } else {
                         add_devs.push((dev, (devnode, dev_size)))
                     }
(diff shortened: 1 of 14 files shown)


mulkieran:
I like "Remove engine::ErrorEnum::Ok..." and "Rename ErrorEnum to DBusErrorEnum" just fine. AFAIAC, they can go in at any time.

I think that in the final form it would be nice if "InternalEngineError" had a commit  message.

I didn't look any further, 'cause this is preliminary, AFAIU.

agrover:
this code is ready for review @mulkieran @trgill but not merging.

As discussed in the meeting today, I was going to wait until #199 goes in and then add some tests for the blockdev stuff before merging. The ErrorEnum commits could go in earlier, if desired.

In any case, please take a look, thanks.

mulkieran:
I do not follow "Refactor engine::ErrorEnum to implement Error trait". I'm satisfied if it is the case that the errors reported on the D-Bus will still contain the kind of error and the specific device, filesystem, or other object involved in the error.

src/engine/strat_engine/blockdev.rs
```diff
@@ -3,36 +3,34 @@
...
 use super::engine::DevOwnership;
-
 pub use super::BlockDevSave;
 
 type DevUuid = Uuid;
 type PoolUuid = Uuid;
 
-const BDA_STATIC_HDR_SIZE: Sectors = Sectors(8);
+
+pub const BDA_STATIC_HDR_SECTORS: u64 = 16;
```
mulkieran: Why are these two public?

src/engine/strat_engine/blockdev.rs
```diff
@@ -76,30 +74,29 @@ pub fn find_all() -> EngineResult<BTreeMap<PoolUuid, BTreeMap<DevUuid, BlockDev>
                                format!("Could not open {}", devnode.display()))
             }));
 
-        let mut buf = [0u8; 4096];
+        let mut buf = [0u8; BDA_STATIC_HDR_SIZE as usize];
```
mulkieran: These casts are potential trouble wherever they happen. What it be any help to make these two constants usize?

src/engine/strat_engine/blockdev.rs
```diff
@@ -151,19 +148,19 @@ pub fn initialize(pool_uuid: &PoolUuid,
 
         let dev_size = try!(blkdev_size(&f));
 
-        let mut buf = [0u8; 4096];
+        let mut buf = [0u8; BDA_STATIC_HDR_SIZE as usize];
         try!(f.seek(SeekFrom::Start(0)));
```
mulkieran: It should be safe not to seek from 0, since the file was just opened.

src/engine/strat_engine/blockdev.rs
```diff
@@ -135,7 +132,7 @@ pub fn initialize(pool_uuid: &PoolUuid,
 
     /// Gets device information, returns an error if problem with obtaining
     /// that information.
-    fn dev_info(dev: &Device) -> EngineResult<(PathBuf, u64, DevOwnership)> {
+    fn dev_info(dev: &Device) -> EngineResult<(PathBuf, u64, DevOwnership, File)> {
```
mulkieran: Please augment the comment to indicate the role of the various return values.

src/engine/strat_engine/blockdev.rs
```diff
@@ -211,28 +207,26 @@ pub fn initialize(pool_uuid: &PoolUuid,
         Ok(add_devs)
     }
 
-    match validate_mda_size(mda_size) {
-        None => {}
-        Some(err) => {
-            return Err(EngineError::Engine(ErrorEnum::Invalid, err));
-        }
-    };
+    try!(validate_mda_size(mda_size));
```
mulkieran: I don't think the temporary convenience is worth the change.

src/engine/strat_engine/blockdev.rs
```diff
@@ -172,30 +169,29 @@ pub fn initialize(pool_uuid: &PoolUuid,
...
                                             MIN_DEV_SIZE);
                 return Err(EngineError::Engine(ErrorEnum::Invalid, error_message));
             };
             match ownership {
-                DevOwnership::Unowned => add_devs.push((dev, (devnode, dev_size))),
+                DevOwnership::Unowned => add_devs.push((dev, (devnode, dev_size, f))),
                 DevOwnership::Theirs => {
                     if !force {
-                        let error_str = format!("First 4K of {} not zeroed", devnode.display());
+                        let error_str = format!("First 8K of {} not zeroed", devnode.display());
```
mulkieran: It would be nice if this 8K were dependent on some constant, and not just a string, but I don't know that it is that important.

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,365 @@
...
 
-use consts::SECTOR_SIZE;
-
 use types::Sectors;
-use types::SectorOffset;
-
 use super::engine::DevOwnership;
+use consts::SECTOR_SIZE;
+use engine::{EngineResult, EngineError, ErrorEnum};
+use super::blockdev::BDA_STATIC_HDR_SIZE;
```
mulkieran: Don't know why BDA_STATIC_HDR_SIZE isn't defined in metadata and used in blockdev.

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,365 @@
...
-    pub fn write(&self, buf: &mut [u8], offset: usize) {
-        let (offset1, offset2, offset3, offset4) = MDA::offsets(offset);
-
-        let last_updated = self.last_updated;
-        LittleEndian::write_u64(&mut buf[offset..offset1], last_updated.sec as u64);
-        LittleEndian::write_u32(&mut buf[offset1..offset2], last_updated.nsec as u32);
-
-        LittleEndian::write_u32(&mut buf[offset2..offset3], self.used);
-        LittleEndian::write_u32(&mut buf[offset3..offset4], self.crc);
+impl BDA {
```
mulkieran: This could use some docs.

I am happy that it got moved out of the blockdev module.

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,365 @@
...
+        Ok(BDA {
+            header: header,
+            regions: regions,
+        })
     }
 
-    /// Read recorded values into buffer at buf_offset.
-    pub fn read(buf: &[u8], buf_offset: usize, offset: SectorOffset) -> MDA {
-        let (offset1, offset2, offset3, offset4) = MDA::offsets(buf_offset);
+    pub fn load(f: &mut File, header: StaticHeader) -> EngineResult<BDA> {
```
mulkieran: Header comment here would be helpful.

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,365 @@
...
-            Some(err) => {
-                return Err(err.into());
-            }
-        };
 
-        Ok(SigBlock {
-            mda: MDAGroup::read(buf, offset + 96, mda_size),
+        let mda_size = Sectors(LittleEndian::read_u64(&buf[96..104]) as u64);
+
+        try!(validate_mda_size(mda_size));
```
mulkieran: Again, probably a mistake.

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,365 @@
...
+            }
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct MDARegions {
+    // Spec defines 4 regions, but regions 2 & 3 are duplicates of 0 and 1 respectively
+    pub region_size: Sectors,
+    mdas: [MDAHeader; 2],
```
mulkieran: Maybe the 2 should be a constant, or a derived number.

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,365 @@
...
+        try!(f.flush());
+
+        Ok(MDARegions {
+            region_size: region_size,
+            mdas: [MDAHeader::new(), MDAHeader::new()],
+            latest_json: None,
+        })
+    }
+
+    pub fn load(header: &StaticHeader, f: &mut File) -> EngineResult<MDARegions> {
```
mulkieran: This method could definitely use a header comment.

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,365 @@
...
+#[derive(Debug, Clone)]
+pub struct MDARegions {
+    // Spec defines 4 regions, but regions 2 & 3 are duplicates of 0 and 1 respectively
+    pub region_size: Sectors,
+    mdas: [MDAHeader; 2],
+    latest_json: Option<Vec<u8>>,
+}
+
+impl MDARegions {
+    pub fn initialize(header: &StaticHeader, f: &mut File) -> EngineResult<MDARegions> {
```
mulkieran: Header comment helpful.

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,365 @@
...
+                                                       "MDA region json CRC".into()));
+                    }
+                    Some(json_buf)
+                }
+            };
+
+            Ok((mda, json_buf))
+        };
+
+        let (mda0, json0) = load_region(0)
```
mulkieran: Should you do a repair or something if you fail on one of the two copies?

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,365 @@
...
+
+impl MDAHeader {
+    pub fn new() -> MDAHeader {
+        MDAHeader {
+            last_updated: Timespec::new(0, 0),
+            used: 0,
+        }
+    }
+
+    pub fn from_buf(buf: &[u8; MDA_REGION_HDR_SIZE]) -> EngineResult<(MDAHeader, u32)> {
```
mulkieran: This could definitely use a header, if only to clarify what the second value returned in the result is.

mulkieran:
Generally "Update blockdev..." commit seems like it's going in the right direction.

src/engine/strat_engine/blockdev.rs
```diff
@@ -55,10 +56,67 @@ pub fn resolve_devices(paths: &[&Path]) -> io::Result<BTreeSet<Device>> {
...
+impl BlockDevs {
+    pub fn new(devs: BTreeMap<DevUuid, BlockDev>) -> BlockDevs {
+        BlockDevs { devs: devs }
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.devs.is_empty()
+    }
+
+    /// Read the latest json across all blockdevs
```
mulkieran: "json" is the format of the data you're reading, not its meaning or significance.

src/engine/strat_engine/blockdev.rs
```diff
@@ -55,10 +56,67 @@ pub fn resolve_devices(paths: &[&Path]) -> io::Result<BTreeSet<Device>> {
...
+        for bd in bds.iter().rev() {
+            let mut f = match OpenOptions::new()
+                .read(true)
+                .open(&bd.devnode) {
+                Ok(f) => f,
+                Err(_) => continue,
+            };
+            match bd.bda.load_state(&mut f) {
+                Ok(Some(json)) => return Some(json),
+                _ => continue,
```
mulkieran: It seems troubling that this continues in the face of so many errors without any indication of the error. Perhaps the return type should be enriched. Especially in the second match it seems like the distinction between ```Ok(None)``` and ```Err(...)``` is kind of significant.

src/engine/strat_engine/blockdev.rs
```diff
@@ -55,10 +56,67 @@ pub fn resolve_devices(paths: &[&Path]) -> io::Result<BTreeSet<Device>> {
     Ok(devices)
 }
 
+#[derive(Debug)]
+pub struct BlockDevs {
```
mulkieran: I think that BlockDevs could probably be better implemented as a module then a struct + impl.

src/engine/strat_engine/metadata.rs
```diff
@@ -74,14 +74,22 @@ impl BDA {
...
+        self.regions.save_state(time, metadata, &mut f)
+    }
 
-        Ok(())
+    /// Read latest metadata from the disk
+    pub fn load_state(&self, mut f: &mut File) -> EngineResult<Option<Vec<u8>>> {
+        self.regions.load_state(&mut f)
+    }
+
+    pub fn last_updated(&self) -> &Option<Timespec> {
```
mulkieran: I think the name could be improved to something like ```last_update_time```. Otherwise, it sounds a bit like you're locating the last updated thing.

src/engine/strat_engine/metadata.rs
```diff
@@ -186,8 +194,7 @@ impl StaticHeader {
 pub struct MDARegions {
     // Spec defines 4 regions, but regions 2 & 3 are duplicates of 0 and 1 respectively
     pub region_size: Sectors,
-    mdas: [MDAHeader; 2],
-    latest_json: Option<Vec<u8>>,
+    pub mdas: [MDAHeader; 2],
```
mulkieran: Still in favor of giving a name to that 2.

src/engine/strat_engine/metadata.rs
```diff
@@ -205,64 +212,46 @@ impl MDARegions {
 
         Ok(MDARegions {
             region_size: region_size,
-            mdas: [MDAHeader::new(), MDAHeader::new()],
-            latest_json: None,
+            mdas: [MDAHeader::new(per_region_size), MDAHeader::new(per_region_size)],
```
mulkieran: Do you really need a ```per_region_size``` for every header. I think they are all the same.

src/engine/strat_engine/metadata.rs
```diff
@@ -205,64 +212,46 @@ impl MDARegions {
...
             try!(f.seek(SeekFrom::Start(offset)));
             try!(f.read_exact(&mut hdr_buf));
+            let mda = try!(MDAHeader::from_buf(&hdr_buf, per_region_size));
 
-            let (mda, json_crc) = try!(MDAHeader::from_buf(&hdr_buf));
-            if mda.used > *region_size * SECTOR_SIZE {
-                return Err(EngineError::Engine(ErrorEnum::Invalid,
-                                               "mda0.used > region_size".into()));
-            }
+            // Loading checks json CRC
```
mulkieran: Please say the name of the thing, not its format.

src/engine/strat_engine/metadata.rs
```diff
@@ -205,64 +212,46 @@ impl MDARegions {
...
-            .unwrap_or_else(|_| load_region(2).unwrap_or_else(|_| (MDAHeader::new(), None)));
-        let (mda1, json1) = load_region(1)
-            .unwrap_or_else(|_| load_region(3).unwrap_or_else(|_| (MDAHeader::new(), None)));
-
-        let latest_json = match mda0.last_updated.cmp(&mda1.last_updated) {
-            Ordering::Less => json0,
-            Ordering::Greater => json1,
-            Ordering::Equal => json0,
-        };
+        let mda0 =
```
mulkieran: Still wondering if you should do some fixing up if you fail on one of the two copies.

src/engine/strat_engine/metadata.rs
```diff
@@ -345,6 +371,29 @@ impl MDAHeader {
...
     }
+
+    /// Given a pre-seek()ed File, load the MDA region and return the contents
+    // MDAHeader cannot seek because it doesn't know which region it's in
+    pub fn load_region(&self, f: &mut File) -> EngineResult<Option<Vec<u8>>> {
+        let mut hdr_buf = [0u8; MDA_REGION_HDR_SIZE];
+        try!(f.read_exact(&mut hdr_buf));
+
+        if self.used > self.region_size {
+            return Err(EngineError::Engine(ErrorEnum::Invalid, "mda0.used > region_size".into()));
```
mulkieran: Why does error message specify "mda0"? Couldn't it be any?

mulkieran:
I don't understand the commit message for "Use Cow...".

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,416 @@
...
-impl MDA {
-    pub fn new(offset: SectorOffset) -> MDA {
-        MDA {
-            crc: 0,
-            last_updated: Timespec::new(0, 0),
-            used: 0,
-            offset: offset,
-        }
+impl BDA {
+    pub fn initialize(mut f: &mut File, header: StaticHeader) -> EngineResult<BDA> {
```
mulkieran: Method header would be helpful.

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,416 @@
...
+        Ok(BDA {
+            header: header,
+            regions: regions,
+        })
     }
 
-    /// Write out recorded values to buf at offset.
-    pub fn write(&self, buf: &mut [u8], offset: usize) {
-        let (offset1, offset2, offset3, offset4) = MDA::offsets(offset);
+    // Wipe metadata
```
mulkieran: This comment doesn't add much.

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,416 @@
...
+    }
 
-        MDA {
-            last_updated: Timespec::new(LittleEndian::read_u64(&buf[buf_offset..offset1]) as i64,
-                                        LittleEndian::read_u32(&buf[offset1..offset2]) as i32),
-            used: LittleEndian::read_u32(&buf[offset2..offset3]),
-            crc: LittleEndian::read_u32(&buf[offset3..offset4]),
-            offset: offset,
-        }
+    pub fn last_updated(&self) -> &Option<Timespec> {
```
mulkieran: I think the name ```last_updated``` is a bit misleading, both for the name of the method and of the field in the struct. ```last_update_time``` or ```most_recent_update_time``` maybe?

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,416 @@
...
+
+        let pool_uuid = try!(Uuid::parse_str(from_utf8(&buf[32..64]).unwrap())
+            .map_err(|_| EngineError::Engine(ErrorEnum::Invalid, "invalid pool uuid".into())));
+        let dev_uuid = try!(Uuid::parse_str(from_utf8(&buf[64..96]).unwrap())
+            .map_err(|_| EngineError::Engine(ErrorEnum::Invalid, "invalid dev uuid".into())));
+
+
+        let mda_size = Sectors(LittleEndian::read_u64(&buf[96..104]) as u64);
+
+        try!(validate_mda_size(mda_size));
```
mulkieran: Changing the return type of ```validate_mda_size``` is convenient, but probably wrong.

src/engine/strat_engine/blockdev.rs
```diff
@@ -172,197 +227,102 @@ pub fn initialize(pool_uuid: &PoolUuid,
...
         Ok(add_devs)
     }
 
-    match validate_mda_size(mda_size) {
-        None => {}
-        Some(err) => {
-            return Err(EngineError::Stratis(ErrorEnum::Invalid(err)));
-        }
-    };
+    try!(validate_mda_size(mda_size));
```
mulkieran: Again, the convenience of changing the signature is probably not worth it.

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,416 @@
...
-    pub total_size: Sectors,
-    pub pool_uuid: Uuid,
-    pub dev_uuid: Uuid,
-    pub mda: MDAGroup,
-    pub mda_sectors: Sectors,
-    pub reserved_sectors: Sectors,
+pub struct MDARegions {
+    // Spec defines 4 regions, but regions 2 & 3 are duplicates of 0 and 1 respectively
+    pub region_size: Sectors,
+    pub mdas: [MDAHeader; 2],
```
mulkieran: This literal 2 should have a name and be derived from NUM_MDA_REGIONS or possibly NUM_MDA_REGIONS should be derived from the name for this 2.

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,416 @@
...
+        for region in 0..NUM_MDA_REGIONS {
+            try!(f.seek(SeekFrom::Start(BDA_STATIC_HDR_SIZE + (region as u64 * per_region_size))));
+            try!(f.write_all(&hdr_buf));
+        }
+
+        try!(f.flush());
+
+        Ok(MDARegions {
+            region_size: region_size,
+            mdas: [MDAHeader::new(per_region_size), MDAHeader::new(per_region_size)],
```
mulkieran: Seems wierd to me that each header has a ```per_region_size``` attribute, since I think they are the same for all headers. But it does look more convenient.

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,416 @@
...
-               -> SigBlock {
-        SigBlock {
-            mda: MDAGroup::new(mda_size),
-            pool_uuid: pool_uuid.clone(),
-            dev_uuid: dev_uuid.clone(),
-            total_size: total_size,
-            mda_sectors: mda_size,
-            reserved_sectors: MDA_RESERVED_SIZE,
+impl MDARegions {
+    pub fn initialize(header: &StaticHeader, f: &mut File) -> EngineResult<MDARegions> {
```
mulkieran: Method header doc would be helpful.

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,416 @@
...
+        let mda = &self.mdas[newer_region];
+
+        let mut load_json = |region| {
+            let offset = BDA_STATIC_HDR_SIZE + (region as u64 * *self.region_size * SECTOR_SIZE) +
+                         MDA_REGION_HDR_SIZE as u64;
+            try!(f.seek(SeekFrom::Start(offset)));
+            mda.load_region(f)
+        };
+
+        Ok(load_json(newer_region)
```
mulkieran: Should you be doing some repairs here if you fail on the first of the two regions?

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,416 @@
...
         }
+    }
+}
 
-        let pool_uuid = try!(Uuid::parse_str(from_utf8(&buf[offset + 32..offset + 64]).unwrap())
-            .map_err(|_| "invalid pool uuid"));
-        let dev_uuid = try!(Uuid::parse_str(from_utf8(&buf[offset + 64..offset + 96]).unwrap())
-            .map_err(|_| "invalid dev uuid"));
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub struct MDAHeader {
```
mulkieran: Have you tried not deriving Eq and PartialEq here. I don't think they're used, and they do not seem like a good idea.

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,416 @@
...
+            try!(f.read_exact(&mut hdr_buf));
+            let mda = try!(MDAHeader::from_buf(&hdr_buf, per_region_size));
+
+            // Loading checks json CRC
+            try!(mda.load_region(f));
+
+            Ok(mda)
+        };
+
+        let mda0 =
```
mulkieran: I'm wondering about repairs here, too, if first member of pair fails.

mulkieran:
Ok. First pass completed. Except for the introduction of the BlockDevs struct and the change in the return type of ```validate_mda_size()```, everything looks like it is headed in a good direction.

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,416 @@
...
+            json_crc: LittleEndian::read_u32(&buf[4..8]),
         })
     }
 
-    /// Read from Stratis magic number area in buf at offset.
-    pub fn read_strat_magic(buf: &[u8], offset: usize) -> &[u8] {
-        &buf[offset + 4..offset + 20]
-    }
+    pub fn to_buf(data: &[u8], timestamp: &Timespec) -> [u8; MDA_REGION_HDR_SIZE] {
+        let mut buf = [0u8; MDA_REGION_HDR_SIZE];
```
mulkieran: It feels like you ought to be setting the crc for the header itself in this method.

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,365 @@
...
 /// Validate MDA size
 /// Return None if MDA size is fine, otherwise a message.
-pub fn validate_mda_size(size: Sectors) -> Option<String> {
-    if *size % NUM_MDA_COPIES != 0 {
-        let error_message = format!("MDA size {} is not divisible by number of copies \
-                                    required {}",
-                                    *size,
-                                    NUM_MDA_COPIES);
-        return Some(error_message);
+pub fn validate_mda_size(size: Sectors) -> EngineResult<()> {
```
agrover: How so? The design doc defines two conditions for a valid mda size: it must be above a certain minimum and it needs to be divisible by 4. I am of the opinion that if an invalid value is passed to `BlockDev::initialize()` or read from the disk in `StaticHeader::from_buf`, something is seriously wrong and we should return an `EngineError`. This allows the rest of the code to assume the mda size is valid, which lets us for example calculate mda region starting offsets.

mulkieran: I agree with you entirely about the behavior of ```BlockDev::initialize()``` and ```StaticHeader::from_buf```. I'm just saying that you should achieve this behavior in a slightly more cumbersome fashion. But I think the whole issue was looming a little too large in my mind yesterday. Feel free to ignore.

src/engine/strat_engine/metadata.rs
mulkieran: Ignore this one, too!

src/engine/strat_engine/blockdev.rs
mulkieran: Ignore!

src/engine/strat_engine/blockdev.rs
agrover: If we do this then there are other spots where we'd end up casting the other way ☹️ . I think conceptually `u64` is better because we're dealing with on-disk values that are not variable across 32 and 64bit CPU archs. But yeah when creating arrays these casts are ugly and get tiresome.

src/engine/strat_engine/blockdev.rs
agrover: 👍

src/engine/strat_engine/blockdev.rs
agrover: 👍

src/engine/strat_engine/blockdev.rs
agrover: I see how we could pull out BlockDevs struct & impl into a separate module, but I don't see how we could avoid the struct + impl altogether. Explain a little more?

I also could see how `{read,write}_metadata()` could be `StratPool` methods...I think creating `BlockDevs` was in anticipation of `StratPool` getting really big and wanting to break it up (this is what happened with Froyo), but since this isn't the case yet, it's probably premature. I'll just put these in `StratPool` and drop `BlockDevs`.

src/engine/strat_engine/blockdev.rs
agrover: What would we do differently? We could log something if `Err` but the real problem arises if there are *no* good copies to be found. I could see us getting fancy and writing more copies if we didn't succeed getting enough, but that is an enhancement. I'll add a TODO.

I also am going to change this so we only look at metadata with the latest timestamp. Anything previous is out of date and pretty much useless...

src/engine/strat_engine/metadata.rs
agrover: 👍

src/engine/strat_engine/metadata.rs
agrover: 👍

src/engine/strat_engine/metadata.rs
agrover: yes this is why.

src/engine/strat_engine/metadata.rs
agrover: I've opened https://github.com/stratis-storage/stratisd/issues/210, there are a few places we could repair stuff, but it's probably good to think a bit more before committing to this.

src/engine/strat_engine/metadata.rs
agrover: leftover, removed, 👍

src/engine/strat_engine/metadata.rs
agrover: oops, yes! 👍

agrover:
I've dropped the Cow patch (too unrelated), made changes based on review feedback, and have gone ahead and hooked things up so it actually should save state to blockdevs in json format now, as well.

src/engine/strat_engine/pool.rs
```diff
@@ -54,6 +56,56 @@ impl StratPool {
             redundancy: redundancy,
         })
     }
+
+    /// Read the latest data across all blockdevs
```
mulkieran: Suggested alternative method header comment:

```
Read most recent pool metadata from a block device in the pool.
Returns None if no metadata obtained.

File: src/engine/strat_engine/pool.rs [hunk 39]
```diff
@@ -54,6 +56,56 @@ impl StratPool {
...
+            };
+            match bd.bda.load_state(&mut f) {
+                Ok(Some(data)) => return Some(data),
+                _ => continue,
+            }
+        }
+        None
+    }
+
+    /// Write the given data to all blockdevs
```
mulkieran: Suggested method comment:

```
Write pool metadata to all block devices in the pool.
```

src/engine/strat_engine/pool.rs
```diff
@@ -54,6 +56,56 @@ impl StratPool {
             redundancy: redundancy,
         })
     }
+
+    /// Read the latest data across all blockdevs
+    pub fn read_metadata(&self) -> Option<Vec<u8>> {
```
mulkieran: Suggested alternative method header comment:

```
Read most recent pool metadata from a block device in the pool.
Returns None if no metadata obtained.

File: src/engine/strat_engine/pool.rs [hunk 41]
```diff
@@ -101,11 +106,22 @@ impl StratPool {
         let time = time::now().to_timespec();
 
         for (_, bd) in &mut self.block_devs {
-            bd.save_state(&time, data).unwrap(); // ignoring failure
+            bd.save_state(&time, data.as_bytes()).unwrap(); // ignoring failure
```
mulkieran: Please change this comment to a TODO or something. As it stands it is deeply ambiguous. I'm assuming that you're against just panicking and crashing if saving the metadata fails on a block device, but it is actually not obvious from the comment.

src/engine/strat_engine/pool.rs
```diff
@@ -127,6 +143,7 @@ impl Pool for StratPool {
         let mut bds = try!(initialize(&self.pool_uuid, devices, MIN_MDA_SIZE, force));
         let bdev_paths = bds.iter().map(|p| p.1.devnode.clone()).collect();
         self.block_devs.append(&mut bds);
+        try!(self.write_metadata());
```
mulkieran: Should you be doing the same for ```add_cachedevs```? Making ```add_cachedevs``` unimplemented would be fine with me, too.

src/engine/strat_engine/blockdev.rs
```diff
@@ -135,7 +108,10 @@ pub fn initialize(pool_uuid: &PoolUuid,
 
     /// Gets device information, returns an error if problem with obtaining
     /// that information.
-    fn dev_info(dev: &Device) -> EngineResult<(PathBuf, u64, DevOwnership)> {
+    /// Returns a tuple with the blockdev's path, its size in bytes,
+    /// its ownership as determined by calling determine_ownership(),
+    /// and an open File handle, all of which are needed later.
+    fn dev_info(dev: &Device) -> EngineResult<(PathBuf, u64, DevOwnership, File)> {
```
mulkieran: Do we have to worry about limits on number of opened files?

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,450 @@
...
-
 use super::engine::DevOwnership;
-
-const MDA_OFFSETS: [usize; 4] = [8, 12, 16, 20];
-const MDA_RESERVED_SIZE: Sectors = Sectors(2048 * 3); // = 3 MiB
+use consts::{SECTOR_SIZE, MEGA};
+use engine::{EngineResult, EngineError, ErrorEnum};
+
+pub const BDA_STATIC_HDR_SECTORS: u64 = 16;
+pub const BDA_STATIC_HDR_SIZE: u64 = BDA_STATIC_HDR_SECTORS * SECTOR_SIZE;
```
mulkieran: I think ```BDA_STATIC_HDR_SIZE``` does not need to be public.

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,450 @@
...
-                }
-            }
+    /// Given a pre-seek()ed File, load the MDA region and return the contents
+    // MDAHeader cannot seek because it doesn't know which region it's in
+    pub fn load_region(&self, f: &mut File) -> EngineResult<Option<Vec<u8>>> {
+        let mut hdr_buf = [0u8; MDA_REGION_HDR_SIZE];
+        try!(f.read_exact(&mut hdr_buf));
+
+        if self.used > self.region_size {
+            return Err(EngineError::Engine(ErrorEnum::Invalid, "mda0.used > region_size".into()));
```
mulkieran: This error message could probably use some bulking up.

mulkieran:
Second pass completed.

mulkieran:
Btw, a bunch of the metadata code lends itself to QuickCheck style tests. I can do these after PR gets pushed. They will not take a lot of time unless they find bugs and will help catch irritating and time consuming regressions in future.

mulkieran:
```
[mulhern@dhcp-25-209 rustproject]$ cargo run
   Compiling rustproject v0.1.0 (file:///home/mulhern/my-projects/rustproject)
    Finished debug [unoptimized + debuginfo] target(s) in 0.13 secs
     Running `target/debug/rustproject`
u64 max value 18446744073709551615
u64 max value as u32 4294967295
u32 max value 4294967295
u32 max value as u64 4294967295
```

This should illustrate why I would rather be casting a usize (which could be u32 or u64) to a u64 than the other way around.

src/engine/strat_engine/blockdev.rs
agrover: maybe eventually but I think that's beyond the scope of this PR.

src/engine/strat_engine/metadata.rs
agrover: 👍

src/engine/strat_engine/metadata.rs
agrover: 👍

agrover:
> This should illustrate why I would rather be casting a usize (which could be u32 or u64) to a u64 than the other way around.

Sorry, sleep deprived, can you spell it out for me?

agrover:
> Still not quite following what "Refactor engine::ErrorEnum to implement Error trait " does. All I really care about is that messages can still convey both the category of the problem and the specific thing involved after this change.

Yeah it's minor. Our errors still do those things. Supporting `Error` just makes handling errors slightly easier (you can just call description() or cause() instead of matching.) It also might make life easier when we check out error_chain at some point, don't know.

src/engine/strat_engine/metadata.rs
```diff
@@ -4,262 +4,454 @@
...
+        let mut buf = [0u8; MDA_REGION_HDR_SIZE];
 
-    /// Read pool uuid from buf at offset.
-    /// Return an error if the data can't be interpreted as a Uuid.
-    pub fn read_pool_uuid(buf: &[u8], offset: usize) -> Result<Uuid, String> {
-        Uuid::parse_str(from_utf8(&buf[offset + 32..offset + 64]).unwrap())
-            .map_err(|_| "invalid pool uuid".into())
+        LittleEndian::write_u32(&mut buf[4..8], crc32::checksum_ieee(data));
+        LittleEndian::write_u64(&mut buf[8..16], data.len() as u64);
+        LittleEndian::write_u64(&mut buf[16..24], timestamp.sec as u64);
```
mulkieran: The timestamp writes should use signed integer methods, i.e., ```write_i64``` and ```write_i32```.

mulkieran:
All casts of u64 to usize are mistakes because it is impossible to guarantee that ```x == x as usize```, unless:

1. The program always runs on 64 bit machines OR
2. It can be proven that wherever such a cast occurs the value of x is no greater than the maximum value of a u32.

I just ran over the code, and (2) seems to be mostly true, but these downcasts are still a bug waiting to happen. Feel free to defer consideration of this problem until a later PR. I will start an issue for it.

src/engine/strat_engine/metadata.rs
agrover: These fields are defined in the design doc as unsigned. I'm not sure why the sec and nsec fields in `time::Timespec` are signed -- it doesn't really have an impact based on their usage -- but unless we decide to change the design doc I believe use of unsigned write methods is correct.

mulkieran: Then the read methods should be changed to unsigned as well, so the operations are inverses of each other, as:

```
let time = {
    match LittleEndian::read_u64(&buf[62..24]) {
      0 => None,
      # signed cast are always safe, as highest order bit of each value read is guaranteed to be 0.
      secs => Some(Timespec::new(secs as i64, LittleEndian::read_u32(&buf[24..28]) as i32),
    }
};
```

It would be nice to comment writes as well, as:

```
# unsigned casts are always safe, as sec and nsec values are never negative
```

Otherwise it looks too much like a bug.

mulkieran: And, to allay all worries:

```
assert!(timestamp.ge(Timespec::new(0, 0));
```
in the to_buf() method, which is pretty cheap.

agrover:
Added commit to make timespec-related changes as suggested. Looking good?

mulkieran:
Go ahead: :+1: .

trgill:
:+1:

PR #209, opened 2017-01-17
Pin rustfmt version to 0.6.3.

mulkieran:
0.7.0 is actually deleting things at this time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 check: fmt build test
 
 ${HOME}/.cargo/bin/cargo-fmt:
-	cargo install rustfmt
+	cargo install rustfmt --vers 0.6.3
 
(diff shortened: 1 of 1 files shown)


mulkieran:
@agrover @trgill review please?

agrover:
This is breaking our CI currently, right? I'm going to go ahead and merge this.

mulkieran:
The two bugs are actually different:
1. The first bug was caused by a change in the underlying syntex library: https://github.com/rust-lang-nursery/rustfmt/issues/1229#issuecomment-273326447.
2. Our bug is probably a bug in rustfmt.

So, we need to wait until our bug is fixed, and until syntex 0.56.0 works its way through Cargo to Travis before we start using a newer format, which will probably be rustfmt 0.7.1.

PR #212, opened 2017-01-25
Changes to make dbus API a little easier to work with

mulkieran:


mulkieran:
@grover, @trgill review?

This is stuff from #193 that could just as well go in now.

Also, rebasing is getting to be boring.

There is a possibility that the ActionQueue can in future be improved so that the ```push_remove()``` method takes a Path and not a String, but the current implementation just preserves the old behavior.

Also, the ```drain()``` method doesn't take a Range, as that would require using an unstable feature, but since we always drain the queue entirely, that doesn't really matter.

src/dbus_api/types.rs
```diff
@@ -0,0 +1,121 @@
...
+
+#[derive(Default, Debug)]
+pub struct TData;
+impl DataType for TData {
+    type ObjectPath = DbusContext;
+    type Property = ();
+    type Interface = ();
+    type Method = ();
+    type Signal = ();
+}
```
mulkieran: I promise to add more stuff to it...in fact PR #193 already does, that stuff just isn't ready yet.

src/dbus_api/types.rs
```diff
@@ -119,3 +118,33 @@ impl DataType for TData {
...
+
+/// An action queue.
+/// Add and remove actions are pushed onto the queue.
+/// The queue can also be drained.
+#[derive(Debug)]
+pub struct ActionQueue {
+    queue: VecDeque<DeferredAction>,
+}
+
+impl ActionQueue {
```
mulkieran: I'm happy to leave the ActionQueue commit on the head of PR #193 instead of on the tail of this PR. I expect to be able to make it more valuable, but not until some other work is done.

agrover:
ok cool. I am 👍 to merge as is, then.

trgill:
:+1:  Ok.

Just a thought... When I moved stratisd.rs into bin, I was thinking we might want move the dbus stuff into bin.

mulkieran:
@trgill that might be a better spot. @agrover what do you think?

agrover:
I have no strong preference..

mulkieran:
rebased.

PR #216, opened 2017-01-26
Master self

mulkieran:


agrover:
(Commenting before asked for, I know.) I like the first commit, but I think what the second does is non-idiomatic. `Self` is for use in traits, when you don't know the type ([reference](http://rustbyexample.com/trait.html)). In an impl you do, and it's better to use the concrete type. (I think I introduced some non-trait usages of Self in #207, I would prefer we just correct those spots.)

mulkieran:
I'm happy to drop the commit. "Programming Rust" sort of recommends the practice generally, but it doesn't back up the recommendation at all.

agrover:
OK yeah, for now at least. Thanks.

mulkieran:
Superceded by #220.

PR #218, opened 2017-01-27
Linear Dev with Integration Test

trgill:
Added -- skip for test_lineardev_setup to Makefile
LinearDev uses DM linear to  concatenate of a list of BlockDevs into continuous sectors.
Integration test for LinearDev - the test takes a list of destroyable block devices and
concatenates them - it then checks the size of the resulting DM device to validate that it
is the sum of the sector ranges allocated on each block device.

### DIFF
--- a/src/engine/strat_engine/lineardev.rs
+++ b/src/engine/strat_engine/lineardev.rs
@@ -1,0 +1,72 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use devicemapper::{DM, DevId, DeviceInfo, DmFlags};
+
+use engine::{EngineError, EngineResult, ErrorEnum};
+use super::blockdev::BlockDev;
+use types::Sectors;
+
+pub struct LinearDev {
+    name: String,
+    dev_info: Option<DeviceInfo>,
+}
+
+/// support use of DM to concatenate blockdevs and creat e a
+/// /dev/mapper/<name> for use as continuous sectors.
+impl LinearDev {
+    pub fn new(name: &str) -> LinearDev {
+        LinearDev {
+            name: name.to_owned(),
+            dev_info: None,
+        }
+    }
+
+    /// Generate a Vec<> to be passed to DM.  The format of the Vec entries is:
+    /// <logical start sec> <length> "linear" /dev/xxx <start offset>
+    fn dm_table(&self, block_devs: &Vec<&BlockDev>) -> Vec<(u64, u64, String, String)> {
+        let mut table = Vec::new();
+        let mut logical_start_sector = Sectors(0);
+        for block_dev in block_devs {
+            let (start, length) = block_dev.avail_range();
+            let dstr = block_dev.dstr();
+            let line = (*logical_start_sector,
+                        *length,
+                        "linear".to_owned(),
+                        format!("{} {}", dstr, *start));
+            debug!("dmtable line : {:?}", line);
+            table.push(line);
+            logical_start_sector = logical_start_sector + length;
+        }
+
+        table
+    }
+
+    /// Use DM to concatenate a set of blockdevs together into a
+    /// /dev/mapper/xxx block device of continuous sectors.
+    pub fn concat(&mut self, dm: &DM, block_devs: &Vec<&BlockDev>) -> EngineResult<()> {
+
+        try!(dm.device_create(&self.name, None, DmFlags::empty()));
+        let table = self.dm_table(block_devs);
+        let id = &DevId::Name(&self.name);
+        self.dev_info = Some(dm.table_load(id, &table).unwrap());
+
+        try!(dm.device_suspend(id, DmFlags::empty()));
+
+        return Ok(());
+    }
+
+    pub fn name(&self) -> EngineResult<&str> {
+        match self.dev_info {
+            Some(ref di) => return Ok(di.name().clone()),
+            None => return Err(EngineError::Engine(ErrorEnum::Invalid, "No dev_info".into())),
+        }
+    }
+
+    pub fn teardown(&self, dm: &DM) -> EngineResult<()> {
+        try!(dm.device_remove(&DevId::Name(&self.name), DmFlags::empty()));
+
+        Ok(())
+    }
+}
--- a/src/engine/strat_engine/lineardev.rs
+++ b/src/engine/strat_engine/lineardev.rs
@@ -50,7 +50,7 @@ impl LinearDev {
         try!(dm.device_create(&self.name, None, DmFlags::empty()));
         let table = self.dm_table(block_devs);
         let id = &DevId::Name(&self.name);
-        self.dev_info = Some(dm.table_load(id, &table).unwrap());
+        self.dev_info = Some(try!(dm.table_load(id, &table)));
 
         try!(dm.device_suspend(id, DmFlags::empty()));
 
--- a/src/types.rs
+++ b/src/types.rs
@@ -2,19 +2,19 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-use std::io;
-use std::fmt;
-use std::error::Error;
-use std::borrow::Cow;
-use std::fmt::Display;
-use std::ops::{Div, Mul, Rem};
-
-use nix;
-use term;
-use dbus;
-use serde;
 
 use consts::SECTOR_SIZE;
+use dbus;
+
+use nix;
+use serde;
+use std::borrow::Cow;
+use std::error::Error;
+use std::fmt;
+use std::fmt::Display;
+use std::io;
+use std::ops::{Div, Mul, Rem};
+use term;
 
 pub type StratisResult<T> = Result<T, StratisError>;
 
@@ -147,7 +147,6 @@ impl serde::Deserialize for Sectors {
         Ok(Sectors(val))
     }
 }
-
 
 // An error type for errors generated within Stratis
 //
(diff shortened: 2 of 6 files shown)


trgill:

I think when @mulkieran makes updates to SectorOffset we should update some of this code to use the updates.  But, we we may want to merge for now.

trgill:
going to put this in after #221

tests/util/blockdev_utils.rs
```diff
@@ -37,14 +47,62 @@ pub fn clean_blockdev_headers(blockdev_paths: &Vec<&Path>) {
...
-                panic!("Failed to clean signature on {:?} : {:?}", path, err);
+            Err(e) => {
+                panic!("Failed to clean signature on {:?} : {:?}", path, e);
             }
         }
     }
     info!("devices cleaned for test");
 }
 
+pub fn get_size(path: &Path) -> TestResult<u64> {
```
mulkieran: ```get_size()``` should probably return Sectors.

tests/lineardev_tests.rs
```diff
@@ -0,0 +1,170 @@
...
+        Ok(blockdev_map) => blockdev_map,
+        Err(e) => {
+            let message = format!("Failed to initialize starting set of blockdevs {:?}", e);
+            return Err(Framework(Error(message)));
+        }
+    };
+
+    let name = "stratis_testing_linear";
+
+    let mut blockdev_vec = Vec::new();
```
mulkieran: This would go better with ```from_iter()```, like:

```let blockdev_vec = Vec::from_iter(blockdev_map.values());``` roughly.

File: tests/lineardev_tests.rs [hunk 3]
```diff
@@ -0,0 +1,170 @@
...
+
+    let linear_dev = match concat_blockdevs(dm, &name, &blockdev_vec) {
+        Ok(dev) => dev,
+        Err(e) => {
+            let message = format!("Failed to concat_blockdevs {:?}", e);
+            return Err(Framework(Error(message)));
+        }
+    };
+
+    match validate_sizes(&name, &blockdev_vec) {
```
mulkieran: This won't pass type-checker, given return type of ```validate_size()```.

mulkieran:
Whoops! I thought it was review-final.

trgill:
I removed it while talking to @agrover. I'll address your feedback and rebase on top of your changes for Sector/SectorOffset.

src/engine/strat_engine/lineardev.rs
```diff
@@ -0,0 +1,66 @@
...
+    pub fn new_lineardev(name: &str) -> LinearDev {
+        LinearDev {
+            name: name.to_owned(),
+            dev_info: None,
+        }
+    }
+
+    fn dm_table(&self, block_devs: &Vec<BlockDev>) -> Vec<(u64, u64, String, String)> {
+        let mut table = Vec::new();
+        let mut logical_start_sector = 0;
```
trgill: yeah, this works.

only quick is line 33 can't dereference start due to the return type.  dstr and start need to be combined into a single string.  the table is passed to the DM bindnig and that is the type of the input.  Ok if we deal with this (if needed) in a later PR?

src/engine/strat_engine/lineardev.rs
```diff
@@ -0,0 +1,66 @@
...
+
+        table
+    }
+
+    pub fn concat(&mut self, dm: &DM, block_devs: &Vec<BlockDev>) -> EngineResult<()> {
+
+        try!(dm.device_create(&self.name, None, DmFlags::empty()));
+        let table = self.dm_table(block_devs);
+        let id = &DevId::Name(&self.name);
+        self.dev_info = Some(dm.table_load(id, &table).unwrap());
```
trgill: you are concerned that DM could be left in a bad state?  Or something not cleaned up?

src/types.rs
```diff
@@ -65,7 +65,11 @@ custom_derive! {
     pub struct SectorOffset(pub u64);
 }
 
-
+impl fmt::Display for SectorOffset {
```
trgill: Ok

src/engine/strat_engine/lineardev.rs
mulkieran: If you dereference start what you get is a u64 which ```format!``` should be able to handle fine, I would think.

src/engine/strat_engine/lineardev.rs
mulkieran: ```table_load()``` is a device mapper method which I don't know much about, but which isn't simple. I'm just wondering if we should be handling the failure, somehow.

src/engine/strat_engine/lineardev.rs
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+
+    fn dm_table(&self, block_devs: &Vec<BlockDev>) -> Vec<(u64, u64, String, String)> {
+        let mut table = Vec::new();
+        let mut logical_start_sector = 0;
+        for block_dev in block_devs {
+            let (start, length) = block_dev.avail_range();
+            let dstr = block_dev.dstr();
+            let line = (logical_start_sector,
```
mulkieran: Somewhere in the device mapper docs in the kernel it says that the very first logical start sector in a table must always be 0.

I'm not sure that this code would cause that to happen, although devicemapper-rs's ```load_table()``` might.

src/engine/strat_engine/lineardev.rs
agrover: Other dm calls are wrapped in `try!`, that seems appropriate here too, yes?

tests/lineardev_tests.rs
trgill: Ok

agrover:
👍

src/engine/strat_engine/lineardev.rs
```diff
@@ -0,0 +1,66 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use devicemapper::{DM, DevId, DeviceInfo, DmFlags};
+
+use engine::{EngineError, EngineResult, ErrorEnum};
+use super::blockdev::BlockDev;
+use types::Sectors;
+
+pub struct LinearDev {
+    name: String,
```
mulkieran: What is this name?

Generally speaking, method header comment for the implementation of LinearDev would be a great help.

tests/util/blockdev_utils.rs
```diff
@@ -37,14 +43,30 @@ pub fn clean_blockdev_headers(blockdev_paths: &Vec<&Path>) {
...
-                panic!("Failed to clean signature on {:?} : {:?}", path, err);
+            Err(e) => {
+                panic!("Failed to clean signature on {:?} : {:?}", path, e);
             }
         }
     }
     info!("devices cleaned for test");
 }
 
+#[allow(dead_code)]
```
mulkieran: Is this really still dead code? Can you remove this?

src/engine/strat_engine/lineardev.rs
```diff
@@ -0,0 +1,66 @@
...
+    dev_info: Option<DeviceInfo>,
+}
+
+impl LinearDev {
+    pub fn new_lineardev(name: &str) -> LinearDev {
```
mulkieran: Might as well just call this function ```new```, it's what we do elsewhere.

tests/lineardev_tests.rs
```diff
@@ -0,0 +1,177 @@
...
+use util::blockdev_utils::get_size;
+use util::test_config::TestConfig;
+use util::test_consts::DEFAULT_CONFIG_FILE;
+use util::test_result::TestError::Framework;
+use util::test_result::TestErrorEnum::Error;
+use util::test_result::TestResult;
+
+use uuid::Uuid;
+
+fn concat_blockdevs(dm: &DM, name: &str, block_devs: &Vec<&BlockDev>) -> TestResult<LinearDev> {
```
mulkieran: Method header comments throughout would be helpful.

tests/lineardev_tests.rs
```diff
@@ -0,0 +1,177 @@
...
+    let mut linear_sectors = Sectors(0);
+
+    for blockdev in block_devs {
+        let (_start_sector, length) = blockdev.avail_range();
+        linear_sectors = linear_sectors + length;
+    }
+
+    debug!("available linear space = {} sectors", linear_sectors);
+
+    // The /dev/mapper/<name> device is not immediately available for use.
```
mulkieran: Is there anything more effective to do than just to wait.

src/engine/strat_engine/lineardev.rs
trgill: Is there a module you consider appropriately commented?  Not sure I'm following the add/remove criteria for comments.  Thinking an example might help.

src/engine/strat_engine/lineardev.rs
trgill: Ok

tests/lineardev_tests.rs
trgill: Yeah, I suspect there is... we could probably wait for an event, poll etc.

@agrover suggested this was good enough for a test.  I'm happy to put a TODO to remind us to fix it.  That work?

tests/util/blockdev_utils.rs
trgill: It is not dead code, but if I remove the directive I get a compiler warning for dead code.

I'm not sure why.

tests/lineardev_tests.rs
mulkieran: TODO is fine for now.

src/engine/strat_engine/lineardev.rs
mulkieran: The traits here: https://github.com/stratis-storage/stratisd/blob/master/src/engine/engine.rs are almost adequately commented.

The tests here: https://github.com/stratis-storage/stratisd/blob/master/src/engine/strat_engine/metadata.rs#L533 state what they're trying to do reasonably well.

trgill:
@mulkieran is this good to merge?

trgill:
Ok, I updated the test in this PR.  I'll update the others in a subsequent PR.

PR #221, opened 2017-01-31
Work on Sectors and new Bytes type

mulkieran:


mulkieran:
@agrover @trgill review?

This addresses a few concerns:
* Suspicious looking downcasts.
* Dereferencing Sectors to do computations, thus losing advantage of type-checking.
* Representing Bytes as a plain number, ditto.
* Allowing some non-sensical operations on Sectors.

The implementation of both Bytes and Sectors is probably incomplete, but any additional traits can be implemented as appropriate and when necessary.

trgill:
This looks good to me, but maybe we could wait so @agrover could have a look before we merge?

mulkieran:
Absolutely.

mulkieran:
rebased.

mulkieran:
@agrover @trgill review?

I'm moving this up to review final because #225 would like to go in, but it needs this to be reviewed first.

agrover:
A big PR that I mostly really like! A couple points for discussion, naturally:

* I'm 👍 with removing `DataBlocks` now (every line of code should justify its existence based upon the existing codebase) but be ready to add it back in once we get to implementing ThinPool, since that's the allocation unit it uses.
* `SectorOffset` - Expressed in sectors but semantically different from `Sectors` (the length of the range of sectors). We should try to let the type system keep us from confusing these as much as possible. Froyo ended up using `(SectorOffset, Sectors)` tuples all over the place. If we don't choose that solution, maybe we have a named `struct` with them as members instead or something, but I do think we need something here even if we choose to skip `SectorOffset` type.
* 1a7fdf7 do/should we have a guideline for when we put the type name as part of a const and when we don't? I swear at some point we changed these the other way...
* Say in commitlog for 60a37b1, why a Bytes newtype rather than just a type alias for u64? Maybe also squash into following commit where Bytes also starts being used?
* 9388551 I find these help readability, I'd prefer we keep them, although maybe with kilo->kibi etc.
* 7e7d9ca please don't do the ellipsis thing to the commit summary. If you really can't shorten it then going over a little isn't the end of the world.
* b90d849 why should the fs quota be in sectors? Sectors are a block layer thing...
* 15cc0d6 what's the benefit?
* 36624df nice catch. not just useless but buggy.

Thanks!

mulkieran:
* About ```DataBlocks```... I understand that we will probably have to add in some code for its intended purpose at some time. I just always try very hard to get rid of unused code...it rots very fast and confuses the eye and brain.
* ```SectorOffset``` was in some sense not used at all because ```avail_range()``` isnn't called. I'm happy to add in something to do its job when the time comes, but right now it really has no job.
* 1a7fdf7 It's not the type, its the units. I don't think we should put the type in the variable name. But I think there's a good deal of justification for the units in constant names, especially when they are bare numbers with no other indicator of what they mean.
* 60a37b1 I guess the why is exactly the same as for Sectors type...to gain the benefits of various kinds of type-safety.
* 9388551 I find they hinder readability. I'll try to come up w/ some alternative.
* 7e7d9ca Those ellipses are a GitHub artifact. I always keep my commit titles <= 80 chars.
* b90d849 I was guessing wildly on this one. I'll change it to Bytes. But this illustrates why it's good to have Bytes _and_ Sectors.
* 15cc0d6 A bit more type-safety.
* 36624df thanks!

agrover:
* Recommend aiming for 70-char max commit summaries.
* I happen to know @trgill is calling `avail_range` in some in-development code he's working on 😃 *and* he had a bug (now fixed) stemming from the values' semantics! Let's keep SectorOffset for now and re-assess if we like it more (or not) once we have Todd's code as context. I know we're going to be doing a lot of manipulation of lists of sector ranges on devices, so we're going to want to do *something* to make them easy to work with safely.
* Please go ahead and add explanatory text based upon the above to the commits that don't have any.
* 15cc0d6 I feel negative vibes towards this. I think it's because we're using incorrect types for things and duplicating consts just because Rust uses usize for array sizes/indexing. Maybe we can instead finesse this by implementing [Into(usize)](https://doc.rust-lang.org/std/convert/trait.Into.html) for `Bytes` and other types that need it? if arrays need to become `vec![]`s (b/c array sizing needs to be const) I'd be ok with that too, even. Other options?

mulkieran:
* Where is the code that calls ```avail_range()```? SectorOffset is buggy right now, as it defines non-sensical operations. If we're going to carry it forward because it will be used, then I would like to have a clear idea of how it is used so that I can fix and improve it and make it interoperate properly with Sectors now.
* We can't implemet ```Into(usize)``` for bytes because ```Into(T)``` must not fail by definition, and conversion of u64 to usize can fail, by definition. All this commit really does is take a problem that was previously widespread and isolate it in a single spot. There may be other alternative, but I doubt ```vec!``` will help.

agrover:
* Talked to Todd. He can rework his PR based on changes in this PR (incl no `SectorOffset`) so it sounds like this going in first is the way to proceed. That's ok for now, and just leaves an issue that I'd like for us to come up with a solution for working with `(Sector, Sector)` data types (where .0 is an offset and .1 is a length). I've opened #230 so we can discuss further out of the context of this PR.
* `panic!` if `into<T>` fails? Shouldn't ever happen, no?

mulkieran:
@agrover review? The final three commits are the only real changes.

agrover:
where are the IEC names coming from? I'm trying to square with https://en.wikipedia.org/wiki/Gibibyte IRC byte units and they seem rather...short?

mulkieran:
They come from the table on this page: https://en.wikipedia.org/wiki/Binary_prefix. They are the standard modifiers for the unit symbols. So Ki means 1024 somethings, while k means 1000 somethings, e.g., KiB, kg. The standard prefixes for the words seem not quite as good because they are part of a word, e.g., kibi-bytes, kilo-grams.

mulkieran:
Superceded by #231.

PR #223, opened 2017-02-01
Add env_logger to stratisd and libstratis

trgill:
Move log dependencies from tests to application level.
Change stratisd.rs to use env_logger.

When user passes --debug to stratisd enable logging for binary crate
and the libstratis crate.  --debug will override RUST_LOG.

RUST_LOG env variable controls per module logging.  For example:

RUST_LOG=stratisd=debug,libstratis=debug
RUST_LOG=libstratis::engine::strat_engine::engine=info

### DIFF
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -4,7 +4,9 @@
 
 #[macro_use]
 extern crate libstratis;
-
+#[macro_use]
+extern crate log;
+extern crate env_logger;
 extern crate devicemapper;
 #[macro_use]
 extern crate clap;
@@ -31,10 +33,13 @@ extern crate enum_derive;
 extern crate quickcheck;
 
 use std::io::Write;
+use std::env;
 use std::error::Error;
 use std::process::exit;
 
 use clap::{App, Arg};
+use log::LogLevelFilter;
+use env_logger::LogBuilder;
 
 use libstratis::engine::Engine;
 use libstratis::engine::sim_engine::SimEngine;
@@ -51,18 +56,8 @@ fn write_err(err: StratisError) -> StratisResult<()> {
     Ok(())
 }
 
-pub static mut DEBUG: bool = false;
-
-macro_rules! dbgp {
-    ($($arg:tt)*) => (
-        unsafe {
-            if DEBUG {
-                println!($($arg)*)
-            }
-        })
-}
-
 fn main() {
+
     let matches = App::new("stratis")
         .version(VERSION)
         .about("Stratis storage management")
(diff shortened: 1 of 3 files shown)


trgill:
@agrover @mulkieran review please...

agrover:
no additional comments, looks good (you sure you didn't want review-final on this?)

trgill:
@mulkieran this good to go?

src/bin/stratisd.rs
```diff
@@ -75,15 +69,21 @@ fn main() {
         .get_matches();
 
     if matches.is_present("debug") {
-        unsafe { DEBUG = true }
-    };
+        let mut builder = LogBuilder::new();
```
mulkieran: I think ```filter()``` and ```parse()``` must be redundant.

Why not:

```
let mut builder = LogBuilder::new();
builder.format(|record: &LogRecord| format!("{} - {}", record.level(), record.args()));
if matches.is_present("debug") {
    builder.filter(Option("stratisd"), LogLevelFilter::Debug);
    builder.filter(Option("libstratis"), LogLevelFilter::Debug);
} else {
    builder.filter(Option("stratisd"), LogLevelFilter::Error);
    builder.filter(Option("libstratis"), LogLevelFilter::Error);
};
builder.init().unwrap();
```

I think we want to know about our, still, hypothetical, error messages even if we haven't passed the ```--debug``` flag. And we want our format on log messages to be uniform, regardless.

trgill: error! is always on... from the docs:

Log levels are controlled on a per-module basis, and by default all logging is disabled except for error!.

trgill: builder.init().unwrap() doesn't read the RUST_LOG env variable.  So, this code would ignore RUST_LOG.

trgill: The format is the same as the default that we get when we call env_logger::init().

Can you explain your objection to parse()?  It is more efficent to call filter()?

mulkieran: We should call one or the other of ```filter()```, ```parse()``` since they do the same thing, AFAIU. I do not like ```parse()``` because it is more heavyweight, as it has to parse a string by means of regular expressions. ```filter()``` is better, because more programmatic. I think ```parse()``` is more appropriate if someone had passed in those strings that is parses from the command line.

mulkieran: I think if you drop the format spec altogether, the format will be the same, at least that's what the code looks like. The global initializer seems to accept the LogBuilder's default format.

mulkieran: Maybe:

```
let mut builder = LogBuilder::new();
if matches.is_present("debug") {
    builder.filter(Option("stratisd"), LogLevelFilter::Debug);
    builder.filter(Option("libstratis"), LogLevelFilter::Debug);
} else {
    if let Ok(s) = env::var("RUST_LOG") {
        builder.parse(&s);
    }
};
builder.init().unwrap();
```

Then it is very clear that the intention is that ```--debug``` flag overrides whatever might be in the environment variable.

Regarding that doc sentence, to me it's meaning is quite uncertain. "on a per-module basis" is just a terminally vague phrase.

mulkieran:
:+1:

PR #226, opened 2017-02-01
Fix MDARegions::older.

mulkieran:
It was reversed.

### DIFF
--- a/src/engine/strat_engine/metadata.rs
+++ b/src/engine/strat_engine/metadata.rs
@@ -341,8 +341,8 @@ impl MDARegions {
 
     pub fn older(&self) -> usize {
         match self.mdas[0].last_updated.cmp(&self.mdas[1].last_updated) {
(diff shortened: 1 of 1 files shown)


mulkieran:
@agrover @trgill review? This should be an easy ack. I noticed it when I noticed that the newer timestamp was None and the older timestamp was Some(...), which was clearly wrong.

agrover:
Please keep the explicit listing of enumeration options

trgill:
:+1: 
Ok.  It is intentional that equal and greater return the same thing?

mulkieran:
It's a don't care in the extremely unlikely situation that the timestamps are equal. 1 is as good as 0.

The other option is to view two equal, non-None, timestamps as a sign that something is wrong, and return an error, but that is probably overkill.

trgill:
Ok.

PR #227, opened 2017-02-03
Set data checksum when saving state.

mulkieran:


mulkieran:
@agrover @trgill review?

The first commit fixes a significant bug...the two others are closely related.

PR #229, opened 2017-02-06
Minor changes from PR #221

mulkieran:
These are all changes from PR #221 that have already been approved and might as well go in now as they are not tangled up with any of the other changes that are still pending.


mulkieran:
@agrover review? This should be synch.

PR #231, opened 2017-02-07
Work on Sectors and Bytes type

mulkieran:


mulkieran:
This supercedes #221.The code is the same, but the commits are a bit cleaned up.

mulkieran:
The PR also implements AddAssign for Bytes and Sectors, because it would be kind of useful for linear dev PR and sometime might be necessary.

It turns out that these Newtype* macros can be parameterized to give correct implementation for non-self operation, but not enough. So, NewtypeMul(u64) would, I think, generate the correct operation for multiplying a Sector by a u64, but would fail to generate the operation for multiplying by a usize. Furthermore, it does not generate the operation for multiplying a u64 by a Sector, which is probably wrong. So, it seems best to stick with my macros where the newtype macros are not adequate.

src/engine/strat_engine/metadata.rs
```diff
@@ -227,17 +234,22 @@ impl StaticHeader {
 pub struct MDARegions {
     // Spec defines 4 regions, but regions 2 & 3 are duplicates of 0 and 1 respectively
     pub region_size: Sectors,
-    pub mdas: [MDAHeader; NUM_PRIMARY_MDA_REGIONS as usize],
+    pub mdas: [MDAHeader; NUM_PRIMARY_MDA_REGIONS],
 }
 
 impl MDARegions {
+    /// Calculate the offset from start of device for an MDARegion.
+    fn mda_offset(index: usize, per_region_size: Bytes) -> u64 {
```
trgill: Shouldn't the return type be Sectors?

mulkieran: The result of the function is used exclusively as an argument to ```SeekFrom()``` which requires a ```u64```, representing bytes, as an argument.

It might make more sense to return it as Bytes and then dereference the value where used in ```SeekFrom()``` calls, but, since this is a private function, I didn't go to the trouble.

PR #236, opened 2017-02-08
Work to improve/expand destroy_pool testing a little

agrover:
`destroy_pool` should succeed if the pool has devs, but should fail if the pool has filesystems.

### DIFF
--- a/src/engine/engine.rs
+++ b/src/engine/engine.rs
@@ -83,6 +83,11 @@ pub trait Pool: Debug {
     fn blockdevs(&mut self) -> Vec<&mut Dev>;
     fn cachedevs(&mut self) -> Vec<&mut Dev>;
 
+    /// Destroy the pool.
+    /// Will fail if filesystems allocated from the pool are in use,
+    /// or even exist.
+    fn destroy(self) -> EngineResult<()>;
+
     /// Ensures that all designated filesystems are gone from pool.
     /// Returns a list of the filesystems found, and actually destroyed.
     /// This list will be a subset of the names passed in fs_names.
--- a/src/engine/sim_engine/pool.rs
+++ b/src/engine/sim_engine/pool.rs
@@ -81,6 +81,11 @@ impl Pool for SimPool {
         destroy_filesystems!{self; fs_names}
     }
 
+    fn destroy(self) -> EngineResult<()> {
+        // Nothing to do here.
+        Ok(())
+    }
+
     fn create_filesystems<'a, 'b, 'c>(&'a mut self,
                                       specs: &[(&'b str, &'c str, Option<Bytes>)])
                                       -> EngineResult<Vec<&'b str>> {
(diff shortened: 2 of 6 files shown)


src/engine/engine.rs
```diff
@@ -83,6 +83,8 @@ pub trait Pool: Debug {
     fn blockdevs(&mut self) -> Vec<&mut Dev>;
     fn cachedevs(&mut self) -> Vec<&mut Dev>;
 
+    fn destroy(self) -> EngineResult<()>;
```
mulkieran: Needs a comment explaining its purpose.

src/engine/sim_engine/pool.rs
```diff
@@ -81,6 +81,12 @@ impl Pool for SimPool {
         destroy_filesystems!{self; fs_names}
     }
 
+    fn destroy(mut self) -> EngineResult<()> {
+        self.block_devs.clear();
```
mulkieran: Why is clearing these data structures necessary for SimPool and not for StratPool?

agrover: true... should this function just do nothing?

agrover: to answer my own q: yes, I believe it should do nothing 😃

src/engine/engine.rs
agrover: yup

src/engine/strat_engine/pool.rs
```diff
@@ -157,6 +157,21 @@ impl Pool for StratPool {
         Ok(bdev_paths)
     }
 
+    fn destroy(self) -> EngineResult<()> {
+
+        // TODO: first, tear down DM mappings
+
+        for (_, mut bd) in self.block_devs {
```
mulkieran: Minor: You could use ```values()``` like:

```
for mut bd in self.block_devs.values_mut() {
```
```

File: src/engine/engine.rs [hunk 4]
```diff
@@ -83,6 +83,11 @@ pub trait Pool: Debug {
     fn blockdevs(&mut self) -> Vec<&mut Dev>;
     fn cachedevs(&mut self) -> Vec<&mut Dev>;
 
+    /// Destroy the pool.
```
mulkieran: What does it mean to destroy a pool?

src/engine/strat_engine/pool.rs
agrover: 👍

mulkieran:
Now that ```wipe_header()``` is implemented in source for BlockDev it can be used in already implemented test code: https://github.com/stratis-storage/stratisd/blob/master/tests/util/blockdev_utils.rs#L35.

mulkieran:
@agrover I'm still hoping for improvements to existing tests in test directroy by the use of newly introduced```BlockDev::wipe_header()``` where appropriate.

agrover:
@mulkieran unfortunately it doesn't just drop in... `wipe_metadata` is a BlockDev method, but  in the tests `test_blockdev_setup` fn we are preparing the raw block devices -- we have a list of paths not a list of BlockDevs.

PR #238, opened 2017-02-14
Use DM to create a thin-pool for thin provisioned devices

trgill:
Added thingpooldev to combine a metadata and data device into a DM thin-pool
Added a test to validate the creation of the thin-pool
Added --skip to the Makefile for the integration test

### DIFF
--- a/src/engine/strat_engine/thinpooldev.rs
+++ b/src/engine/strat_engine/thinpooldev.rs
@@ -1,0 +1,88 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use devicemapper::{DM, DevId, DeviceInfo, DmFlags};
+use engine::EngineResult;
+
+use std::path::Path;
+
+use super::blockdev::BlockDev;
+
+use types::DataBlocks;
+use types::Sectors;
+
+pub struct ThinPoolDev {
+    pub name: String,
+    dev_info: Option<DeviceInfo>,
+    data_block_size: Option<Sectors>,
+    pub low_water_mark: Option<DataBlocks>, // How close to full before we are worried?
+    pub meta_dev: Option<BlockDev>,
+    pub data_dev: Option<BlockDev>,
+}
+
+/// support use of DM to create pools for thin provisioned devices
+impl ThinPoolDev {
+    pub fn new(name: &str) -> ThinPoolDev {
+        ThinPoolDev {
+            name: name.to_owned(),
+            dev_info: None,
+            data_block_size: None,
+            low_water_mark: None,
+            meta_dev: None,
+            data_dev: None,
+        }
+    }
+    /// Generate a Vec<> to be passed to DM.  The format of the Vec entries is:
+    /// <start sec> <length> "thin-pool" /dev/meta /dev/data <block size> <low water mark>
+    fn dm_table(&self,
+                length: Sectors,
+                data_block_size: Sectors,
+                low_water_mark: DataBlocks,
+                meta_dev: &Path,
+                data_dev: &Path)
+                -> Vec<(u64, u64, String, String)> {
+        let mut table = Vec::new();
+        let params = format!("{} {} {} {} 1 skip_block_zeroing",
+                             meta_dev.to_string_lossy(),
+                             data_dev.to_string_lossy(),
+                             *data_block_size,
+                             *low_water_mark);
+        table.push((0u64, length.0, "thin-pool".to_owned(), params));
+        debug!("dmtable line : {:?}", table);
+        table
+    }
+
+    /// Use DM to create a "thin-pool".  A "thin-pool" is shared space for
+    /// other thin provisioned devices to use.
+    ///
+    /// See section "Setting up a fresh pool device":
+    /// https://www.kernel.org/doc/Documentation/device-mapper/thin-provisioning.txt
+    pub fn setup(&mut self,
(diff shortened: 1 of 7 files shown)


agrover:
:+1:

I'd like us to not wait, and consider this for final review and merging.

src/engine/strat_engine/thinpooldev.rs
```diff
@@ -0,0 +1,88 @@
...
+                data_block_size: &Sectors,
+                low_water_mark: &DataBlocks,
+                meta_dev: &PathBuf,
+                data_dev: &PathBuf)
+                -> Vec<(u64, u64, String, String)> {
+        let mut table = Vec::new();
+        let params = format!("{} {} {} {} 1 skip_block_zeroing",
+                             meta_dev.to_string_lossy(),
+                             data_dev.to_string_lossy(),
+                             data_block_size.0,
```
mulkieran: Everywhere ```.0``` appears should be a dereference instead. (except maybe DataBlocks values, see comment below).

src/engine/strat_engine/thinpooldev.rs
```diff
@@ -0,0 +1,88 @@
...
+            low_water_mark: None,
+            meta_dev: None,
+            data_dev: None,
+        }
+    }
+    /// Generate a Vec<> to be passed to DM.  The format of the Vec entries is:
+    /// <start sec> <length> "thin-pool" /dev/meta /dev/data <block size> <low water mark>
+    fn dm_table(&self,
+                length: &Sectors,
+                data_block_size: &Sectors,
```
mulkieran: Sectors and DataBlocks are Copy, so there's probably no point in passing as references. This point applies to the method below as well.

src/types.rs
```diff
@@ -148,6 +148,32 @@ impl serde::Deserialize for Sectors {
     }
 }
 
+// A type for Data Blocks as used by the thin pool.
+custom_derive! {
+    #[derive(NewtypeFrom, NewtypeAdd, NewtypeSub, NewtypeDeref,
```
mulkieran: Most of these ```Newtype*``` macro calls should be deleted. It is probably possible to use the same subset as are used for Bytes and Sectors. It would probably be fine to use the ```unsigned_mul``` macros, although I can't see anywhere that DataBlocks are multiplied by any unsigned numbers.

AFAICT, the size of a DataBlock in Sectors is not a constant? In that case, you're probably better off parameterizing the DataBlock on its scaling factor in Sectors.

src/engine/strat_engine/thinpooldev.rs
```diff
@@ -0,0 +1,88 @@
...
+use engine::EngineResult;
+
+use std::path::PathBuf;
+
+use super::blockdev::BlockDev;
+
+use types::DataBlocks;
+use types::Sectors;
+
+pub struct ThinPoolDev {
```
mulkieran: These various fields need some more explanation.

tests/thinpooldev_tests.rs
```diff
@@ -0,0 +1,217 @@
...
+    Ok(())
+}
+
+#[test]
+/// Get list of safe to destroy devices.
+/// Clean any headers from the devices.
+/// Construct meta and data devices for use in the thin-pool
+/// Test creating a thin-pool device
+/// Teardown the DM device
+pub fn test_thinpoolsetup_setup() {
```
mulkieran: This test can not fail. Perhaps it should be using ```assert!``` somewhere?

src/engine/strat_engine/thinpooldev.rs
agrover: What would be good here?

src/types.rs
agrover: opened #241. Should be addressed but in a later PR.

src/engine/strat_engine/thinpooldev.rs
mulkieran: Whatever you know that I don't. For example, what does low-water-mark really mean? I know what it means when I'm at the shore, but that's not really enough for me to figure out what it means in this context.

src/engine/strat_engine/thinpooldev.rs
```diff
@@ -0,0 +1,88 @@
...
+                data_dev: &Path)
+                -> Vec<(u64, u64, String, String)> {
+        let mut table = Vec::new();
+        let params = format!("{} {} {} {} 1 skip_block_zeroing",
+                             meta_dev.to_string_lossy(),
+                             data_dev.to_string_lossy(),
+                             *data_block_size,
+                             *low_water_mark);
+        table.push((0u64, length.0, "thin-pool".to_owned(), params));
+        debug!("dmtable line : {:?}", table);
```
mulkieran: It's nice to have the ability to write out debug info...but I feel we shouldn't put it in in a miscellaneous fashion. It really needs judgement, otherwise you end up with horribly repetitive, inscrutable logs that only experts can read. Also, we might want to strategize about putting information in the journal, in a useful way.

IOW, is it Ok to remove the calls to the ```debug!``` macro now?

agrover: leaving as-is for now, opened https://github.com/stratis-storage/stratis-docs/issues/21 so we can come up with a policy and then do a global sweep

src/engine/strat_engine/thinpooldev.rs
agrover: I added a comment for low_water_mark but I'm not clear beyond that how to divide up fields. They *all* seem DM-necessary to me.

tests/thinpooldev_tests.rs
agrover: fixed

PR #240, opened 2017-02-15
Master gmo prelim

mulkieran:

PR #244, opened 2017-02-21
GetManagedObjects refactoring

mulkieran:


mulkieran:
Supercedes #240.

src/dbus_api/api.rs
```diff
@@ -252,26 +216,25 @@ fn create_filesystems(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
...
 
     let mut b_engine = dbus_context.engine.borrow_mut();
-    let ref mut pool = get_pool!(b_engine; pool_name; default_return; return_message);
+    let ref mut pool = get_pool!(b_engine; pool_uuid; default_return; return_message);
 
     let specs = filesystems.map(|x| (x.0, x.1, tuple_to_option(x.2).map(|x| Bytes(x))))
         .collect::<Vec<(&str, &str, Option<Bytes>)>>();
     let result = pool.create_filesystems(&specs);
 
     let msg = match result {
```
agrover: style: so many `let`s... maybe unavoidable but I think you can at least do without `result` since it's only used in the `match`.

src/dbus_api/api.rs
```diff
@@ -371,33 +333,8 @@ fn rename_filesystem(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
...
-                    }
-                }
-                None => {
-                    let error_message = format!("no dbus object path for renamed filesystem");
-                    let (rc, rs) = code_to_message_items(DbusErrorEnum::INTERNAL_ERROR,
-                                                         error_message);
-                    return_message.append3(return_value, rc, rs)
-                }
-            }
+            let (rc, rs) = ok_message_items();
```
agrover: wow, really? Amazing. love seeing code get deleted :smiley:

src/dbus_api/api.rs
```diff
@@ -82,18 +82,18 @@ fn object_path_to_pair(dbus_context: &DbusContext,
 }
 
 /// Get name for pool from object path
-fn object_path_to_pool_uuid(dbus_context: &DbusContext,
+fn pool_object_path_to_pair(dbus_context: &DbusContext,
                             path: &dbus::Path)
-                            -> Result<Uuid, (MessageItem, MessageItem)> {
-    let uuid = match dbus_context.pools.borrow().get(path) {
+                            -> Result<(dbus::Path<'static>, Uuid), (MessageItem, MessageItem)> {
+    let pair = match dbus_context.pools.borrow().get(path) {
```
agrover: This is just for uniformity? Maybe if they were sharing a function, but pool already has its own function, so adding an unuseful return parameter may cause confusion. I vote to drop this, unless there are other reasons...

src/dbus_api/api.rs
```diff
@@ -65,9 +65,9 @@ fn get_next_arg<'a, T>(iter: &mut Iter<'a>, loc: u16) -> Result<T, MethodErr>
 }
 
 /// Get filesystem name from object path
-fn object_path_to_pair(dbus_context: &DbusContext,
-                       fs_object_path: &dbus::Path)
-                       -> Result<(Uuid, Uuid), (MessageItem, MessageItem)> {
+fn fs_object_path_to_pair(dbus_context: &DbusContext,
+                          fs_object_path: &dbus::Path)
+                          -> Result<(dbus::Path<'static>, Uuid), (MessageItem, MessageItem)> {
```
agrover: comments need updating.

style: Hmm, do you think it's worthwhile to use `type`s to help clarify what exactly the `Path<'static>` *is*? We could make it more clear *which* uuid we're returning in a similar manner. Of course the comments can describe but `type`s may be less likely to get stale. Thoughts?

src/engine/engine.rs
```diff
@@ -101,6 +101,9 @@ pub trait Pool: HasName + HasUuid {
 
     /// Rename this pool.
     fn rename(&mut self, name: &str) -> ();
+
+    /// Get the filesystem in this pool with this UUID.
+    fn get_filesystem(&mut self, uuid: &Uuid) -> Option<&mut Filesystem>;
```
agrover: why'd we want this, again? plz describe in commitlog.

src/dbus_api/macros.rs
```diff
@@ -6,16 +6,16 @@
...
-                let (rc, rs) = code_to_message_items(DbusErrorEnum::POOL_NOTFOUND,
-                                                     format!("no pool for uuid {}", $uuid));
-                return Ok(vec![$message.append3($default, rc, rs)]);
-            }
+        if let Some(pool) = $engine.get_pool(&$uuid) {
+            pool
+        } else {
+            let (rc, rs) = code_to_message_items(DbusErrorEnum::POOL_NOTFOUND,
+                                                 format!("no pool for uuid {}", $uuid));
+            return Ok(vec![$message.append3($default, rc, rs)]);
```
agrover: style: `if let` is better if you just care about one case, but in this case you care about both. Not much shorter. Maybe try `Option::unwrap_or_else`?

src/engine/engine.rs
```diff
@@ -51,6 +51,12 @@ pub trait Dev: Debug {
 pub trait Filesystem: HasName + HasUuid {
     /// Rename this filesystem.
     fn rename(&mut self, name: &str) -> ();
+
+    /// Get this filesystem's mount point
+    fn mountpoint(&self) -> &String;
+
+    /// Get this filesystem's quota
+    fn quota(&self) -> &Bytes;
```
agrover: why are we adding these when a previous commit removed the dbus bits for these, and made the comment they are scheduled for a future release?

src/dbus_api/filesystem.rs
```diff
@@ -55,6 +55,11 @@ pub fn create_dbus_filesystem<'a>(dbus_context: &DbusContext) -> dbus::Path<'a>
         .emits_changed(EmitsChangedSignal::False)
         .on_get(get_filesystem_name);
 
+    let pool_property = f.property::<&dbus::Path, _>("Pool", ())
```
agrover: Please put reason for this in the commitlog

src/engine/structures.rs
```diff
@@ -0,0 +1,156 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use std::collections::HashMap;
+use std::slice::IterMut;
+
+use uuid::Uuid;
+
+use super::engine::{HasName, HasUuid};
```
mulkieran: It is quite possible, but we should defer that to a later PR and when we've looked at a few more cases.

src/engine/structures.rs
```diff
@@ -0,0 +1,156 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
mulkieran: I added some, I hope it helps.

src/engine/structures.rs
```diff
@@ -0,0 +1,156 @@
...
+use super::engine::{HasName, HasUuid};
+
+
+/// Map UUID and name to T items.
+#[derive(Debug)]
+pub struct Table<T: HasName + HasUuid> {
+    items: Vec<T>,
+    name_map: HashMap<String, usize>,
+    uuid_map: HashMap<Uuid, usize>,
+}
```
mulkieran: I think I've answered that in my Table commit message. Essentially, there are lots of benefits to encapsulation that it is worthwhile realizing here. I considered a variety of two-map approaches. None of them were as consistent or clear as the current approach. Working through what is necessary to remove an entry given a UUID and then working through what it necessary to find one given a name will show that these two-map approaches are surprisingly hard to support. A bidirectional map would ease things, but the one we have been using is both O(n), which could become a problem, and buggy (in its insertion function). I would _much_ rather use standard library data structures.

src/dbus_api/api.rs
mulkieran: I was able to get rid of the method entirely in a later commit, so the outdated header comment is pretty much moot, and very hard to fix.

I don't think it would help. It's what the dbus-rs developer has chosen to use to represent the data view of an ObjectPath. It's hard to put that in a type. If you want to distinguish it from a standard library Path, the code forces use to do that with the ```dbus``` prefix.

src/dbus_api/api.rs
mulkieran: I really want to keep it. It is a very low-cost uniformity, quite meaningful, and will be very convenient when we drop these look-aside tables entirely (when the new dbus-rs release arrives). When this happens we will decorate the object path with exactly one type of data, what the dbus library allows. If that data is the same for all object paths it will reduce error-handling code by a lot.

src/dbus_api/api.rs
```diff
@@ -64,10 +66,10 @@ fn get_next_arg<'a, T>(iter: &mut Iter<'a>, loc: u16) -> Result<T, MethodErr>
 
 /// Get filesystem name from object path
 fn object_path_to_pair(dbus_context: &DbusContext,
-                       fs_object_path: &str)
-                       -> Result<(String, String), (MessageItem, MessageItem)> {
-    let fs_pool_pair = match dbus_context.filesystems.borrow().get_by_first(fs_object_path) {
-        Some(fs_name) => fs_name.clone(),
+                       fs_object_path: &dbus::Path)
+                       -> Result<(Uuid, Uuid), (MessageItem, MessageItem)> {
```
mulkieran: The method will go away. This was an intermediate commit, to get to a compiling state. The final version returned the object path of the pool and the uuid of the filesystem.

src/dbus_api/api.rs
```diff
@@ -80,28 +82,28 @@ fn object_path_to_pair(dbus_context: &DbusContext,
 }
 
 /// Get name for pool from object path
-fn object_path_to_pool_name(dbus_context: &DbusContext,
+fn object_path_to_pool_uuid(dbus_context: &DbusContext,
```
mulkieran: The method will be gone....

src/dbus_api/api.rs
```diff
@@ -80,28 +82,28 @@ fn object_path_to_pair(dbus_context: &DbusContext,
...
+    let uuid = match dbus_context.pools.borrow().get(path) {
         Some(pool) => pool.clone(),
         None => {
             let items = code_to_message_items(DbusErrorEnum::INTERNAL_ERROR,
                                               format!("no pool for object path {}", path));
             return Err(items);
         }
     };
-    Ok(pool_name)
+    Ok(uuid)
```
mulkieran: The particular construction to which I believe you are referring went away in the macros that were derived from it.

src/dbus_api/api.rs
mulkieran: I've removed a few that can be easily removed in a new commit.

src/dbus_api/api.rs
mulkieran: Yes, the new, simplified one-directional tables are a huge win when renaming pools and filesystems.

src/engine/engine.rs
mulkieran: These enable the setting of Quota and Mountpoint properties in the D-Bus API.

src/dbus_api/macros.rs
mulkieran: I'ld like to hold onto it. It saves a level of indentation, which is nice. I think ```unwrap_or_else``` would be kind of awkward.

mulkieran:
@agrover All Dbus tests are now passing again. I would like to abstract the filesystem lookups for filesystem properties in filesystem.rs, but I continue to be defeated by lifetime errors. I'm ready to let this one go, so long as the CI passes.

src/engine/structures.rs
agrover: ok. btw I don't think I'd heard about the bug in bidirmap insert, can you elaborate?

src/engine/engine.rs
agrover: Maybe we should not have put these in the DBus API yet then. We should remove them there, and also here, for now.

agrover:
* Rebase/edit all commits and remove periods from end of all commit summary lines
* Ensure all commits have commit messages beyond the summary. A lot of the later ones do but many of the early ones don't. This is important to do so that once we start getting more contributors, they can understand our motivation for each change.

Definitely getting close to there. Thanks. This is a lot of work.

src/engine/structures.rs
mulkieran: _This_ is the work of a person who has worked out the theory of bidirectional mappings pretty well, as far as I can tell: https://bidict.readthedocs.io/en/latest/. An important invariant of a bidirectional map is a unique set of keys _and_ a unique set of values. The implementation of ```bidir_map``` does not ensure this property, i.e., it allows insertion of a new item to leave duplicate values. The documentation does not specify what happens on duplicates, but it is pretty obvious from reading the code.

agrover: thanks for the explanation!

agrover:
bidir-map usage is zero? then we can drop it from Cargo.toml I think.

mulkieran:
Removed ```bidir-map``` crate requirement.

mulkieran:
Superceded by #246.

PR #247, opened 2017-02-24
Call Stratis lib "stratis" instead of "libstratis"

agrover:
I think this is more in line with Rust naming conventions -- the names of
other crates we use don't start with "lib". Or, it might have the potential
to cause confusion.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,5 +1,5 @@
 [root]
-name = "libstratis"
+name = "stratis"
 version = "0.1.0"
 dependencies = [
  "byteorder 0.3.13 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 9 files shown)


agrover:
@mulkieran @trgill curious to hear your thoughts on this little PR.

mulkieran:
rustc source tree is packed with lib* subdirectories which are exported as plain '*' crates which supports your statement about the naming conventions.

On the other, we already are using the name "stratis" for the CLI, which could promote confusion.

If we followed the rustc approach, we would probably call the directory "libstratisd_daemon" and the crate "stratisd_daemon".

trgill:
I think you are right - the "lib" prefix does not follow rust conventions (only based on looking around).

Not a lot of guidance here: https://doc.rust-lang.org/1.0.0/style/features/crates.html

agrover:
didn't reach a consensus this would be entirely beneficial, due to too many things named stratis.

PR #248, opened 2017-02-24
Pull event loop out of DBus and check pool status

agrover:
stratisd will not be entirely driven by DBus requests -- it may need to
respond to periodic timers, DM events, or udev events, for example. This
patch hoists the event loop out of the dbus::run function and back into
main() where we call poll() ourselves. It doesn't actually add support for
polling on other fds besides DBus fds yet.

For now, based on dbus-rs Watch example code, which works but requires
`unsafe`.

Divide dbus_api::run() into connect() and handle() functions.

### DIFF
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -28,6 +28,8 @@ extern crate quickcheck;
 use std::io::Write;
 use std::env;
 use std::error::Error;
+use std::rc::Rc;
+use std::cell::RefCell;
 
 use clap::{App, Arg};
 use log::LogLevelFilter;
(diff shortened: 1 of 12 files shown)


src/bin/stratisd.rs
```diff
@@ -74,21 +77,45 @@ fn main() {
...
+                    if let Err(e) = write_err(r) {
+                        panic!("Unable to write to stderr: {}", e)
+                    }
+                }
+            }
         }
 
-        exit(1);
+        // Ask the engine to check its pools
+        engine.borrow_mut().check()
```
trgill: Just a question.

This check() will need to be fairly quick, right?  dbus won't respond to user requests while this is running?

agrover: yes, correct.

PR #252, opened 2017-03-07
Updates to D-Bus API

mulkieran:


mulkieran:
If it passes, it might as well go in, modulo any reservations about that Cargo.lock change.

trgill:
:+1:

PR #254, opened 2017-03-09
Connect API to thin devs

trgill:
Connect API to thin devices - pools and thin provisioned filesystems

Added Debug to ThinPoolDev and ThinDev because I added them to structs that derived
Debug.

Moved wait_for_dm out of tests and into ThinPoolDev for use by structs that create DM
devices.

Added size() to LinearDev so we can make thin upper limit of the ThinPoolDev proportionate
to the underlying data LinearDev.

StratPool now creates a backing ThinPoolDev.

Create Filesystem now creates a thindev in the pool and puts and XFS filesystem on it.

Added destroy() to cleanup filesystems

Note: There are a few TODO markers in this PR.
    1. Currently thin_id is just a random number
    2. Pool requires at least 2 devices (meta and data)
    3. Hard coded data blocksize and low water mark.
    4. We sleep for a time waiting for DM devices
    5. Need to decide on the ratio of real storage to the size of the thin pool. (1:1 currently)
    6. We don't require a size to be provided for create_filesystem - but devicemapper requires
        an initial size for a thin provisioned device.

### DIFF
--- a/src/dbus_api/util.rs
+++ b/src/dbus_api/util.rs
@@ -44,6 +44,7 @@ pub fn engine_to_dbus_err(err: &EngineError) -> (DbusErrorEnum, String) {
                 engine::ErrorEnum::Busy => DbusErrorEnum::BUSY,
                 engine::ErrorEnum::Invalid => DbusErrorEnum::ERROR,
                 engine::ErrorEnum::NotFound => DbusErrorEnum::NOTFOUND,
+                engine::ErrorEnum::FailedToOpen => DbusErrorEnum::ERROR,
             }
         }
         EngineError::Io(_) => DbusErrorEnum::IO_ERROR,
--- a/src/engine/macros.rs
+++ b/src/engine/macros.rs
@@ -23,10 +23,11 @@ macro_rules! destroy_filesystems {
     ( $s:ident; $fs:expr ) => {
         let mut removed = Vec::new();
         for uuid in $fs.iter().map(|x| *x) {
-            if $s.filesystems.remove_by_uuid(&uuid).is_some() {
+            if let Some(fs) = $s.filesystems.remove_by_uuid(&uuid) {
+                try!(fs.destroy());
                 removed.push(uuid);
-            };
-        };
+            }
+        }
         Ok(removed)
     }
 }
--- a/src/engine/strat_engine/blockdev.rs
+++ b/src/engine/strat_engine/blockdev.rs
@@ -189,6 +189,13 @@ pub fn initialize(pool_uuid: &PoolUuid,
 
     let add_devs = try!(filter_devs(dev_infos, pool_uuid, force));
 
+    // TODO: Fix this code.  We should deal with any number of blockdevs
+    //
+    if add_devs.len() <= 2 {
+        return Err(EngineError::Engine(ErrorEnum::Error,
+                                       "Need at least 2 blockdevs to create a pool".into()));
+    }
+
     let mut bds = BTreeMap::new();
     for (dev, (devnode, dev_size, mut f)) in add_devs {
 
--- a/src/engine/strat_engine/blockdev.rs
+++ b/src/engine/strat_engine/blockdev.rs
@@ -11,6 +11,8 @@ use std::fs::{OpenOptions, read_dir};
 use std::os::unix::prelude::AsRawFd;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
+use std::thread;
+use std::time::Duration;
 
 use time::Timespec;
 use devicemapper::Device;
@@ -249,5 +251,11 @@ impl BlockDev {
         assert!(start <= self.bda.header.blkdev_size);
         let length = self.bda.header.blkdev_size - start;
         (start, length)
+    }
+
+    /// The /dev/mapper/<name> device is not immediately available for use.
+    /// TODO: Implement wait for event or poll.
+    pub fn wait_for_dm() {
+        thread::sleep(Duration::from_millis(500))
     }
 }
--- a/src/engine/strat_engine/blockdev.rs
+++ b/src/engine/strat_engine/blockdev.rs
@@ -191,7 +191,7 @@ pub fn initialize(pool_uuid: &PoolUuid,
 
     // TODO: Fix this code.  We should deal with any number of blockdevs
     //
-    if add_devs.len() <= 2 {
+    if add_devs.len() < 2 {
         return Err(EngineError::Engine(ErrorEnum::Error,
                                        "Need at least 2 blockdevs to create a pool".into()));
     }
--- a/src/engine/strat_engine/filesystem.rs
+++ b/src/engine/strat_engine/filesystem.rs
@@ -8,13 +8,12 @@ use uuid::Uuid;
 use devicemapper::DM;
 
 use consts::IEC;
+
 use engine::EngineResult;
 use engine::Filesystem;
 use engine::strat_engine::thindev::ThinDev;
 use engine::strat_engine::thinpooldev::ThinPoolDev;
-
 use super::super::engine::{HasName, HasUuid};
-
 use types::Bytes;
 
 
--- a/src/engine/strat_engine/filesystem.rs
+++ b/src/engine/strat_engine/filesystem.rs
@@ -68,4 +68,9 @@ impl Filesystem for StratFilesystem {
     fn rename(&mut self, name: &str) {
         self.name = name.to_owned();
     }
+
+    fn destroy(self) -> EngineResult<()> {
+        let dm = try!(DM::new());
+        self.thin_dev.teardown(&dm)
+    }
 }
--- a/src/engine/strat_engine/filesystem.rs
+++ b/src/engine/strat_engine/filesystem.rs
@@ -1,12 +1,22 @@
 // This Source Code Form is subject to the terms of the Mozilla Public
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
+extern crate rand;
 
 use uuid::Uuid;
 
+use devicemapper::DM;
+
+use consts::IEC;
+use engine::EngineResult;
 use engine::Filesystem;
+use engine::strat_engine::thindev::ThinDev;
+use engine::strat_engine::thinpooldev::ThinPoolDev;
 
 use super::super::engine::{HasName, HasUuid};
+
+use types::Bytes;
+
 
 #[derive(Debug)]
 pub struct StratFilesystem {
@@ -28,9 +38,12 @@ impl StratFilesystem {
         let thin_id = rand::random::<u16>();
         // TODO We don't require a size to be provided for create_filesystems -
         // but devicemapper requires an initial size for a thin provisioned
-        // device - currently hard coded to Sectors(300000).
-        let mut new_thin_dev =
-            try!(ThinDev::new(name, dm, thin_pool, thin_id as u32, Sectors(300000)));
+        // device - currently hard coded to 1GB.
+        let mut new_thin_dev = try!(ThinDev::new(name,
+                                                 dm,
+                                                 thin_pool,
+                                                 thin_id as u32,
+                                                 Bytes(IEC::Gi).sectors()));
         try!(new_thin_dev.create_fs());
         Ok(StratFilesystem {
             fs_id: fs_id,
--- a/src/engine/strat_engine/filesystem.rs
+++ b/src/engine/strat_engine/filesystem.rs
@@ -12,6 +12,32 @@ use super::super::engine::{HasName, HasUuid};
 pub struct StratFilesystem {
     fs_id: Uuid,
     name: String,
+    thin_dev: ThinDev,
+}
+
+impl StratFilesystem {
+    pub fn new(fs_id: Uuid,
+               name: &str,
+               dm: &DM,
+               thin_pool: &mut ThinPoolDev)
+               -> EngineResult<StratFilesystem> {
+        // TODO should replace with proper id generation. DM takes a 24 bit
+        // number for the thin_id.  Generate a u16 to avoid the possibility of
+        // "too big". Should this be moved into the DM binding (or lower)?
+        // How can a client be expected to avoid collisions?
+        let thin_id = rand::random::<u16>();
+        // TODO We don't require a size to be provided for create_filesystems -
+        // but devicemapper requires an initial size for a thin provisioned
+        // device - currently hard coded to Sectors(300000).
+        let mut new_thin_dev =
+            try!(ThinDev::new(name, dm, thin_pool, thin_id as u32, Sectors(300000)));
+        try!(new_thin_dev.create_fs());
+        Ok(StratFilesystem {
+            fs_id: fs_id,
+            name: name.to_owned(),
+            thin_dev: new_thin_dev,
+        })
+    }
 }
 
 impl HasName for StratFilesystem {
--- a/src/engine/strat_engine/pool.rs
+++ b/src/engine/strat_engine/pool.rs
@@ -2,7 +2,9 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+use devicemapper::DM;
 use std::collections::BTreeMap;
+use std::iter::FromIterator;
 use std::path::Path;
 use std::path::PathBuf;
 use std::vec::Vec;
@@ -19,6 +21,8 @@ use engine::Filesystem;
 use engine::Pool;
 use engine::RenameAction;
 use engine::engine::Redundancy;
+use engine::strat_engine::lineardev::LinearDev;
+use engine::strat_engine::thinpooldev::ThinPoolDev;
 
 use super::super::engine::{HasName, HasUuid};
 use super::super::structures::Table;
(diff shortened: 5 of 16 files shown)


trgill:
@agrover @mulkieran please review.

src/engine/strat_engine/thindev.rs
```diff
@@ -2,23 +2,31 @@
...
 use std::path::Path;
 use std::path::PathBuf;
 use std::process::Command;
+
 use types::Sectors;
 
 #[derive(Clone)]
 pub struct ThinDev {
-    name: String,
+    pub name: String,
```
mulkieran: Are you sure any of this has to be public? Within a module, even private fields are public.

src/engine/strat_engine/thinpooldev.rs
```diff
@@ -109,4 +112,10 @@ impl ThinPoolDev {
         try!(self.meta_dev.teardown(dm));
         Ok(())
     }
+
```
mulkieran: I don't see why this wait should belong to ThinPool, it seems highly generic.

src/engine/strat_engine/pool.rs
```diff
@@ -48,12 +58,38 @@ impl StratPool {
         let devices = try!(resolve_devices(paths));
         let bds = try!(initialize(&pool_uuid, devices, MIN_MDA_SECTORS, force));
 
+        // TODO: Fix this code.  We should deal with any number of blockdevs
+        if bds.len() < 2 {
+            return Err(EngineError::Engine(ErrorEnum::MinBlockDev,
+                                           "Need at least 2 blockdevs to create a pool".into()));
+        }
+
+        let bds_copy = bds.clone();
```
mulkieran: I'm not sure all this copying is necessary.

src/engine/strat_engine/pool.rs
```diff
@@ -143,7 +179,9 @@ impl Pool for StratPool {
 
     fn destroy(mut self) -> EngineResult<()> {
 
-        // TODO: first, tear down DM mappings
+        // TODO Do we want to create a new File each time we interact with DM?
+        let dm = DM::new().unwrap();
```
mulkieran: Why is it unwrapped here, but ```try!``` elsewhere?

src/engine/strat_engine/blockdev.rs
```diff
@@ -207,7 +207,7 @@ pub fn initialize(pool_uuid: &PoolUuid,
 }
 
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
mulkieran: Why do we need to be able to clone BlockDevs?

src/engine/strat_engine/filesystem.rs
```diff
@@ -1,17 +1,42 @@
...
+}
+
+impl StratFilesystem {
+    pub fn new(fs_id: Uuid,
+               name: &str,
+               dm: &DM,
+               thin_pool: &mut ThinPoolDev)
+               -> EngineResult<StratFilesystem> {
+        // TODO should replace with proper id generation.
+        let thin_id = rand::random::<u16>();
```
mulkieran: Why defer this? Can't we just use UUID?

src/engine/strat_engine/filesystem.rs
```diff
@@ -1,17 +1,42 @@
...
+        let mut new_thin_dev =
+            try!(ThinDev::new(name, dm, thin_pool, thin_id as u32, Sectors(300000)));
```
mulkieran: Is this another arbitrary constant that needs a TODO?

src/engine/strat_engine/metadata.rs
```diff
@@ -34,7 +34,7 @@ pub const MIN_MDA_SECTORS: Sectors = Sectors(2032);
 
 const STRAT_MAGIC: &'static [u8] = b"!Stra0tis\x86\xff\x02^\x41rh";
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
mulkieran: What's driving all this cloning?

src/engine/strat_engine/pool.rs
```diff
@@ -163,9 +163,25 @@ impl StratPool {
 
 impl Pool for StratPool {
     fn create_filesystems<'a, 'b>(&'a mut self,
-                                  _specs: &[&'b str])
+                                  specs: &[&'b str])
                                   -> EngineResult<Vec<(&'b str, Uuid)>> {
-        unimplemented!()
+        let names = BTreeSet::from_iter(specs);
+        for name in names.iter() {
```
mulkieran: It's possible you can just use ```for name in names```.

src/engine/strat_engine/pool.rs
```diff
@@ -163,9 +163,25 @@ impl StratPool {
...
+            if self.filesystems.contains_name(name) {
+                return Err(EngineError::Engine(ErrorEnum::AlreadyExists, name.to_string()));
+            }
+        }
+
+        let mut result = Vec::new();
+        for name in names.iter() {
```
mulkieran: Possibly just ```for name in names``` here as well.

src/engine/macros.rs
```diff
@@ -23,10 +23,14 @@ macro_rules! destroy_filesystems {
     ( $s:ident; $fs:expr ) => {
         let mut removed = Vec::new();
         for uuid in $fs.iter().map(|x| *x) {
+            match $s.filesystems.get_mut_by_uuid(&uuid) {
```
mulkieran: I _think_ this match case is redundant. Probably you can call ```remove()``` on the thing returned by ```remove_by_uuid()```

src/dbus_api/util.rs
```diff
@@ -44,6 +44,8 @@ pub fn engine_to_dbus_err(err: &EngineError) -> (DbusErrorEnum, String) {
                 engine::ErrorEnum::Busy => DbusErrorEnum::BUSY,
                 engine::ErrorEnum::Invalid => DbusErrorEnum::ERROR,
                 engine::ErrorEnum::NotFound => DbusErrorEnum::NOTFOUND,
+                engine::ErrorEnum::MinBlockDev => DbusErrorEnum::ERROR,
```
mulkieran: Is MinBlockDev error necessary? It is only used in a line that is a TODO.

src/engine/macros.rs
trgill: You mean call destroy() on the thing returned by remove_by_uuid()?

If so - I tried that, but what is returned by remove_by_uuid() is not mutable.

src/engine/strat_engine/thindev.rs
trgill: You are right.  Will fix.

src/engine/macros.rs
mulkieran: Possibly there is something that should be done about the fact that the return value is immutable. I will take a look.

src/engine/strat_engine/filesystem.rs
trgill: It can't be a UUID.  We need to pass it to the devimapper binding as u32.

Not sure how to defer.

src/engine/macros.rs
mulkieran: But, it's also the case that none of the ```teardown()``` methods require a mutable self, which means that Filesystem's ```destroy()``` method does not require a mutable self either. I think those changes should be made, and then you can certainly use result of ```remove_by_uuid()``` method, regardless of any changes I can make.

src/engine/strat_engine/filesystem.rs
mulkieran: I see, it needs to be passed to the dm-table. Can you make line above:

```let thin_id = rand::random::<u32>();``` to avoid the widening cast below? It _ought_ to be more random that way.

If devicemapper really requires u32s, then I guess I don't see anything particularly wrong with just generating random ones. What else should we do?

Comment (trgill) [inline src/engine/strat_engine/filesystem.rs]: This updated comment ok?

        // TODO should replace with proper id generation. DM takes a 24 bit number 
        // for the thin_id.  Generate a u16 to avoid the possibility of "too big".
        // Should this be moved into the DM binding (or lower)?  How can a client
        // be expected to avoid collisions?

Comment (mulkieran) [inline src/engine/strat_engine/filesystem.rs]: Looks good. Thanks for explaining what is going on there.

File: src/engine/strat_engine/filesystem.rs (hunk 7 above)
Comment (trgill) [inline src/engine/strat_engine/filesystem.rs]: Yes.  How about I add the following?

        // TODO We don't require a size to be provided for create_filesystems -
        // but devicemapper requires an initial size for a thin provisioned
        // device - currently hard coded to Sectors(300000).

I also updated the issues list in the PR header.

Comment (mulkieran) [inline src/engine/strat_engine/filesystem.rs]: Looks good.

File: src/engine/macros.rs (hunk 11 above)
Comment (mulkieran) [inline src/engine/macros.rs]: If you really needed the removed thing to be mutable  you can use some variant of ```let mut```, like in https://github.com/stratis-storage/stratisd/pull/256.

File: src/engine/strat_engine/blockdev.rs (hunk 5 above)
Comment (trgill) [inline src/engine/strat_engine/blockdev.rs]: I will take another look at this and ask @agrover if he has any ideas.

I wasn't able to get StratPool::new() to work without making a clone of the BlockDevs in the Vec.

I think the alternative would be constructing the ThinPoolDev outside of new().  Hopefully I'm missing something.

Comment (trgill) [inline src/engine/strat_engine/blockdev.rs]: Looks like we can avoid the clone by: 

        let meta_dev;
        let data_dev;
        let length;
        {
            let mut data_devs = Vec::from_iter(bds.values());
            let metadata_blockdev = data_devs.remove(0);

            data_dev = try!(LinearDev::new(&format!("stratis_{}_data", name), dm, &data_devs));
            length = try!(data_dev.size()).sectors();

            let meta_blockdevs = vec![metadata_blockdev];
            meta_dev = try!(LinearDev::new(&format!("stratis_{}_meta", name), dm, &meta_blockdevs));
        }
        // TODO Fix hard coded data blocksize and low water mark.
        let thinpool_dev = try!(ThinPoolDev::new(&format!("stratis_{}_thinpool", name),
                                                 dm,
                                                 length,
                                                 Sectors(1024),
                                                 DataBlocks(256000),
                                                 meta_dev,
                                                 data_dev));

Comment (trgill) [inline src/engine/strat_engine/blockdev.rs]: I will remove the clone code.

File: src/engine/strat_engine/pool.rs (hunk 9 above)
Comment (trgill) [inline src/engine/strat_engine/pool.rs]: I get a compiler error if I make that change.   The error is related to:

 result = Vec::new(). (and the return of the vec)

the trait bound `str: std::marker::Sized` is not satisfied [E0277]:
`str` does not have a constant size known at compile-time
tuple elements must have `Sized` type
the trait `std::marker::Sized` is not implemented for `str`

File: src/engine/strat_engine/pool.rs (hunk 4 above)
Comment (trgill) [inline src/engine/strat_engine/pool.rs]: will fix

File: src/engine/macros.rs (hunk 11 above)
Comment (trgill) [inline src/engine/macros.rs]: Ok

The root was that teardown() had an un-needed "&mut self".  Thanks for the help.

File: src/dbus_api/util.rs (hunk 12 above)
Comment (trgill) [inline src/dbus_api/util.rs]: You'd prefer I use one of the other errors?

File: src/engine/strat_engine/pool.rs [hunk 13]
```diff
@@ -48,12 +58,38 @@ impl StratPool {
         let devices = try!(resolve_devices(paths));
         let bds = try!(initialize(&pool_uuid, devices, MIN_MDA_SECTORS, force));
 
+        // TODO: Fix this code.  We should deal with any number of blockdevs
+        if bds.len() < 2 {
```
trgill: ok

src/engine/strat_engine/pool.rs
```diff
@@ -48,12 +58,38 @@ impl StratPool {
         let devices = try!(resolve_devices(paths));
         let bds = try!(initialize(&pool_uuid, devices, MIN_MDA_SECTORS, force));
 
+        // TODO: Fix this code.  We should deal with any number of blockdevs
```
trgill: done

trgill:
@mulkieran you are thinking I should put wait_for_dm() in blockdev?  My thinking for thinpooldev was that it is somewhat of a parent for all the other devices.  but I agree with your comment.

src/engine/strat_engine/pool.rs
mulkieran: Ok, it doesn't really matter that much.

src/engine/macros.rs
```diff
@@ -23,10 +23,12 @@ macro_rules! destroy_filesystems {
     ( $s:ident; $fs:expr ) => {
         let mut removed = Vec::new();
         for uuid in $fs.iter().map(|x| *x) {
-            if $s.filesystems.remove_by_uuid(&uuid).is_some() {
+            let fs = $s.filesystems.remove_by_uuid(&uuid);
```
mulkieran:
```
if let Some(fs) = $s.filesystems.remove_by_uuid(&uuid) {
     try!(fs.destroy());
     removed.push(uuid);
}
```

is more idiomatic here.

src/engine/strat_engine/pool.rs
```diff
@@ -48,12 +59,35 @@ impl StratPool {
         let devices = try!(resolve_devices(paths));
         let bds = try!(initialize(&pool_uuid, devices, MIN_MDA_SECTORS, force));
 
+        let meta_dev;
+        let data_dev;
+        let length;
+        {
+            let mut data_devs = Vec::from_iter(bds.values());
+            let metadata_blockdev = data_devs.remove(0);
```
mulkieran: You don't yet know that ```data_devs```is non-empty, so ```remove()``` could panic.

src/engine/strat_engine/pool.rs
```diff
@@ -127,9 +161,25 @@ impl StratPool {
...
+            }
+        }
+
+        let mut result = Vec::new();
+        for name in names.iter() {
+            let uuid = Uuid::new_v4();
+            let dm = try!(DM::new());
+            let new_filesystem = try!(StratFilesystem::new(uuid, name, &dm, &mut self.thin_pool));
+            self.filesystems.insert(new_filesystem);
+            result.push((**name, uuid));
```
mulkieran: Might name only need one dereference operator, ```*name``` instead of ```**name```? I can't figure out why it should be two.

src/engine/strat_engine/pool.rs
trgill: I moved the check to initialize.  (?)

src/engine/strat_engine/pool.rs
trgill: with a single dereference:

mismatched types [E0308]:
expected type `std::vec::Vec<(&str, uuid::Uuid)>`
   found type `std::vec::Vec<(&&str, uuid::Uuid)>`
expected str, found &str

src/engine/macros.rs
trgill: Ok.

src/engine/strat_engine/blockdev.rs
```diff
@@ -181,6 +183,12 @@ pub fn initialize(pool_uuid: &PoolUuid,
         Ok(add_devs)
     }
 
+    // TODO: Fix this code.  We should deal with any number of blockdevs
```
mulkieran: You don't want to check the length of ```devices```. It's the length of ```add_devs``` that matters.

src/engine/strat_engine/pool.rs
mulkieran: OK. Currently it is guaranteed that bds.len() is at least 2. That's tricky, though. Seems like its worth an assert and another TODO linking it to the check in ```initialize()```.

Also,  maybe tighten up block every so slightly, as:

```
let mut data_devs = Vec::from_iter(bds.values());
meta_dev = try!(LinearDev::new(..., ..., &vec![data_devs.remove(0)]);
data_dev = ...
length = ...
```

src/engine/strat_engine/pool.rs
mulkieran: Puzzling, but it is probably nothing to worry about.

src/engine/strat_engine/blockdev.rs
trgill: Ok. Thanks.

src/engine/strat_engine/metadata.rs
```diff
@@ -34,7 +34,7 @@ pub const MIN_MDA_SECTORS: Sectors = Sectors(2032);
```
mulkieran: Is it possible to drop the cloning of the metadata structures?

src/engine/strat_engine/thindev.rs
```diff
@@ -68,11 +77,15 @@ impl ThinDev {
...
     }
 
-    pub fn teardown(&mut self, dm: &DM) -> EngineResult<()> {
+    pub fn teardown(&self, dm: &DM) -> EngineResult<()> {
         try!(dm.device_remove(&DevId::Name(&self.name), DmFlags::empty()));
         Ok(())
     }
 
+    pub fn name(&self) -> &str {
+        self.dev_info.name().clone()
```
mulkieran: Why is it necessary to clone the result of ```name()``` in this and equivalent methods?

trgill: Ok.  I was thinking it would be best to not return references to strings.  I think it is like returning a pointer to a string inside a C structure - typically not a good idea for an API in C.  But, I think you are indicating  rust lifetimes will protect us... (?)

src/engine/strat_engine/metadata.rs
trgill: done

agrover:
👍

PR #264, opened 2017-03-15
Adjust Table implementation

mulkieran:
Take into account that name may change, but UUID is constant.
Make T not implement name(), and by implication, not store its own name.
By this means, a variety of bugs of the kind where there are two things
with one name in a Table are prevented. The bugs would not arise by
insertion or removal, which works properly in previous version, but by renaming
of an element within the table.

This preserves theoretical O(1) of all operations. It is possible to
come up with simpler implementations that are O(n), but they do not
avoid the problem of duplicate names on rename any better.

### DIFF
--- a/src/engine/structures.rs
+++ b/src/engine/structures.rs
@@ -12,17 +12,15 @@ use super::engine::{HasName, HasUuid};
 
 /// Map UUID and name to T items.
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


mulkieran:
This is WIP, expected to fail.

agrover:
closing due to inactivity.

PR #266, opened 2017-03-15
Fail on warnings in tests as well as for regular builds

mulkieran:
Otherwise, we don't catch errors in our test-only code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ build:
 	RUSTFLAGS='-D warnings' cargo build
 
 test:
+	RUSTFLAGS='-D warnings' \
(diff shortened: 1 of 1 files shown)


mulkieran:
Sound, we hope...but not exactly complete is the Rust analysis engine.

mulkieran:
There seems to be a class of unnecessary mut modifiers that the Rust compiler can not detect. This isn't necessarily a bug. It would be a bug, if we needed a mut modifier and the compiler couldn't detect that. So, I don't feel there is anything to report.

On the other hand, we want tests to fail on all warnings that the compiler can detect, and that makes this happen.

So, we might as well put it in.

agrover:
👍, merging as last reviewer

PR #268, opened 2017-03-15
Put helpful comments on all unconditional unwraps

mulkieran:
Either explain why they must succeed, or leave a TODO for further work.

### DIFF
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -77,6 +77,8 @@ fn main() {
             builder.parse(&s);
         }
     };
+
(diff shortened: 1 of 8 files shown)


agrover:
The places where we know the call must succeed (e.g. logged, claiming dbus) could we use `expect()`? Does the same thing as unwrap, but with a message.

mulkieran:
That's not how I see ```expect``` used in examples. We should propose a new method for Option and Result, ```surprise()``` or maybe ```surprised_by()```.

mulkieran:
Superceded by #269.

PR #269, opened 2017-03-16
unwrap cleanup in source code

mulkieran:
Change all unwrap() calls to expect() calls wherever it is possible to
explain why the unwrap() must succeed.

Propagate an error in some cases where unwrap() is inappropriate.

Leave a TODO on any remaining unwraps().

Do a small amount of tidying.


mulkieran:
Found one more in structures.rs.

PR #275, opened 2017-03-22
Zero the superblock on meta_data device before it is passed to thin-pool

trgill:
When constructing a thin-pool, Stratis reserves the first n sectors on a block device by creating a
DM linear device with a starting offset.  DM writes the super block in the first block.  DM requires 
this first block to be zeros when the meta data for the thin-pool is initially created.  If we don't zero
the superblock DM issues very confusing error messages because it triggers code paths that are trying to re-adopt the device with the attributes that have been passed.

Add method to zero a sector range at a given offset.
Removed un-needed "mut" for &self parameter on wipe_metadata.
Added constant DATA_BLOCK_SIZE

### DIFF
--- a/src/engine/strat_engine/blockdev.rs
+++ b/src/engine/strat_engine/blockdev.rs
@@ -242,7 +242,7 @@ impl BlockDev {
         }
     }
 
-    pub fn wipe_metadata(&mut self) -> EngineResult<()> {
+    pub fn wipe_metadata(&self) -> EngineResult<()> {
         let mut f = try!(OpenOptions::new().write(true).open(&self.devnode));
         BDA::wipe(&mut f)
     }
--- a/src/engine/strat_engine/blockdev.rs
+++ b/src/engine/strat_engine/blockdev.rs
@@ -7,6 +7,7 @@ use std::collections::BTreeSet;
 use std::collections::HashMap;
 use std::fs::File;
 use std::io::ErrorKind;
+use std::io::{Seek, Write, SeekFrom};
 use std::fs::{OpenOptions, read_dir};
 use std::os::unix::prelude::AsRawFd;
 use std::path::{Path, PathBuf};
@@ -95,7 +96,22 @@ pub fn find_all() -> EngineResult<HashMap<PoolUuid, HashMap<DevUuid, BlockDev>>>
     Ok(pool_map)
 }
 
+/// Zero sectors at the given offset
+pub fn wipe_sectors(path: &Path, offset: Sectors, sector_count: Sectors) -> EngineResult<()> {
+    let mut f = try!(OpenOptions::new()
+        .write(true)
+        .open(path));
 
+    let zeroed = [0u8; SECTOR_SIZE];
+
+    // set the start point to the offset
+    try!(f.seek(SeekFrom::Start(*offset)));
+    for _ in 0..*sector_count {
+        try!(f.write_all(&zeroed));
+    }
+    try!(f.flush());
+    Ok(())
+}
 
 /// Initialize multiple blockdevs at once. This allows all of them
 /// to be checked for usability before writing to any of them.
--- a/src/engine/strat_engine/pool.rs
+++ b/src/engine/strat_engine/pool.rs
@@ -203,13 +203,13 @@ impl Pool for StratPool {
         Ok(bdev_paths)
     }
 
-    fn destroy(mut self) -> EngineResult<()> {
+    fn destroy(self) -> EngineResult<()> {
 
         // TODO Do we want to create a new File each time we interact with DM?
         let dm = try!(DM::new());
         try!(self.thin_pool.teardown(&dm));
 
-        for bd in self.block_devs.values_mut() {
+        for bd in self.block_devs.values() {
             try!(bd.wipe_metadata());
         }
 
--- a/src/engine/strat_engine/pool.rs
+++ b/src/engine/strat_engine/pool.rs
@@ -22,6 +22,7 @@ use engine::Filesystem;
 use engine::Pool;
 use engine::RenameAction;
 use engine::engine::Redundancy;
+use engine::strat_engine::blockdev::wipe_sectors;
 use engine::strat_engine::lineardev::LinearDev;
 use engine::strat_engine::thinpooldev::ThinPoolDev;
 
@@ -33,8 +34,9 @@ use super::blockdev::{BlockDev, initialize, resolve_devices};
 use super::filesystem::StratFilesystem;
 use super::metadata::MIN_MDA_SECTORS;
 
-use types::DataBlocks;
-use types::Sectors;
+use types::{DataBlocks, Sectors};
+
+pub const DATA_BLOCK_SIZE: Sectors = Sectors(2048);
 
 #[derive(Debug)]
 pub struct StratPool {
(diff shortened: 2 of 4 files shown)


src/engine/strat_engine/blockdev.rs
```diff
@@ -231,7 +248,7 @@ impl BlockDev {
         }
     }
 
-    pub fn wipe_metadata(&mut self) -> EngineResult<()> {
+    pub fn wipe_metadata(&self) -> EngineResult<()> {
```
agrover: I'm wondering if we actually want this to take `self`, i.e. consume the blockdev. Would this work?

src/engine/strat_engine/blockdev.rs
```diff
@@ -97,7 +98,23 @@ pub fn find_all() -> EngineResult<BTreeMap<PoolUuid, BTreeMap<DevUuid, BlockDev>
...
+    let mut f = try!(OpenOptions::new()
+        .read(true)
+        .write(true)
+        .open(path));
 
+    let zeroed = [0u8; SECTOR_SIZE];
+
+    // set the start point to the offset
+    try!(f.seek(SeekFrom::Start(offset.0)));
+    for _i in 0..sector_count.0 {
```
agrover: can this just be `_` instead of `_i`?

src/engine/strat_engine/blockdev.rs
```diff
@@ -97,7 +98,23 @@ pub fn find_all() -> EngineResult<BTreeMap<PoolUuid, BTreeMap<DevUuid, BlockDev>
     Ok(pool_map)
 }
 
+/// Zero sectors at the given offset
+pub fn wipe_sectors(path: &PathBuf, offset: &Sectors, sector_count: &Sectors) -> EngineResult<()> {
+    let mut f = try!(OpenOptions::new()
+        .read(true)
```
mulkieran: You don't need to read, do you?

src/engine/strat_engine/blockdev.rs
```diff
@@ -97,7 +98,23 @@ pub fn find_all() -> EngineResult<BTreeMap<PoolUuid, BTreeMap<DevUuid, BlockDev>
...
+        .write(true)
+        .open(path));
 
+    let zeroed = [0u8; SECTOR_SIZE];
+
+    // set the start point to the offset
+    try!(f.seek(SeekFrom::Start(offset.0)));
```
mulkieran: These ```x.0```s should be ```*x``` s.

src/engine/strat_engine/blockdev.rs
```diff
@@ -97,7 +98,23 @@ pub fn find_all() -> EngineResult<BTreeMap<PoolUuid, BTreeMap<DevUuid, BlockDev>
     Ok(pool_map)
 }
 
+/// Zero sectors at the given offset
+pub fn wipe_sectors(path: &PathBuf, offset: &Sectors, sector_count: &Sectors) -> EngineResult<()> {
```
mulkieran: The method will be more general if it takes a reference to a Path, rather than a PathBuf.

src/consts.rs
```diff
@@ -15,3 +17,4 @@ pub mod IEC {
 }
 
 pub const SECTOR_SIZE: usize = 512;
+pub const DATA_BLOCK_SIZE: Sectors = Sectors(2048);
```
mulkieran: Please define ```DATA_BLOCK_SIZE``` in the only module where it is used. I've been trying to limit the contents of this file to globally used constants.

src/engine/strat_engine/pool.rs
```diff
@@ -71,19 +73,20 @@ impl StratPool {
         {
             let mut data_devs = Vec::from_iter(bds.values());
 
-            meta_dev = try!(LinearDev::new(&format!("stratis_{}_meta", name),
-                                           dm,
-                                           &vec![data_devs.remove(0)]));
+            let meta_blockdev = data_devs.remove(0);
+            meta_dev =
+                try!(LinearDev::new(&format!("stratis_{}_meta", name), dm, &vec![meta_blockdev]));
+            try!(wipe_sectors(&try!(meta_dev.path()), &Sectors(0), &DATA_BLOCK_SIZE));
```
mulkieran: You probably need a comment here, to inform posterity why you had to wipe these sectors.

src/engine/strat_engine/blockdev.rs
mulkieran: ```sector_count: &Sectors``` -> ```length: Sectors```

mulkieran: There's currently no point in using references to Sectors for parameter types, since they are implicitly Copy.

src/engine/strat_engine/blockdev.rs
mulkieran: I dunno if that would be a good idea or not.

But it would definitely be correct to update ```StratPool::destroy()``` to accompany this change. ```self.block_devs.values_mut()``` can surely go to ```self.block_devs.values()```. Maybe ```StratPool::destroy()``` doesn't need to take a mutable reference either.

src/engine/strat_engine/blockdev.rs
mulkieran: There's a ```wipe_header()``` method in tests/utiils/blockdev_utils.rs that can probably be replaced by this method.

src/engine/strat_engine/blockdev.rs
trgill: Ok.  I made the updates.

src/engine/strat_engine/blockdev.rs
trgill: yup.  done.

src/engine/strat_engine/blockdev.rs
mulkieran: I bet if you empty ```self.block_devs``` in ```StratPool::destroy()```, i.e., consume the elements instead of iterating through them it will pass the compiler.

src/engine/strat_engine/pool.rs
```diff
@@ -2,6 +2,8 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+pub const DATA_BLOCK_SIZE: Sectors = Sectors(2048);
```
mulkieran: Please put const below all the ```use``` directives.

src/engine/strat_engine/blockdev.rs
```diff
@@ -225,7 +241,7 @@ impl BlockDev {
         }
     }
 
-    pub fn wipe_metadata(&mut self) -> EngineResult<()> {
+    pub fn wipe_metadata(&self) -> EngineResult<()> {
```
mulkieran: I bet if you empty ```self.block_devs``` in ```StratPool::destroy()```, i.e., consume the elements instead of iterating through them it will pass the compiler.

src/engine/strat_engine/pool.rs
trgill: Ok.

src/engine/strat_engine/blockdev.rs
trgill: block_devs is immutable because &self is immutable.  so I can't move items out of block_devs, right?

mulkieran: ```Pool::destroy()``` consumes self, not a reference to self, so I think you can do a whole lot in the body of the method.

trgill: Opened issue https://github.com/stratis-storage/stratisd/issues/282 to track the issue.

agrover:
👍

mulkieran:
Please separate out update to Cargo.lock. It should have gone in with a previous commit in a different PR.

PR #276, opened 2017-03-23
Refactor D-Bus API to use decorated object paths

mulkieran:


mulkieran:
Some commits picked from #274.

This PR is principally just a refactoring to use object path data instead of look-aside table. It probably does that pretty well.

But, I think I"ve realized that a redesign of some of the architecture would be desirable, and if that is the case, it might be better to add the redesign before merging the PR.

Here are the points:
* The first item in the OPContext, the parent object path, is probably bad. The reason is that it assumes that their must be a unique parent in some sort of tree structure. We may find other relationships that objects have to each other, and we should allow for that. It would be better for a filesystem to publish it's pool's UUID, than to publish it's pool's object path as it's one unique parent.

What is the solution:
* Have the filesystem's Pool property be the pool UUID, not the pool object path.
* In the engine, use a global table of filesystems, not a per-pool table. This table is single-key, UUID to filesystem.
* Have each pool store a two-key table that just, essentially maps names to UUIDs and vice-versa.
* Do not have the Filesystem struct store a name, but do have it store a pool UUID.
* Make the OPContext just be Uuid. The base tree gets None, because it needs no UUID.

Below are illustrations of how this would work from the client side and the engine side:
* CreatePool
  - client side - nothing to do, the client locates the manager object and passes a bunch of arguments
  - server-side - the pool object path, with pool UUID, is added to tree
* DestroyPool
  - client side - locate pool based on its Name and get its object path, easy.
  - server side - look up the pool object path in the tree and get it's UUID from the data. Call engine method with UUID.
* Filesystme's Rename
  - client side - locate the filesystem by pool name and filesystem name. First get the pool object path by its name. Then get the pool  UUID by its object path. Then locate the filesystem object path by name and pool UUID.
  - server side - from the filesystem object path get its UUID. Look the filesystem up in the global table and gets its pool UUID. Use the pool UUID to lookup the pool in the engine. Invoke the pool rename method passing it the filesystem UUID. This method need only change the in-pool data structures that map UUID to name.
* Filesystem's Name property
  - server side - Get the filesystems UUID from the object path. Look up the filesystem in the global table to get its Pool UUID. Use the pool UUID to get the pool object from the engine. Invoke a pool method that returns the filesystem name.
* Pool's CreateFilesystems
  - client side - Just locate the pool object path by its name.
  - server side - get the pool UUID from the pool object path. Invoke an engine method passing the pool UUID and the filsystem specs. Insert items into the global table and the pool specific table. Build the apprpriate filesystem objects.
* Pool's DestroyFilesystems:
  - client side - Locate the pool object path by its name. Get the pool UUID. Use the pool UUID and the filesystem names to get the filesystem object paths.
  - server side - Use the pool object path to get the pool UUID. Use the filesystem object paths to get the filesystem UUIDs. Invoke an engine method passing the pool UUID and filesystem UUIDs. The engine method can remove filesystems from the global table and invoke pool methods to update the pool table. It needs to do something if the filesystems to be removed do not belong to the pool, like error.
There is an alternative or addition that would lift DestroyFilesystems up to the engine. This method would only take Filesystem UUIDs and could destroy filesystems belonging to any pool.

* Pool's Name property:
  - server side - no change

mulkieran:
The engine would also store a global device table, again mapping from UUIDs to device information. In the possible event where we have devices that do not belong to a particular pool, this would be handy. (The metadata has a pool id, but a special value could be used for no pool.)

mulkieran:
@agrover @trgill review? Since this is just a refactoring it might as well go in now, and the other questions can be resolved in subsequent PRs.

PR #278, opened 2017-03-23
Master object path rest

mulkieran:
Supercedes #274. Extends #276.


mulkieran:
This is some preliminary work on adding devices to the D-Bus API.

mulkieran:
This is almost entirely obsoleted, except for the work on Tables, which can be set aside, possibly indefinitely.

PR #281, opened 2017-03-27
No longer pin to lower rustfmt version

mulkieran:
The bug that we were triggering was fixed in the newest version.

https://github.com/rust-lang-nursery/rustfmt/issues/1280


mulkieran:
still broken.

PR #285, opened 2017-03-28
Move Sectors, Bytes and DataBlocks from Stratis into devicemapper-rs

trgill:
Updated the "use" statements for moved Sectors, Bytes and DataBlocks
The removed references to newtype_derive. newtype_derive is not
required after changes.
Bumped the devicemapper crate version from "0.4.0" to "0.4.1".
The "0.4.1" version has the changes required.

PR #288, opened 2017-03-31
Put all unshared functions in the files where they are used

mulkieran:
That way it is not necessary to call them dead.
Reorder imports, in files where it is necessary to add or remove them
anyway.


mulkieran:
@trgill review? If you are OK with this, it can go in today. It'll probably conflict w/  #285, because of ordering of imports.

PR #291, opened 2017-04-03
Switch to DBus system bus

agrover:
This is a 1-line change to the code, but won't work unless we tell dbus-daemon we're allowed to be on the system bus by copying a special file. Add this file and describe where it needs to be copied in the README.

First, also add some additional development info to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,11 @@ licensed under the MPL 2.0 to be merged by us.
 
 ### Setting up for development
 
+Stratisd runs as root, and requires access to the D-Bus system bus. Thus in
+order to work properly, a D-Bus conf file must exist to grant access, either
+installed by distribution packaging; or manually, by copying `stratisd.conf`
+to `/etc/dbus-1/system.d/`.
+
 Stratisd requires Rust 1.15.1+ and Cargo to build. These may be available via
 your distribution's package manager. If not, [Rustup](https://www.rustup.rs/)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -40,3 +38,18 @@ question.
...
+Stratisd runs as root, and requires access to the D-Bus system bus. Thus in
+order to work properly, a D-Bus conf file must exist to grant access, either
+installed by distribution packaging; or manually, by copying `stratisd.conf`
+to `/etc/dbus-1/system.d/`.
+
+Stratisd requires Rust 1.15.1+ and Cargo to build. These may be available via
+your distribution's package manager. If not, [Rustup](https://www.rustup.rs/)
+is available to install and update the Rust toolchain.
+
+Once toolchain is in place, run `cargo build` to build, and then run the
```
mulkieran: What about ```cargo run -- --help``` instead of instructions for explicitly starting stratisd from debug path?

agrover: stratisd need to be run as root. Running through Cargo as root (e.g. `sudo cargo run -- --debug`) recompiles everything as root: uses /root/.cargo/config, generated files are owned by root etc., which is not desired. I think it's better to compile as a user, then only run as root.

mulkieran: That makes sense. You might still want to tell them to use the ```--help``` flag instead of the ```--debug``` flag, you never know what we might change about the flags, but help should stick.

PR #292, opened 2017-04-04
Respect force flag w/ Stratis pools

mulkieran:
If force flag is true, we should initialize the blockdev belonging
to some other Stratis pool.

### DIFF
--- a/src/engine/strat_engine/blockdev.rs
+++ b/src/engine/strat_engine/blockdev.rs
@@ -181,10 +181,14 @@ pub fn initialize(pool_uuid: &PoolUuid,
                 }
                 DevOwnership::Ours(uuid) => {
                     if *pool_uuid != uuid {
-                        let error_str = format!("Device {} already belongs to Stratis pool {}",
-                                                devnode.display(),
-                                                uuid);
-                        return Err(EngineError::Engine(ErrorEnum::Invalid, error_str));
+                        if !force {
+                            let error_str = format!("Device {} already belongs to Stratis pool {}",
+                                                    devnode.display(),
+                                                    uuid);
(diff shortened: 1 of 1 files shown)


mulkieran:
@agrover @trgill review? The code does what it is supposed to, but the question is, am I fixing or introducing a bug?

The theory is that these Stratis owned devices being initialized for the pool may have come from somewhere else entirely. Stratis can not reconstruct their pool and then uninitialize them, so Stratis should be allowed to force initialize them.

Another take is that Stratis should have the ability to reconstruct a partial pool and in deleting that pool, remove the Stratis info from the disk. That's tricky, though.

trgill:
We can't allow force to work on a disk that is in a valid/working Stratis pool, right?  We'd need to clean up DM if we allowed it.  

What does this code do when you try to force a new pool when it is already part of one?  I'd think we don't currently deal with the errors that would be returned from DM.

mulkieran:
Stealing a disk from an existing Stratis pool would cause chaos. But, when you're using force, you're kind of insisting that you know what you're doing and have evaluated the consequences.

agrover:
I think that Stratis should never reuse a blockdev that's currently part of an active pool, but if a blockdev has a Stratis signature but is not part of an active pool on this system, then --force should allow it to be added to an active pool.

Checking this likely needs to happen at a higher level than Blockdev, so this code *could* be correct if the checking is happening elsewhere.

mulkieran:
@agrover That seems like the correct approach. It will require moving ```Pool::add_blockdevs``` up to engine, which I don't expect will be trivial. In the meantime, I'll let this PR sit.

mulkieran:
There's still an issue here, but this PR is entirely obsolete.

PR #296, opened 2017-04-04
add lib to create loopback devices and rewrite all tests to use loopback

mulkieran:
Use it to implement a few suitable tests. Remove tests that have been
made redundant because they're being tested by means of loopbacked devices.


mulkieran:
The array of LoopDeviceSpecs doesn't contain any information and could currently be turned into a u8 without ill effects. Other than that, I think that this is ready to go.

Makefile
```diff
@@ -9,10 +10,20 @@ fmt: ${HOME}/.cargo/bin/cargo-fmt
 build:
 	RUSTFLAGS='-D warnings' cargo build
 
+# Tests are in order of complexity, from least to greatest.
+test-loop:
+	sudo env "PATH=${PATH}" RUSTFLAGS='-D warnings' cargo test -- --test test_force_flag_stratis
+	sudo env "PATH=${PATH}" RUSTFLAGS='-D warnings' cargo test -- --test test_linear_device
+	sudo env "PATH=${PATH}" RUSTFLAGS='-D warnings' cargo test -- --test test_thinpool_device
+	sudo env "PATH=${PATH}" RUSTFLAGS='-D warnings' cargo test -- --test test_pool_blockdevs
```
agrover: can you put in the commitlog how the four old tests match up with the newly named four tests? Also, just generally discuss how the new test implementations differ from the old tests. They should differ in style but still cover exactly the same cases, is this true?

mulkieran:
@agrover @trgill review?

Makefile
```diff
@@ -9,10 +10,20 @@ fmt: ${HOME}/.cargo/bin/cargo-fmt
...
+
 test:
 	RUSTFLAGS='-D warnings' \
-	cargo test -- --skip test_pools --skip test_blockdev_setup \
-		--skip test_lineardev_setup --skip test_thinpool
+	cargo test -- \
+		--skip test_force_flag_stratis \
+		--skip test_linear_device \
+		--skip test_pool_blockdevs \
+		--skip test_thinpool_device
```
mulkieran: I put a comment in the Makefile:

```Do loop tests last, since they require sudo (and change permissions)```

mulkieran: And I amended the comment to be more explanatory.

Makefile
mulkieran: Done.

mulkieran:
@agrover We can do as you suggest, but I think we should defer solving that problem until we have one concrete example to work with.

mulkieran:
There are actually two issues.
The first is making it possible to run the existing tests either on loopbacked devices or on real devices. The existing tests have a property that they are not particular about what kind of devices they get, they just need "enough". So, setting something up so that they can be run on either real or loopbacked ought to be pretty simple. I'm trying out an idea where the code that tests the properties is shared, but the frameworks are separated. If the tests are to be run using loopbacked devices they have a name like ```loop_<name>``` if on real devices they get asked for explicitly as ```real_<name>```. Anything more complicated, like looking for real devices and then falling back on loopbacked devices if no real devices are found, seems like unhelpful complexity to me.

The second is devices that may _require_ to be run on real devices. Those are quite possible, but I don't expect would be handled by this framework, and worrying about them should be deferred until later.

mulkieran:
Superceded by #298.

Issue #282, opened 2017-03-27
fn wipe_metadata(&self) -> EngineResult<()> 


trgill:
Should this consume self rather than take a reference?

PR #301, opened 2017-04-07
Make final methods consume self

mulkieran:
Closes: #282.

These final methods are teardown() for DM struct and wipe_metadata() for
blockdevs. We believe that with both these methods, once the method action
is performed that we should not be using the struct anymore.

Remove some actions in a test that did a use after a method call.

Update StratPool::destroy() for consuming wipe_metadata() method.

### DIFF
--- a/src/engine/strat_engine/pool.rs
+++ b/src/engine/strat_engine/pool.rs
@@ -201,13 +201,13 @@ impl Pool for StratPool {
         Ok(bdev_paths)
     }
 
-    fn destroy(self) -> EngineResult<()> {
+    fn destroy(mut self) -> EngineResult<()> {
 
         // TODO Do we want to create a new File each time we interact with DM?
         let dm = try!(DM::new());
         try!(self.thin_pool.teardown(&dm));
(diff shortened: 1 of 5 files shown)


mulkieran:
Rebased.

src/engine/strat_engine/pool.rs
```diff
@@ -207,7 +207,8 @@ impl Pool for StratPool {
         let dm = try!(DM::new());
         try!(self.thin_pool.teardown(&dm));
 
-        for bd in self.block_devs.values() {
+        let mut blockdevs = self.block_devs;
+        for (_, bd) in blockdevs.drain() {
```
agrover: style: don't need the let binding if you change to `fn destroy(mut self)` above?

In any case 👍 I think moving self into destroy/teardown() makes intuitive sense.

mulkieran:
mut on self parameter seems to work as well...

mulkieran:
@trgill review? This is ready to go and agrover has already acked it.

PR #304, opened 2017-04-13
Explore what happens when logging is enabled in tests

mulkieran:
This is only a test.


mulkieran:
This can be closed. The purpose was to get an easily referred to glimpse of what logging actually looks like.

PR #312, opened 2017-04-25
Update to use devicemapper-rs 4.7

trgill:
Minor adjustment to the use statements for re-organization of
devicemapper-rs.


agrover:
the just-released devicemapper-rs 0.4.7 should require fewer changes, suggest moving to that version.

trgill:
Ok.  Yes - that required fewer changes.   Now, there is just one change to the 
ThinPoolStatus::Good() constructor. (and the Cargo.toml update for the new version).

PR #314, opened 2017-04-26
Block allocation improvements

agrover:
This PR lets us more easily track the used segments of each blockdev, and also adds a BlockDevMgr, to let us more easily ask for segments across all the disks from which to construct lineardevs.

There is some work that builds on this to use this in StratPool::new, but there was a small change to devicemapper API that I was hoping to land that makes things a little easier.

### DIFF
--- a/src/engine/strat_engine/blockdev.rs
+++ b/src/engine/strat_engine/blockdev.rs
@@ -26,6 +26,7 @@ use engine::{DevUuid, EngineResult, EngineError, ErrorEnum, PoolUuid};
 use super::metadata::{StaticHeader, BDA, validate_mda_size};
 use super::engine::DevOwnership;
 pub use super::BlockDevSave;
+use engine::strat_engine::range_alloc::RangeAllocator;
 
 const MIN_DEV_SIZE: Bytes = Bytes(IEC::Gi as u64);
 
@@ -66,10 +67,15 @@ pub fn find_all() -> EngineResult<HashMap<PoolUuid, HashMap<DevUuid, BlockDev>>>
 
         if let Some(bda) = BDA::load(&mut f).ok() {
             let dev = try!(Device::from_str(&devnode.to_string_lossy()));
+            // TODO: Parse MDA and also initialize RangeAllocator with
+            // in-use regions
+            let allocator = RangeAllocator::new_with_used(bda.dev_size(),
+                                                          &[(Sectors(0), bda.size())]);
             Ok(Some(BlockDev {
                 dev: dev,
                 devnode: devnode.to_owned(),
                 bda: bda,
+                used: allocator,
             }))
         } else {
             Ok(None)
@@ -219,11 +225,13 @@ pub fn initialize(pool_uuid: &PoolUuid,
                                        &Uuid::new_v4(),
                                        mda_size,
                                        dev_size.sectors()));
+        let allocator = RangeAllocator::new_with_used(bda.dev_size(), &[(Sectors(0), bda.size())]);
 
         let bd = BlockDev {
             dev: dev,
             devnode: devnode.clone(),
             bda: bda,
+            used: allocator,
         };
         bds.push(bd);
     }
@@ -236,6 +244,7 @@ pub struct BlockDev {
     dev: Device,
     pub devnode: PathBuf,
     bda: BDA,
+    used: RangeAllocator,
 }
 
 impl BlockDev {
@@ -307,5 +316,15 @@ impl BlockDev {
     /// Last time metadata was written to this device.
     pub fn last_update_time(&self) -> Option<&Timespec> {
         self.bda.last_update_time()
+    }
+
+    pub fn available(&self) -> Sectors {
+        self.used.available()
+    }
+
+    // Find some sector ranges that could be allocated. If more
+    // sectors are needed than our capacity, return partial results.
+    pub fn request_space(&mut self, size: Sectors) -> (Sectors, Vec<(Sectors, Sectors)>) {
+        self.used.request(size)
     }
 }
--- a/src/engine/strat_engine/blockdev.rs
+++ b/src/engine/strat_engine/blockdev.rs
@@ -2,6 +2,8 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+// Code to handle a single block device.
+
 use std::io;
 use std::collections::{HashMap, HashSet};
 use std::fs::File;
@@ -241,7 +243,7 @@ pub fn initialize(pool_uuid: &PoolUuid,
 
 #[derive(Debug)]
 pub struct BlockDev {
-    dev: Device,
+    pub dev: Device,
     pub devnode: PathBuf,
     bda: BDA,
     used: RangeAllocator,
--- a/src/engine/strat_engine/blockdev.rs
+++ b/src/engine/strat_engine/blockdev.rs
@@ -212,13 +212,6 @@ pub fn initialize(pool_uuid: &PoolUuid,
 
     let add_devs = try!(filter_devs(dev_infos, pool_uuid, force));
 
-    // TODO: Fix this code.  We should deal with any number of blockdevs
-    //
-    if add_devs.len() < 2 {
-        return Err(EngineError::Engine(ErrorEnum::Error,
-                                       "Need at least 2 blockdevs to create a pool".into()));
-    }
-
     let mut bds = Vec::new();
     for (dev, (devnode, dev_size, mut f)) in add_devs {
 
--- a/src/engine/strat_engine/blockdev.rs
+++ b/src/engine/strat_engine/blockdev.rs
@@ -199,6 +199,10 @@ pub fn initialize(pool_uuid: &PoolUuid,
                                                 devnode.display(),
                                                 uuid);
                         return Err(EngineError::Engine(ErrorEnum::Invalid, error_str));
+                    } else {
+                        // Already in this pool (according to its header)
+                        // TODO: Check we already know about it
+                        // if yes, ignore. If no, add it w/o initializing?
                     }
                 }
             }
--- a/src/engine/strat_engine/blockdevmgr.rs
+++ b/src/engine/strat_engine/blockdevmgr.rs
@@ -1,0 +1,135 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// Code to handle a collection of block devices.
+
+use std::collections::HashMap;
+use std::path::{Path, PathBuf};
+
+use devicemapper::{Sectors, Segment};
+use time::Timespec;
+
+use engine::{EngineResult, PoolUuid};
+use super::metadata::MIN_MDA_SECTORS;
+pub use super::BlockDevSave;
+use engine::strat_engine::blockdev::{BlockDev, resolve_devices, initialize};
+
+#[derive(Debug)]
+pub struct BlockDevMgr {
+    pub block_devs: HashMap<PathBuf, BlockDev>,
+}
+
+impl BlockDevMgr {
+    pub fn new(block_devs: Vec<BlockDev>) -> BlockDevMgr {
+        BlockDevMgr {
+            block_devs: block_devs.into_iter().map(|bd| (bd.devnode.clone(), bd)).collect(),
+        }
+    }
+
+    pub fn add(&mut self,
+               pool_uuid: &PoolUuid,
+               paths: &[&Path],
+               force: bool)
+               -> EngineResult<Vec<PathBuf>> {
+        let devices = try!(resolve_devices(paths));
+        let bds = try!(initialize(pool_uuid, devices, MIN_MDA_SECTORS, force));
+        let bdev_paths = bds.iter().map(|p| p.devnode.clone()).collect();
+        for bd in bds {
+            self.block_devs.insert(bd.devnode.clone(), bd);
+        }
+        Ok(bdev_paths)
+    }
+
+    pub fn destroy_all(mut self) -> EngineResult<()> {
+        for (_, bd) in self.block_devs.drain() {
+            try!(bd.wipe_metadata());
+        }
+        Ok(())
+    }
+
+    // Unused space left on blockdevs
+    pub fn avail_space(&self) -> Sectors {
+        self.block_devs.values().map(|bd| bd.available()).sum()
+    }
+
+    /// If available space is less than size, return None, else return
+    /// the segments allocated.
+    pub fn alloc_space(&mut self, size: Sectors) -> Option<Vec<Segment>> {
+        let mut needed: Sectors = size;
+        let mut segs = Vec::new();
+
+        if self.avail_space() < size {
+            return None;
+        }
+
+        for mut bd in self.block_devs.values_mut() {
+            if needed == Sectors(0) {
+                break;
+            }
+
+            let (gotten, r_segs) = bd.request_space(needed);
+            segs.extend(r_segs.iter()
+                .map(|&(start, len)| Segment::new(bd.dev, start, len)));
+            needed = needed - gotten;
+        }
+
+        assert_eq!(needed, Sectors(0));
+
+        Some(segs)
+    }
+
+    pub fn devnodes(&self) -> Vec<PathBuf> {
+        self.block_devs.keys().map(|p| p.clone()).collect()
+    }
+
+    pub fn to_save(&self) -> HashMap<String, BlockDevSave> {
+        self.block_devs
+            .iter()
+            .map(|(_, bd)| (bd.uuid().simple().to_string(), bd.to_save()))
+            .collect()
+    }
+
+    /// Write the given data to all blockdevs marking with specified time.
+    // TODO: Cap # of blockdevs written to, as described in SWDD
+    pub fn save_state(&mut self, time: &Timespec, metadata: &[u8]) -> EngineResult<()> {
+        // TODO: Do something better than panic when saving to blockdev fails.
+        // Panic can occur for a the usual IO reasons, but also:
+        // 1. If the timestamp is older than a previously written timestamp.
+        // 2. If the variable length metadata is too large.
+        for bd in self.block_devs.values_mut() {
+            bd.save_state(time, metadata).unwrap();
+        }
+        Ok(())
+    }
+
+    /// Return the metadata from the first blockdev with up-to-date, readable
+    /// metadata.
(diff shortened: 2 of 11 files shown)


mulkieran:
I am in favor of implementing Step for Bytes and Sectors: https://github.com/stratis-storage/devicemapper-rs/pull/37.

agrover:
@mulkieran see my comments over there. While the implementation of RangeAllocator is in flux, I hope we can start using it now. This would allow us to proceed with development that builds upon it, while we also work to improve its implementation in parallel.

src/engine/strat_engine/range_alloc.rs
```diff
@@ -0,0 +1,230 @@
...
+        (amount - needed, segs)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_allocator() {
```
mulkieran: It is nice when you have a test to state the property it is supposed to verify.

src/engine/strat_engine/range_alloc.rs
```diff
@@ -0,0 +1,230 @@
...
+    pub fn available(&self) -> Sectors {
+        Sectors((0..self.max).filter(|val| !self.used.contains(val)).count() as u64)
+    }
+
+    pub fn used(&self) -> Sectors {
+        Sectors((0..self.max).filter(|val| self.used.contains(val)).count() as u64)
+    }
+
+    /// Get a list of (offset, length) segments that are in use
+    fn used_ranges(&self) -> Vec<(u64, u64)> {
```
trgill: You didn't use the devicemapper::Segment type for ranges because it has an un-needed path?  Maybe create a Range type?

src/engine/strat_engine/range_alloc.rs
```diff
@@ -0,0 +1,230 @@
...
+                };
+
+                (vec, new_range)
+            });
+
+        used_ranges
+    }
+
+    /// Get a list of (offset, length) segments that are not in use
+    fn avail_ranges(&self) -> Vec<(u64, u64)> {
```
trgill: The ranges are in sectors, right?  Why not a Vec of Sectors?

src/engine/strat_engine/blockdev.rs
```diff
@@ -328,3 +328,125 @@ impl BlockDev {
...
+    // Unused space left on blockdevs
+    pub fn avail_space(&self) -> Sectors {
+        self.block_devs
+            .values()
+            .map(|bd| bd.available())
+            .sum()
+    }
+
+    /// Allocate segments equal to the requested sectors from the
+    /// blockdevs. All or None.
```
mulkieran: The header comment is a bit cryptic. Maybe:

If available space is less than size, return None, else return the segments allocated.

src/engine/strat_engine/blockdev.rs
```diff
@@ -328,3 +328,125 @@ impl BlockDev {
         self.used.request(size)
     }
 }
+
```
mulkieran: I would prefer if BlockDevMgr got its own file.

src/engine/strat_engine/pool.rs
```diff
@@ -59,18 +63,23 @@ impl StratPool {
...
-        let meta_dev = try!(LinearDev::new(&format!("stratis_{}_meta", name),
-                                           dm,
-                                           &[bds[0].avail_range_segment()]));
+        let mut block_mgr =
+            BlockDevMgr::new(try!(initialize(&pool_uuid, devices, MIN_MDA_SECTORS, force)));
+
+        if block_mgr.avail_space() < INITIAL_META_SIZE + INITIAL_DATA_SIZE {
+            let short_bytes = *(INITIAL_META_SIZE + INITIAL_DATA_SIZE - block_mgr.avail_space()) *
+                              SECTOR_SIZE as u64;
+            block_mgr.remove_all().expect("cannot de-initialize blockdevs");
```
mulkieran: There is no reason to be confident that there is no IO error, so this should be either an unwrap()/TODO or some kind of try!().

src/engine/strat_engine/pool.rs
```diff
@@ -59,18 +63,23 @@ impl StratPool {
...
+            return Err(EngineError::Engine(ErrorEnum::Invalid,
+                                           format!("Initial total usable pool capacity too \
+                                                    small by {} bytes",
+                                                   short_bytes)));
+        }
+
+        let meta_regions = block_mgr.alloc_space(INITIAL_META_SIZE)
+            .expect("blockmgr must not fail, already checked for space");
+        let meta_dev = try!(LinearDev::new(&format!("stratis_{}_meta", name), dm, &meta_regions));
```
mulkieran: Can we formalize this naming scheme in the design doc?

tests/util/simple_tests.rs
```diff
@@ -99,9 +99,9 @@ pub fn test_force_flag_stratis(paths: &[&Path]) -> () {
     initialize(&uuid, unique_devices.clone(), MIN_MDA_SECTORS, false).unwrap();
     assert!(initialize(&uuid2, unique_devices.clone(), MIN_MDA_SECTORS, false).is_err());
 
-    // FIXME: once requirement that number of devices added be at least 2 is removed
-    // this should succeed.
-    assert!(initialize(&uuid, unique_devices.clone(), MIN_MDA_SECTORS, false).is_err());
+    // Attempting to initialize existing blockdevs into the pool they
```
mulkieran: Feel free to drop these comment lines, since the header is fixed. They're unnecessary.

agrover:
hi, here's a rebase with fixes just to the first commit (range_alloc.rs): changes are
* add docstrings
* use Sectors for params instead of u64
* call ::new from ::new_with_used
* use asserts instead of panic!() inside conditionals
* make insert_range into insert_ranges (taking a slice like remove_ranges does)
* only allow limit up to 2^64-1 (otherwise overflow causes bug in used_ranges()
* fix test name

I left the asserts around length = 0 and removing items not present, not using saturate because I believe these are indicators of bugs in the usage of this struct that we want to proactively catch. We should discuss further.

src/engine/strat_engine/range_alloc.rs
agrover: funny, there is Range support in unstable Rust, so we probably want to wait until that hits stable and then use it.

src/engine/strat_engine/range_alloc.rs
agrover: yup

src/engine/strat_engine/range_alloc.rs
agrover: improved it a little, if it still could be improved what do you suggest?

src/engine/strat_engine/range_alloc.rs
```diff
@@ -0,0 +1,230 @@
...
+
+    pub fn available(&self) -> Sectors {
+        Sectors((0..self.max).filter(|val| !self.used.contains(val)).count() as u64)
+    }
+
+    pub fn used(&self) -> Sectors {
+        Sectors((0..self.max).filter(|val| self.used.contains(val)).count() as u64)
+    }
+
+    /// Get a list of (offset, length) segments that are in use
```
agrover: I just looked at the code. Its focus seems to be on efficiently storing a specific numerical use case rather than providing ranges.

src/engine/strat_engine/range_alloc.rs
```diff
@@ -0,0 +1,230 @@
...
+        }
+    }
+
+    pub fn remove_ranges(&mut self, to_free: &[(Sectors, Sectors)]) -> () {
+        for &(off, len) in to_free {
+            self.check_for_overflow(*off, *len);
+            for val in *off..*off + *len {
+                let removed = self.used.remove(&val);
+                if removed == false {
+                    panic!(format!("tried to remove value {} not present in RangedSet", val));
```
agrover: see below, worried about this masking bugs.

src/engine/strat_engine/range_alloc.rs
```diff
@@ -0,0 +1,230 @@
...
+            let inserted = self.used.insert(val);
+            if inserted == false {
+                panic!(format!("inserted value {} already present in RangedSet", val));
+            }
+        }
+    }
+
+    pub fn remove_ranges(&mut self, to_free: &[(Sectors, Sectors)]) -> () {
+        for &(off, len) in to_free {
+            self.check_for_overflow(*off, *len);
```
agrover: see below, worried about this masking bugs.

src/engine/strat_engine/range_alloc.rs
```diff
@@ -0,0 +1,230 @@
...
+        if let None = off.checked_add(len) {
+            panic!("off+len overflows u64");
+        }
+        if off + len > self.max {
+            panic!("off+len greater than max allocator range");
+        }
+    }
+
+    pub fn insert_range(&mut self, off: Sectors, len: Sectors) -> () {
+        if len == Sectors(0) {
```
agrover: see below, worried about this masking bugs.

src/engine/strat_engine/range_alloc.rs
```diff
@@ -0,0 +1,230 @@
...
+        }
+
+        allocator
+    }
+
+    fn check_for_overflow(&self, off: u64, len: u64) {
+        if let None = off.checked_add(len) {
+            panic!("off+len overflows u64");
+        }
+        if off + len > self.max {
```
agrover: yup!

src/engine/strat_engine/range_alloc.rs
```diff
@@ -0,0 +1,230 @@
...
+
+        for &(off, len) in initial_used {
+            allocator.insert_range(off, len)
+        }
+
+        allocator
+    }
+
+    fn check_for_overflow(&self, off: u64, len: u64) {
+        if let None = off.checked_add(len) {
```
agrover: 👍

src/engine/strat_engine/range_alloc.rs
```diff
@@ -0,0 +1,230 @@
...
+impl RangeAllocator {
+    pub fn new(max: Sectors) -> RangeAllocator {
+        RangeAllocator {
+            max: *max,
+            used: RangedSet::new(),
+        }
+    }
+
+    pub fn new_with_used(max: Sectors, initial_used: &[(Sectors, Sectors)]) -> RangeAllocator {
+        let mut allocator = RangeAllocator {
```
agrover: aha yup

src/engine/strat_engine/range_alloc.rs
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    /// Allocated sectors
+    pub fn used(&self) -> Sectors {
+        Sectors((0..*self.limit).filter(|val| self.used.contains(val)).count() as u64)
+    }
+
+    /// Get a list of (offset, length) segments that are in use
+    fn used_ranges(&self) -> Vec<(Sectors, Sectors)> {
+
```
mulkieran: Hoist the ```used_ranges``` vec out of the fold, just like ```free``` in ```avail_ranges()``` and this will be clearer.

src/engine/strat_engine/range_alloc.rs
```diff
@@ -0,0 +1,232 @@
...
+        allocator.insert_ranges(initial_used);
+        allocator
+    }
+
+    fn check_for_overflow(&self, off: Sectors, len: Sectors) {
+        assert_ne!(off.checked_add(*len), None);
+        assert!(off + len <= self.limit, "off+len greater than range limit");
+    }
+
+    fn insert_ranges(&mut self, ranges: &[(Sectors, Sectors)]) -> () {
```
mulkieran: This function could use a header comment to explain it better. I think panicking when inserted value is already there is correct thing to do, but it needs justification.

src/engine/strat_engine/range_alloc.rs
```diff
@@ -0,0 +1,232 @@
...
+#[derive(Debug)]
+pub struct RangeAllocator {
+    limit: Sectors,
+    used: RangedSet<u64>,
+}
+
+impl RangeAllocator {
+    /// Create a new RangeAllocator with all entries in the range unused.
+    pub fn new(limit: Sectors) -> RangeAllocator {
+        assert!(*limit < MAX, "limit must be less than 2^64");
```
agrover: yeah, limit=25 means there are 25 entries in the range, numbered 0 to 24. What's a better word to use here?

src/engine/strat_engine/range_alloc.rs
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    fn check_for_overflow(&self, off: Sectors, len: Sectors) {
+        assert_ne!(off.checked_add(*len), None);
+        assert!(off + len <= self.limit, "off+len greater than range limit");
+    }
+
+    fn insert_ranges(&mut self, ranges: &[(Sectors, Sectors)]) -> () {
+        for &(off, len) in ranges {
+            assert_ne!(len, Sectors(0));
```
agrover: not 100% in agreement but ok.

src/engine/strat_engine/range_alloc.rs
agrover: ok. yes this seems to help.

src/engine/strat_engine/range_alloc.rs
```diff
@@ -0,0 +1,232 @@
...
+        let mut needed = amount;
+
+        for (start, len) in self.avail_ranges() {
+            if needed == Sectors(0) {
+                break;
+            }
+
+            let to_use = min(needed, len);
+
+            segs.push((start, to_use));
```
agrover: yup!

agrover:
Here's more fixes to the first commit based on @mulkieran 's feedback. Will dig into comments on later commits next. Thanks!

src/engine/strat_engine/pool.rs
agrover: opened https://github.com/stratis-storage/stratis-docs/issues/33

tests/util/simple_tests.rs
agrover: okay.

src/engine/strat_engine/blockdev.rs
agrover: heh ok.

src/engine/strat_engine/blockdev.rs
agrover: sounds good.

src/engine/strat_engine/pool.rs
agrover: ok, will `try!()`

src/engine/strat_engine/range_alloc.rs
mulkieran: Can't think of one.

src/engine/strat_engine/range_alloc.rs
```diff
@@ -0,0 +1,232 @@
...
+        assert_eq!(request.1.len(), 2);
+
+        let good_remove_ranges = [(Sectors(21), Sectors(20)), (Sectors(41), Sectors(40))];
+        allocator.remove_ranges(&good_remove_ranges);
+        assert_eq!(allocator.used(), Sectors(68));
+        assert_eq!(allocator.available(), Sectors(60));
+    }
+
+    #[test]
+    #[should_panic]
```
mulkieran:
```
/// Verify that ```remove_ranges()``` panics if ranges to be removed share elements.
```

src/engine/strat_engine/range_alloc.rs
```diff
@@ -0,0 +1,232 @@
...
+
+        let _request = allocator.request(Sectors(128));
+
+        // overlap
+        let bad_remove_ranges = [(Sectors(21), Sectors(20)), (Sectors(40), Sectors(40))];
+        allocator.remove_ranges(&bad_remove_ranges);
+    }
+
+    #[test]
+    #[should_panic]
```
agrover: Hm? I thought we were in agreement now that setting a used range should panic?

agrover: Arrgh, I updated this from the version you commented on, and am now lost trying to tie your comments to the appropriate test case. Adding these comments is 👍. I think it might be quicker for you to push a commit to this PR that makes the additions, would you be so kind, please? 😃

agrover:
Applied all recommended changes except the suggested test comments in range_alloc.rs.
Rebased against master.

mulkieran:
It'll be easier for me to put in test method header comments in a follow-up commit...

mulkieran:
This is starting to block my metadata work a fair amount...

src/engine/strat_engine/blockdev.rs
```diff
@@ -66,10 +67,15 @@ pub fn find_all() -> EngineResult<HashMap<PoolUuid, HashMap<DevUuid, BlockDev>>>
 
         if let Some(bda) = BDA::load(&mut f).ok() {
             let dev = try!(Device::from_str(&devnode.to_string_lossy()));
+            // TODO: Parse MDA and also initialize RangeAllocator with
```
agrover: `flex_devs` gives the segments that meta dev, thin meta dev, and thin data dev are using across the blockdevs. Putting this info into the RangeAllocator will let us extend these without stomping on their existing allocations.

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -0,0 +1,138 @@
...
+        let devices = try!(resolve_devices(paths));
+        let bds = try!(initialize(pool_uuid, devices, MIN_MDA_SECTORS, force));
+        let bdev_paths = bds.iter().map(|p| p.devnode.clone()).collect();
+        for bd in bds {
+            self.block_devs.insert(bd.devnode.clone(), bd);
+        }
+        Ok(bdev_paths)
+    }
+
+    pub fn remove_all(mut self) -> EngineResult<()> {
```
agrover: 👍

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -0,0 +1,138 @@
...
+                .map(|&(start, len)| Segment::new(bd.dev, start, len)));
+            needed = needed - gotten;
+        }
+
+        assert_eq!(needed, Sectors(0));
+
+        Some(segs)
+    }
+
+    pub fn paths(&self) -> Vec<PathBuf> {
```
agrover: 👍

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -0,0 +1,138 @@
...
+        self.block_devs.iter().map(|p| p.1.devnode.clone()).collect()
```
agrover: 👍

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    pub fn to_save(&self) -> HashMap<String, BlockDevSave> {
```
agrover: yep.

PR #315, opened 2017-04-27
Update to use devicemapper 0.5

agrover:
Some things are now only exported from the top level.

LinearDev::new changed to take a &[Segment] instead of a &[&Segment] param.


trgill:
I think Bytes and Sectors are now at the top level.

agrover:
oops! forgot to fixup tests. one sec...

PR #316, opened 2017-04-28
Do not wipe files backing loopback devices

agrover:
Previously, we wanted them wiped, but this also served to ensure they were
a particular size. Unfortunately, some distros such as Fedora use in-memory
tmpfs for /tmp, and it will run out of space if multiple GiB-size files are
created there.

Luckily, we don't really need to wipe or fully back our files in tmp. If
we want a file to be a certain size but not write to it, ideally
we would use fallocate(2) to do this. The file doesn't actually take up
space on disk until it is written-to. Since nix crate doesn't yet
support fallocate(), a workaround is just to seek() and then write a little
to set the size but leave most of it not actually allocated by the fs.

### DIFF
--- a/tests/loopbacked_tests.rs
+++ b/tests/loopbacked_tests.rs
@@ -10,14 +10,14 @@ extern crate tempdir;
 mod util;
 
 use std::fs::OpenOptions;
+use std::io::{Seek, Write, SeekFrom};
 use std::path::{Path, PathBuf};
 
-use devicemapper::{Bytes, Sectors};
+use devicemapper::Bytes;
 use loopdev::{LoopControl, LoopDevice};
 use tempdir::TempDir;
 
 use libstratis::consts::IEC;
-use libstratis::engine::strat_engine::blockdev::wipe_sectors;
 
 use util::simple_tests::test_force_flag_dirty;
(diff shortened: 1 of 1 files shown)


tests/loopbacked_tests.rs
```diff
@@ -33,12 +33,18 @@ fn get_devices(count: u8, dir: &TempDir) -> Vec<LoopDevice> {
...
         let subdir = TempDir::new_in(dir, &index.to_string()).unwrap();
         let path = subdir.path().join("store");
-        OpenOptions::new().read(true).write(true).create(true).open(&path).unwrap();
-        wipe_sectors(&path, Sectors(0), length).unwrap();
+        let mut f = OpenOptions::new().read(true).write(true).create(true).open(&path).unwrap();
+
+        // the proper way to do this is fallocate, but nix doesn't implement yet.
+        // TODO: see https://github.com/nix-rust/nix/issues/596
+        f.seek(SeekFrom::Start(*length)).unwrap();
+        f.write(" ".as_bytes()).unwrap();
```
mulkieran: Couldn't you just write a zero here, as ```&[0u8]```?

PR #302, opened 2017-04-10
Master loopbacked 2

mulkieran:

PR #318, opened 2017-05-02
Wipe the 1st 1 MiB of file backing loopbacked device

mulkieran:
Otherwise, data may remain from the previous test.
1 MiB is smaller than 1 GiB, so it should not crowd tmpfs.

### DIFF
--- a/tests/loopbacked_tests.rs
+++ b/tests/loopbacked_tests.rs
@@ -13,11 +13,12 @@ use std::fs::OpenOptions;
 use std::io::{Seek, Write, SeekFrom};
 use std::path::{Path, PathBuf};
 
-use devicemapper::Bytes;
+use devicemapper::{Bytes, Sectors};
 use loopdev::{LoopControl, LoopDevice};
 use tempdir::TempDir;
 
 use libstratis::consts::IEC;
+use libstratis::engine::strat_engine::blockdev::wipe_sectors;
 
 use util::simple_tests::test_force_flag_dirty;
 use util::simple_tests::test_force_flag_stratis;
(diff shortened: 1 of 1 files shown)


mulkieran:
I believe that this is necessary due to observing a surprising test failure recently: https://travis-ci.org/stratis-storage/stratisd/builds/227749794.

1 MiB is an arbitrary value, and it may be necessary to refine this, but I think it should do for now.

mulkieran:
We may be addressing the completely wrong problem, though...

trgill:
I'm not sure this is the root of the problem, but I don't have a problem with merging it if you think it will move us forward. 👍

agrover:
I'd be fine with this if we understood why it fixed the failure. If this is blocking progress and this fixes it then I'm fine with applying this for now. Please open an issue and assign it to me and I can look into it out of the critical path.

mulkieran:
Since I applied this to my working branch: https://github.com/stratis-storage/stratisd/pull/302 I haven't seen this failure again. I don't know if other people are reliably seeing this particular failure with this change not applied. I know I never saw this failure until your PR that didn't write the whole 1 GiB was merged. All the evidence I have is circumstantial.

mulkieran:
I'ld like to hold off a bit longer and see what https://github.com/stratis-storage/stratisd/pull/324 does.

trgill:
I still see failures with this merged.  I'm starting to think that merging #318 is the right move.

agrover:
👍 , please just also open an issue for this and assign to me, and I'll maybe follow up.

trgill:
#331 has been opened and assigned to @agrover

PR #324, opened 2017-05-04
Explicitly delete the temporary directory for loop devices

mulkieran:
To be better alerted to a problem with the test infrastructure.


mulkieran:
I think this change is benign and might as well go in regardless. So far, though, it seems to have fixed that intermittent failure as well. Right now, at 11:05, I'm running tests for the third time, and they keep succeeding. Wierd, yet disturbing.

mulkieran:
4th run.

PR #326, opened 2017-05-05
No longer pin to lower rustfmt version

mulkieran:
The bug that we were triggering was fixed in the newest version.

https://github.com/rust-lang-nursery/rustfmt/issues/1280

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 check: fmt build docs test test-loop
 
 ${HOME}/.cargo/bin/cargo-fmt:
-	cargo install rustfmt --vers 0.6.3
+	cargo install rustfmt
 
 fmt: ${HOME}/.cargo/bin/cargo-fmt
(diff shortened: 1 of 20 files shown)


mulkieran:
Giving this another try, since it has been a few weeks.

mulkieran:
Let's go ahead. But I still wouldn't try to do any configuration, I expect anything but the defaults will still be funky:
https://github.com/rust-lang-nursery/rustfmt/issues/1507.

@agrover @trgill review?

agrover:
I think I'd feel more comfortable if we kept the pin but just updated to the newer version. It still seems prudent to evaluate changes before adopting them. This keeps us all in sync with each other and helps avoid lost productivity if some version starts breaking.

mulkieran:
I think the process could work like this...don't pin the version. If a PR fails due to rustfmt doing something dumb, pin to the previous version and wait until rustfmt gets better again. Otherwise, the coordination overhead is actually higher and the CI doesn't help us as much as it should.

agrover:
OK, if you think this is best.

PR #327, opened 2017-05-05
RangeAllocator rework

agrover:
Reimplement RangeAllocator using BTreeMap instead of ranged_set. This nets us some efficiency advantages and lets us drop an external dependency. No API changes.

### DIFF
--- a/src/engine/strat_engine/range_alloc.rs
+++ b/src/engine/strat_engine/range_alloc.rs
@@ -3,25 +3,23 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 use std::cmp::min;
-use std::u64::MAX;
+use std::collections::Bound::{Included, Unbounded};
+use std::collections::BTreeMap;
 
 use devicemapper::Sectors;
-use ranged_set::RangedSet;
 
 #[derive(Debug)]
 pub struct RangeAllocator {
     limit: Sectors,
-    used: RangedSet<u64>,
+    used: BTreeMap<Sectors, Sectors>,
 }
 
 impl RangeAllocator {
     /// Create a new RangeAllocator with all entries in the range unused.
     pub fn new(limit: Sectors) -> RangeAllocator {
-        // Implementation of used_ranges means we can't handle limit=MAX
-        assert!(*limit < MAX, "limit must be less than 2^64");
         RangeAllocator {
             limit: limit,
-            used: RangedSet::new(),
+            used: BTreeMap::new(),
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/engine/strat_engine/range_alloc.rs
```diff
@@ -57,44 +97,58 @@ impl RangeAllocator {
...
-            for val in *off..*off + *len {
-                let removed = self.used.remove(&val);
-                if removed == false {
-                    panic!(format!("tried to remove value {} not present in RangedSet", val));
+            let (prev_off, prev_len) = self.used
+                .range((Unbounded, Included(off)))
+                .rev()
+                .next()
+                .map(|(k, v)| (*k, *v))
+                .expect("must match an existing range");
```
mulkieran: I don't believe this expect must always succeed. A panic! if its None would be consistent w/ the rest, I think.

agrover: I don't see how that differs in function from the PR's current implementation: If we don't find an existing range, panic. Can you elaborate?

mulkieran: Why is ```remove_ranges()``` public? This affects how you should deal with it. If it were private, then public methods would control how it was invoked. But since it's public, different rules. I realize that at this point, nothing is invoking it.

Speaking of which, I just noticed that ```new_with_used``` can also panic if its ranges overlap. Given that it is public, that is kind of wrong, so something should be done. The most correct thing would be for ```insert_ranges()``` to return a Result and for ```new_with_used``` to pass on that result, I think.

Difference between expect() and a straight up panic!() is that expect means that you have demonstrated by an argument that the unwrap must always succeed. panic() means you haven't, but you're not going to handle the error, regardless. But given that the method in question is public, switching from expect to panic may not be the right thing, anyway.

mulkieran:
See #328. It is relevant to this discussion because it raises the question of whether ```insert_ranges``` should be public. At first, I thought it would be necessary to make it public, because it would have to be called in order to add ranges discovered after all BlockDevs had been read. Then, I realized that ```find_all``` would almost certainly be improved by just building a map of Devices. But, it is possible that it will still be necessary to make ```insert_ranges``` public.

src/engine/strat_engine/range_alloc.rs
agrover: Let's discuss API issues separate from this PR, since this PR doesn't change the API.

I still don't understand the objection here. The old code panic()ed if you tried to remove a non-present range, the new code does too -- the expect fails and panic()s.

agrover:
* rebase against latest head
* Change `remove_ranges` to avoid expect() for style and clarity

src/engine/strat_engine/range_alloc.rs
```diff
@@ -43,10 +41,52 @@ impl RangeAllocator {
...
+                        None
+                    }
+                });
+
+            let contig_next = self.used
+                .range(off..)
+                .next()
+                .map(|(k, v)| (*k, *v))
+                .and_then(|(next_off, next_len)| {
+                    assert!(off + len >= next_off,
```
mulkieran: I'm finding it hard to convince myself that this shouldn't be ```<=``` (not ```>=```). If I'm right, it's probably worth another test.

src/engine/strat_engine/range_alloc.rs
```diff
@@ -43,10 +41,52 @@ impl RangeAllocator {
...
+                    self.used.insert(off, len + next_len);
+                    self.used.remove(&next_off);
+                }
+                (Some((prev_off, prev_len)), None) => {
+                    // Contig with prev, just extend prev
+                    *self.used.get_mut(&prev_off).expect("must exist") = prev_len + len;
+                }
+                (Some((prev_off, prev_len)), Some((next_off, next_len))) => {
+                    // Contig with both, remove next and extend prev
+                    self.used.remove(&next_off);
```
mulkieran: An ```expect()``` here, since it must remove or something is wrong with our logic? This is pretty minor.

src/engine/strat_engine/range_alloc.rs
```diff
@@ -57,47 +97,62 @@ impl RangeAllocator {
...
+            // switch based on if the to-remove range starts or ends
+            // at the same point as the existing range
+            match (prev_off == off, prev_off + prev_len == off + len) {
+                (true, true) => {
+                    // Exactly matches existing range
+                    self.used.remove(&prev_off);
+                }
+                (true, false) => {
+                    // A tail segment remains
+                    self.used.remove(&prev_off);
```
mulkieran: Maybe an expect here, for the same reason as above.

src/engine/strat_engine/range_alloc.rs
agrover: I think you're right it's buggy, I think it might properly be `>` rather than `>=` esp since we test for `==` below.

agrover: ah yes, should be `<=`.

agrover:
* Fix logic error in insert_range
* Add `expect` on `remove()`s
* Add more unit tests

PR #328, opened 2017-05-08
Rewrite find_all()

mulkieran:
Rewrite ```find_all()``` and add a test of its correct operation.


mulkieran:
A bunch of these commits are pulled from #302.

mulkieran:
@agrover @trgill review? I think all non-test commits could probably be squashed into one if acked.

agrover:
Please add a brief PR description.

Also, if `find_all` just returns Devices, where do we instantiate BlockDevs? Does that get re-added in a later PR?

mulkieran:
We instantiate the BlockDevs in a subsequent step, when we have obtained more complete information.

mulkieran:
rebased

PR #329, opened 2017-05-08
Master teardown

mulkieran:


mulkieran:
@agrover @trgill review? This is necessarily incomplete. But it will have to be implemented completely, and it will be useful for testing reconstruction from metadata.

src/engine/structures.rs
```diff
@@ -33,6 +33,11 @@ impl<T: HasName + HasUuid> Table<T> {
 
     }
 
+    /// Empty this table of all its items, returning them in a vector.
+    pub fn empty(mut self) -> Vec<T> {
+        self.items.drain(..).collect()
```
agrover: Does it let you skip the drain/collect and just do `self.items` here?

src/engine/strat_engine/pool.rs
```diff
@@ -199,6 +199,21 @@ impl StratPool {
             }
         }
     }
+
+    /// Teardown a pool.
+    /// Take down the device mapper devices belonging to the pool.
+    /// This method and destroy() must keep their DM teardown operations
+    /// in sync.
+    pub fn teardown(self) -> EngineResult<()> {
+        // TODO: sync files
```
agrover: Not sure -- we may find all filesystems need to be unmounted to teardown, not just synced.

src/engine/structures.rs
mulkieran: yes, it does. all that cleverness for nothing.

mulkieran:
Made changes based on comments. @trgill review?

trgill:
I think you need to run rustfmt on simple_tests.rs before merge.

PR #334, opened 2017-05-09
Continued work on setup

mulkieran:
Additions and revisions to existing setup code.

Changes are:
* Revisions and additions to code in setup.rs. Removal of some obsoleted code.
* Return type changes for metadata methods for loading the variable length metadata. This allows to distinguish better between an error when reading metadata and there being no Stratis metadata to read.
* Addition of new tests and removal of obsoleted tests.

### DIFF
--- a/src/engine/strat_engine/serde_structs.rs
+++ b/src/engine/strat_engine/serde_structs.rs
@@ -18,8 +18,11 @@ use std::path::PathBuf;
 
 use devicemapper::Sectors;
 
-/// Implements saving and restoring from metadata.
-pub trait DSerializable<T> {
+/// Implements saving struct data to a serializable form and reconstructing
+/// a struct from that form.
+/// Assuming the context of the existing devices this must be an isomorphism,
+/// i.e., setup(x.to_save()) == x and setup(x).to_save() == x or it's a bug.
+pub trait Isomorphism<T> {
     fn to_save(&self) -> T;
     fn setup(T) -> Self
         where Self: Sized
--- a/src/engine/strat_engine/serde_structs.rs
+++ b/src/engine/strat_engine/serde_structs.rs
@@ -11,14 +11,22 @@
 // structs. These contain simple, serde-friendly data types, and we
 // can convert to or from them when saving our current state, or
 // restoring state from saved metadata.
-//
-// Look for "to_save" and "setup" methods, that either return or take
-// the below structs as parameters.
 
 use std::collections::HashMap;
+use std::marker::Sized;
 use std::path::PathBuf;
 
 use devicemapper::Sectors;
+
+/// Implements saving and restoring from metadata.
+pub trait DSerializable<T> {
+    fn to_save(&self) -> T;
+    fn setup(T) -> Self
+        where Self: Sized
+    {
+        unimplemented!()
+    }
+}
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct BlockDevSave {
--- a/src/engine/strat_engine/serde_structs.rs
+++ b/src/engine/strat_engine/serde_structs.rs
@@ -12,11 +12,7 @@
 // can convert to or from them when saving our current state, or
 // restoring state from saved metadata.
 
-use std::collections::HashMap;
 use std::marker::Sized;
-use std::path::PathBuf;
-
-use devicemapper::Sectors;
 
 /// Implements saving struct data to a serializable form and reconstructing
 /// a struct from that form.
@@ -32,14 +28,6 @@ pub trait Isomorphism<T> {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct BlockDevSave {
-    pub devnode: PathBuf,
-    pub total_size: Sectors,
-}
-
-#[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct StratSave {
+pub struct PoolSave {
     pub name: String,
-    pub id: String,
-    pub block_devs: HashMap<String, BlockDevSave>,
 }
--- a/src/engine/strat_engine/setup.rs
+++ b/src/engine/strat_engine/setup.rs
@@ -71,15 +71,11 @@ pub fn find_all() -> EngineResult<HashMap<PoolUuid, Vec<PathBuf>>> {
         }
 
         let mut f = f.expect("unreachable if f is err");
-
-        match try!(StaticHeader::determine_ownership(&mut f)) {
-            DevOwnership::Ours(uuid) => {
-                pool_map
-                    .entry(uuid)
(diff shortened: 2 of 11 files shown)


mulkieran:
@agrover @trgill preliminary review? The changes here get the code to the point where it is reasonably certain that a struct of the correct type can be re-constructed from the metadata.

src/engine/strat_engine/serde_structs.rs
```diff
@@ -11,24 +11,20 @@
...
-    pub devnode: PathBuf,
-    pub total_size: Sectors,
+/// Implements saving and restoring from metadata.
+pub trait DSerializable<T> {
+    fn to_save(&self) -> T;
+    fn setup(T) -> Self
+        where Self: Sized
+    {
+        unimplemented!()
+    }
```
agrover: 1. Why is this trait needed?
2. Current trait name is a little too close to serde trait names

src/engine/strat_engine/setup.rs
```diff
@@ -0,0 +1,182 @@
...
+        // 1. The device does not exist anymore.
+        if f.is_err() {
+            let err = f.unwrap_err();
+            match err.kind() {
+                ErrorKind::NotFound => {
+                    continue;
+                }
+                _ => {
+                    if let Some(errno) = err.raw_os_error() {
+                        if Errno::from_i32(errno) == Errno::ENXIO {
```
agrover: comment why ENXIO is singled out here?

src/engine/strat_engine/setup.rs
```diff
@@ -0,0 +1,182 @@
...
+        let mut f = f.expect("unreachable if f is err");
+
+        match try!(StaticHeader::determine_ownership(&mut f)) {
+            DevOwnership::Ours(uuid) => {
+                pool_map
+                    .entry(uuid)
+                    .or_insert_with(Vec::new)
+                    .push(devnode)
+            }
+            _ => continue,
```
agrover: `if let` instead of match? 1. just interested in one case 2. continuing right before the end of the loop is weird.

agrover: hmm I guess this is all just moved from another file by this PR so it's a little unfair to get all style-picky with it now...feel free to disregard.

src/engine/strat_engine/setup.rs
mulkieran: Because I observed it happening...there's probably other things that should not be treated as errors, but they can be filled in as they are observed. Some of them are probably pretty unlikely and this approach is at least conservative.

src/engine/strat_engine/setup.rs
mulkieran: Might as well...

src/engine/strat_engine/setup.rs
```diff
@@ -0,0 +1,182 @@
...
+
+    // No device nodes means no metadata
+    if devnodes.is_empty() {
+        return Ok(None);
+    }
+
+    fn get_bda(devnode: &Path) -> EngineResult<Option<BDA>> {
+        let mut f = try!(OpenOptions::new().read(true).open(devnode));
+        BDA::load(&mut f)
+    }
```
mulkieran: Gone.

src/engine/strat_engine/setup.rs
```diff
@@ -0,0 +1,182 @@
...
+        }
+        bdas.push((devnode, bda));
+    }
+
+    // We may have had no devices with BDAs for this pool, so return if no BDAs.
+    if bdas.is_empty() {
+        return Ok(None);
+    }
+
+
```
mulkieran: Gone.

src/engine/strat_engine/setup.rs
```diff
@@ -0,0 +1,182 @@
...
+    // Most recent time should never be None if this was a properly
+    // created pool; this allows for the method to be called in other
+    // circumstances.
+    let most_recent_time = most_recent_bda.last_update_time();
+    if most_recent_time.is_none() {
+        return Ok(None);
+    }
+
+    // Try to read from all available devnodes that could contain most
+    // recent metadata. In the even of errors, continue to try until all are
```
mulkieran: Done.

src/engine/strat_engine/serde_structs.rs
mulkieran: 1. It makes it clear in the code what structs are saveable and what are not. This is better than grepping for ```to_save``` and ```setup``` methods as the header comment requires. It also clearly enforces what the signatures of these methods should be.

2. Do you have a suggestion?

mulkieran: How about Summarized or -able?

mulkieran:
I'm setting this to review-final because the code is in good shape, and I may start to block if it doesn't get committed soon. @agrover @trgill review?

src/engine/strat_engine/serde_structs.rs
agrover: 1. ok cool thanks. Maybe add some version of this as a comment in the source?
2. `Saveable` or `StratSerializable`? `Persistable`? `Summarized` doesn't seem accurate since the info being saved is not shortened or simplified, which is what summary connotes to me. 🚲 🏠

src/engine/strat_engine/setup.rs
agrover: add a comment to the code with some version of this explanation?

mulkieran: Done.

src/engine/strat_engine/serde_structs.rs
mulkieran: I'm going with "Isomorphic".

tests/util/simple_tests.rs
```diff
@@ -304,3 +282,43 @@ pub fn test_empty_pool(paths: &[&Path]) -> () {
...
+
+    let (paths1, paths2) = paths.split_at(2);
+
+    let mut engine = StratEngine::new();
+
+    let name1 = "name1";
+    let (uuid1, _) = engine.create_pool(&name1, paths1, None, false).unwrap();
+
+    let pools = find_all().unwrap();
+    let metadata = get_pool_metadata(&pools).unwrap();
```
trgill: Are you using pools here due to the formatting issue you mentioned?  Since you only use it once, it could be replaced with the call?

mulkieran: No, the formatting would be fine. The issue I saw was about inserting a match expression as an argument in a function call. So far, rustfmt hasn't done anything wierd for anything else. I just like having just one unwrap() per line, at the end, for these tests. It makes it clearer for me to read.

mulkieran:
Requested changes made.

agrover:
👍

PR #337, opened 2017-05-11
Changes to get ready for invoking engine's setup() method from main()

mulkieran:
The purpose of this PR is to prepare to be calling setup() functions from main().

Mostly this PR is about tidying up our error handling, removing unused error constructors, and so forth.
It also rearranges main so that it can return an exit code in the recommended fashion. It rearranges some other methods in stratis.rs for a similar purpose. It moves some files at the top-level to ~~runner~~ stratis subdirectory for better encapsulation and to more readily access the engine::EngineError type.

PR #340, opened 2017-05-12
Small table stuff

agrover:
very small stuff that gave me a chance to learn about implementing IntoIterator.


agrover:
Hmm, how about if you cherry-picked these onto your #337 instead?

PR #341, opened 2017-05-12
Insert a call to the engine's setup method in main

mulkieran:
This PR does a bit of refactoring and cleanup in order to be able to insert a call to the engine's ```setup()``` method in main. The main bits are:

* Removing a lot of unused error codes
* Rearranging some error types
* Rearranging the code in stratis.rs so that main() can return an exit code
* Some general cleanup
* Defining an incomplete ```setup()``` method for the engine and invoking it in main(). The method is incomplete because it is currently not possible to setup. Therefore, if it sees evidence that it needs to do any setup, the program just exits with an error.


mulkieran:
rebased

mulkieran:
Supercedes #337. The only change of any significance is the very final commit.

agrover:
👍

I'm guessing that setup()'s incompleteness will be remedied very shortly. But if it weren't, I would say you'd probably want to mention this not just in the PR summary but in the actual commit's log or as a code comment. (Looking at the code, I first assumed it was broken, not placeholder/incomplete)

mulkieran:
Added a TODO. @trgill review?

PR #343, opened 2017-05-15
Master setup

mulkieran:


mulkieran:
rebased.

mulkieran:
expected to fail.

mulkieran:
rebased.

mulkieran:
@agrover: What happens if the actual size of the block device is different from the size recorded in the BDA? Will it be necessary to overwrite the BDA or at least the size value at some point?

Should we have chosen the other option, to store the size in the metadata, and not in the BDA?

agrover:
Here's what my thinking is:
```
If bda.size > actual size // what? disk shrank!
    emit warning
    don't use disk (possibly causing pool to be unusable)

if bda.size < actual size // more space than expected
    Just use space up to bda.size
    possibly have some command to tell Stratis it can expand

if bda.size == actual size // we're good
    use the disk
```

mulkieran:
Superceded by #370, #371.

PR #347, opened 2017-05-16
Add a method for constructing DM device names

mulkieran:

### DIFF
--- a/src/engine/strat_engine/dmdevice.rs
+++ b/src/engine/strat_engine/dmdevice.rs
@@ -1,0 +1,76 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// Functions for dealing with device mapper devices.
+
+use std::fmt;
+use std::fmt::Display;
+
+use super::super::super::engine::{FilesystemUuid, PoolUuid};
+
+const FORMAT_VERSION: u16 = 1;
+
+pub enum FlexRole {
+    MetadataVolume,
+    ThinData,
+    ThinMeta,
+}
+
+impl Display for FlexRole {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            FlexRole::MetadataVolume => write!(f, "mdv"),
+            FlexRole::ThinData => write!(f, "thindata"),
+            FlexRole::ThinMeta => write!(f, "thinmeta"),
+        }
+    }
+}
+
+pub enum ThinRole {
+    Filesystem(FilesystemUuid),
+}
+
+impl Display for ThinRole {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            ThinRole::Filesystem(uuid) => write!(f, "fs-{}", uuid.simple().to_string()),
+        }
+    }
+}
+
+pub enum ThinPoolRole {
+    Pool,
+}
+
+impl Display for ThinPoolRole {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            ThinPoolRole::Pool => write!(f, "pool"),
+        }
+    }
+}
+
+/// Format a name for the flex layer.
+pub fn format_flex_name(pool_uuid: &PoolUuid, role: FlexRole) -> String {
+    return format!("stratis-{}-{}-flex-{}",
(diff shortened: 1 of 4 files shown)


mulkieran:

mulkieran:
I do not know where the unique id for the thin layer is supposed to come from.

src/engine/strat_engine/dmdevice.rs
```diff
@@ -0,0 +1,142 @@
...
+        match *self {
+            FlexRole::MetadataVolume => write!(f, "mdv"),
+            FlexRole::ThinData => write!(f, "thindata"),
+            FlexRole::ThinMeta => write!(f, "thinmeta"),
+        }
+    }
+}
+
+pub enum ThinRole {
+    Filesystem,
```
agrover: How about `Filesystem(FilesystemUuid)`?

src/engine/strat_engine/dmdevice.rs
```diff
@@ -0,0 +1,142 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// Functions for dealing with device mapper devices.
```
agrover: ensuring consistency of device mapper device *names*

src/engine/strat_engine/dmdevice.rs
```diff
@@ -0,0 +1,142 @@
...
+                   role);
+}
+
+/// Format a name for the thin pool layer.
+fn format_thinpool_name(pool_uuid: &PoolUuid, role: ThinPoolRole) -> String {
+    return format!("stratis-{}-{}-thinpool-{}",
+                   FORMAT_VERSION,
+                   pool_uuid.simple().to_string(),
+                   role);
+}
```
agrover: these three functions only differ in the first param to format. Suggest having the match in `format_name` get the string that differs and then do the format.

src/engine/strat_engine/dmdevice.rs
```diff
@@ -0,0 +1,142 @@
...
+
+/// Format a name for the given pool UUID, layer, and role.
+/// Return an error if argument values are incompatible.
+/// It might, for example, be possible to pass a role value that is not
+/// appropriate for the givne layer.
+pub fn format_name(pool_uuid: &PoolUuid, layer: Layer, role: Role) -> EngineResult<String> {
+    match layer {
```
agrover: I think `Layer` is not needed. Removing it also means it's impossible to have it in conflict with `Role` 😃

agrover: Also, just return `String`? I don't see a good reason why this func should ever fail.

src/engine/strat_engine/dmdevice.rs
mulkieran: It's a whole module...I left open the possibility that Stratis might need to do something else with DM devices besides construct their names.

src/engine/strat_engine/dmdevice.rs
mulkieran: The problem with that is that you loose certain benefits of type-checking. I've committed another approach.

mulkieran:
Some changes made. I expect to squash the commits, but I'm keeping them separate for now.

src/engine/strat_engine/dmdevice.rs
mulkieran: If you were ever using it dynamically, with unknown values, it could certainly fail. I've decided to assume, for now, that that will never happen, in my new commit.

src/engine/strat_engine/dmdevice.rs
trgill: @agrover do we want to mix Filesystem with DM specific functions?  you are thinking these are "layers" - not "DM layers"

mulkieran: I think that the problem is that we _must_ disambiguate between different thin devices that Stratis might have set up, and the filesystem UUID is the only way to do it. This our Stratis filesystem UUID, not any UUID that the filesystem itself has.

mulkieran: Right now, the docs say that the UUID is "layer-specific". If what is really meant is "role-specific", then the suggested change is probably the right thing to do.

mulkieran:
@agrover review?

src/engine/strat_engine/dmdevice.rs
agrover: Ah yes, I think the doc needs fixing.

PR #349, opened 2017-05-17
Cleanups related to encapsulation and imports

mulkieran:


mulkieran:
@agrover @trgill review?

agrover:
> Make all imports relative
> 
> This helps us differentiate between imports from ourselves and imports
> from other crates. It also makes it clear by ordering which are more and
> which are less specific.

I don't care for `use super::super::super::blah` syntax. Move code up/down a level and everything breaks. We already know which imports are ours because of grouping of use statements.

I get that right now we have both relative and non-relative imports.. what would tidying things up by going the other way and eliminating relative imports look like?

> Do not make strat_engine modules public if unused 

I don't see a benefit. The downside is if someone does want to add tests, they now have to change another spot.

src/engine/sim_engine/blockdev.rs
```diff
@@ -15,7 +15,6 @@ use super::randomization::Randomizer;
 /// A list of very basic states a SimDev can be in.
 pub enum State {
     OK,
-    FAILED,
 }
```
agrover: `State` only has one value left! Does it still serve a purpose at this point?

mulkieran: I've gone ahead and dropped it.

mulkieran:
> > Do not make strat_engine modules public if unused 

> I don't see a benefit. The downside is if someone does want to add tests, they now have to change another spot.

The upside is that the compiler has a little conversation with them about the change they are making and they can decide if that's what they really want.

mulkieran:
> > Make all imports relative
> > 
> > This helps us differentiate between imports from ourselves and imports
> > from other crates. It also makes it clear by ordering which are more and
> > which are less specific.

> I don't care for `use super::super::super::blah` syntax. Move code up/down a level and everything breaks. We already know which imports are ours because of grouping of use statements.

We won't be moving code up and down a level too terribly often, and as edits go, it's an unusually easy one to do. Probably an easy ripgrep exercise.

> I get that right now we have both relative and non-relative imports.. what would tidying things up by going the other way and eliminating relative imports look like?

If we used only absolute imports the sort order would be much less obvious and local imports would look long and complicated. The compiler would report more confusing error messages if we got imports wrong.

mulkieran:
rebased

mulkieran:
rebased

mulkieran:
@agrover @trgill review? This cleanup is not actually complete. Efforts at further encapsulation lead to more exposure of many more unused methods that should probably be removed. However, this seems enough for now, and I would be happy for it to go in.

agrover:
I think this PR looks `super`. 👍

PR #352, opened 2017-05-19
Do not export the *Pools

mulkieran:
They are unneeded by the binary, which operates only on the engine.

Get rid of some very unnused code thus revealed.


mulkieran:
This is the non-controversial commit cherry-picked from #349. @agrover review, please?

agrover:
👍

PR #359, opened 2017-05-24
Add spare metadata device and remove setup_thinpooldev

agrover:
This PR is waiting on restore work to land, since check/repair is only applicable if we are setting up a pre-existing thinpool.

The logical place to do thin repair is wherever `ThinPool::setup()` ends up being called from.

### DIFF
--- a/src/engine/strat_engine/pool.rs
+++ b/src/engine/strat_engine/pool.rs
@@ -50,6 +50,7 @@ pub struct StratPool {
     pub filesystems: Table<StratFilesystem>,
     redundancy: Redundancy,
     thin_pool: ThinPoolDev,
+    thin_pool_meta_spare: Vec<Segment>,
     mdv: MetadataVol,
 }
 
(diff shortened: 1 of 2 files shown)


src/engine/strat_engine/pool.rs
```diff
@@ -135,7 +141,8 @@ impl StratPool {
 
     /// Minimum initial size for a pool.
     pub fn min_initial_size() -> Sectors {
-        INITIAL_META_SIZE + INITIAL_DATA_SIZE + INITIAL_MDV_SIZE
+        // One extra meta for spare
+        (INITIAL_META_SIZE * 2u64) + INITIAL_DATA_SIZE + INITIAL_MDV_SIZE
```
mulkieran: This ought not to need to be 2u64, plain 2 should do.

agrover: it said
```
the trait `std::ops::Mul<i32>` is not implemented for `devicemapper::Sectors`
```

mulkieran: Yeah, it wouldn't be. I guess you do have to specify it to be _some_ unsigned type, and it won't infer one of the many acceptable types. I wonder why...

agrover:
rebased and dropped thin_repair commit from this PR.

PR #363, opened 2017-05-25
Extend MDV if it fills up

agrover:
The initial MDV size is 16MiB (~12MiB usable). Given that we're writing a tiny amount per filesystem, it would take a HUGE number of tracked filesystems to require growing the MDV. But, theoretically possible.

This also includes `filesystem::grow_fs`, which may be useful for growing our thin filesystems when the time comes.

### DIFF
--- a/src/engine/strat_engine/filesystem.rs
+++ b/src/engine/strat_engine/filesystem.rs
@@ -153,6 +153,21 @@ pub fn create_fs(dev_path: &Path) -> EngineResult<()> {
     Ok(())
 }
 
+pub fn grow_fs(dev_path: &Path) -> EngineResult<()> {
+
+    debug!("Grow filesystem for : {:?}", dev_path);
+    let output = try!(Command::new("xfs_growfs").arg(&dev_path).output());
+
+    if output.status.success() {
+        debug!("Grew xfs filesystem on {:?}", dev_path)
+    } else {
+        let message = String::from_utf8_lossy(&output.stderr);
+        debug!("stderr: {}", message);
+        return Err(EngineError::Engine(ErrorEnum::Error, message.into()));
+    }
+    Ok(())
+}
+
 pub fn mount_fs(dev_path: &Path, mount_point: &Path) -> EngineResult<()> {
 
     debug!("Mount filesystem {:?} on : {:?}", dev_path, mount_point);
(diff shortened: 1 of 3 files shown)


agrover:
review @mulkieran @trgill ?

src/engine/strat_engine/filesystem.rs
```diff
@@ -153,6 +153,21 @@ pub fn create_fs(dev_path: &Path) -> EngineResult<()> {
     Ok(())
 }
 
+pub fn grow_fs(dev_path: &Path) -> EngineResult<()> {
```
mulkieran: At some point, I was given to understand that these ```*_fs``` functions were junk functions that would be eliminated at some time in the future. Are they still junk functions or are they expected to be as much permanent parts of our code base as the other parts?

agrover: They all appear to be calling out to other executables. Some of these like `mount` and `umount` are syscalls that we could probably call ourselves using nix. Others like `xfs_growfs` we will need until XFS devs give us an API we can wrap and call directly.

mulkieran: It they are non-junk methods, they should be held to the same standard as our other "permanent" code.

agrover: I agree, but that's a bigger issue. What does that mean for this PR? `grow_fs` matches the existing style of similar functions. Can we merge this and then fix it up with the rest of them in a separate PR?

src/engine/strat_engine/mdv.rs
```diff
@@ -131,6 +136,21 @@ impl MetadataVol {
             }
         }
 
+        let fsinfo = try!(Statvfs::for_path(&self.mount_pt));
```
mulkieran: These ```check``` functions are causing our code to deteriorate and are becoming a maintenance burden. It is not possible to make any sort of useful review of the code that we are adding to them at this point. Our best course is to rip out the check functions, saving the useful stuff that has already been added, and start again in a better organized fashion.

src/engine/strat_engine/filesystem.rs
mulkieran: Matching the existing style of a lot of functions that shouldn't even exist doesn't seem like a virtue.

agrover:
ok, on hold pending discussions on check() and reworking _fs functions.

agrover:
this is so old, I'm going to open a fresh PR.

PR #364, opened 2017-05-25
Change RangeAllocator methods so they return a Result

mulkieran:
The motivation for this is that during setup, RangeAllocators must be instantiated from metadata in order to reconstruct the BlockDevs.

### DIFF
--- a/src/engine/strat_engine/blockdevmgr.rs
+++ b/src/engine/strat_engine/blockdevmgr.rs
@@ -236,7 +236,8 @@ pub fn initialize(pool_uuid: &PoolUuid,
         }
 
         let bda = bda.expect("!bda.is_err()");
-        let allocator = RangeAllocator::new(bda.dev_size(), &[(Sectors(0), bda.size())]);
+        let allocator = RangeAllocator::new(bda.dev_size(), &[(Sectors(0), bda.size())])
+            .expect("bda.size() < bda.dev_size() and single range");
 
         bds.push(BlockDev::new(dev, devnode, bda, allocator));
     }
--- a/src/engine/strat_engine/range_alloc.rs
+++ b/src/engine/strat_engine/range_alloc.rs
@@ -30,9 +30,19 @@ impl RangeAllocator {
         Ok(allocator)
     }
 
-    fn check_for_overflow(&self, off: Sectors, len: Sectors) {
-        assert_ne!(off.checked_add(*len), None);
-        assert!(off + len <= self.limit, "off+len greater than range limit");
+    fn check_for_overflow(&self, off: Sectors, len: Sectors) -> EngineResult<()> {
+        if off.checked_add(*len).is_none() {
+            let err_msg = format!("elements in range ({}, {}) inexpressible in u64", off, len);
+            return Err(EngineError::Engine(ErrorEnum::Invalid, err_msg));
+        }
+        if off + len > self.limit {
+            return Err(EngineError::Engine(ErrorEnum::Invalid,
+                                           format!("elements in range ({}, {}) exceed limit {}",
+                                                   off,
+                                                   len,
+                                                   self.limit)));
+        }
+        Ok(())
     }
 
     /// Mark ranges previously marked as unused as now used.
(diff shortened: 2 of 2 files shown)


mulkieran:
Cherry-picked from #343.

mulkieran:
The motivation for this is that during setup, RangeAllocators must be instantiated from metadata in order to reconstruct the BlockDevs.

mulkieran:
@agrover @trgill review?

agrover:
I don't think it makes sense to use `Result` for RangeAllocator. Result is for runtime errors that may occur for reasons other than code errors, like I/O. That's not the same case as for example disallowing overlap in ranges given to `insert_ranges`, which is rather a usage guarantee that RangeAllocator is requiring be upheld. Just like `Vec` panics if the array index is out of bounds. If there is a risk of overlap happening then users need to check for this beforehand.

mulkieran:
I think that it is likely that it will be the case that we will be able to argue that any call which tries to insert overlapping ranges into a RangeAllocator is logically impossible.

But, I don't think that makes it the case that RangeAllocator should be built with assertions. This forces all developers to do a form of implicit global reasoning. I'm not against doing reasoning about the behavior of the entire program all at once where it is necessary, but I think it is a mistake to do that where it is unnecessary. I think it is unnecessary here.

I do not believe there is much difference in execution cost between assertions and return a Result type.

I agree that in the case where a BlockDev is first initialized, in ```blockdevmgr::initialize()``` it is impossible for the RangeAllocator instantiation to fail, so I have changed the call from a ```try!``` to an ```expect()```. Note that that is local, not global reasoning, so I'm good with it.

I also believe that in the case where I'm setting up from metadata an argument can be made that the ranges must be non-overlapping and legal. But it is much more complicated. I will rework the relevant code in #343 so it too can validly call expect(). I think this is possible, although I haven't worked through all the details, of which the most important is, I think, the possibility that the size of the device may have changed.

I think the analogy between indexing outside the bounds of the vector is not a very close one. It is easy to check the length of the vector. It is much harder to check that none of your ranges overlap with each other or with the ranges of the range allocator being added to. The client would have to duplicate something like what the insert method does, which is costly.

Speaking of costly, the ```insert_ranges()``` algorithm looks to me like it's at least O((m + n)(2n)), i.e., quadratic. It has crossed my mind that it might be much more efficient to sort the argument ranges, O(n logn) and then step through existing and inserted in parallel O(m + n) which would bring down the algorithm cost to somewhere between linear and O(n log n), which would be a nice improvement. I've put in a TODO about this.

mulkieran:
Actually, I could not justify an ```expect()``` in calling code in #343, i.e., it ddefinitly should return the error.

agrover:
I'd like to see how #343 uses `insert_ranges`. I think the TODO to make it atomic is important in the context that it *is* atomic now (either it 100% succeeds or it panics). if it doesn't panic, now we might possibly recover and continue. But does it make sense to even do that?? Who exactly is using `insert_ranges` and where are they getting their ranges from that the ranges might not be correct? And how can we continue if somebody's not getting the exact ranges they (perhaps wrongly) expected? So I'd love to see the actual usage get fleshed out to understand where recovering from invalid ranges makes sense.

mulkieran:
See ```setup::get_pool_blockdevs()``` in #343 PR. The likely problem is not that the ranges overlap in this case, but that the device size has changed and the ranges are specified for Sectors that don't even exist on the now smaller device.

mulkieran:
Suggested algorithm for ```insert_ranges```.

1. Sort ranges argument into a BTreeMap, called ranges.
2. Verify that last element does not overflow and does not exceed limit. If it does, return an error.
3. Iterate through sorted ranges verifying no overlap, and no overflow. Limit is allready taken care of by step (2). If overlap or overflow, return an error.
4. If self.used does not exist, set self.used to ranges. Return Ok.
5. Iterate in parallel through ranges and self.used checking for overlap. If overlap found, return an error.
6. Iterate again, inserting and not checking for overlap. Return Ok.

agrover:
👍 . And you were going to also open issues for the TODOs? 😃

mulkieran:
One issue in devicemapper: https://github.com/stratis-storage/devicemapper-rs/issues/61.

mulkieran:
rebased.

PR #365, opened 2017-05-26
Substitute OpenOptions for File in all calls 

mulkieran:
In order to open files with the minimum permissions for the task and in
pursuit of consistancy.

We no longer use File methods to open files anywhere.


mulkieran:
This is a test...

mulkieran:
that failed. Why do we need the file to be readable to get its size?

mulkieran:
@trgill review?

PR #366, opened 2017-05-26
Do not store device size in BDA

mulkieran:
Use the RangeAllocator limit field as our source of this information.

### DIFF
--- a/src/engine/strat_engine/blockdev.rs
+++ b/src/engine/strat_engine/blockdev.rs
@@ -70,7 +70,7 @@ impl BlockDev {
 
     /// The device's size.
     pub fn size(&self) -> Sectors {
-        self.bda.dev_size()
+        self.used.size()
     }
 
     /// Last time metadata was written to this device.
--- a/src/engine/strat_engine/blockdevmgr.rs
+++ b/src/engine/strat_engine/blockdevmgr.rs
@@ -210,12 +210,8 @@ pub fn initialize(pool_uuid: &PoolUuid,
     let mut bds = Vec::new();
     for (dev, (devnode, dev_size, mut f)) in add_devs {
 
-        let bda = try!(BDA::initialize(&mut f,
-                                       pool_uuid,
-                                       &Uuid::new_v4(),
-                                       mda_size,
-                                       dev_size.sectors()));
-        let allocator = RangeAllocator::new(bda.dev_size(), &[(Sectors(0), bda.size())]);
+        let bda = try!(BDA::initialize(&mut f, pool_uuid, &Uuid::new_v4(), mda_size));
+        let allocator = RangeAllocator::new(dev_size.sectors(), &[(Sectors(0), bda.size())]);
 
         bds.push(BlockDev::new(dev, devnode, bda, allocator));
     }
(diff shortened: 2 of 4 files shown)


mulkieran:
@agrover @trgill: Do we need to store the size of the device in the Static Header at all? This PR demonstrates that we don't actually do much with it. In #343, I've discovered that it is almost an inconvenience. When we initialize a BlockDev, it has a RangeAllocator which has a limit value, which is the size of the device. When we reconstruct the BlockDev during setup, we must get the RangeAllocator, and must at that time, find out the actual size of the device (which may differ from that on the BDA). When we instantiate the RangeAllocator, we are then in a position where the bda we have loaded is wrong about the size, the RangeAllocator is correct, and the value of the size actually on the BDA on the device is incorrect. We can update the value of the BDA in the BlockDev, but to get it correct on the device, we actually have to write it. We have a lot of ability for this to go wrong, so, the question is, do we actually need it at all?

mulkieran:
@agrover @trgill review? I believe this could be ready to go. It leaves a chunk of unused space that is 12 bytes long, from the 20th to the 32nd byte. The alternative is to shift some data up, but it's not obvious to me that that would be useful.

agrover:
> We can update the value of the BDA in the BlockDev, but to get it correct on the device, we actually have to write it. 

This field is not the "actual size of the blockdev", it's the "size of the blockdev that we expect to be able to use". If the two values differ, then that doesn't mean we need to correct the copy in the BDA, but rather that -- especially if the actual space is less than what we expect -- an assumption we were relying on has been violated and we probably need to fail ASAP so the admin can fix things. (We just had our data truncated, the sooner the admin knows, the more likely they can restore it intact.)

mulkieran:
We already have a value for the size of the blockdev that we expect to be able to use, it's the largest sector yet allocated and we are keeping track of that via the data that is used to instantiate the blockdev mgr.

What is the problem if the device is smaller but no data is lost? Is that really an indication of anything bad? What precisely is the assumption that has been violated?

trgill:
should we remove the "review final" tag on this PR?

mulkieran:
Yes.

agrover:
Storing the size seems very common. Therefore I think we should feel ok with keeping it in the BDA. See below:

LVM
64 bit size (I think in sectors)
in both MDA_header and raw_locn struct
source: lvm2 git lib/format_text/layout.h

ZFS
Name: “asize”
Value: DATA_TYPE_UINT64
Description: Amount of space that can be allocated from this top
level vdev
source: ZFS on-disk format PDF

XFS (xfs_rfsblock_t  sb_dblocks)
64bit count of blocks (blocks are usually 4K)
source: http://xfs.org/docs/xfsdocs-xml-dev/XFS_Filesystem_Structure/tmp/en-US/html/Allocation_Groups.html#Superblocks

ext4 (s_blocks_count_lo, s_blocks_count_hi)
64bit count of blocks (blocks are usually 4K)
source: https://ext4.wiki.kernel.org/index.php/Ext4_Disk_Layout

mulkieran:
I'ld like to know why they do this...

mulkieran:
Opened issue #435, instead.

PR #367, opened 2017-05-30
Make StaticHeader::sigblock_to_buf private

mulkieran:
It is only used within the module, so it can be made private to the module.


mulkieran:
@trgill review?

This isn't going to get any more commits, so it might as well get merged.

Issue #188, opened 2016-12-20
setup::initialize method should possibly clean-up if failure to write sigblock occurs.


mulkieran:
Currently, it errors out in the middle of initialization.

It might be better to back out and zero out devices that were initialized before the error.

mulkieran:
This is related to #203, and should probably be handled together with it.

PR #368, opened 2017-05-30
If batch initialization fails, clean up

mulkieran:
Wipe the headers of all block devs that had been initialized
in the method before the failure occurred.

Resolves: #188


mulkieran:

PR #370, opened 2017-05-31
Serialize variable length metadata

mulkieran:
This PR achieves the serialization of all the variable length metadata fields that the current implementation of Stratis supports.


mulkieran:
@agrover @trgill review?

PR #373, opened 2017-05-31
Add a function to obtain BlockDevs from metadata

mulkieran:
This PR adds a function to obtain BlockDevs belonging to a pool from the pool's metadata. It also adds methods to distinguish between two BlockDev sizes, its current capacity and the size that it was recorded to have in the metadata. A difference between these two numbers may be cause for a warning or other action at a later stage.

The function ```get_pool_blockdevs``` is quite conservative. This is correct, but at a future date it might be reasonable to make it less conservative. For example, if any blockdev belonging to any pool can not be instantiated, the entire method returns an error. In future, it might make sense to modify the method so that a failure for one pool's blockdevs still allows it to discover the BlockDevs for other pools. This would allow Stratis to be set up with a subset of the pools it thinks it should have.

### DIFF
--- a/src/engine/strat_engine/blockdev.rs
+++ b/src/engine/strat_engine/blockdev.rs
@@ -56,7 +56,7 @@ impl BlockDev {
     /// List the available-for-upper-layer-use range in this blockdev.
     pub fn avail_range(&self) -> Segment {
         let start = self.bda.size();
-        let size = self.size();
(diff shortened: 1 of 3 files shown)


mulkieran:
rebased

mulkieran:
@agrover @trgill review?

agrover:
Looks great! 👍

PR #374, opened 2017-05-31
Do not cast IEC constants to u64

mulkieran:
They are u64 already.


mulkieran:
@agrover review?

PR #375, opened 2017-06-01
Resolve workaround and rely on FromIterator

agrover:
I noticed the comment in `BlockDevMgr::record` and spent some time trying to resolve the worked-around issue. This patch also takes advantage of some slick Rust support for turning a `Vec<Result<T>>` into `Result<Vec<T>>` where it will return the first `Err`, or the collection. I found one other spot where this feature helped. Other spots where we have a `Vec<Result<T>>` didn't seem to benefit as much, in my opinion, so I left them.

### DIFF
--- a/src/engine/strat_engine/blockdevmgr.rs
+++ b/src/engine/strat_engine/blockdevmgr.rs
@@ -142,24 +142,13 @@ impl BlockDevMgr {
 
 impl Recordable<HashMap<String, BlockDevSave>> for BlockDevMgr {
     fn record(&self) -> EngineResult<HashMap<String, BlockDevSave>> {
-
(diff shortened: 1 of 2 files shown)


src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -142,24 +142,13 @@ impl BlockDevMgr {
...
-                }
-                Err(err) => return Err(err),
-            }
-        }
-        Ok(result)
+        self.block_devs
+            .iter()
+            .map(|bd| {
+                     bd.record()
+                         .and_then(|bds| Ok((bd.uuid().simple().to_string(), bds)))
```
mulkieran: ```bds``` looks like a plural. ```bdsave``` is better.

PR #380, opened 2017-06-01
update pool::uuid() to return a PoolUuid

trgill:
Patch fixes what I think was just a typo.


trgill:
@agrover @mulkieran I think this is trivial - but I seem to curse myself whenever I use that label. :-)

PR #395, opened 2017-06-06
set thindev allocated for filesystem to a default size of 1 TiB

trgill:
Also updated StratPool::initialize() to use the  DATA_LOWATER constant.


agrover:
👍

PR #397, opened 2017-06-07
Abstract the action of putting filesystem properties on the D-Bus

mulkieran:
Now that we put two normal properties belonging to the fileystem, it is
time to abstract getting a handle on the filesystem into some kind of
generic method. Passing a function that actually gets the property seems
like a good approach.


mulkieran:
@agrover @trgill review?

agrover:
looks great! merging.

PR #401, opened 2017-06-09
Get rid of a stray 'extern' directive

mulkieran:
We already indicate that rand is external at top level.


mulkieran:
@trgill review?

PR #407, opened 2017-06-13
Change real space backing thin devs to 1 GiB

trgill:
Updated the initial size of real space allocated to back
a thin pool to 768 MiB.  Stratis currently gives thin devices a
virtual size of 1 TiB.  The XFS filesystem consumes ~547 MiB for
meta data when provisioned over a 1 TiB device.  If the
INITIAL_DATA_SIZE is only 512 MiB, Stratis will need to extend the
pool as the filesystem is writing the meta data.  Thin pool
expansion is not yet in place.  This update will allow provisioning
XFS without running out of real space in the pool.

### DIFF
--- a/src/engine/strat_engine/pool.rs
+++ b/src/engine/strat_engine/pool.rs
@@ -37,7 +37,7 @@ const META_LOWATER: u64 = 512;
 const DATA_LOWATER: DataBlocks = DataBlocks(512);
 
 const INITIAL_META_SIZE: Sectors = Sectors(16 * Mi / SECTOR_SIZE as u64);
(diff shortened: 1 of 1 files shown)


agrover:
I'm ok with this change. However, either in this PR or a follow-on I think we need to make sure things are consistent between the INITIAL_*_SIZE consts and MIN_DEV_SIZE. If we create a pool with a single blockdev of MIN_DEV_SIZE, we should not run out of room in carving it up.

mulkieran:
I'm not certain that the new value shouldn't be 768 MiB, i.e., 256 + 512. That leaves plenty of space, but should also. allow the MIN_DEV_SIZE to remain at 1 GiB.

PR #408, opened 2017-06-13
Complete a simplified version of setup

mulkieran:


mulkieran:
rebased.

mulkieran:
@agrover @trgill I'm seeking preliminary feedback on this one.

src/engine/strat_engine/pool.rs
```diff
@@ -105,22 +109,46 @@ impl StratPool {
...
+        let blockdevs = try!(get_blockdevs(&metadata, devnodes));
+        let (thinpool, mdv) = try!(get_dmdevs(uuid, &blockdevs, &metadata));
+        let filesystems = try!(get_filesystems(uuid, &thinpool, &mdv));
+        let thindev_ids =
+            ThinDevIdPool::new_from_ids(filesystems.iter().map(|x| x.thin_id()).collect());
+        Ok(StratPool {
+               name: metadata.name,
+               pool_uuid: uuid,
+               block_devs: BlockDevMgr::new(blockdevs),
+               filesystems: Table::new(),
```
mulkieran: Whoops! I forgot, need to get the filesystems into the table. Otherwise, shoudl be good.

agrover:
setup_thinpooldev seems to unconditionally wipe the meta device's sectors, which seems good for the initialize path, but not so good for the setup path.

mulkieran:
Added a test. Maybe it will fail?

mulkieran:
The very first initialization step in the new test returns an error. Good test!

mulkieran:
rebased

mulkieran:
I think wiping the sectors is good for the setup path, assuming a clean shutdown previously, which is what were doing. So there's no point in messing with it now.

agrover:
If the sectors are wiped on setup, the info on which blocks in the data area map to which thin devs is lost.

mulkieran:
When we tear down the thinpool dev we tear down its meta and its data dev. When we setup from a clean shutdown, we assume that the metadev is not around and make a new one.

It makes sense that the metadev handles metadata about the the thin devices in the thinpool. That's what it is for. But I don't understand how that metadata is preserved when we are tearing down the metadata device when we tear down the engine.

agrover:
Tearing down removes the DM device but any data that was previously written to its sectors remain on the disk. When (re)created, the thinpool target will look for its metadata in the meta device, and if it's zeroed out then it will think it is a new thinpool rather than a recreated existing one. Stratis would then read the MDV, see it has a filesystem with thin id 7, try to create that thin dev, and the thin pool would return an error since its internal metadata no longer tracks that such a thin dev was ever created.

mulkieran:
rebased. However, this is incomplete since reconstructed ```thin_pool_meta_spare``` is just an empty vec.

mulkieran:
Superceded by #415.

Issue #259, opened 2017-03-13
revisit hard coded data blocksize and low water mark


trgill:
Should be more adaptive to what make sense.

mulkieran:
I think these means that these values should not be constants, but instead configurable.

agrover:
low water mark will not be using a hardcoded value, see https://github.com/stratis-storage/stratisd/issues/1083.

That leaves hardcoded data blocksize as a question. For 1.0, the answer is surely no. Beyond that.. maybe someday?

Going to close this issue, we can revisit in the future.

PR #411, opened 2017-06-15
Get rid of a TODO which is certainly in the wrong place

mulkieran:
At this point, the values, although hard-coded are not literals, they're
constants, defined elsewhere. There is an issue for the problem:
https://github.com/stratis-storage/stratisd/issues/259. We can remove the
TODO.


mulkieran:
@trgill review?

PR #412, opened 2017-06-15
Move cleanup code into a separate module

mulkieran:
There will be very much more as time goes on.

Use newly created wipe_blockdevs() method where appropriate.

Tidy up some cleanup code adding TODOs where appropriate.

### DIFF
--- a/src/engine/strat_engine/blockdevmgr.rs
+++ b/src/engine/strat_engine/blockdevmgr.rs
@@ -18,6 +18,7 @@ use super::super::consts::IEC;
 use super::super::errors::{EngineError, EngineResult, ErrorEnum};
 use super::super::types::{DevUuid, PoolUuid};
 
+use super::cleanup::wipe_blockdevs;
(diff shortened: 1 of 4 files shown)


mulkieran:
@agrover @trgill review?

This is starting to block #408 .

Also, note the TODOs. These mark where we should be updating a hypothetical state machine, or, by some other means, communicating the problem in a programmatic way.

Note, however, that the TODOs are in very distinct locations, one in blockdevmgr and the other in pool. That means that any state machine we update must be global, or perhaps we need to enrich the return type of a lot of methods, or perhaps we actually need to be writing this information in some kind of file somewhere, or some combination thereof.

mulkieran:
One option is to pass the state machine, owned by the engine as a mutable reference. That would be a helpful indication in the signature of the method of its effects.

agrover:
👍 for this PR without the environment state commit, based on our discussion in team mtg.

mulkieran:
rebased and whatnot.

PR #413, opened 2017-06-15
StratFilesystem::initialize should call ThinDev::new()

trgill:
StratFilesystem::initialize shouldn't call StratFilesytem::setup
since ::setup calls ThinDev::setup.  On the ::initialize call,
we need to use ThinDev::new().


agrover:
lgtm 👍

PR #414, opened 2017-06-15
Run thin_repair

agrover:
If ThinPoolDev::setup fails due to CheckFailed, run thin_repair. If that
succeeds then swap meta and meta_spare and continue.

Some work here to track meta_spare segments in on-disk metadata.

Relies on some dm-rs changes to expose device rename, as well as returning
moved lineardevs in CheckFailed error so we can try again.


mulkieran:
Superceded by #415.

PR #417, opened 2017-06-16
Add thin pool expansion when a thin-dev exceeds the low water mark

trgill:
Attempt to expand the physical space allocated behind a thin-pool when filesystem
space consumption exceeds the low water mark.

Add tests for expansion.

Remove obsolete comments from Makefile about test order/complexity.

### DIFF
--- a/src/engine/strat_engine/pool.rs
+++ b/src/engine/strat_engine/pool.rs
@@ -43,8 +43,7 @@ use super::setup::{get_blockdevs, get_metadata};
 
 const DATA_BLOCK_SIZE: Sectors = Sectors(2048);
 const META_LOWATER: u64 = 512;
-const DATA_LOWATER: DataBlocks = DataBlocks(512);
-
+pub const DATA_LOWATER: DataBlocks = DataBlocks(512);
 const INITIAL_META_SIZE: Sectors = Sectors(16 * Mi / SECTOR_SIZE as u64);
 const INITIAL_DATA_SIZE: Sectors = Sectors(768 * Mi / SECTOR_SIZE as u64);
 const INITIAL_MDV_SIZE: Sectors = Sectors(16 * Mi / SECTOR_SIZE as u64);
@@ -212,6 +211,25 @@ impl StratPool {
         self.block_devs
             .save_state(&now().to_timespec(), data.as_bytes())
     }
+    /// Return an extend size for the physical space backing a pool
+    /// TODO: returning the current size will double the space provisoned to
+    /// back the pool.  We should determine if this is a reasonable value.
+    fn extend_size(&self, current_size: Sectors) -> Sectors {
+        current_size
+    }
+
+    /// Expand the physical space allocated to a pool by the value from extend_size()
+    /// Return tne number of Sectors added
+    fn extend_data(&mut self, dm: &DM, current_size: Sectors) -> EngineResult<Sectors> {
+        let extend_size = self.extend_size(current_size);
+        if let Some(new_data_regions) = self.block_devs.alloc_space(extend_size) {
+            try!(self.thin_pool.extend_data(dm, new_data_regions));
+        } else {
+            return Err(EngineError::Engine(ErrorEnum::Error,
+                                           format!("No free data regions for pool expansion")));
+        }
+        Ok(extend_size)
+    }
 
     pub fn check(&mut self) -> () {
         #![allow(match_same_arms)]
@@ -255,7 +273,11 @@ impl StratPool {
                 }
 
                 if usage.used_data > usage.total_data - DATA_LOWATER {
-                    // TODO: Extend data device
+                    // Request expansion of physical space allocated to the pool
+                    match self.extend_data(&dm, usage.total_data.0 * DATA_BLOCK_SIZE) {
+                        Ok(_) => {}
+                        Err(_) => {} // TODO: Take pool offline?
+                    }
                 }
             }
             ThinPoolStatus::Fail => {
--- a/src/engine/strat_engine/pool.rs
+++ b/src/engine/strat_engine/pool.rs
@@ -41,11 +41,11 @@ use super::metadata::MIN_MDA_SECTORS;
 use super::serde_structs::{FilesystemSave, FlexDevsSave, PoolSave, Recordable, ThinPoolDevSave};
 use super::setup::{get_blockdevs, get_metadata};
 
-const DATA_BLOCK_SIZE: Sectors = Sectors(2048);
+pub const DATA_BLOCK_SIZE: Sectors = Sectors(2048);
 const META_LOWATER: u64 = 512;
 pub const DATA_LOWATER: DataBlocks = DataBlocks(512);
 const INITIAL_META_SIZE: Sectors = Sectors(16 * Mi / SECTOR_SIZE as u64);
-const INITIAL_DATA_SIZE: Sectors = Sectors(768 * Mi / SECTOR_SIZE as u64);
+pub const INITIAL_DATA_SIZE: Sectors = Sectors(768 * Mi / SECTOR_SIZE as u64);
 const INITIAL_MDV_SIZE: Sectors = Sectors(16 * Mi / SECTOR_SIZE as u64);
 
 #[derive(Debug)]
--- a/tests/util/pool_tests.rs
+++ b/tests/util/pool_tests.rs
@@ -1,0 +1,63 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+/// Test the functionality of stratis pools.
+extern crate devicemapper;
+extern crate env_logger;
+
+use std::fs::OpenOptions;
+use std::io::Write;
+use std::path::Path;
+
+use self::devicemapper::DM;
+use self::devicemapper::Sectors;
+use self::devicemapper::consts::SECTOR_SIZE;
+
+use libstratis::engine::Pool;
+use libstratis::engine::strat_engine::pool::{DATA_BLOCK_SIZE, DATA_LOWATER, INITIAL_DATA_SIZE,
+                                             StratPool};
+use libstratis::engine::strat_engine::StratEngine;
+use libstratis::engine::types::Redundancy;
+
+/// Verify that a the physical space allocated to a pool is expanded when
+/// the nuber of sectors written to a thin-dev in the pool exceeds the
+/// INITIAL_DATA_SIZE.  If we are able to write more sectors to the filesystem
+/// than are initially allocated to the pool, the pool must have been expanded.
+pub fn test_thinpool_expand(paths: &[&Path]) -> () {
+    StratEngine::initialize().unwrap();
+    let mut pool = StratPool::initialize("stratis_test_pool",
+                                         &DM::new().unwrap(),
+                                         paths,
+                                         Redundancy::NONE,
+                                         true)
+            .unwrap();
+    let &(_, fs_uuid) = pool.create_filesystems(&vec!["stratis_test_filesystem"])
+        .unwrap()
+        .first()
+        .unwrap();
+
+    let devnode = pool.get_filesystem(&fs_uuid)
+        .unwrap()
+        .devnode()
+        .unwrap();
+    // Braces to ensure f is closed before destroy
+    {
+        let mut f = OpenOptions::new().write(true).open(devnode).unwrap();
+        // Write 1 more sector than is initially allocated to a pool
+        let write_size = INITIAL_DATA_SIZE + Sectors(1);
+        let buf = &[1u8; SECTOR_SIZE];
+        for i in 0..*write_size {
+            f.write_all(buf).unwrap();
+            // Simulate handling a DM event by running a pool check at the point where
+            // the amount of free space in pool has decreased to the DATA_LOWATER value.
+            // TODO: Actually handle DM events and possibly call extend() directly,
+            // depending on the specificity of the events.
+            if i == *(INITIAL_DATA_SIZE - Sectors(*DATA_LOWATER * *DATA_BLOCK_SIZE)) {
+                pool.check();
+            }
+        }
+    }
+    pool.destroy_filesystems(&[&fs_uuid]).unwrap();
+    pool.teardown().unwrap();
+}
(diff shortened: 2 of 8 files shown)


trgill:
The CI will fail until ThinPool::extend_data() is added to devicemapper-rs.

src/engine/strat_engine/pool.rs
```diff
@@ -168,6 +168,21 @@ impl StratPool {
             .save_state(&now().to_timespec(), data.as_bytes())
     }
 
+    /// Expand the physical space allocated to a pool by the DATA_EXPANSION_PERCENT
+    fn extend_data(&mut self, current_size: Sectors) -> EngineResult<Sectors> {
+        let expand_size: Sectors =
+            Sectors((current_size.0 as f64 * (DATA_EXPANSION_PERCENT as f64 / 100.0 as f64))
+                        .round() as u64);
+        if let Some(new_data_regions) = self.block_devs.alloc_space(expand_size) {
+            let dm = DM::new().expect("Could not get DM handle");
```
agrover: Any reason not to use `try!` here?

Overall LGTM. 👍

agrover: oops! PR wasn't yet tagged for review! Sometimes I just get carried away 😃

trgill: It is fine.  I was thinking I should wait until it can pass the CI.  It needs the new version of device-mapper-rs.

Not sure I follow your comment.  try! for getting the DM handle or alloc_space()?  alloc_space() returns an Option, right?

agrover:
Yeah just the DM handle.

trgill:
Ok.  Thanks.  Your comment made me realize I should have just passed in the dm handle from check().  So, I made that update.

agrover:
👍😀

agrover:
Rebased, resolved conflicts and fixed a trivial CI failure. CI is still failing, but in a nontrivial manner. Something needs work.

trgill:
test_setup leaves a some DM devices around and a couple filesystems mounted.  It causes the find_all() call this test uses to fail.

@mulkieran should I put together a PR to fix that test?

mulkieran:
@trgill Please do. I _think_ that if ```engine.teardown()``` doesn't solve the problem, we'll have more work to do in the actual code. But I'm optimistic that it will.

mulkieran:
These days we're not explicitly updating the Cargo.toml. Instead, we leave devicemapper at 0.5 and do a ```cargo update --package devicemapper" which update the Cargo.lock file.

tests/util/simple_tests.rs
```diff
@@ -25,22 +25,50 @@ use self::tempdir::TempDir;
...
+/// INITIAL_DATA_SIZE.  If we are able to write more sectors to the filesytem
+/// than are initially allocated to the pool, the pool must have been expanded.
+fn verify_pool_expand(pool: &mut StratPool, fs_uuid: &FilesystemUuid) {
+    let devnode = pool.get_filesystem(&fs_uuid)
+        .unwrap()
+        .devnode()
+        .unwrap();
+    let mut f = OpenOptions::new().write(true).open(devnode).unwrap();
+    // Write 10 percent more sectors than is initially allocated to a pool
+    let write_size: u64 = (INITIAL_DATA_SIZE.0 as f64 +
```
mulkieran: Why not just make ```write_size``` double the initial? If the amount is just doubled, that should test expansion once. If the algorithm is refined, so that the increase is less than double, then the test will just test increase over multiple iterations, which ought to be good.

tests/util/simple_tests.rs
```diff
@@ -25,22 +25,50 @@ use self::tempdir::TempDir;
...
+                           (INITIAL_DATA_SIZE.0 as f64 * (0.10 as f64 / 100.0 as f64)))
+            .round() as u64;
+    let buf = &[1u8; SECTOR_SIZE];
+    for i in 0..write_size {
+        f.write_all(buf).is_ok();
+        // Run a pool.check() for each block written.  This simulates an event that would be
+        // handled by Stratis.
+        if i % DATA_BLOCK_SIZE.0 == 0 {
```
mulkieran: I understand that you're simulating an event, but I don't see why it should be simulated every ```i % DATA_BLOCK_SIZE.0``` times. What sort of event would it be?

src/engine/strat_engine/pool.rs
```diff
@@ -41,12 +41,12 @@ use super::metadata::MIN_MDA_SECTORS;
 use super::serde_structs::{FilesystemSave, FlexDevsSave, PoolSave, Recordable, ThinPoolDevSave};
 use super::setup::{get_blockdevs, get_metadata};
 
-const DATA_BLOCK_SIZE: Sectors = Sectors(2048);
+pub const DATA_BLOCK_SIZE: Sectors = Sectors(2048);
 const META_LOWATER: u64 = 512;
 const DATA_LOWATER: DataBlocks = DataBlocks(512);
-
+const DATA_EXPANSION_PERCENT: f32 = 20.0;
```
mulkieran: With proposed change ```DATA_EXPANSION_PERCENT``` can probably go away.

trgill: I was thinking that slow growth would make things easier on a storage admin.  I think doubling or moving to the next power of 2 is fine for very small filesystems, but I don't think it scales.  I think in the long run - we will want to make this a modifiable setting.  How about if I put it in a function, open an issue and we can discuss when @agrover gets back?

mulkieran: It's not the amount I object to, it's the use of percentages at all, and the use of floats to represent percentages.

Shoving the calculation into a function with a TODO seems like the right thing to do. But please leave out any use of percentages or floats. Doubling is just the simplest thing to do, so I'ld recommend it.

tests/util/simple_tests.rs
```diff
@@ -25,22 +25,50 @@ use self::tempdir::TempDir;
...
 fn dirty_sectors(path: &Path, offset: Sectors, length: Sectors) {
     write_sectors(path, offset, length, &[1u8; SECTOR_SIZE]).unwrap();
 }
 
+/// Verify that a the physical space allocated to a pool is expanded when
+/// the nuber of sectors written to a thin-dev in the pool exceeds the
+/// INITIAL_DATA_SIZE.  If we are able to write more sectors to the filesytem
+/// than are initially allocated to the pool, the pool must have been expanded.
+fn verify_pool_expand(pool: &mut StratPool, fs_uuid: &FilesystemUuid) {
+    let devnode = pool.get_filesystem(&fs_uuid)
```
trgill: ok

tests/util/simple_tests.rs
trgill: We need to implement something that listens to DM events.  When a "low water mark crossed" event is triggered - Stratis will field the event and call check on the pool.  

Until we've got a listener - I'm manually calling check in the test.  How about if I open an issue, add a TODO to the code?

I could also change the code to only call the check() function when we are within X (say 100) sectors of the INITIAL_DATA_SIZE?

src/engine/strat_engine/pool.rs
trgill: FYI - LVM uses a percentage for expansion when low water mark is crossed.  But, I'll remove it.

tests/util/simple_tests.rs
```diff
@@ -212,6 +240,21 @@ pub fn test_thinpool_device(paths: &[&Path]) -> () {
...
+    StratEngine::initialize().unwrap();
+    let mut pool: StratPool = StratPool::initialize("stratis_test_pool",
+                                                    &DM::new().unwrap(),
+                                                    paths,
+                                                    Redundancy::NONE,
+                                                    true)
+            .unwrap();
+    let fs_vec = pool.create_filesystems(&vec!["stratis_test_filesystem"])
+        .unwrap();
+    let &(_, fs_uuid): &(&str, FilesystemUuid) = fs_vec.first().unwrap();
```
trgill: ok

tests/util/simple_tests.rs
trgill: The intent was to save time.  Writing sector by sector is slow and 10 percent more is enough.  If you feel strongly - I'll change it.

mulkieran:
Do you know why cargo.lock got so huge?

trgill:
I ran a cargo update to get the latest devicemapper-rs.

looks like ~15 lines were added... it didn't get much bigger, right?  the changes concern you?

tests/util/pool_tests.rs
```diff
@@ -0,0 +1,66 @@
...
+use libstratis::engine::strat_engine::device::wipe_sectors;
+use libstratis::engine::strat_engine::metadata::BDA_STATIC_HDR_SECTORS;
+use libstratis::engine::strat_engine::pool::{DATA_BLOCK_SIZE, DATA_LOWATER, INITIAL_DATA_SIZE,
+                                             StratPool};
+use libstratis::engine::strat_engine::StratEngine;
+use libstratis::engine::types::Redundancy;
+
+/// Verify that a the physical space allocated to a pool is expanded when
+/// the nuber of sectors written to a thin-dev in the pool exceeds the
+/// INITIAL_DATA_SIZE.  If we are able to write more sectors to the filesytem
```
mulkieran: Missing 's' in "filesytem".

tests/util/pool_tests.rs
```diff
@@ -0,0 +1,66 @@
...
+        .unwrap();
+    let &(_, fs_uuid) = fs_vec.first().unwrap();
+    let devnode = pool.get_filesystem(&fs_uuid)
+        .unwrap()
+        .devnode()
+        .unwrap();
+    // Braces to ensure f is closed before destroy
+    {
+        let mut f = OpenOptions::new().write(true).open(devnode).unwrap();
+        // Write 10 percent more sectors than is initially allocated to a pool
```
mulkieran: Just write one more sector than the initial size, so:

```
/// Write 1 more sector than is initially allocated to a pool
let write_size: INITIAL_DATA_SIZE + Sectors(1); // this is what lines 51 - 53 are actually doing AFAICT.
let buf = &[1u8; SECTOR_SIZE];
for i in 0..write_size { // maybe a cast to u64 here as ```*write_size as u64```
    f.write_all(buf).unwrap();
    // Simulate handling a DM event by running a pool check at the point where the amount of free space in pool has decreased
    // to the DATA_LOWATER value.
    // TODO: Actually handle DM events and possibly call extend() directly, depending on the specificity of the events.
    if i == *(INITIAL_DATA_SIZE - Sectors(*DATA_LOWATER * DATA_BLOCK_SIZE)) {
        pool.check();
    }
}
```

When the pool is allocated, it gets exactly INITIAL_BLOCK_SIZE for data, so the extra ```- 100``` seems unnecessary.

mulkieran:
@trgill @grover would prefer that we didn't do a wholesale update. Try resetting the Cargo.lock to its previous value and running ```cargo update --package devicemapper``` if that's the only change you need. Hopefully the Cargo.lock file changes will be a bit smaller. There is a possibility that we will need to update clippy as well, but try w/out doing that, first.

trgill:
Ok - I ran "cargo update --package devicemapper" and updated the commit with Cargo.lock

trgill:
@mulkieran this ok?

tests/util/pool_tests.rs
```diff
@@ -0,0 +1,66 @@
...
```
mulkieran: Fix comment to match code.

tests/util/pool_tests.rs
```diff
@@ -0,0 +1,66 @@
...
+                                             StratPool};
+use libstratis::engine::strat_engine::StratEngine;
+use libstratis::engine::types::Redundancy;
+
+/// Verify that a the physical space allocated to a pool is expanded when
+/// the nuber of sectors written to a thin-dev in the pool exceeds the
+/// INITIAL_DATA_SIZE.  If we are able to write more sectors to the filesystem
+/// than are initially allocated to the pool, the pool must have been expanded.
+pub fn test_thinpool_expand(paths: &[&Path]) -> () {
+    for path in paths {
```
mulkieran: This should be taken care of in the test setup infrastructure, so it should be unneeded here.

tests/util/pool_tests.rs
```diff
@@ -0,0 +1,66 @@
...
+        .unwrap()
+        .devnode()
+        .unwrap();
+    // Braces to ensure f is closed before destroy
+    {
+        let mut f = OpenOptions::new().write(true).open(devnode).unwrap();
+        // Write 10 percent more sectors than is initially allocated to a pool
+        let write_size = INITIAL_DATA_SIZE + Sectors(1);
+        let buf = &[1u8; SECTOR_SIZE];
+        for i in 0..write_size.0 {
```
mulkieran: Use ```*``` here, not ```.0```.

tests/util/pool_tests.rs
```diff
@@ -0,0 +1,66 @@
...
+            // Simulate handling a DM event by running a pool check at the point where
+            // the amount of free space in pool has decreased to the DATA_LOWATER value.
+            // TODO: Actually handle DM events and possibly call extend() directly,
+            // depending on the specificity of the events.
+            if i == *(INITIAL_DATA_SIZE - Sectors(*DATA_LOWATER * *DATA_BLOCK_SIZE)) {
+                pool.check();
+            }
+        }
+    }
+    pool.destroy_filesystems(&vec![&fs_uuid]).unwrap();
```
mulkieran: No need for ```vec!```. I think yo ucan just do ```&[&fs_uuid]```.

mulkieran:
The large amount of changes in Cargo.lock may be required to get clippy working, though.

tests/util/pool_tests.rs
```diff
@@ -0,0 +1,61 @@
...
+/// than are initially allocated to the pool, the pool must have been expanded.
+pub fn test_thinpool_expand(paths: &[&Path]) -> () {
+    StratEngine::initialize().unwrap();
+    let mut pool: StratPool = StratPool::initialize("stratis_test_pool",
+                                                    &DM::new().unwrap(),
+                                                    paths,
+                                                    Redundancy::NONE,
+                                                    true)
+            .unwrap();
+    let fs_vec = pool.create_filesystems(&vec!["stratis_test_filesystem"])
```
mulkieran: You might as well compress this line and the one below into a single line.

tests/util/pool_tests.rs
```diff
@@ -0,0 +1,61 @@
...
+use libstratis::engine::strat_engine::StratEngine;
+use libstratis::engine::types::Redundancy;
+
+/// Verify that a the physical space allocated to a pool is expanded when
+/// the nuber of sectors written to a thin-dev in the pool exceeds the
+/// INITIAL_DATA_SIZE.  If we are able to write more sectors to the filesystem
+/// than are initially allocated to the pool, the pool must have been expanded.
+pub fn test_thinpool_expand(paths: &[&Path]) -> () {
+    StratEngine::initialize().unwrap();
+    let mut pool: StratPool = StratPool::initialize("stratis_test_pool",
```
mulkieran: This line shouldn't need the type annotation.

trgill:
fixed cargo.lock and made updates as requested.

agrover:
👍

PR #421, opened 2017-06-19
Put devicemapper use directive below other crate use directives

mulkieran:
It is more specific than other non-standard crates because it is a Stratis
project.


mulkieran:
See https://github.com/stratis-storage/stratis-docs/issues/43.

PR #427, opened 2017-06-22
Add a Default implementation for Table, ActionQueue and SimEngine

mulkieran:
Also for Randomizer, in order to make implementation for SimEngine easier.
Get rid of new(), and use default() where appropriate.


mulkieran:
This is just barely an improvement, but it fixes a previous mistake, so it is going in.

PR #430, opened 2017-06-27
Get rid of some clippy complaints

mulkieran:
These are all clippy motivated updates.

This gets rid of a few clippy errors and runs clippy target verbosely for greater detail.

Getting rid of these particular errors should expose the remaining, where as leaving these particular errors in obscures all the other errors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ docs:
 	cargo doc --no-deps
 
 clippy:
-	RUSTFLAGS='-D warnings' cargo build --features "clippy"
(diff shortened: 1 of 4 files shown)


mulkieran:
@trgill review?

mulkieran:
stratisd no longer compiles on nightly, but I don't know why. This might be something that the Rust people would like to know, but, in the mean time, we shouldn't try to work on these clippy errors, since we can only run clippy on nightly.

mulkieran:
@trgill review, please?

mulkieran:
@trgill review?

PR #432, opened 2017-06-29
Use AddAssign instead of Assign

mulkieran:
This change gets rid of clippy error assign_op_pattern.


mulkieran:
@trgill review?

Issue #433, opened 2017-06-29
determine best way to expand physical space backing a pool


trgill:
Currently when data consumption crosses the low water mark, we double the physical space.  Not sure this approach will scale well...

mulkieran:
We also may want to, if allocation fails, try for a smaller number, and report the problem. It may be better to have a partial success in this case then to fail dramatically.

trgill:
related to #494

mulkieran:
it was noted that we should take into account the granularity of the allocation. For example, if allocating for data device in thinpool, anything less than a full data block is no use. There is plenty to do in this area.

mulkieran:
I feel like I might as well close this in favor of more recent #693 and #697.

PR #431, opened 2017-06-28
Package the thinpool device and extra segments together

mulkieran:
For better encapsulation.

Signed-off-by: mulhern <amulhern@redhat.com>

PR #436, opened 2017-07-05
Enrich return type of BlockDevMgr::alloc_space

mulkieran:


mulkieran:
This will start to block #431 pretty soon.

mulkieran:

src/engine/strat_engine/pool.rs
```diff
@@ -211,24 +215,25 @@ impl StratPool {
...
-    /// TODO: returning the current size will double the space provisoned to
-    /// back the pool.  We should determine if this is a reasonable value.
-    fn extend_size(&self, current_size: Sectors) -> Sectors {
-        current_size
-    }
 
-    /// Expand the physical space allocated to a pool by the value from extend_size()
-    /// Return tne number of Sectors added
+    /// Expand the physical space allocated to a pool.
+    /// Return tne number of Sectors added.
```
trgill: typo

src/engine/strat_engine/pool.rs
```diff
@@ -211,24 +215,25 @@ impl StratPool {
...
+    // TODO: Improve this function:
+    // * If there is enough space available the size of the new region is
+    // double the size of the old. This is most likely too greedy, we need a
+    // more sophisticated policy.
+    // * This method succeeds if the pool can be extended at all. This is
+    // most likely too generous a policy, as a very small extension may be
+    // no help at all.
```
trgill: I agree this might be no help at all.  It may just require the caller to deal with a case where a useless amount of space was allocated.  Let's see what @agrover thinks.  I'd suggest that it would be better to figure out a better allocation policy.

mulkieran: And it might not...if the allocation policy says allocate X amount of space, and only Y amount of space is left, and Y is only a tiny bit less than X, then Y may be just as good as X. The allocation policy and the amount actually allocated and what to do about it really can be dealt with separately. The tricky part arises if Y is actually very small, and then it is tricky for a number of reasons.

src/engine/strat_engine/pool.rs
mulkieran: Fixed.

src/engine/strat_engine/pool.rs
agrover: There's a granularity: if the thinpool DataBlock size is 4096, there is no benefit by extending the data device by 4095 sectors. Can we return a multiple of a given granularity?

But besides granularity, if we ask for 100 DataBlock equivalents and only get one, we'd certainly want to use that one, I'd think.

mulkieran: Yeah. 1 DataBlock is a way better starting point for lower bound than 1 Sector.

mulkieran:
So, we know that the initial size is divisible by the size of a datablock. And we know that if we always expand by doubling, the new size will also be divisible by the size of a datablock. If we always get the amount we ask for, then we will never have a problem. If we sometimes get less, then we should round down to the nearest datablock, get rid of some previously allocated blocks, and allocate the reduced amount. The rounding should definitely be implemented in the Sectors class. I'm pushing this back to "in progress" until that can be done.

mulkieran:
The issue is covered in #433 and there's not really anything in this old PR worth keeping around.

PR #424, opened 2017-06-21
Simplify and document metadata

mulkieran:
* Move MDAHeader and MDARegions into a separate submodule.
* Add a lot of header docs.
* Refactor MDARegions so it stores a list of ```Option<MDAHeader>``` instead of storing a list of ```MDAHeader```s which have almost all Option fields.
* Do not check crc of variable length metadata when initially loading BDA.

PR #438, opened 2017-07-06
Simplify, extend, and document metadata

mulkieran:
Also, make use of some extensions to enhance ```BlockDevMgr::save_state()```. The enhancements allow it to conform, more or less entirely, to the specifications in the design doc.

A principle step is to move MDARegions and MDAHeader into their own separate module internal to metadata.
This controls the access to MDAHeader implementation so that it is only through MDARegions methods. All MDAHeader methods are private to the module.

An important step is that the region size is no longer stored in each MDAHeader, but only in the containing MDARegions struct. Since access to MDAHeader methods is completely controlled MDARegions methods can verify that the data is not too big to be written. This change enable the simplification of MDARegions methods, since MDAHeader do not have Option fields, but can themselves be Option fields of their MDARegions. Moreover, ```MDAHeader::to_buf()``` can simply take a reference to an MDAHeader, rather than a selection of its fields.

An important change in the behavior is that loading an MDAHeader does not check the crc of the variable length metadata.

The maximum size of variable length metadata that can be written to a particular blockdev is now exported for the use of ```BlockDevMgr::save_state()```. ```BlockDevMgr::save_state()``` now conforms to its specification in the design doc with the caveat that the maximum number of blockdevs written to is designated by a constant, and perhaps should be configurable instead.

### DIFF
--- a/src/engine/strat_engine/blockdev.rs
+++ b/src/engine/strat_engine/blockdev.rs
@@ -92,6 +92,11 @@ impl BlockDev {
         self.used.available()
     }
 
+    /// The maximum size of variable length metadata that can be accommodated.
+    pub fn max_metadata_size(&self) -> Sectors {
+        self.bda.max_data_size()
+    }
+
     // Find some sector ranges that could be allocated. If more
     // sectors are needed than our capacity, return partial results.
     pub fn request_space(&mut self, size: Sectors) -> (Sectors, Vec<Segment>) {
--- a/src/engine/strat_engine/blockdevmgr.rs
+++ b/src/engine/strat_engine/blockdevmgr.rs
@@ -10,7 +10,7 @@ use std::fs::{File, OpenOptions};
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 
-use time::Timespec;
+use time::{Duration, Timespec};
 use uuid::Uuid;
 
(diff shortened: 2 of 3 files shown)


mulkieran:
Supercedes #424.

mulkieran:
@agrover @trgill review? The only significant code changes from previous version are the two last commits.

agrover:
👍 thanks!

src/engine/strat_engine/blockdev.rs
```diff
@@ -92,6 +92,11 @@ impl BlockDev {
         self.used.available()
     }
 
+    /// The maximum size of variable length metadata that can be accomodated.
```
trgill: accommodated

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -126,13 +132,32 @@ impl BlockDevMgr {
 
     /// Write the given data to all blockdevs marking with specified time.
     /// Return an error if data was not written to any blockdev.
-    // TODO: Cap # of blockdevs written to, as described in SWDD
+    /// Omit blockdevs which do not have sufficient space in BDA to accomodate
```
trgill: same here... accommodate

mulkieran:
Typos fixed.

PR #444, opened 2017-07-07
Style changes, mostly in setup.rs

agrover:
These are style changes that do not change the meaning of the code, and attempt to use Rust higher-level idioms (e.g. `filter_map` and `and_then`) to minimize the use of lower-level things (e.g. `continue` and `expect`).

### DIFF
--- a/src/engine/strat_engine/setup.rs
+++ b/src/engine/strat_engine/setup.rs
@@ -113,30 +113,24 @@ pub fn get_metadata(pool_uuid: PoolUuid, devnodes: &[PathBuf]) -> EngineResult<O
         }
     }
 
-    // We may have had no devices with BDAs for this pool, so return if no BDAs.
-    if bdas.is_empty() {
-        return Ok(None);
-    }
-
-    // Get a most recent BDA
-    let &(_, ref most_recent_bda) = bdas.iter()
-        .max_by_key(|p| p.1.last_update_time())
-        .expect("bdas is not empty, must have a max");
-
     // Most recent time should never be None if this was a properly
     // created pool; this allows for the method to be called in other
     // circumstances.
-    let most_recent_time = most_recent_bda.last_update_time();
-    if most_recent_time.is_none() {
-        return Ok(None);
-    }
+    let most_recent_time = {
+        match bdas.iter()
+                  .filter_map(|&(_, ref bda)| bda.last_update_time())
+                  .max() {
+            Some(time) => time,
+            None => return Ok(None),
+        }
+    };
 
     // Try to read from all available devnodes that could contain most
     // recent metadata. In the event of errors, continue to try until all are
     // exhausted.
     for &(devnode, ref bda) in
         bdas.iter()
-            .filter(|&&(_, ref bda)| bda.last_update_time() == most_recent_time) {
+            .filter(|&&(_, ref bda)| bda.last_update_time() == Some(most_recent_time)) {
 
         let poolsave = OpenOptions::new()
             .read(true)
--- a/src/engine/strat_engine/setup.rs
+++ b/src/engine/strat_engine/setup.rs
@@ -195,10 +195,9 @@ pub fn get_blockdevs(pool_save: &PoolSave, devnodes: &[PathBuf]) -> EngineResult
         let device = try!(Device::from_str(&dev.to_string_lossy()));
 
         // If we've seen this device already, skip it.
-        if devices.contains(&device) {
+        if !devices.insert(device) {
             continue;
         }
-        devices.insert(device);
 
         let bda = try!(BDA::load(&mut try!(OpenOptions::new().read(true).open(dev))));
         let bda = try!(bda.ok_or(EngineError::Engine(ErrorEnum::NotFound,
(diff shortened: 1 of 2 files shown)


src/engine/strat_engine/setup.rs
```diff
@@ -147,23 +147,18 @@ pub fn get_metadata(pool_uuid: PoolUuid, devnodes: &[PathBuf]) -> EngineResult<O
...
-        } else {
-            continue;
+            .filter(|&&(_, ref bda)| bda.last_update_time() == most_recent_time) {
+
+        let poolsave = OpenOptions::new()
+            .read(true)
+            .open(devnode)
+            .ok()
+            .and_then(|mut f| bda.load_state(&mut f).ok())
+            .and_then(|opt| opt)
```
mulkieran: Is this line really necessary?

src/engine/strat_engine/setup.rs
```diff
@@ -206,6 +208,20 @@ pub fn get_blockdevs(pool_save: &PoolSave, devnodes: &[PathBuf]) -> EngineResult
         let actual_size = try!(blkdev_size(&try!(OpenOptions::new().read(true).open(dev))))
             .sectors();
 
+        match actual_size.cmp(&bda.dev_size()) {
```
mulkieran: We should handle this elsewhere...we have functions to do so. Look for ```recorded_size```.

src/engine/strat_engine/setup.rs
```diff
@@ -120,30 +120,21 @@ pub fn get_metadata(pool_uuid: PoolUuid, devnodes: &[PathBuf]) -> EngineResult<O
...
-    let most_recent_time = most_recent_bda.last_update_time();
-    if most_recent_time.is_none() {
-        return Ok(None);
-    }
+    let most_recent_time = {
+        match bdas.iter()
+                  .filter_map(|&(_, ref bda)| bda.last_update_time())
+                  .max() {
+            Some(time) => time,
+            None => return Ok(None), // No max means no devices with metadata
```
mulkieran: Please drop the comment.

src/engine/strat_engine/setup.rs
```diff
@@ -120,30 +120,21 @@ pub fn get_metadata(pool_uuid: PoolUuid, devnodes: &[PathBuf]) -> EngineResult<O
...
-    if bdas.is_empty() {
-        return Ok(None);
-    }
-
-    // Get a most recent BDA
-    let &(_, ref most_recent_bda) = bdas.iter()
-        .max_by_key(|p| p.1.last_update_time())
-        .expect("bdas is not empty, must have a max");
-
-    // Most recent time should never be None if this was a properly
```
mulkieran: Please retain comment, it is still true.

src/engine/strat_engine/setup.rs
agrover: yes, bda.load_state gives us a `Result<Option<Vec<u8>>>` and this stops the chain if the option is None. I must admit staring at this for a while for a better way to express this but none came to mind. Any suggestions?

src/engine/strat_engine/setup.rs
agrover: ok.

src/engine/strat_engine/setup.rs
agrover: yup will do.

src/engine/strat_engine/setup.rs
agrover: ah yes I remember seeing some work you did on this, but I'm only seeing Blockdev::recorded_size in master branch. Where'd it go?

mulkieran: We do not have any code that uses it yet. The way we should use it is run some sort of check on the pool after it has been reconstructed.

My point is that we do not have to act like frantic physicists who are trying to make certain that we actually just, however briefly, obtained the 125th element. Each blockdev necessarily knows its actual capacity and its recorded size. We can use this information at any point, to take whatever action we desire. This information is not transient, it is not important that it was first obtained at this point in the setup code, it is a constant known fact about every blockdev. We should trreat it that way.

src/engine/strat_engine/setup.rs
```diff
@@ -207,6 +185,20 @@ pub fn get_blockdevs(pool_save: &PoolSave, devnodes: &[PathBuf]) -> EngineResult
         let actual_size = try!(blkdev_size(&try!(OpenOptions::new().read(true).open(dev))))
             .sectors();
 
+        match actual_size.cmp(&bda.dev_size()) {
+            Ordering::Less => {
+                error!("Actual device size for {:?} less than expected by {} bytes",
```
mulkieran: For _future_ reference, Sectors and Bytes both implement Display.

src/engine/strat_engine/setup.rs
agrover: ok, just going to drop this commit then and address later.

agrover:
rebased and updated based on comments.

src/engine/strat_engine/setup.rs
mulkieran: ```unwrap_or(None)``` probably, but it is not important.

PR #445, opened 2017-07-10
Improve *_fs methods

mulkieran:
Improvements are simplification, use of nix crate (instead of Command), and elimination.

### DIFF
--- a/src/engine/strat_engine/filesystem.rs
+++ b/src/engine/strat_engine/filesystem.rs
@@ -3,7 +3,6 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 use std::error::Error;
-use std::ffi::OsStr;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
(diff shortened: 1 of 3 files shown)


mulkieran:
@agrover @trgill review?

src/engine/strat_engine/filesystem.rs
```diff
@@ -147,39 +149,14 @@ pub fn create_fs(dev_path: &Path) -> EngineResult<()> {
...
-        return Err(EngineError::Engine(ErrorEnum::Error, message.into()));
-    }
-    Ok(())
-}
-
-pub fn unmount_fs<I, S>(mount_point: &Path, flags: I) -> EngineResult<()>
-    where I: IntoIterator<Item = S>,
-          S: AsRef<OsStr>
+/// Mount the filesystem located at devnode at mountpoint.
+pub fn mount_fs<P1, P2>(devnode: &P1, mountpoint: &P2) -> EngineResult<()>
```
agrover: Feels strange to have `mount_fs` with no `unmount_fs`. Do we really need `mount_fs` any more? Can MDV just use the nix method directly? Or if we do want `mount_fs`, then we need a trivial `unmount_fs` wrapping umount, just for the sake of consistency 😃

src/engine/strat_engine/filesystem.rs
```diff
@@ -152,7 +152,7 @@ pub fn mount_fs<P1, P2>(devnode: &P1, mountpoint: &P2) -> EngineResult<()>
 {
     Ok(try!(mount(Some(devnode),
                   mountpoint,
-                  None as Option<&Path>,
+                  Some("xfs"),
```
agrover: If this does actually fix the intermittent issue then I think the commitmsg can be a little more assertive in saying so. Does it?

src/engine/strat_engine/filesystem.rs
mulkieran: We don't need ```mount_fs```.

src/engine/strat_engine/filesystem.rs
```diff
@@ -131,22 +131,18 @@ impl Recordable<FilesystemSave> for StratFilesystem {
...
+                .arg("-f")
+                .arg(&devnode)
+                .status())
+               .success() {
+        Ok(())
     } else {
-        let message = String::from_utf8_lossy(&output.stderr);
-        debug!("stderr: {}", message);
-        return Err(EngineError::Engine(ErrorEnum::Error, message.into()));
+        let err_msg = format!("Failed to create new filesystem at {:?}", devnode);
```
mulkieran: It very well could be. But let's wait until we're actually using it to have it.

mulkieran:
```mount_fs``` has now also been removed and changes folded into first commit.

mulkieran:
We're still having these wierd ENODEV errors, but now in the source, rather than in the test. So, I've added "xfs" as filesystemtype in the source code as well.

mulkieran:
and now it passes...

mulkieran:
@agrover @trgill review?

I'ld like to stick with this. I agree that it might turn out to be useful to have the text of the error output from the command on failure, but I'ld rather hold out for a library, and expect that we won't need any further detail before that library comes into being.

trgill:
👍 we can update later if the error output is needed I suppose.  I liked being able to turn on the debug! stuff, but I'm in the minority.  :)

agrover:
okey dokey 👍

PR #446, opened 2017-07-10
Rewrite the building of bdas vec using collect() and and_then()

agrover:
This is a style change but also a behavior change: keep going if file
open() or BDA::load fails, instead of exiting with a failure. Yes, the
failed device may have contained the latest metadata, but if it can't be
read then it is akin to a missing device, and we should still proceed.

This seems in line with the later code that continues if the variable
metadata can't be read.

### DIFF
--- a/src/engine/strat_engine/setup.rs
+++ b/src/engine/strat_engine/setup.rs
@@ -98,20 +98,25 @@ pub fn find_all() -> EngineResult<HashMap<PoolUuid, Vec<PathBuf>>> {
 /// Returns None if no metadata found for this pool.
 pub fn get_metadata(pool_uuid: PoolUuid, devnodes: &[PathBuf]) -> EngineResult<Option<PoolSave>> {
 
-    // Get pairs of device nodes and matching BDAs
-    // If no BDA, or BDA UUID does not match pool UUID, skip.
-    // If there is an error reading the BDA, error. There could have been
-    // vital information on that BDA, for example, it may have contained
-    // the newest metadata.
-    let mut bdas = Vec::new();
-    for devnode in devnodes {
-        let bda = try!(BDA::load(&mut try!(OpenOptions::new().read(true).open(devnode))));
-        if let Some(bda) = bda {
-            if *bda.pool_uuid() == pool_uuid {
-                bdas.push((devnode, bda));
-            }
-        }
-    }
+    // Get pairs of device nodes and matching BDAs from readable and
+    // valid devices. Skip if no BDA, or BDA UUID does not match pool
+    // UUID.
+    let bdas = devnodes
+        .iter()
+        .filter_map(|devnode| {
+            OpenOptions::new()
+                .read(true)
+                .open(devnode)
+                .ok()
+                .and_then(|mut f| BDA::load(&mut f).ok())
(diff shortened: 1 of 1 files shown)


agrover:
I also want to mention https://github.com/stratis-storage/stratis-docs/issues/46 as we probably want to take a more systematic approach at some point to how we handle errors.

src/engine/strat_engine/setup.rs
```diff
@@ -98,20 +98,25 @@ pub fn find_all() -> EngineResult<HashMap<PoolUuid, Vec<PathBuf>>> {
...
-    let mut bdas = Vec::new();
-    for devnode in devnodes {
-        let bda = try!(BDA::load(&mut try!(OpenOptions::new().read(true).open(devnode))));
-        if let Some(bda) = bda {
-            if *bda.pool_uuid() == pool_uuid {
-                bdas.push((devnode, bda));
-            }
-        }
-    }
+    // Get pairs of device nodes and matching BDAs from readable and
```
mulkieran: I thought the semantics through very carefully when I wrote this initially, and I continue to believe that it is correct. If there is an error, then nothing is known about the BDA, and this method can not guarantee that it is returning the most recent metadata available from the list of blockdevs passed. Therefore it should return an error. If for all of the devices the BDA does not exist or it does not match the pool, then his method can confidently return None. There's a big distinction between "I don't know" and "I know there aren't any" and that's what this code was capturing.

mulkieran:
If by "later code", you mean the code that starts at 131-136, we do not error unless we fail to get metadata from any device that is supposed to contain the most recent metadata. That makes sense, since if we get metadata from at least one device then we have obtained the most recent metadata for that list of devices, which is what the method guarantees to do.

I am nothing if not systematic.

mulkieran:
This method's semantics are correct according to its specification. If something needs to be changed, the place to start is with the specification. It now has a simple specification: return the most recent metadata corresponding to this pool from this list of devices. It returns an error exactly when it cannot guarantee that it is doing that.

There are alternate specifications, like:
1. Return the most recent metadata from this list of devices. This would take just one parameter, a list of devices, and would assume that all devices are from the same pool. This is a very general approach, which is nice, but has an important and unenforceable pre-condition: that all devices must belong to the same pool. That's why I didn't like it.
2. Return the most recent metadata from a list of devices all of which must belong to the given pool. This is a more restrictive approach than mine, and I didn't like it for that reason. It forces you to return an error in more situations, which is generally a bad sign. The upside is that it might catch an error in the calling code, but I'ld prefer just to make the calling code correct.
3. etc. ...

mulkieran:
This made me realize that ```get_blockdevs()``` is actually a bit sloppier than ```get_metadata()```, so I wrote a fix to make it consistent: PR: #449.

agrover:
This is weird, it looks like in this context BDA::load will never fail. Is this a moot issue?

* `BDA::load` consists of StaticHeader::Setup and MdaRegions::load
* StaticHeader::setup will always succeed b/c we ran it in find_all->determine_ownership()
* MdaRegions::load will always succeed b/c failing to read the mdaregion just means the Option<MdaRegion> is set to None.

agrover:
Based on discussion, closing this PR

PR #449, opened 2017-07-11
Make get_blockdevs() consistent with get_metadata()

mulkieran:
It now ignores devices without a BDA or devices which do not belong to the
specified pool just like get_metadata() does.

This fixes a bug where the method could fail if it was given a devnode that
did not belong to the pool. It would fail in the last stage, when it found that
it had constructed a blockdev that didn't match the metadata.

### DIFF
--- a/src/engine/strat_engine/pool.rs
+++ b/src/engine/strat_engine/pool.rs
@@ -153,7 +153,7 @@ impl StratPool {
                                 .ok_or(EngineError::Engine(ErrorEnum::NotFound,
                                                            format!("no metadata for pool {}",
                                                                    uuid))));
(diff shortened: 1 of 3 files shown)


agrover:
For both these functions, they should be able to assume that the devnodes they're being passed all belong to the same pool, since `find_all` is organizing them so. Unless `find_all` breaks, all devices will have the same pool uuid. Instead of filtering on uuid, how about an assert that ensures this precondition is always true?

agrover:
`get_blockdevs` is currently a standalone function and can't assume its parameters are pre-filtered to a single pool so I think this technically makes sense. (but wouldn't mind if someday things were different so it didn't have to check so much)

👍

PR #450, opened 2017-07-12
Get rid of unnecessary method

mulkieran:
We can use standard functions on Option instead.


mulkieran:
clippy was complaining about errors in ```ref_ok_or```...turns out we shouldn't really have the method in the first place.

mulkieran:
@agrover @trgill review?

Issue #457, opened 2017-07-12
Pool rename should work


agrover:
It doesn't look like it is saving the new name to on-disk metadata.

mulkieran:
The code looks like it is. But there is an easy test to write...

mulkieran:
AFAICT, the solution is to write the pool metadata once the rename is accomplished, i.e., at the very end of ```StratEngine::rename_pool()```. @agrover @trgill, do you agree?

mulkieran:
See #459.

agrover:
agree.

trgill:
👍

PR #459, opened 2017-07-13
Update pool metadata when a pool is renamed

mulkieran:
Resolves: #457.


mulkieran:
This is expected to fail...if we wrote the pool metadata, we would write the correct (new) name, but we don't write the pool metadata at all.

mulkieran:
Diagnosis confirmed.

mulkieran:
@agrover @trgill review?

PR #462, opened 2017-07-13
Expose total physical size (in sectors) and number of sectors used on D-Bus

### LINKED ISSUE #458: Listing pools should also show remaining capacity
Right now we just show the name, but it would also be very useful to have a property that exposed either used/free (DataBlocks?) or a percentage of the total space used. This would need to include both space in the pool unallocated to a Flex Layer device, as well as the amount of free blocks remaining in the flex layer data device.

mulkieran:
Resolves #458.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@ test-loop:
 	sudo env "PATH=${PATH}" RUSTFLAGS='-D warnings' RUST_BACKTRACE=1 cargo test -- --test loop_test_setup
 	sudo env "PATH=${PATH}" RUSTFLAGS='-D warnings' RUST_BACKTRACE=1 cargo test -- --test loop_test_thinpool_expand
 	sudo env "PATH=${PATH}" RUSTFLAGS='-D warnings' RUST_BACKTRACE=1 cargo test -- --test loop_test_pool_rename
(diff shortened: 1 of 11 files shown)


mulkieran:
Resolves #458.

mulkieran:

mulkieran:
I want to run a test on it from other side of D-Bus first, actually.

trgill:
Ok - please ping when it is ready for review.

mulkieran:
Holding this, until we're quite sure what type the values should be when put on the D-Bus.

The reason for this is that the logical sizes could exceed the number of Sectors that can be expressed in a u64 fairly easily and we want to be consistent with how we represent Sector values on the D-Bus.

Our options are:
1. as a Str (that's the current choice)
2. as a pair of u64s, giving us a 128 bit value, which is going to take a lot longer to overflow than a u64
3. as an array of bytes
   * of arbitrary length
   * of fixed length

The disadvantage of (1) is that you have to translate back and forth between the number and the string representation of the number. But it tolerated unbounded length.

The disadvantage of (2) is that it is bounded and it requires unpacking and what not.

If you're going to go with (3), it might as well be of unbounded length. But, we will want the lowest order bytes to be at the 0th element in the array, and so forth.

mulkieran:
rebased.

mulkieran:
(1) is looking not so bad. It will be easily understood by human eyes when inspecting the output of ```GetManagedObjects()```. It is easily interpreted by the client.

@trgill @agrover review?

PR #469, opened 2017-07-18
Update extend_data() to take current_size as DataBlocks

trgill:
Since extension of the data region should be done as a multiple of
the DataBlock Size - add a comment as a reminder.

### DIFF
--- a/src/engine/strat_engine/pool.rs
+++ b/src/engine/strat_engine/pool.rs
@@ -285,15 +285,17 @@ impl StratPool {
     /// Return an extend size for the physical space backing a pool
     /// TODO: returning the current size will double the space provisoned to
     /// back the pool.  We should determine if this is a reasonable value.
-    fn extend_size(&self, current_size: Sectors) -> Sectors {
+    fn extend_size(&self, current_size: DataBlocks) -> DataBlocks {
         current_size
     }
 
     /// Expand the physical space allocated to a pool by the value from extend_size()
-    /// Return tne number of Sectors added
-    fn extend_data(&mut self, dm: &DM, current_size: Sectors) -> EngineResult<Sectors> {
+    /// Return the number of DataBlocks added
+    fn extend_data(&mut self, dm: &DM, current_size: DataBlocks) -> EngineResult<DataBlocks> {
         let extend_size = self.extend_size(current_size);
-        if let Some(new_data_regions) = self.block_devs.alloc_space(extend_size) {
+        if let Some(new_data_regions) =
+            self.block_devs
+                .alloc_space(*extend_size * DATA_BLOCK_SIZE) {
             try!(self.thin_pool.extend_data(dm, new_data_regions));
         } else {
             return Err(EngineError::Engine(ErrorEnum::Error,
@@ -345,7 +347,7 @@ impl StratPool {
 
                 if usage.used_data > usage.total_data - DATA_LOWATER {
                     // Request expansion of physical space allocated to the pool
-                    match self.extend_data(&dm, usage.total_data.0 * DATA_BLOCK_SIZE) {
+                    match self.extend_data(&dm, usage.total_data) {
                         Ok(_) => {}
                         Err(_) => {} // TODO: Take pool offline?
                     }


src/engine/strat_engine/pool.rs
```diff
@@ -289,9 +289,11 @@ impl StratPool {
...
 
     /// Expand the physical space allocated to a pool by the value from extend_size()
-    /// Return tne number of Sectors added
-    fn extend_data(&mut self, dm: &DM, current_size: Sectors) -> EngineResult<Sectors> {
-        let extend_size = self.extend_size(current_size);
+    /// Return the number of Sectors added
+    fn extend_data(&mut self, dm: &DM, current_size: DataBlocks) -> EngineResult<Sectors> {
+        // The space backing the pool should be extended by a multiple of the
+        // DATA_BLOCK_SIZE.
+        let extend_size = self.extend_size(current_size.0 * DATA_BLOCK_SIZE);
```
agrover: I think `DataBlocks` has a sectors() method that you can use here?

mulkieran: DataBlocks doesn't have a ```sectors()``` method because the number of sectors in a DataBlock is not known at the point of definition (unlike MetaBlocks).

src/engine/strat_engine/pool.rs
```diff
@@ -289,9 +289,11 @@ impl StratPool {
     }
 
     /// Expand the physical space allocated to a pool by the value from extend_size()
-    /// Return tne number of Sectors added
-    fn extend_data(&mut self, dm: &DM, current_size: Sectors) -> EngineResult<Sectors> {
-        let extend_size = self.extend_size(current_size);
+    /// Return the number of Sectors added
+    fn extend_data(&mut self, dm: &DM, current_size: DataBlocks) -> EngineResult<Sectors> {
```
mulkieran: I feel like the result type should also be changed to DataBlocks.

src/engine/strat_engine/pool.rs
agrover: oops, my mistake.

mulkieran: The function called ```extend_size()``` should take a DataBlocks argument and have a DataBlocks return type. You can transform it to Sectors when passing it to ```alloc_space()```.

src/engine/strat_engine/pool.rs
```diff
@@ -284,15 +284,19 @@ impl StratPool {
...
+    fn extend_size(&self, current_size: DataBlocks) -> DataBlocks {
         current_size
     }
 
     /// Expand the physical space allocated to a pool by the value from extend_size()
-    /// Return tne number of Sectors added
-    fn extend_data(&mut self, dm: &DM, current_size: Sectors) -> EngineResult<Sectors> {
+    /// Return the number of DataBlocks added
+    fn extend_data(&mut self, dm: &DM, current_size: DataBlocks) -> EngineResult<DataBlocks> {
+        // The space backing the pool should be extended by a multiple of the
```
mulkieran: You can drop this comment, it is implicit in the return type and explicit in the header comment.

src/engine/strat_engine/pool.rs
```diff
@@ -284,15 +284,19 @@ impl StratPool {
...
+        // DATA_BLOCK_SIZE.
         let extend_size = self.extend_size(current_size);
-        if let Some(new_data_regions) = self.block_devs.alloc_space(extend_size) {
+        if let Some(new_data_regions) =
+            self.block_devs
+                .alloc_space(extend_size.0 * DATA_BLOCK_SIZE) {
```
mulkieran: ```*extend_size``` instead of ```extend_size.0```.

trgill: what rule do you apply for using * vs .0?  I'll update the style guide with it.

mulkieran:
```.0``` if inside the implementation...```*``` if outside the implementation.

mulkieran:
There are some constraints that complicate the use of Sectors, Bytes, etc. Ideally, the single field belonging to Sectors would be private, and so the compiler would tell you to dereference, as accessing the field using ```.0``` would be a compiler error. The reason this field is public is because we like to use Sectors, Bytes, etc. as consts, and there are rules that restrict the complexity of expressions that can be evaluated during constant assignment. In particular, we can't have a ```new()``` constructor, and must instead use this primitive struct constructor, which in turn requires that field to be public. The ```lazy_static``` crate might help to get around this, but at some cost in additional complexity.

There's a move afoot to allow more complex expressions for constant assignment, but we may be forced to come up with an alternative solution before that has made any progress.

Issue #454, opened 2017-07-12
Should be able to recover from a termination


agrover:
It should be possible for the stratisd process to be killed, and then restarted. Upon doing so, it should not teardown/recreate existing DM mappings, but discover and resume managing the existing DM devices.

mulkieran:
Closed by #490.

PR #472, opened 2017-07-19
Refactor MDV & filesystems into ThinPool, and enable destroying filesystems

### LINKED ISSUE #453: Destroying a filesystem doesn't work
From looking at filesystem.rs, we call `thin_dev.teardown` instead of `destroy`. This isn' trivially fixed because `destroy` needs a reference to the ThinPoolDev, to send the "delete" message to. I think we may need to refactoring to make this possible.

agrover:
Previously, StratEngine called thindev::teardown, which removes the DM
mapping but does not free references to the thindev's blocks in the
thinpool. This required:

* Un-macro-ing destroy_filesystems
* Removing destroy() from trait Filesystem

fixes #453


agrover:
still needs integration tests, working on it.

src/engine/sim_engine/pool.rs
```diff
@@ -73,7 +73,13 @@ impl Pool for SimPool {
     fn destroy_filesystems<'a, 'b>(&'a mut self,
                                    fs_uuids: &[&'b FilesystemUuid])
                                    -> EngineResult<Vec<&'b FilesystemUuid>> {
-        destroy_filesystems!{self; fs_uuids}
+        let mut removed = Vec::new();
+        for uuid in fs_uuids.iter().map(|x| *x) {
```
mulkieran: This looks like an opportunity for ```cloned()```, maybe for ```filter_map```.

src/engine/strat_engine/pool.rs
```diff
@@ -448,10 +448,18 @@ impl Pool for StratPool {
...
                                    -> EngineResult<Vec<&'b FilesystemUuid>> {
-        for fsid in fs_uuids {
-            try!(self.mdv.rm_fs(fsid));
+        let dm = try!(DM::new());
+
+        let mut removed = Vec::new();
+        for uuid in fs_uuids {
+            if let Some(fs) = self.filesystems.remove_by_uuid(&uuid) {
+                try!(self.mdv.rm_fs(uuid));
+                try!(fs.destroy(&dm, &self.thin_pool.thin_pool));
```
mulkieran: This is very iffy. We should consider two alternatives:

1. Storing filesystems in ThinPool.
2. Storing thindevs in ThinPool and storing only thindev ids in StratFilesystem.

agrover:
Hi @mulkieran please take a look.

src/engine/strat_engine/filesystem.rs
```diff
@@ -58,14 +57,14 @@ impl StratFilesystem {
                  name: &str,
                  size: Sectors,
                  dm: &DM,
-                 thin_pool: &ThinPool)
+                 thin_pool: &ThinPoolDev)
                  -> EngineResult<StratFilesystem> {
         let device_name = format_thin_name(&pool_uuid, ThinRole::Filesystem(fs_id));
-        let thin_dev = try!(thin_pool.setup_thin_device(dm, &device_name, thindev_id, size));
+        let thin_dev = try!(ThinDev::setup(name, dm, &thin_pool, thindev_id, size));
```
mulkieran: You can lose the ```&``` on ```thin_pool```.

src/engine/strat_engine/thinpool.rs
```diff
@@ -73,46 +92,89 @@ impl ThinPool {
...
+
+        let filesystems: Vec<StratFilesystem> = {
+            /// Set up a filesystem from its metadata.
+            let get_filesystem = |fssave: &FilesystemSave| -> EngineResult<StratFilesystem> {
+                Ok(try!(StratFilesystem::setup(pool_uuid,
+                                               fssave.uuid,
+                                               fssave.thin_id,
+                                               &fssave.name,
+                                               fssave.size,
+                                               &dm,
```
mulkieran: You can drop the ```&``` on ```dm```.

src/engine/strat_engine/thinpool.rs
```diff
@@ -128,19 +190,16 @@ impl ThinPool {
...
-                             -> EngineResult<ThinDev> {
-        Ok(try!(ThinDev::setup(name, dm, &self.thin_pool, id, size)))
-    }
-
     /// Tear down the thin pool.
-    pub fn teardown(self, dm: &DM) -> DmResult<()> {
-        self.thin_pool.teardown(dm)
+    pub fn teardown(self, dm: &DM) -> EngineResult<()> {
+        for fs in self.filesystems.empty() {
+            try!(fs.teardown(&dm));
```
mulkieran: You can drop the ```&```.

src/engine/sim_engine/pool.rs
```diff
@@ -73,7 +73,13 @@ impl Pool for SimPool {
     fn destroy_filesystems<'a, 'b>(&'a mut self,
                                    fs_uuids: &[&'b FilesystemUuid])
                                    -> EngineResult<Vec<&'b FilesystemUuid>> {
-        destroy_filesystems!{self; fs_uuids}
+        let mut removed = Vec::new();
+        for uuid in fs_uuids {
+            if self.filesystems.remove_by_uuid(&uuid).is_some() {
```
mulkieran: You can drop the ```&``` on ```uuid```.

src/engine/engine.rs
```diff
@@ -74,6 +72,9 @@ pub trait Pool: HasName + HasUuid {
     /// Get the filesystem in this pool with this UUID.
     fn get_filesystem(&mut self, uuid: &FilesystemUuid) -> Option<&mut Filesystem>;
 
+    /// Does the pool have any created filesystems?
+    fn has_filesystems(&self) -> bool;
```
mulkieran: You can leave this out of the interface, because the only place it is used is in a macro, and therefore your can leave out its implementations in the Pool implementation for StratPool and SimPool. I'm trying to restrict the Pool interface to things that the D-Bus API requires, and I think that's a good principle. You may need to put implementations of this method in the implementations for StratPool and SimPool.

agrover:
@mulkieran changed based on feedback. Still seeing failing CI, not sure why.

mulkieran:
It might make sense to swap this for #454, as I'm very familiar with this code, having worked on it a good deal, but still working on understanding the relevant behavior of devicemapper for #454.

agrover:
@mulkieran review?

mulkieran:
Description of PR seems obsolete.

agrover:
Yeah true.. the first two commits do things that the big 3rd commit undoes in favor of a different approach. Let me squash and give a better description...

mulkieran:
"Move /dev/stratis creation from StratEngine::initialize to MDV::setup" looks like a good idea to me and unrelated to the rest of this PR. Suggest putting that into a separate PR, along with the change of moving the declaration of ```DEV_PATH``` into the mdv module, since it is only used there after the commit, and allowing that to clearly separately.

tests/util/pool_tests.rs
```diff
@@ -97,3 +98,33 @@ pub fn test_filesystem_rename(paths: &[&Path]) {
...
+    pool.destroy_filesystems(&[&fs_uuid]).unwrap();
+
+    // Try to setup a thindev that has been destroyed
+    let dm = DM::new().unwrap();
+    let thindev = ThinDev::setup("stratis_test_thin_dev",
+                                 &dm,
+                                 pool.thinpooldev(),
+                                 fs_id,
+                                 Sectors(128u64));
+    assert!(thindev.is_err());
```
mulkieran: Would be good to figure out how to check if the filesystem entry has been removed from the mdv.

agrover:
I've created #483 for the /dev/stratis commit. Once that's merged I'll rebase this (CI for this PR won't pass without it, so I need to leave it here until then)

tests/util/pool_tests.rs
agrover: sounds doable.

tests/util/pool_tests.rs
```diff
@@ -97,3 +98,33 @@ pub fn test_filesystem_rename(paths: &[&Path]) {
 
     engine.teardown().unwrap();
 }
+
+/// Verify that destroy_filesystems actually deallocates the space
+/// from the thinpool, by that attempting to reinstantiate it using
```
agrover: deleting "that".

src/engine/strat_engine/thinpool.rs
```diff
@@ -185,7 +254,69 @@ impl ThinPool {
...
+        try!(self.mdv.save_fs(&new_filesystem));
+        self.filesystems.insert(new_filesystem);
+
+        Ok(())
+    }
+
+    pub fn destroy_filesystem(&mut self, dm: &DM, uuid: &FilesystemUuid) -> EngineResult<()> {
+        if let Some(fs) = self.filesystems.remove_by_uuid(uuid) {
+            try!(fs.destroy(dm, &self.thin_pool));
+            try!(self.mdv.rm_fs(uuid));
```
agrover: MDV is metadata storage for filesystems. When filesystems moved into the thinpool it made sense for the MDV to move with it.

tests/util/pool_tests.rs
```diff
@@ -127,4 +130,17 @@ pub fn test_thinpool_thindev_destroy(paths: &[&Path]) -> () {
...
+    // thinpool, ::setup() will fail.
+    let paths2 = paths
+        .into_iter()
+        .map(|x| x.to_path_buf())
+        .collect::<Vec<_>>();
+    let mut pool2 = StratPool::setup(pool_uuid, &paths2).unwrap();
+
+    // This also should never happen, given the previous two parts of
+    // this test.
+    assert!(pool2.get_filesystem(&fs_uuid).is_none())
```
mulkieran: Unfortunately, are tests have to clean up after themselves, including doing a teardown of any pools that were setup. Right now this is causing the subsequent test too fail.

agrover:
Oh yes.

src/engine/strat_engine/pool.rs
```diff
@@ -325,51 +289,48 @@ impl StratPool {
...
-                    ThinPoolWorkingStatus::Good => {}
-                    ThinPoolWorkingStatus::ReadOnly => {
-                        // TODO: why is pool r/o and how do we get it
-                        // rw again?
-                    }
-                    ThinPoolWorkingStatus::OutOfSpace => {
-                        // TODO: Add more space if possible, or
-                        // prevent further usage
-                        // Should never happen -- we should be extending first!
+            ThinPoolStatus::Good => {}
```
mulkieran: This is related to the problem of ```thin_pool_status()``` behaviour. It is hard to comment, since I think that the type being matched on is wrong.

src/engine/strat_engine/filesystem.rs
```diff
@@ -58,10 +57,10 @@ impl StratFilesystem {
                  name: &str,
                  size: Sectors,
                  dm: &DM,
-                 thin_pool: &ThinPool)
+                 thin_pool: &ThinPoolDev)
                  -> EngineResult<StratFilesystem> {
         let device_name = format_thin_name(&pool_uuid, ThinRole::Filesystem(fs_id));
-        let thin_dev = try!(thin_pool.setup_thin_device(dm, &device_name, thindev_id, size));
+        let thin_dev = try!(ThinDev::setup(&device_name, dm, thin_pool, thindev_id, size));
```
mulkieran: Maybe better to move some of this method into ThinPool. Same with ```initialize()``` above.

src/engine/mod.rs
```diff
@@ -30,7 +30,7 @@ pub mod strat_engine;
 
 mod consts;
 #[allow(module_inception)]
-mod engine;
+pub mod engine;
```
mulkieran: Maybe unnecessary at this point.

tests/util/pool_tests.rs
```diff
@@ -97,3 +99,49 @@ pub fn test_filesystem_rename(paths: &[&Path]) {
...
+    let dm = DM::new().unwrap();
+    let thindev = ThinDev::setup("stratis_test_thin_dev",
+                                 &dm,
+                                 pool.thinpooldev(),
+                                 fs_id,
+                                 Sectors(128u64));
+    assert!(thindev.is_err());
+    pool.teardown().unwrap();
+
+    // Check that destroyed fs is not present in MDV. If the record
```
mulkieran: I've realized that relying on this behavior of ```setup()``` might make for too unstable a test. Perhaps it would be better just to take a look at the filesystem metadata on the mdv before tearing down the pool.

tests/util/pool_tests.rs
```diff
@@ -31,7 +33,7 @@ pub fn test_thinpool_expand(paths: &[&Path]) -> () {
                                               Redundancy::NONE,
                                               true)
             .unwrap();
-    let &(_, fs_uuid) = pool.create_filesystems(&vec!["stratis_test_filesystem"])
+    let &(_, fs_uuid) = pool.create_filesystems(&["stratis_test_filesystem"])
```
agrover: yeah I'm going to drop this and address separately.

tests/util/pool_tests.rs
```diff
@@ -97,3 +99,49 @@ pub fn test_filesystem_rename(paths: &[&Path]) {
...
+                                 pool.thinpooldev(),
+                                 fs_id,
+                                 Sectors(128u64));
+    assert!(thindev.is_err());
+    pool.teardown().unwrap();
+
+    // Check that destroyed fs is not present in MDV. If the record
+    // had been left on the MDV that didn't match a thin_id in the
+    // thinpool, ::setup() will fail.
+    let paths2 = paths
```
agrover: We need `to_path_buf` here because `setup` takes a `&[PathBuf]`, which is not ideal. (A different PR will fix that, and also remove `paths2` entirely.

src/engine/structures.rs
```diff
@@ -38,6 +38,15 @@ impl<'a, T: HasName + HasUuid> IntoIterator for &'a mut Table<T> {
     }
 }
 
+impl<'a, T: HasName + HasUuid> IntoIterator for &'a Table<T> {
```
agrover: The existing one is for `&mut`, this one is for just `&`, which we need to iterate when we don't have a mutable reference.

src/engine/strat_engine/thinpool.rs
```diff
@@ -31,6 +39,14 @@ pub struct ThinPool {
     thin_pool: ThinPoolDev,
     meta_spare: Vec<Segment>,
     id_gen: ThinDevIdPool,
+    filesystems: Table<StratFilesystem>,
+    mdv: MetadataVol,
+}
+
+pub enum ThinPoolStatus {
```
agrover: I don't understand, can you clarify?

src/engine/strat_engine/thinpool.rs
```diff
@@ -73,46 +92,89 @@ impl ThinPool {
...
-                                                          data_block_size,
-                                                          low_water_mark,
-                                                          new_meta_dev,
-                                                          data_dev)),
-                       meta_spare: new_spare_segments,
-                       id_gen: ThinDevIdPool::new_from_ids(thin_ids),
-                   })
+            Err(e) => Err(e.into()),
+        };
+        let (thinpool_dev, spare_segments) = try!(res);
```
agrover: doing so causes rustfmt to generate an extremely unpleasant result.

src/engine/strat_engine/thinpool.rs
```diff
@@ -73,46 +92,89 @@ impl ThinPool {
...
+
+        let filesystems: Vec<StratFilesystem> = {
```
agrover: ok

src/engine/strat_engine/thinpool.rs
```diff
@@ -73,46 +92,89 @@ impl ThinPool {
...
+           })
     }
 
     /// The status of the thin pool as calculated by DM.
     pub fn thin_pool_status(&self, dm: &DM) -> EngineResult<ThinPoolStatus> {
-        Ok(try!(self.thin_pool.status(dm)))
+        try!(self.thin_pool.status(dm));
+        try!(self.mdv.check());
+
+        for fs in &self.filesystems {
```
agrover: So it's called by Pool::check, it calls mdv and filesystem.check(), this is basically a check() function, which is a big deferred todo that won't be fixed by this PR. We could rename it `check()` to emphasize its placeholder nature?

src/engine/strat_engine/thinpool.rs
```diff
@@ -128,19 +190,16 @@ impl ThinPool {
...
-                             id: ThinDevId,
-                             size: Sectors)
-                             -> EngineResult<ThinDev> {
-        Ok(try!(ThinDev::setup(name, dm, &self.thin_pool, id, size)))
-    }
-
     /// Tear down the thin pool.
-    pub fn teardown(self, dm: &DM) -> DmResult<()> {
-        self.thin_pool.teardown(dm)
+    pub fn teardown(self, dm: &DM) -> EngineResult<()> {
```
agrover: ok

src/engine/strat_engine/thinpool.rs
```diff
@@ -185,7 +254,69 @@ impl ThinPool {
             .map(|s| s.length)
             .sum();
 
-        Ok(data_dev_used + spare_total + meta_dev_total)
+        let mdv_total = self.mdv.segments().iter().map(|s| s.length).sum();
+
+        Ok(data_dev_used + spare_total + meta_dev_total + mdv_total)
+    }
+
+    pub fn get_filesystem_by_uuid(&mut self,
```
agrover: ok

src/engine/strat_engine/thinpool.rs
```diff
@@ -185,7 +254,69 @@ impl ThinPool {
...
+
+    pub fn get_filesystem_by_name(&mut self, name: &str) -> Option<&mut StratFilesystem> {
+        self.filesystems.get_mut_by_name(name)
+    }
+
+    pub fn has_filesystems(&self) -> bool {
+        !self.filesystems.is_empty()
+    }
+
+    pub fn create_filesystem(&mut self,
```
agrover: ok

src/engine/strat_engine/thinpool.rs
```diff
@@ -185,7 +254,69 @@ impl ThinPool {
...
+                             pool_uuid: &Uuid,
+                             fs_uuid: FilesystemUuid,
+                             name: &str,
+                             dm: &DM,
+                             size: Option<Sectors>)
+                             -> EngineResult<()> {
+        let new_filesystem =
+            try!(StratFilesystem::initialize(pool_uuid, fs_uuid, name, dm, self, size));
```
agrover: deferring this to another PR, this results in `StratFilesystem::setup` and `initialize` having much less functionality.

src/engine/strat_engine/pool.rs
```diff
@@ -423,61 +395,50 @@ impl Pool for StratPool {
     /// Precondition: All filesystems belonging to this pool must be
     /// unmounted.
     fn destroy(self) -> EngineResult<()> {
-        teardown_pool!(self);
+        let dm = try!(DM::new());
```
agrover: I don't understand, this method is in the Pool trait impl.

src/engine/strat_engine/pool.rs
```diff
@@ -423,61 +395,50 @@ impl Pool for StratPool {
...
         try!(self.block_devs.destroy_all());
         Ok(())
     }
 
     fn destroy_filesystems<'a, 'b>(&'a mut self,
                                    fs_uuids: &[&'b FilesystemUuid])
                                    -> EngineResult<Vec<&'b FilesystemUuid>> {
-        for fsid in fs_uuids {
-            try!(self.mdv.rm_fs(fsid));
+        let dm = try!(DM::new());
```
agrover: ok.

src/engine/strat_engine/pool.rs
```diff
@@ -423,61 +395,50 @@ impl Pool for StratPool {
...
     }
 
     fn total_physical_used(&self) -> EngineResult<Sectors> {
-        self.thin_pool
-            .total_physical_used()
-            .and_then(|v| {
-                          Ok(v + self.block_devs.metadata_size() +
-                             self.mdv.segments().iter().map(|s| s.length).sum())
-                      })
+        self.thin_pool.total_physical_used()
```
agrover: oops! ok.

src/engine/strat_engine/thinpool.rs
mulkieran: ahh.

src/engine/strat_engine/filesystem.rs
agrover: deferred to a later PR.

src/engine/mod.rs
agrover: without this, the added integration test doesn't compile (it needs to import HasUuid to call pool.uuid().

tests/util/pool_tests.rs
agrover: Need to find a way to do this, since this involves peering inside the implementation.

src/engine/structures.rs
mulkieran: Still better if it were right next to the other one.

tests/util/pool_tests.rs
mulkieran: You think it should take something like ```&[&Path]```.

src/engine/structures.rs
agrover: Yes it's right under the other one.

tests/util/pool_tests.rs
agrover: yeah, either a `&[&Path]`, or is a `&[AsRef<Path>]` a thing?. In any case I think this PR can just leave it for later.

src/engine/strat_engine/pool.rs
mulkieran: Yes. My plan is to make the header comments in the definition of the Pool trait really useful. If you look at the header comment for this method in the Pool trait definition, you will see that it has gotten bad.

src/engine/strat_engine/thinpool.rs
mulkieran: The last two weeks the ```check()``` methods only consumed an hour of my time, which was hardly worth mentioning. But this week it looks like it is going to get much worse.

src/engine/structures.rs
mulkieran: Oh, good.

src/engine/strat_engine/pool.rs
agrover: ah ok. I'll move it there.

src/engine/strat_engine/thinpool.rs
mulkieran: This is supposed to be an overall status for the Stratis ThinPool struct. It feels like it ought to be a product type rather than a sum type, like:

```
struct ThinPoolStatus {
  thinpool_status: dm::ThinPoolStatus,
  ...other statuses
}
```

src/engine/strat_engine/thinpool.rs
mulkieran: The only value this method can return is ```ThinPoolStatus::Good``` or an error of some kind. That makes the entire match starting at line 328 useless, AFAICT.

agrover:
@mulkieran final review?

src/engine/strat_engine/pool.rs
```diff
@@ -420,51 +390,43 @@ impl Pool for StratPool {
...
+        self.thin_pool.rename_filesystem(uuid, new_name)
     }
 
     fn rename(&mut self, name: &str) {
         self.name = name.to_owned();
     }
 
     fn get_filesystem(&mut self, uuid: &FilesystemUuid) -> Option<&mut Filesystem> {
-        get_filesystem!(self; uuid)
+        self.get_strat_filesystem(uuid)
```
mulkieran: Might as well stick w/ macro, that was what happened with ```get_pool```.

src/engine/strat_engine/thinpool.rs
```diff
@@ -73,74 +91,100 @@ impl ThinPool {
...
+        Ok(ThinPool {
+               thin_pool: thinpool_dev,
+               meta_spare: spare_segments,
+               id_gen: ThinDevIdPool::new_from_ids(thin_ids),
+               filesystems: fs_table,
+               mdv: mdv,
+           })
     }
 
     /// The status of the thin pool as calculated by DM.
```
mulkieran: The header comment needs updating.

mulkieran: And you might as well change the name to ```check()```.

src/engine/strat_engine/thinpool.rs
```diff
@@ -73,74 +91,100 @@ impl ThinPool {
...
-    pub fn teardown(self, dm: &DM) -> DmResult<()> {
-        self.thin_pool.teardown(dm)
+    /// Tear down the components managed here: filesystems, the MDV,
+    /// and the actual thinpool device itself.
+    pub fn teardown(self, dm: &DM) -> EngineResult<()> {
+        for fs in self.filesystems.empty() {
+            try!(fs.teardown(dm));
+        }
+
+        try!(self.thin_pool.teardown(dm));
```
mulkieran: I think these two should be swapped on a sort of FIFO principle.

src/engine/strat_engine/thinpool.rs
```diff
@@ -185,7 +239,78 @@ impl ThinPool {
...
+
+    pub fn get_mut_filesystem_by_name(&mut self, name: &str) -> Option<&mut StratFilesystem> {
+        self.filesystems.get_mut_by_name(name)
+    }
+
+    pub fn has_filesystems(&self) -> bool {
+        !self.filesystems.is_empty()
+    }
+
+    /// Create a filesystem within the thin pool.
```
mulkieran: You should either have this check that the name is not already in the pool or explicitly state the precondition that this must be a new name.

tests/util/pool_tests.rs
```diff
@@ -97,3 +99,49 @@ pub fn test_filesystem_rename(paths: &[&Path]) {
...
+        .into_iter()
+        .map(|x| x.to_path_buf())
+        .collect::<Vec<_>>();
```
mulkieran: If possible,

```let paths2: Vec<_> = paths.into_iter().map(|x| x.to_path_buf()).collect();```

File: tests/util/pool_tests.rs (hunk 15 above)
Comment (mulkieran) [inline tests/util/pool_tests.rs]: You could look for the file on the mdv filesystem, maybe.

File: tests/util/pool_tests.rs (hunk 34 above)
Comment (agrover) [inline tests/util/pool_tests.rs]: ok

File: src/engine/strat_engine/pool.rs (hunk 30 above)
Comment (agrover) [inline src/engine/strat_engine/pool.rs]: I don't follow, we can't use the macro any more because it assumes self has a `filesystems` member.

File: src/engine/strat_engine/thinpool.rs (hunk 32 above)
Comment (agrover) [inline src/engine/strat_engine/thinpool.rs]: I think this is best addressed by explicit comments in this method on which teardowns depend on other things succeeding, and which don't, so that's the approach I've taken.

File: src/engine/strat_engine/thinpool.rs (hunk 31 above)
Comment (agrover) [inline src/engine/strat_engine/thinpool.rs]: ok

File: src/engine/strat_engine/thinpool.rs [hunk 35]
```diff
@@ -41,7 +56,8 @@ impl ThinPool {
                low_water_mark: DataBlocks,
                spare_segments: Vec<Segment>,
                meta_dev: LinearDev,
-               data_dev: LinearDev)
+               data_dev: LinearDev,
```
agrover: ok

src/engine/strat_engine/thinpool.rs
```diff
@@ -31,6 +39,13 @@ pub struct ThinPool {
     thin_pool: ThinPoolDev,
     meta_spare: Vec<Segment>,
     id_gen: ThinDevIdPool,
+    filesystems: Table<StratFilesystem>,
+    mdv: MetadataVol,
+}
+
+pub struct ThinPoolStatus {
```
agrover: header and member docstrings added

src/engine/strat_engine/pool.rs
mulkieran: Right. Must as well get rid of macro entirely, in that case. Since SimPool and StratPool have diverged it now serves no purpose.

src/engine/strat_engine/filesystem.rs
```diff
@@ -91,6 +68,11 @@ impl StratFilesystem {
     pub fn teardown(self, dm: &DM) -> EngineResult<()> {
         Ok(try!(self.thin_dev.teardown(dm)))
     }
+
+    /// Destroy the filesystem.
```
agrover: it'll fail if it's still mounted. My take on listing preconditions was that it was for things that needed to be true for correct operation, but that weren't checked (maybe b/c checked at a higher level), and didn't return an Err, in the function itself. In this case, trying to destroy the thindev under a mounted filesystem will fail so I think that means we don't need to list it explicitly as a precondition. Thoughts?

src/engine/strat_engine/filesystem.rs
```diff
@@ -31,43 +27,24 @@ pub enum FilesystemStatus {
...
-            thin_dev: thin_dev,
-        };
+        let fs = StratFilesystem::setup(fs_id, name, thin_dev);
 
         try!(create_fs(try!(fs.devnode()).as_path()));
         Ok(fs)
     }
 
     /// Setup a filesystem, setting up the thin device as necessary.
     // FIXME: Check for still existing device mapper devices.
```
mulkieran: Move this FIXME to the ThinPool method where it is relevant now, as ```setup()```.

src/engine/strat_engine/filesystem.rs
```diff
@@ -31,43 +27,24 @@ pub enum FilesystemStatus {
...
-            name: name.to_owned(),
-            thin_dev: thin_dev,
-        };
+        let fs = StratFilesystem::setup(fs_id, name, thin_dev);
 
         try!(create_fs(try!(fs.devnode()).as_path()));
         Ok(fs)
     }
 
     /// Setup a filesystem, setting up the thin device as necessary.
```
mulkieran: Header comment is obsolete.

PR #473, opened 2017-07-19
Tear down thindevs as part of StratPool::teardown

### LINKED ISSUE #471: Need to properly sync filesysystems in StratPool::teardown()
Whatever is necessary to make sure that the user data is where it needs to be before thin pools and metadata vols go away must be done here.

agrover:
This should let tearing down the thinpool itself succeed.

fixes #471

### DIFF
--- a/src/engine/strat_engine/filesystem.rs
+++ b/src/engine/strat_engine/filesystem.rs
@@ -86,6 +86,11 @@ impl StratFilesystem {
     pub fn thin_id(&self) -> ThinDevId {
         self.thin_dev.id()
     }
+
+    /// Tear down the filesystem.
+    pub fn teardown(self, dm: &DM) -> EngineResult<()> {
+        Ok(try!(self.thin_dev.teardown(dm)))
+    }
 }
 
 impl HasName for StratFilesystem {
--- a/src/engine/strat_engine/pool.rs
+++ b/src/engine/strat_engine/pool.rs
@@ -376,13 +376,15 @@ impl StratPool {
     /// Take down the device mapper devices belonging to the pool.
     /// This method and destroy() must keep their DM teardown operations
     /// in sync.
+    /// Precondition: All filesystems belonging to this pool must be
+    /// unmounted.
     pub fn teardown(self) -> EngineResult<()> {
-        // TODO: any necessary clean up of filesystems
-        if !self.filesystems.is_empty() {
-            return Err(EngineError::Engine(ErrorEnum::Busy,
-                                           format!("May be unsynced files on device.")));
-        }
         let dm = try!(DM::new());
+
+        for fs in self.filesystems.empty() {
+            try!(fs.teardown(&dm));
+        }
+
         try!(self.thin_pool.teardown(&dm));
         try!(self.mdv.teardown(&dm));
         Ok(())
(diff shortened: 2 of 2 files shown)


mulkieran:
What's against putting a ```teardown()``` method in ```StratFilesystem```?

agrover:
sure let's try that, one sec

agrover:
updated.

mulkieran:
Why will tearing down a thindev fail if the filesystem that is on it is still mounted?

src/engine/strat_engine/pool.rs
```diff
@@ -377,12 +377,13 @@ impl StratPool {
...
     /// in sync.
     pub fn teardown(self) -> EngineResult<()> {
-        // TODO: any necessary clean up of filesystems
-        if !self.filesystems.is_empty() {
-            return Err(EngineError::Engine(ErrorEnum::Busy,
-                                           format!("May be unsynced files on device.")));
-        }
         let dm = try!(DM::new());
+
+        // Note: Will fail if filesystems are still mounted
```
mulkieran: Is this note really correct? Or will the teardown succeed but make some sort of mess?

src/engine/strat_engine/filesystem.rs
```diff
@@ -86,6 +86,11 @@ impl StratFilesystem {
     pub fn thin_id(&self) -> ThinDevId {
         self.thin_dev.id()
     }
+
+    /// Tear down the filesystem.
```
mulkieran: Is there more that this method should be doing, i.e., is there a TODO that belongs here?

agrover:
The filesystem opens the block device and takes a reference to it in the kernel until it's unmounted. This prevents the block device from being destroyed until the filesystem is unmounted.

src/engine/strat_engine/pool.rs
agrover: if you have filesystems A-C and B is still mounted, then it'll cause an error and teardown will return, leaving things in a partially torn-down state, yes.

src/engine/strat_engine/filesystem.rs
agrover: I dunno, like what?

mulkieran: I don't know much about this area. You are saying that you are pretty sure there is nothing?

src/engine/strat_engine/pool.rs
mulkieran: Is there anyway we can detect if any are mounted before tearing down the filesystems?

agrover: we can look at /proc/mounts ...

But, what's the use case here? Can we say teardown is only going to be called when the system is shutting down? (Teardown is not exposed via API.)

src/engine/strat_engine/filesystem.rs
agrover: I can't think of anything, or else I would've added it or added a TODO 😃

src/engine/strat_engine/pool.rs
mulkieran: We can not say that.

src/engine/strat_engine/filesystem.rs
mulkieran: Ok!

src/engine/strat_engine/pool.rs
mulkieran: Ok, please get rid of the note and put something like this in the function header comment:

```Precondition: All filesystems belonging to this pool must be unmounted.```

**Comment by agrover** [comment]
Made changes based on feedback.

File: src/engine/strat_engine/pool.rs [hunk 3]
```diff
@@ -376,13 +376,15 @@ impl StratPool {
...
+    /// Precondition: All filesystems belonging to this pool must be
+    /// unmounted.
     pub fn teardown(self) -> EngineResult<()> {
-        // TODO: any necessary clean up of filesystems
-        if !self.filesystems.is_empty() {
-            return Err(EngineError::Engine(ErrorEnum::Busy,
-                                           format!("May be unsynced files on device.")));
-        }
         let dm = try!(DM::new());
+
```
mulkieran: You'll need to add the same action to ```destroy()```, I think, if so, same precondition.

agrover: ok will do

agrover:
updated.

PR #475, opened 2017-07-20
No longer specify trusty distribution

mulkieran:
It is now the default.


mulkieran:
@agrover @trgill review?

PR #476, opened 2017-07-20
Some metadata improvements

mulkieran:
This PR does the following:

* Adds some more tests.
* Fixes a bug in ```MDARegions::initialize()```.
* Improves StaticHeader implementation by abstracting some methods, simplifying some others, and refining ```sigblock_from_buf()``` so that it takes into account the possibility of a partial read.
* Improves ```MDARegions::load()``` by having it return an error if there is truly an error in reading an MDA region.
* Gets rid of one TODO.
* Gets rid of two clippy errors.

### DIFF
--- a/src/engine/strat_engine/metadata.rs
+++ b/src/engine/strat_engine/metadata.rs
@@ -177,17 +177,24 @@ impl StaticHeader {
         }
     }
 
+
+    /// Read data from f into buffer of length _BDA_STATIC_HDR_SIZE.
+    fn read_into_buf<F>(f: &mut F) -> EngineResult<[u8; _BDA_STATIC_HDR_SIZE]>
+        where F: Read + Seek
+    {
+        f.seek(SeekFrom::Start(0))?;
+        let mut buf = [0u8; _BDA_STATIC_HDR_SIZE];
+        f.read(&mut buf)?;
+        Ok(buf)
+    }
+
     /// Try to find a valid StaticHeader on a device.
     /// If there is no StaticHeader on the device, return None.
     /// If there is a problem reading a header, return an error.
     fn setup<F>(f: &mut F) -> EngineResult<Option<StaticHeader>>
         where F: Read + Seek
     {
-        f.seek(SeekFrom::Start(0))?;
-        let mut buf = [0u8; _BDA_STATIC_HDR_SIZE];
-        f.read(&mut buf)?;
-
-        StaticHeader::setup_from_buf(&buf)
+        StaticHeader::setup_from_buf(&StaticHeader::read_into_buf(f)?)
     }
 
     /// Try to find a valid StaticHeader in a buffer.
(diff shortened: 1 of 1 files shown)


mulkieran:
rebased.

mulkieran:
@agrover review?

mulkieran:
It might actually be better to have an ```MDAHeader::initialize()``` function which initializes the MDAHeader with all 0s except for the correct crc. I think that would be slightly more correct.

mulkieran:
Ok, I think this is ready to go now. @agrover review?

src/engine/strat_engine/metadata.rs
```diff
@@ -184,13 +184,14 @@ impl StaticHeader {
...
-    fn read_into_buf<F>(f: &mut F) -> EngineResult<[u8; _BDA_STATIC_HDR_SIZE]>
+    /// Return a newly constructed buffer and the number of bytes read.
+    fn read_into_buf<F>(f: &mut F) -> EngineResult<([u8; _BDA_STATIC_HDR_SIZE], usize)>
         where F: Read + Seek
     {
         try!(f.seek(SeekFrom::Start(0)));
         let mut buf = [0u8; _BDA_STATIC_HDR_SIZE];
-        try!(f.read(&mut buf));
-        Ok(buf)
+        let bytes = try!(f.read(&mut buf));
```
agrover: I think `read_exact` may work well here? https://doc.rust-lang.org/std/io/trait.Read.html#method.read_exact

mulkieran: That's what clippy suggested, so I tried it, then, when the existing tests failed I realized what was wrong. (I guess it is an archaism that the branch is called ```master-read-exact``` still). 

What is wrong is that ```setup()``` is called in a context where it is not certain whether there is a static header or not. One perfectly good reason for there not being a static header is if we have some peculiar block device which we can't read a lot of data from and which does not belong to Stratis. The result of using ```read_exact()``` is to cause engine initialization to promptly fail, presumably in ```find_all``` when it tries to determine ownership on some slightly surprising device.

agrover: Match on read_into_buf returning Err in determine_ownership and return DevOwnership::Theirs instead?

If you do stick with this PR's current approach, it really is a change from "read all expected bytes into a buffer or fail" semantics to "read as much as I can and return that", yes? If so, it's likely that functions shouldn't be taking or returning fixed size arrays any longer, but Vec<u8>s and slices, which track their own length. Shouldn't need `bytes`, instead just call `.len()` on the vec or slice.

mulkieran: It never had that semantics in the first place. It read all the bytes it found, left the rest of the buffer filled in with zeros, and then fired away its various discovery mechanisms.

mulkieran: It could be changed to use ```Read```'s ```take()``` function which returns a ```Take``` adaptor which can then be manipulated, although the adaptor ultimately reads into a buffer. That would be a bit more rustic.

Or ```read_into_buf()``` could be changed so that it returned a vector which is the length of bytes read, as you suggest.

mulkieran:
I think that it is probably time to formalize the metadata reading behavior with a state machine, as the scenarios that are being considered are fairly sophisticated.

So, this depends on: https://github.com/stratis-storage/stratis-docs/issues/44 and I'm moving it back to "in progress".

agrover:
close this? obsoleted by #482?

mulkieran:
#482 was just the easy parts.

mulkieran:
rebased.

agrover:
closing due to inactivity.

mulkieran:
It is active again.

PR #477, opened 2017-07-20
Expand XFS when filesystem space goes below low water mark

trgill:
The core of the PR is focused on expanding the logical space for XFS when the low water mark is crossed.  The expansion is requested via xfs_growfs.

To facilitate testing the changes a size parameter was added to Pool:: create_filesystems().

PR #479, opened 2017-07-21
Add size parameter to Pool::create_filesystems

agrover:
This can be very handy for testing purposes. If it's not given than the
pool implementation will pick a default size.

### DIFF
--- a/src/dbus_api/pool.rs
+++ b/src/dbus_api/pool.rs
@@ -56,7 +56,7 @@ fn create_filesystems(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
     let mut engine = dbus_context.engine.borrow_mut();
     let pool = get_pool!(engine; pool_uuid; default_return; return_message);
 
-    let result = pool.create_filesystems(&filesystems.collect::<Vec<&str>>());
+    let result = pool.create_filesystems(&filesystems.collect::<Vec<&str>>(), None);
 
     let msg = match result {
         Ok(ref infos) => {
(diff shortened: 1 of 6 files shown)


mulkieran:
This means that all the filesystems created at one time will have to be the same size. Originally, each spec included a size, so that each filesystem could have its own size.

agrover:
Do we need that level of flexibility at this point? This is so we can develop/test fs expand using less than terabyte virtual device sizes.

trgill:
@agrover do you see a problem with the way I'm testing fs expand in PR #477 ?

agrover:
@trgill I don't know, I haven't looked at that PR yet. Testing fs expand was the impetus for *this* PR of course, so you still want what this PR aims to do?

trgill:
@agrover I think having this PR work would be great (and a better test), but we can work around it by calling the filesystem check directly.

mulkieran:
The level of flexibility is no harder to implement and more uniform in conception. Is the problem that we actually really don't want to set the size ever in our real code, but only ever for testing?

agrover:
@mulkieran yep, pretty much.

mulkieran:
I'ld like to see an instance of a test where it was supposed to be used...

agrover:
@trgill ok.. well we have this change here if you want it. You could either 1) add (cherry-pick) this commit into your PR #477 and use it in the tests and then close this PR, or 2) stick with using check() in #477, in which case we can also close this PR. Up to you 😃

agrover:
I don't know if this commit might end up being merged someday, but it won't be via this PR. Closing.

PR #480, opened 2017-07-21
Use BACKTRACE=full

agrover:


agrover:
see https://users.rust-lang.org/t/improvement-of-rust-backtrace-merged-on-latest-nightlies/9735. Actually I think we want to keep using 1, so closing.

PR #483, opened 2017-07-24
Move /dev/stratis creation from StratEngine::initialize to MDV::setup

agrover:
MDV setup will fail unless /dev/stratis exists. It should be possible to
setup a pool without setting up the engine. Have each pool ensure
/dev/stratis exists as part of MDV setup, rather than relying on
StratEngine::initialize().


agrover:
updated PR to move `DEV_PATH` definition into mdv.rs.

PR #490, opened 2017-07-27
Make setup when devicemapper devices exist succeed

mulkieran:
If the metadata filesystem is already mounted do not remount it.

### DIFF
--- a/tests/util/setup_tests.rs
+++ b/tests/util/setup_tests.rs
@@ -7,16 +7,26 @@
 
 extern crate devicemapper;
 extern crate libstratis;
+extern crate nix;
+extern crate tempdir;
 extern crate uuid;
 
+use std::fs::OpenOptions;
+use std::io::Write;
 use std::path::Path;
 
+use self::nix::mount::{MsFlags, MNT_DETACH, mount, umount2};
+use self::tempdir::TempDir;
 use self::uuid::Uuid;
 
-use libstratis::engine::Engine;
-use libstratis::engine::types::RenameAction;
+use self::devicemapper::DM;
+
+use libstratis::engine::{Engine, Pool};
+use libstratis::engine::engine::HasUuid;
+use libstratis::engine::types::{Redundancy, RenameAction};
 use libstratis::engine::strat_engine::blockdevmgr::{initialize, resolve_devices};
 use libstratis::engine::strat_engine::metadata::MIN_MDA_SECTORS;
+use libstratis::engine::strat_engine::pool::StratPool;
 use libstratis::engine::strat_engine::serde_structs::Recordable;
 use libstratis::engine::strat_engine::setup::{find_all, get_blockdevs, get_metadata};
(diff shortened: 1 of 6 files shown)


mulkieran:
If the test does not fail, we have some investigating to do.

mulkieran:
Yes, the test fails.

mulkieran:
With old devicemapper:
https://travis-ci.org/stratis-storage/stratisd/jobs/259396315
thread 'loop_test_pool_setup' panicked at 'called `Result::unwrap()` on an `Err` value: DM(Io(Error { repr: Os { code: 16, message: "Device or resource busy" } }))', /checkout/src/libcore/result.rs:860

With new devicemapper:
https://travis-ci.org/stratis-storage/stratisd/jobs/259375804
thread 'loop_test_pool_setup' panicked at 'called `Result::unwrap()` on an `Err` value: Nix(Sys(EBUSY))', /checkout/src/libcore/result.rs:860

mulkieran:
Things are busy, apparently. It is rational to conclude that the new error arises because the device that is already there is busy for some reason, simply from a look at the changes in behaviour introduced in devicemapper.

mulkieran:
Sleeping for half a second makes no difference.

mulkieran:
The good news is that we're failing while setting up the metadata volume which means that likely the devicemapper changes are correct.

mulkieran:
We're failing precisely when we try to mount the mdv filesystem.

mulkieran:
Depend on new devicemapper PR.

mulkieran:
@agrover This is obviously preliminary, because we would expect that the new version of stratisd would pull from the new devicemapper release. There are also some oddities in the test framework changes for faster feedback which I wouldn't commit.

Feel free to take a look, and, in particular, if you have any suggestions for better testing that would be helpful.

agrover:
Looks good to me (minus the just-for-development stuff). The one thought on testing I had was testing a *partially* existing stack of dm devices, but thinking about it more, it would be hard to construct and wouldn't tell us much more than the full setup-pre-existing case.

mulkieran:
Now using devicemapper master branch.

mulkieran:
Now using stratisd devicemapper master branch.

mulkieran:
This is ready to go, except that the testing commit must be removed before final merge.

@agrover @trgill review?

tests/util/setup_tests.rs
```diff
@@ -176,3 +186,45 @@ pub fn test_pool_rename(paths: &[&Path]) {
     assert_eq!(pool_name, name2);
     engine.teardown().unwrap();
 }
+
+/// Verify that setting up a pool when the pool has not been previously torn
+/// down does not fail. Clutter the orignally set up pool with a filesystem
```
agrover: Looks very good to me, especially this test. 👍 from me once temp stuff is removed, and fix a typo in the above comment line.

mulkieran:
Final rebase. Should be ready to go.

PR #504, opened 2017-08-02
Use expects() more appropriately

mulkieran:
Use ```expects()``` to convey information and if there is no information to convey, turn them into ```unwraps()```.

https://github.com/rust-lang-nursery/rust-clippy/issues/1921#issuecomment-319693657


mulkieran:
@tasleson @trgill review?

agrover:
seems good 👍

Issue #478, opened 2017-07-21
time crate is being obsoleted


mulkieran:
We will need to switch to the chrono crate (https://docs.rs/chrono/0.4.0/chrono/index.html) at some point. This will be a bit tricky, since the chrono crate is a lot more ornate.

Kintaro:
Sounds like something easy to start contributing. I'll give it a shot.

agrover:
this is fixed. Thanks!

PR #506, opened 2017-08-03
Remove the time crate and replace with chrono crate

Kintaro:
Resolve issue https://github.com/stratis-storage/stratisd/issues/478


agrover:
Hi! Thanks for the PR, I think you're our first non-RH contributor! 💯 

Yeah, we had problems with versions of rustfmt after 0.8.3 so that's what we're still pinned to, looks like you've figured this out 😃

src/engine/strat_engine/metadata.rs
```diff
@@ -504,29 +504,29 @@ mod mda {
...
+        last_updated: DateTime<Local>,
 
         /// Size of region used for pool metadata.
         used: Bytes,
 
         data_crc: u32,
     }
 
-    // Implementing Default explicitly because Timespec does not implement
+    // Implementing Default explicitly because DateTime<Local> does not implement
```
mulkieran: You can drop the second sentence in this comment, because this PR will have made it obsolete! Do you think there might be any point in submitting a PR against the chrono crate to implement Default for ```DateTime<Local>```.

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -133,9 +133,12 @@ impl BlockDevMgr {
...
     /// write to.
     pub fn save_state(&mut self, metadata: &[u8]) -> EngineResult<()> {
-        let current_time = now().to_timespec();
+        let current_time = Local::now();
         let stamp_time = if Some(current_time) <= self.last_update_time {
-            self.last_update_time.expect("> Some(time)") + Duration::nanoseconds(1)
+            self.last_update_time
+                .expect("> Some(time)")
+                .checked_add_signed(Duration::nanoseconds(1))
+                .expect("> checked_add_signed")
```
mulkieran: ```.expect("self.last_update_time <<  maximum representable DateTime")```

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -212,8 +215,10 @@ pub fn initialize(pool_uuid: &PoolUuid,
     /// and an open File handle, all of which are needed later.
     pub fn dev_info(dev: &Device) -> EngineResult<(PathBuf, Bytes, DevOwnership, File)> {
         let devnode = try!(dev.devnode().ok_or_else(|| {
-            EngineError::Engine(ErrorEnum::NotFound,
-                                format!("could not get device node from dev {}", dev.dstr()))
+            EngineError::Engine(
```
mulkieran: I think this is just a reformat.

src/engine/strat_engine/metadata.rs
```diff
@@ -560,7 +560,7 @@ mod mda {
 
                     Ok(Some(MDAHeader {
                                 used: used,
-                                last_updated: Timespec::new(secs as i64, nsecs as i32),
+                                last_updated: Local.timestamp(secs as i64, nsecs),
```
mulkieran: Because of this change, you can also omit assert about nsecs above. Maybe then you can get rid of ```nsecs``` name entirely and just use the expression here.

src/engine/strat_engine/metadata.rs
```diff
@@ -697,7 +697,7 @@ mod mda {
                 let region_size = (MIN_MDA_SECTORS / 4usize).bytes() +
                                   Bytes(region_size_ext as u64);
                 let header = MDAHeader {
-                    last_updated: Timespec::new(sec, nsec),
+                    last_updated: Local.timestamp(sec, nsec as u32),
```
mulkieran: Better to change ```nsec``` type in ```mda_header``` method signature to ```u32``` and make other changes as necessary (rather than cast here).

src/engine/strat_engine/blockdev.rs
```diff
@@ -7,7 +7,7 @@
 use std::fs::OpenOptions;
 use std::path::PathBuf;
 
-use time::Timespec;
+use chrono::{DateTime, Local};
```
Kintaro: Before the change, the code in `BlockDevMgr::save_state` was using `time::now()`, which returns the local time, so I assumed from there that the local time was a requirement. 

I don't know all the stratisd details and cannot make that decision. But local time seems to make more sense to me. Should UTC be preferred, I can easily make that change.

Kintaro:
Ok, requested changes are done

src/engine/strat_engine/blockdev.rs
mulkieran: Yes. I'm worried that that might have been an oversight. It's nice that the chrono crate makes that decision more explicit.

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -11,7 +11,7 @@ use std::path::{Path, PathBuf};
 use std::str::FromStr;
 
 use rand::{thread_rng, sample};
-use time::{Duration, Timespec, now};
+use chrono::{Duration, Local, DateTime};
```
mulkieran: In theory we alphabetize our use directives, by package and then by name.

So, ```chrono``` directive should go above ```rand```, and the order of names is ```DateTime, Duration, Local```.

I know that ```rand``` directive is not setting the best example, but it's best to leave that alone for this PR.

src/engine/strat_engine/metadata.rs
```diff
@@ -298,7 +298,7 @@ mod mda {
 
     use byteorder::{ByteOrder, LittleEndian};
     use crc::crc32;
-    use time::Timespec;
+    use chrono::{TimeZone, DateTime, Local};
```
mulkieran: Same remark about alphabetization here as above.

src/engine/strat_engine/metadata.rs
```diff
@@ -649,7 +647,7 @@ mod mda {
         use std::io::Cursor;
 
         use quickcheck::{QuickCheck, TestResult};
-        use time::{now, Timespec};
+        use chrono::Local;
```
mulkieran: Alphabetization again.

src/engine/strat_engine/metadata.rs
```diff
@@ -682,22 +680,18 @@ mod mda {
...
-            fn mda_header(data: Vec<u8>, sec: i64, nsec: i32, region_size_ext: u32) -> TestResult {
-                // unwritable timestamp
-                if sec < 0 || nsec < 0 {
-                    return TestResult::discard();
-                }
-
-                // sec value of 0 is interpreted as no timestamp when read
-                if sec == 0 {
+            fn mda_header(data: Vec<u8>, sec: i64, nsec: u32, region_size_ext: u32) -> TestResult {
+                // sec < 0: unwritable timestamp pr
```
mulkieran: typo here.

mulkieran:
@agrover We still need to resolve the UTC vs. Local time question...

agrover:
The consensus seems to be UTC. Can we do that with this PR? I'll open an issue against the design doc to update it as well.

Kintaro:
Ok, will update it to UTC

mulkieran:
@agrover This all looks good to me.

agrover:
Thanks @Kintaro and @mulkieran, merging!

Issue #492, opened 2017-07-28
It should be possible to call `iter()` on a Table


agrover:
Why must we call `into_iter` on `Table` (such as `self.filesystems`) instead of `iter`? Something is weird.

mulkieran:
I think this is resolved by #678.

PR #509, opened 2017-08-03
Add `.iter()` to Table

Kintaro:
This commit adds `.iter()` to `Table` and also adds an
additional version of IntoIterator. This brings it completely
in line with the `Vec` interface of the standard library.

This is referring to https://github.com/stratis-storage/stratisd/issues/492

### DIFF
--- a/src/engine/structures.rs
+++ b/src/engine/structures.rs
@@ -5,6 +5,7 @@
 use std::collections::HashMap;
 use std::iter::IntoIterator;
 use std::slice::{Iter, IterMut};
+use std::vec::IntoIter;
 
 use uuid::Uuid;
 
@@ -33,7 +34,7 @@ impl<'a, T: HasName + HasUuid> IntoIterator for &'a mut Table<T> {
     type Item = &'a mut T;
     type IntoIter = IterMut<'a, T>;
 
-    fn into_iter(mut self) -> IterMut<'a, T> {
+    fn into_iter(mut self) -> Self::IntoIter {
         self.items.iter_mut()
     }
 }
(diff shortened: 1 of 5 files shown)


src/engine/structures.rs
```diff
@@ -178,6 +188,14 @@ impl<T: HasName + HasUuid> Table<T> {
             (Some(name_item), Some(uuid_item)) => vec![name_item, uuid_item],
         }
     }
+
```
mulkieran: My reading of ```Programming Rust``` was that by implementing ```IntoIterator``` in various ways that methods like these would be unnecessary. Could you explain more?

Kintaro: The consuming version was basically missing. So, the first two implementations only work on a reference of a `Table`. But calling `into_iter` on a non-reference `Table` was impossible so far.

The consuming version then can _exploit_ the consuming version of `Vec`'s `into_iter`

src/engine/structures.rs
```diff
@@ -178,6 +188,14 @@ impl<T: HasName + HasUuid> Table<T> {
...
+    pub fn iter(&self) -> Iter<T> {
```
mulkieran: I think we do not want to be able to mutate the items while they are in the container. The container has an invariant which is tested in the tests. But, essentially, the problem is that there are two keys, and each items also contains its two key values. If it is possible to mutate the items while they are in the table, it is possible to break the invariant, regarding the relationship of the keys to the items they point to.

Kintaro:
So, shall I remove the `iter_mut`?

mulkieran:
Almost certainly...but there may be larger issues. I'm afraid this may take a while to resolve.

Kintaro:
The consuming version should be there, imho.
The normal `iter` version also. To be able to iterate on a non-reference table without consuming it.

trgill:
Can one of the admins verify this patch?

mulkieran:
We need to review the design of Table as a whole, unfortunately.

agrover:
closing due to inactivity. See also #678.

PR #511, opened 2017-08-03
Make clippy happy

mulkieran:
It doesn't seem to reach a fixed point, so one change that it asks for may
require another.


agrover:
👍

PR #513, opened 2017-08-03
Experimental-ignore

mulkieran:


mulkieran:
Next time we update clippy, it should be to at least 0.9.0, so this can go away.

PR #522, opened 2017-08-09
Put more StratPool functionality into ThinPool

mulkieran:
This PR moves a significant amount of ThinPool related functionality into the ThinPool implementation. In particular,

1. It generates the appropriate FlexDevsSave struct to return to the ```ThinPool::record()``` method. This makes sense at this point, since all the data required for the FlexDevsSave struct is part of the ThinPool now. It does not implement Recordable for FlexDevsSave, because it requires a closure as an argument.
2. The calculation of ```min_initial_size()``` is moved down to ThinPool, as it is really the ThinPool's minimum initial size that is calculated. The name is changed to ```initial_size()```, its the checking of the relation that makes it a minimum.
3. ```ThinPool::new()``` now handles allocating the segments it requires and setting up necessary device mapper devices. To do this, it takes a mut reference to a BlockDevMgr.
4. ```ThinPool::setup()``` now handles locating the segments corresponding to each device mapper device and setting up the devices as appropriate. To do this, it takes a reference to its corresponding FlexDevsSave struct.

Much of this is enabled by the previous refactoring, which moved all the data that is required to build a FlexDevsSave struct into the the ThinPool. Consequently, passing a FlexDevsSave struct as an argument or returning it from a function is a meaningful thing for ThinPool to do.

The principle motivation of this refactoring is better encapsulation and consolidation of the handling of segment lists into a single implementation.

### DIFF
--- a/src/engine/strat_engine/blockdevmgr.rs
+++ b/src/engine/strat_engine/blockdevmgr.rs
@@ -65,14 +65,9 @@ impl BlockDevMgr {
         Ok(BlockDevMgr::new(initialize(pool_uuid, devices, mda_size, force)?))
     }
 
-    /// Obtain a BlockDev by its Device.
-    pub fn get_by_device(&self, device: Device) -> Option<&BlockDev> {
-        self.block_devs.iter().find(|d| d.device() == &device)
-    }
-
-    // Obtain a BlockDev by its UUID.
-    pub fn get_by_uuid(&self, uuid: &DevUuid) -> Option<&BlockDev> {
-        self.block_devs.iter().find(|d| d.uuid() == uuid)
+    /// Get a function that maps blockdev device numbers to their UUIDs
+    pub fn device_number_to_uuid(&self) -> Box<Fn(&Device) -> Option<DevUuid>> {
+        device_number_to_uuid(&self.block_devs)
     }
 
     pub fn add(&mut self,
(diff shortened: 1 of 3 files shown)


mulkieran:
I feel that this is basically ready. I expect to squash the final six commits before merging, as each one tends to overwrite the previous. For this reason, it is probably best to read this PR as a whole, rather than as individual commits.

mulkieran:
@agrover @trgill review?

mulkieran:
Whoops, I want to try a little experiment.

mulkieran:
rebased.

mulkieran:
Ok. This is ready for preliminary review. I expect to squash the six middle commits and I think it will work best if you mostly look at the whole set of changes rather than each indiividudal commit. I plan to write additional code in ```ThinPool::new()``` and ```ThinPool::setup()``` to teardown previously build devices if there is a failure to build subsequent devices, but I'm waiting on initial feedback before doing that work.

mulkieran:
@agrover @trgill review?

agrover:
I think it might go more smoothly if each flex layer segment kept a shared reference to the `BlockDev` it was allocated from, avoiding lookups. (Would have to create a new type instead of reusing `devicemapper::Segment`):

```
struct BlockDevSegment {
    blockdev: Rc<RefCell<BlockDev>>,
    start: Sectors,
    length: Sectors,
}
```
This would make `mapper` argument to `flexdevssave` go away, so `ThinPool` could implement `Recordable`. This then could lead to `Recordable::record()` not needing to return an `EngineResult<T>` but just `T`.

trgill:
PR looks to move in the right direction to me.

mulkieran:
It would be nice if Recordable trait didn't require ```record()``` to return an EngineResult...

mulkieran:
rebased.

mulkieran:
@agrover I don't think the mapping problem can be solved with your suggestion. We obtain the segments from the underlying devices defined in device mapper, and they don't know about Blockdev's.

agrover:
@mulkieran ok. Let's shelve that idea for now then. This looks 👍 to me.

mulkieran:
I added three more commits, but the commits before are unchanged.

I don't think squashing the commits will repay the effort, so I'm leaving them as is.

I'm leaving cleanup of devicemapper devices until a separate PR, I think it will go better with some work in devicemapper that should be done first. I'll put in an issue for that.

@agrover @trgill final review?

mulkieran:
Just fixed two minor clippy errors, ```map(|x| *x)``` to ```cloned()```.

agrover:
I'm 👍 except for 8ed1a75. I think we make the code simpler if we just keep references to which blockdev from each segment is within.. in each segment (using a new type). Please see last two commits of https://github.com/stratis-storage/stratisd/pull/537 for what I'm thinking.

This is a bigger and somewhat separate change worthy of further discussion, so I recommend holding back 8ed1a75 from this PR so the rest can be merged sooner.

mulkieran:
Superceded by #537.

mulkieran:
"Unify order in which ..." can wait until a separate commit. It's really more connected to cleanup issues.

PR #526, opened 2017-08-09
Update term package to newer version

mulkieran:
Having trouble installing on CI from previous version.


mulkieran:
Something went wrong with the CI and output logs were not avaiable for any of the tests.

mulkieran:
Closed. We don't think term is the real culprit.

PR #531, opened 2017-08-10
Do not cache cargo for CI

mulkieran:
This seems to result in odd compilation errors for our test build.
The evidence is that if the cache is manually cleared, generally the loop
tests succeed, but if not there is a strange compilation error about a
missing item in a manifest file for one of our dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,6 @@ addons:
 
 language: rust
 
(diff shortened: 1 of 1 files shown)


mulkieran:
Merging immediately after manually clearing all caches.

PR #537, opened 2017-08-16
Transfer functionality from StratPool to ThinPool where appropriate

agrover:


src/engine/strat_engine/blockdev.rs
```diff
@@ -42,7 +42,7 @@ impl BlockDev {
         &self.dev
     }
 
-    pub fn wipe_metadata(self) -> EngineResult<()> {
+    pub fn wipe_metadata(&mut self) -> EngineResult<()> {
```
mulkieran: I don't think self reference needs to be mutable.

src/engine/strat_engine/cleanup.rs
```diff
@@ -14,12 +16,13 @@ use super::pool::StratPool;
...
 /// If an error occurs while wiping a blockdev, attempt to wipe all remaining.
-pub fn wipe_blockdevs(mut blockdevs: Vec<BlockDev>) -> EngineResult<()> {
+pub fn wipe_blockdevs(mut blockdevs: Vec<Rc<RefCell<BlockDev>>>) -> EngineResult<()> {
     let mut unerased_devnodes = Vec::new();
 
     for bd in blockdevs.drain(..) {
-        let bd_devnode = bd.devnode.clone();
-        bd.wipe_metadata()
+        let bd_devnode = bd.borrow().devnode.to_owned();
+        bd.borrow_mut()
```
mulkieran: If ```wipe_metadata``` doesn't require a mutable reference, then this can just be a ```borrow()```.

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -40,15 +42,41 @@ pub fn resolve_devices(paths: &[&Path]) -> io::Result<HashSet<Device>> {
...
 }
 
+#[derive(Debug)]
+pub struct BlkDevSegment {
+    pub blockdev: Rc<RefCell<BlockDev>>,
+    pub start: Sectors,
+    pub length: Sectors,
+}
+
+impl BlkDevSegment {
```
mulkieran: ```BlkDevSegment``` is strictly more informative than ```Segment```.

You can not implement ```From/Into``` for the two, obviously, but you should be able to implement a method on ```BlkDevSegment```, like ```into_segment()```. This ought to simplify ```convert_to_dm_segs()``` as well as a bunch of test code and maybe some other things as well.

src/engine/strat_engine/thinpool.rs
```diff
@@ -333,25 +333,25 @@ impl ThinPool {
...
-               thin_data_dev: self.data_segments
-                   .iter()
-                   .map(|bseg| (*bseg.blockdev.borrow().uuid(), bseg.start, bseg.length))
-                   .collect::<Vec<_>>(),
-               thin_meta_dev_spare: self.meta_spare_segments
-                   .iter()
-                   .map(|bseg| (*bseg.blockdev.borrow().uuid(), bseg.start, bseg.length))
-                   .collect::<Vec<_>>(),
-           })
+    pub fn flexdevssave(&self) -> FlexDevsSave {
```
mulkieran: Does it not work out to make this an implementation of ```Recordable<FlexDevsSave>```?

agrover: Ah yes, this works!

We end up with 
```
            flex_devs: self.thin_pool.record(),
            thinpool_dev: self.thin_pool.record(),
```
which is a little strange, but seems to DTRT...

agrover:
Changes based on review feedback, and also added "Don't drain Vecs" 0fe8982.

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -65,14 +97,12 @@ impl BlockDevMgr {
...
 
-    /// Obtain a BlockDev by its Device.
-    pub fn get_by_device(&self, device: Device) -> Option<&BlockDev> {
-        self.block_devs.iter().find(|d| d.device() == &device)
-    }
-
     // Obtain a BlockDev by its UUID.
-    pub fn get_by_uuid(&self, uuid: &DevUuid) -> Option<&BlockDev> {
-        self.block_devs.iter().find(|d| d.uuid() == uuid)
+    pub fn get_by_uuid(&self, uuid: &DevUuid) -> Option<Rc<RefCell<BlockDev>>> {
```
mulkieran: Probably you can lose this method, too. See my comment at use below.

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -108,8 +138,12 @@ impl BlockDevMgr {
                 break;
             }
 
-            let (gotten, r_segs) = bd.request_space(needed);
-            segs.extend(r_segs);
+            let (gotten, r_segs) = bd.borrow_mut().request_space(needed);
+            segs.extend(r_segs
```
mulkieran: Would be nice to have a distinct function for the action of building up ```BlkDevSegments```.

src/engine/strat_engine/thinpool.rs
```diff
@@ -55,16 +66,62 @@ pub struct ThinPoolStatus {
...
-               spare_segments: Vec<Segment>,
-               meta_dev: LinearDev,
-               data_dev: LinearDev,
-               mdv: MetadataVol)
+               block_mgr: &mut BlockDevMgr)
                -> EngineResult<ThinPool> {
+        if block_mgr.avail_space() < ThinPool::initial_size() {
+            let avail_size = block_mgr.avail_space().bytes();
+            return Err(EngineError::Engine(ErrorEnum::Invalid,
+                                           format!("Space on pool must be at least {} bytes, \
```
mulkieran: This is minor, but I'ld prefer it if you assigned the error message to an identifier, say ```err_msg```, and used that as an argument. All this crowding right is a bit annoying.

src/engine/strat_engine/thinpool.rs
```diff
@@ -89,46 +149,92 @@ impl ThinPool {
...
-                 spare_segments: Vec<Segment>,
-                 meta_dev: LinearDev,
-                 data_dev: LinearDev,
-                 mdv: MetadataVol,
-                 fs_save: Vec<FilesystemSave>)
+                 flex_devs: &FlexDevsSave,
+                 bd_mgr: &BlockDevMgr)
                  -> EngineResult<ThinPool> {
+
+        let mapper = |triple: &(Uuid, Sectors, Sectors)| -> EngineResult<BlkDevSegment> {
```
mulkieran: I think you should call a function on ```bd_mgr``` that returns a mapper function, as I do in my addendum to previous PR, rather than use ```get_by_uuid()``` to construct this mapper function. Then ```BlockDevMgr``` has control over how mapping is constructed, and efficiency is also gained.

src/engine/strat_engine/thinpool.rs
```diff
@@ -197,34 +330,42 @@ impl ThinPool {
...
     /// Get the devicemapper::ThinPoolDev for this pool. Used for testing.
     pub fn thinpooldev(&self) -> &ThinPoolDev {
         &self.thin_pool
     }
 
     /// Extend the thinpool with new data regions.
-    pub fn extend_data(&mut self, dm: &DM, segs: Vec<Segment>) -> EngineResult<()> {
-        Ok(self.thin_pool.extend_data(dm, segs)?)
+    pub fn extend_data(&mut self, dm: &DM, new_segs: Vec<BlkDevSegment>) -> EngineResult<()> {
+        self.thin_pool
```
mulkieran: Add a new method to ```LinearDev``` that just resets the table and call it at the bottom with the coalesced segments?

tests/util/dm_tests.rs
```diff
@@ -40,12 +40,18 @@ pub fn test_linear_device(paths: &[&Path]) -> () {
...
+        .iter()
+        .map(|i| i.borrow().avail_range().1)
+        .sum();
 
     let segments = initialized
         .iter()
-        .map(|block_dev| block_dev.avail_range())
-        .collect::<Vec<Segment>>();
+        .map(|block_dev| {
+                 let (start, length) = block_dev.borrow().avail_range();
```
mulkieran: This actions happens repeatedly. Can it be combined into a single utility function?

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -40,15 +42,45 @@ pub fn resolve_devices(paths: &[&Path]) -> io::Result<HashSet<Device>> {
...
 #[derive(Debug)]
 pub struct BlockDevMgr {
-    block_devs: Vec<BlockDev>,
+    block_devs: Vec<Rc<RefCell<BlockDev>>>,
     last_update_time: Option<DateTime<Utc>>,
 }
 
 impl BlockDevMgr {
-    pub fn new(block_devs: Vec<BlockDev>) -> BlockDevMgr {
+    pub fn new(block_devs: Vec<Rc<RefCell<BlockDev>>>) -> BlockDevMgr {
```
mulkieran: I think it would be nicer if the type of ```block_devs``` continued to be just ```Vec<BlockDev>``` but converted the contents to reference counted ref cells here. That way, ```initialize()``` can keep its simple return type.

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -40,15 +42,45 @@ pub fn resolve_devices(paths: &[&Path]) -> io::Result<HashSet<Device>> {
...
+    }
+
+    pub fn to_segment(&self) -> Segment {
+        Segment::new(*self.blockdev.borrow().device(), self.start, self.length)
+    }
+}
+
+/// Build a Vec<Segment> from BlkDevSegments. This is useful for calls
+/// to the devicemapper library.
+pub fn convert_to_dm_segs(bsegs: &[BlkDevSegment]) -> Vec<Segment> {
```
mulkieran: I like name ```project_to_dm``` better.

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -82,7 +112,7 @@ impl BlockDevMgr {
                -> EngineResult<Vec<PathBuf>> {
         let devices = resolve_devices(paths)?;
         let bds = initialize(pool_uuid, devices, MIN_MDA_SECTORS, force)?;
-        let bdev_paths = bds.iter().map(|p| p.devnode.clone()).collect();
+        let bdev_paths = bds.iter().map(|p| p.borrow().devnode.clone()).collect();
```
mulkieran: If initialize returns ```Vec<BlockDev>``` no need to borrow here, but below, need to extend with reference counted blockdevs.

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -206,7 +247,7 @@ pub fn initialize(pool_uuid: &PoolUuid,
                   devices: HashSet<Device>,
                   mda_size: Sectors,
                   force: bool)
-                  -> EngineResult<Vec<BlockDev>> {
+                  -> EngineResult<Vec<Rc<RefCell<BlockDev>>>> {
```
mulkieran: As suggested above, maybe stick to previous return type.

src/engine/strat_engine/cleanup.rs
```diff
@@ -14,12 +16,13 @@ use super::pool::StratPool;
 /// Wipe a Vec of blockdevs of their identifying headers.
 /// Return an error if any of the blockdevs could not be wiped.
 /// If an error occurs while wiping a blockdev, attempt to wipe all remaining.
-pub fn wipe_blockdevs(mut blockdevs: Vec<BlockDev>) -> EngineResult<()> {
+pub fn wipe_blockdevs(blockdevs: Vec<Rc<RefCell<BlockDev>>>) -> EngineResult<()> {
```
mulkieran: If ```initialize``` can stick w/ old return type, then so can this method.

src/engine/strat_engine/setup.rs
```diff
@@ -158,7 +160,7 @@ pub fn get_metadata(pool_uuid: PoolUuid, devnodes: &[PathBuf]) -> EngineResult<O
 pub fn get_blockdevs(pool_uuid: PoolUuid,
                      pool_save: &PoolSave,
                      devnodes: &[PathBuf])
-                     -> EngineResult<Vec<BlockDev>> {
+                     -> EngineResult<Vec<Rc<RefCell<BlockDev>>>> {
```
mulkieran: Maybe this could stay with previous return type also.

tests/util/dm_tests.rs
```diff
@@ -40,12 +40,18 @@ pub fn test_linear_device(paths: &[&Path]) -> () {
                                  MIN_MDA_SECTORS,
                                  false)
             .unwrap();
-    let total_blockdev_size: Sectors = initialized.iter().map(|i| i.avail_range().length).sum();
+    let total_blockdev_size: Sectors = initialized
```
mulkieran: If ```initialize``` retained old return type, no need to borrow here.

tests/util/dm_tests.rs
```diff
@@ -40,12 +40,18 @@ pub fn test_linear_device(paths: &[&Path]) -> () {
...
+        .iter()
+        .map(|i| i.borrow().avail_range().1)
+        .sum();
 
     let segments = initialized
         .iter()
-        .map(|block_dev| block_dev.avail_range())
-        .collect::<Vec<Segment>>();
+        .map(|block_dev| {
```
mulkieran: If intialize retained old type, no need to borow here.

tests/util/dm_tests.rs
```diff
@@ -72,10 +78,12 @@ pub fn test_thinpool_device(paths: &[&Path]) -> () {
                                               initialized.last().unwrap());
 
     let dm = DM::new().unwrap();
-    let metadata_dev = LinearDev::new("stratis_testing_thinpool_metadata",
-                                      &dm,
-                                      vec![metadata_blockdev.avail_range()])
-            .unwrap();
+    let (meta_start, meta_length) = metadata_blockdev.borrow().avail_range();
```
mulkieran: I think that, in theory, all the newly introduced borrowing could be dropped in this file.

agrover:
rebased and added 1 commit to implement your suggestion.

src/engine/strat_engine/blockdevmgr.rs
agrover: Let's come up with some more name options, maybe we can find a mutually tolerable^Hagreeable one. 😃 I think "segs" and/or "convert" needs to be in the name somewhere to be clear what it's doing.

mulkieran: The signature already tells us the domain and range of the function, so that information doesn't need to be in the name.

projection is a particular kind of conversion, it just carries more information than generic "convert".

agrover: Ah. It didn't click with me it referred to mathematical projection, I thought it was a noun -- i.e. it meant converting a type associated with this project to a dm type!

More options: `map_to_dm`, `map_to_dm_segs`, `to_dm_segs`, `make_dm_segs`, `build_vec_of_segments`.

mulkieran: ```map_to_dm```, I guess.

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -40,17 +42,50 @@ pub fn resolve_devices(paths: &[&Path]) -> io::Result<HashSet<Device>> {
...
+
+/// Build a Vec<Segment> from BlkDevSegments. This is useful for calls
+/// to the devicemapper library.
+pub fn convert_to_dm_segs(bsegs: &[BlkDevSegment]) -> Vec<Segment> {
+    bsegs
+        .iter()
+        .map(|bseg| bseg.to_segment())
+        .collect::<Vec<_>>()
+}
+
```
mulkieran: Proposed fun

```
fn inject_bd(bd: Rc<RefCell<BlockDev>>, segs: &[(Sectors, Sectors)]) -> Vec<BlkDevSegment> {
     segs.into_iter().map(|(start, length)| { BlkDevSegment::new(Rc::clone(bd), start, length) }))
}
```
used below.

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -106,8 +140,12 @@ impl BlockDevMgr {
                 break;
             }
 
-            let (gotten, r_segs) = bd.request_space(needed);
-            segs.extend(r_segs);
+            let (gotten, r_segs) = bd.borrow_mut().request_space(needed);
+            segs.extend(r_segs
+                            .into_iter()
```
mulkieran: ```segs.extend(inject_bd(bd, r_segs));```

src/engine/strat_engine/blockdevmgr.rs
agrover: If this was used in more than one spot I'd be in favor, but as it is I don't think a one-line function is quite justified.

However, maybe it's clearer if we have a let binding that does the map, and then pass that to extend() instead of a single big expression? Let's try that.

agrover:
added two commits to address review comments.

src/engine/strat_engine/blockdevmgr.rs
mulkieran: It could be used three times in test code, which is a hint.

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -298,7 +343,9 @@ pub fn initialize(pool_uuid: &PoolUuid,
         } else {
             // TODO: check the return values and update state machine on failure
             let _ = BDA::wipe(&mut f);
-            let _ = wipe_blockdevs(bds);
+            let _ = wipe_blockdevs(bds.into_iter()
```
mulkieran: I don't think ```wipe_blockdevs()``` needs a ```Vec``` of reference counted BlockDevs.

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -40,17 +42,50 @@ pub fn resolve_devices(paths: &[&Path]) -> io::Result<HashSet<Device>> {
         .collect()
 }
 
+#[derive(Debug)]
+pub struct BlkDevSegment {
```
mulkieran: It's not obvious to me that:

```
pub struct BlkDevSegment {
  pub uuid: Uuid,
  pub device: Device,
  pub start: Sectors,
  pub length: Sectors,
}
```

wouldn't be just as useful and would also eliminate ref counting requirement.

src/engine/strat_engine/blockdevmgr.rs
agrover: Oh, just a `&[Rc<RefCell<BlockDev>>]`? Ok.

agrover:
ok, removed the `Rc<RefCell<BlockDev>>` in BlkDevSegment in favor of duplicating the blkdev's uuid and Device (just put a devicemapper::Segment in there, which includes start and length.)

Obviously simpler. I have a feeling BlockDev may become very handy in the future, but in the spirit of YAGNI this does the trick for the present!

Lots of rebasing and squashing to do, but here is the latest.

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -40,6 +40,40 @@ pub fn resolve_devices(paths: &[&Path]) -> io::Result<HashSet<Device>> {
...
+    pub segment: Segment,
+}
+
+impl BlkDevSegment {
+    pub fn new(uuid: DevUuid, segment: Segment) -> BlkDevSegment {
+        BlkDevSegment { uuid, segment }
+    }
+
+    pub fn to_segment(&self) -> Segment {
+        // TODO: dm-rs Segment should implement Clone & Copy.
```
mulkieran: It should definitely implement ```Clone```, maybe not ```Copy```. I don't know where you should draw the line at ```Copy```, but I think it should be pretty small.

src/engine/strat_engine/cleanup.rs
```diff
@@ -14,11 +13,11 @@ use super::pool::StratPool;
 /// Wipe a Vec of blockdevs of their identifying headers.
 /// Return an error if any of the blockdevs could not be wiped.
 /// If an error occurs while wiping a blockdev, attempt to wipe all remaining.
-pub fn wipe_blockdevs(mut blockdevs: Vec<BlockDev>) -> EngineResult<()> {
+pub fn wipe_blockdevs(blockdevs: &[BlockDev]) -> EngineResult<()> {
```
mulkieran: The header comment will need updating...

src/engine/strat_engine/thinpool.rs
mulkieran: I think you should call a function on ```bd_mgr``` that returns a mapper function, as I do in my addendum to previous PR, rather than use ```get_by_uuid()``` to construct this mapper function. Then ```BlockDevMgr``` has control over how mapping is constructed, and efficiency is also gained.

The mapper function that ```BlockDevMgr``` returns just maps a UUID to a device.

src/engine/strat_engine/thinpool.rs
```diff
@@ -197,34 +330,40 @@ impl ThinPool {
...
     /// Get the devicemapper::ThinPoolDev for this pool. Used for testing.
     pub fn thinpooldev(&self) -> &ThinPoolDev {
         &self.thin_pool
     }
 
     /// Extend the thinpool with new data regions.
-    pub fn extend_data(&mut self, dm: &DM, segs: Vec<Segment>) -> EngineResult<()> {
-        Ok(self.thin_pool.extend_data(dm, segs)?)
+    pub fn extend_data(&mut self, dm: &DM, new_segs: Vec<BlkDevSegment>) -> EngineResult<()> {
+        self.thin_pool.extend_data(dm, map_to_dm(&new_segs))?;
```
mulkieran: Move this call to end of method? It would need to be to a method other than ```extend_data```, one that reset the data. That is kind of scary, obviously, but we should be prepared to be able to stand by the correctness of this code. I think such an approach would be more robust to maintain.

The drawback is that you could coalesce the segments in this method, and then fail, and then, ideally, do the operation of uncoalescing them, which would be a pain. But that can be solved by making a local copy of the coalesced segments, doing the coalescing, making the call with the copy, and then setting the field.

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -40,6 +40,40 @@ pub fn resolve_devices(paths: &[&Path]) -> io::Result<HashSet<Device>> {
...
+        Segment::new(self.segment.device, self.segment.start, self.segment.length)
+    }
+}
+
+impl Recordable<Vec<(Uuid, Sectors, Sectors)>> for Vec<BlkDevSegment> {
```
mulkieran: ```DevUuid``` here and on line below?

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -190,10 +224,10 @@ impl BlockDevMgr {
 }
 
 impl Recordable<HashMap<Uuid, BlockDevSave>> for BlockDevMgr {
-    fn record(&self) -> EngineResult<HashMap<Uuid, BlockDevSave>> {
+    fn record(&self) -> HashMap<Uuid, BlockDevSave> {
```
mulkieran: ```DevUuid``` here?

src/engine/strat_engine/thinpool.rs
mulkieran: DevUuid?

mulkieran:
Ok. I have yet another point to make. It looks to me like most/all of ```StratPool::check()``` now really belongs in ```ThinPool::check()```, so long as you feel Ok about passing a mutable reference to a blockdevmgr to ```ThinPool::check()```. Much tidier.

src/engine/strat_engine/blockdevmgr.rs
agrover: We can address this if/when dm-rs makes this change.

src/engine/strat_engine/cleanup.rs
agrover: ok

src/engine/strat_engine/thinpool.rs
agrover: I think we want this at the top, so we don't update our state until we know the lower-level operation has succeeded.

agrover:
rebased, squashed. Nothing changed with `check()` since I think that should get its own PR.

src/engine/strat_engine/thinpool.rs
mulkieran: My point is that actually changing the state should not be necessary, especially now that segments can be cloned.

mulkieran:
```StratPool::check()``` is about 35 lines, most of it empty space. Not much rewriting is needed, it would mostly require deleting some now useless enums. I don't think that moving it requires an extra PR.

agrover:
Changes in a PR should be related, and `check()` isn't related to the changes in this PR.

(I actually did try to whip this up but it's more involved than it looks.. after 15 minutes I still wasn't done. maybe the PR ends up being trivial but let's please give it its own PR, separate from this one.)

src/engine/strat_engine/thinpool.rs
agrover: I'm still not comprehending an alternative implementation that is better. Maybe it'd be faster if you can add a commit to this PR that demonstrates?

mulkieran:
The reason I'm proposing moving the functionality in ```StratPool::check()``` in this PR is that I think that action _is_ naturally part of this PR. If it's hard to move, that may be a bad sign.

src/engine/strat_engine/thinpool.rs
mulkieran: I will give it a try.

mulkieran:
Superceded by #546.

PR #546, opened 2017-08-31
Transfer some functionality from StratPool to ThinPool

mulkieran:
By moving functionality that properly belongs with the components of ThinPool into ThinPool methods the implementation can be tightened up. Significant changes are:

* Generate FlexDevsSave struct in ThinPool rather than using accessor methods in ThinPool to build the struct in StratPool.
* Transfer functionality from StratPool::setup() to ThinPool::setup(), StratPool::new() to ThinPool::new(), StratPool::check() to ThinPool::check().
* Use a new type, called BlkDevSegment to store the segments for the various devices that comprise the ThinPool. Transform them to DM Segments when necessary to invoke DM methods. Make use of the fact that BlkDevSegment contains its BlockDev UUID to change Recordable::record() to it returns a bare type and not a Result.

### DIFF
--- a/src/engine/strat_engine/blockdev.rs
+++ b/src/engine/strat_engine/blockdev.rs
@@ -115,7 +115,7 @@ impl BlockDev {
 }
 
 impl Recordable<BlockDevSave> for BlockDev {
-    fn record(&self) -> EngineResult<BlockDevSave> {
-        Ok(BlockDevSave { devnode: self.devnode.clone() })
+    fn record(&self) -> BlockDevSave {
+        BlockDevSave { devnode: self.devnode.clone() }
     }
 }
--- a/src/engine/strat_engine/blockdev.rs
+++ b/src/engine/strat_engine/blockdev.rs
@@ -9,7 +9,7 @@ use std::path::PathBuf;
 
 use chrono::{DateTime, Utc};
 
-use devicemapper::{Device, Sectors, Segment};
+use devicemapper::{Device, Sectors};
 
 use super::super::errors::EngineResult;
 use super::super::types::{DevUuid, PoolUuid};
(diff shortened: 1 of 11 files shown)


mulkieran:
Intended to superceded #537.

mulkieran:
Fixed a few clippy errors.

mulkieran:
@agrover @trgill review?

agrover:
Reading 2edab1b's commitlog, I think we do care about efficiency, but less so for its own sake and more for how blatantly inefficient code highlights the wrong data structures may be in use. Having `uuid_to_devno` -- a function returning a function -- here is mainly to allow conversion to a collection more suited to lookups -- the HashMap, and only do this conversion work once. OK that makes sense, but can we go further and ask if it makes sense to just use a HashMap (or our own Table) for BlockDevs in BlockDevMgr from the start?

That said, this solution elegantly solves the current issue. I can envision some future feature (removing blockdevs?) that would more strongly push us towards a different base collection type in BlockDevMgr. But YAGNI -- we shouldn't do that before we really need to!

So I'm 👍 for 2edab1b and the rest of the series. I think this is a good approach at this point in time. Just wanted to share some additional thoughts on the matter. 😃 

BTW I also noticed that pool.rs is getting pretty short at this point. I think that's positive and interesting, how StratPool sets things up, but then the block and thinpool pieces each manage their own affairs, with the thinpool making space requests into the blockdevmgr as needed.

mulkieran:
rebased.

mulkieran:
@agrover @trgill any last words?

mulkieran:
@agrover We will need a more sophisticated data structure than a Vec for our blockdevs in ```BlockDevMgr```. But we should avoid a HashMap, because there are more than one crucially useful things on which we may find that we need to index, e.g., Device, UUID, Devnode. I think that will be some serious data structure design work, and we should iron out the issues with our current Table before we do it.

PR #549, opened 2017-08-31
Make write_sectors() faster by using BufWriter

agrover:
Seems like it could reduce the number of system calls we're making.


agrover:
Bigger writes help `dd`, I think they'd help us too. The default buffer size is only 8K so I'm actually going to try some larger buffer sizes and see if they help more, so not merging yet. We're just writing megs of zeroes...

PR #553, opened 2017-09-06
Improve tests with real devices

agrover:
The real tests currently require four block devices, and also take a long time because one test completely wipes one blockdev. This PR improves some tests' ability to use fewer blockdevs in the first few commits.

Later commits make it possible for tests to both retain a minimum number of
devices that they need, as well as optionally running with whatever is made available, by moving to specifying "at least" some number of blockdevs should be given, instead of some absolute number. (Absolute numbers and giving min/max ranges are also supported.)

### DIFF
--- a/tests/real_tests.rs
+++ b/tests/real_tests.rs
@@ -151,51 +151,49 @@ fn test_with_spec<F>(limits: DeviceLimits, test: F) -> ()
 
 #[test]
 pub fn real_test_force_flag_stratis() {
-    test_with_spec(DeviceLimits::Exactly(2), test_force_flag_stratis);
-    test_with_spec(DeviceLimits::Exactly(3), test_force_flag_stratis);
+    test_with_spec(DeviceLimits::AtLeast(1), test_force_flag_stratis);
 }
 
 
 #[test]
 pub fn real_test_linear_device() {
-    test_with_spec(DeviceLimits::Exactly(2), test_linear_device);
-    test_with_spec(DeviceLimits::Exactly(3), test_linear_device);
+    test_with_spec(DeviceLimits::AtLeast(1), test_linear_device);
 }
 
 
 #[test]
 pub fn real_test_thinpool_device() {
-    test_with_spec(DeviceLimits::Exactly(3), test_thinpool_device);
+    test_with_spec(DeviceLimits::AtLeast(1), test_thinpool_device);
 }
 
 #[test]
 pub fn real_test_thinpool_expand() {
-    test_with_spec(DeviceLimits::Exactly(3), test_thinpool_expand);
+    test_with_spec(DeviceLimits::AtLeast(1), test_thinpool_expand);
 }
 
 #[test]
 pub fn real_test_thinpool_thindev_destroy() {
-    test_with_spec(DeviceLimits::Exactly(3), test_thinpool_thindev_destroy);
+    test_with_spec(DeviceLimits::AtLeast(1), test_thinpool_thindev_destroy);
 }
 
 #[test]
 pub fn real_test_pool_blockdevs() {
-    test_with_spec(DeviceLimits::Exactly(3), test_pool_blockdevs);
+    test_with_spec(DeviceLimits::AtLeast(1), test_pool_blockdevs);
 }
 
 #[test]
 pub fn real_test_force_flag_dirty() {
-    test_with_spec(DeviceLimits::Exactly(3), test_force_flag_dirty);
+    test_with_spec(DeviceLimits::AtLeast(1), test_force_flag_dirty);
 }
 
 #[test]
 pub fn real_test_teardown() {
-    test_with_spec(DeviceLimits::Exactly(2), test_teardown);
+    test_with_spec(DeviceLimits::AtLeast(1), test_teardown);
 }
 
 #[test]
 pub fn real_test_initialize() {
-    test_with_spec(DeviceLimits::Exactly(4), test_initialize);
+    test_with_spec(DeviceLimits::AtLeast(2), test_initialize);
 }
 
 #[test]
@@ -205,35 +203,35 @@ pub fn real_test_empty_pool() {
 
 #[test]
 pub fn real_test_basic_metadata() {
-    test_with_spec(DeviceLimits::Exactly(4), test_basic_metadata);
+    test_with_spec(DeviceLimits::AtLeast(2), test_basic_metadata);
 }
 
 #[test]
 pub fn real_test_setup() {
-    test_with_spec(DeviceLimits::Exactly(4), test_setup);
+    test_with_spec(DeviceLimits::AtLeast(2), test_setup);
 }
 
 #[test]
 pub fn real_test_pool_rename() {
-    test_with_spec(DeviceLimits::Exactly(2), test_pool_rename);
+    test_with_spec(DeviceLimits::AtLeast(1), test_pool_rename);
 }
 
 #[test]
 pub fn real_test_blockdevmgr_used() {
-    test_with_spec(DeviceLimits::Exactly(2), test_blockdevmgr_used);
+    test_with_spec(DeviceLimits::AtLeast(1), test_blockdevmgr_used);
 }
 
 #[test]
 pub fn real_test_filesystem_rename() {
-    test_with_spec(DeviceLimits::Exactly(2), test_filesystem_rename);
+    test_with_spec(DeviceLimits::AtLeast(1), test_filesystem_rename);
 }
 
 #[test]
 pub fn real_test_pool_setup() {
-    test_with_spec(DeviceLimits::Exactly(2), test_pool_setup);
+    test_with_spec(DeviceLimits::AtLeast(1), test_pool_setup);
 }
 
 #[test]
 pub fn real_test_xfs_expand() {
-    test_with_spec(DeviceLimits::Exactly(3), test_xfs_expand);
+    test_with_spec(DeviceLimits::AtLeast(1), test_xfs_expand);
 }
--- a/tests/real_tests.rs
+++ b/tests/real_tests.rs
@@ -10,6 +10,7 @@ extern crate serde_json;
 
 mod util;
 
+use std::cmp;
 use std::fs::OpenOptions;
 use std::path::{Path, PathBuf};
 
@@ -61,9 +62,16 @@ impl Drop for RealTestDev {
     }
 }
 
-/// Set up count devices from configuration file.
-/// Wipe first MiB on each device.
-fn get_devices(count: u8) -> Option<Vec<RealTestDev>> {
+enum DeviceLimits {
+    Exactly(usize),
+    AtLeast(usize),
+    #[allow(dead_code)]
+    Range(usize, usize), // inclusive
+}
+
+/// Return one or more lists of device nodes to use, based upon the
+/// constraints. Returns None if constraints can't be met.
+fn get_devices(limits: DeviceLimits) -> Option<Vec<Vec<String>>> {
 
     let file = OpenOptions::new()
         .read(true)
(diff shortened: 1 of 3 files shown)


tests/util/dm_tests.rs
```diff
@@ -74,17 +75,17 @@ pub fn test_thinpool_device(paths: &[&Path]) -> () {
...
-    let (metadata_blockdev, data_blockdev) = (initialized.first().unwrap(),
-                                              initialized.last().unwrap());
+    let mut bd_mgr = BlockDevMgr::new(initialized);
 
     let dm = DM::new().unwrap();
 
-    let (meta_start, meta_length) = metadata_blockdev.avail_range();
-    let meta_segment = Segment::new(*metadata_blockdev.device(), meta_start, meta_length);
+    let meta_segs = bd_mgr
+        .alloc_space(Sectors(16 * IEC::Mi / SECTOR_SIZE as u64))
```
mulkieran: ```Bytes(16 * IEC::Mi).sectors()```

tests/util/dm_tests.rs
```diff
@@ -93,12 +94,13 @@ pub fn test_thinpool_device(paths: &[&Path]) -> () {
     // the same approach when constructing a thin pool.
     wipe_sectors(&metadata_dev.devnode(), Sectors(0), metadata_dev.size()).unwrap();
 
-    let (data_start, data_length) = data_blockdev.avail_range();
-    let data_segment = Segment::new(*data_blockdev.device(), data_start, data_length);
+    let data_segs = bd_mgr
+        .alloc_space(Sectors(IEC::Gi / SECTOR_SIZE as u64))
```
mulkieran: ```Bytes(IEC::Gi).sectors()```

tests/util/setup_tests.rs
```diff
@@ -45,9 +45,10 @@ use libstratis::engine::strat_engine::StratEngine;
 /// 6. Verify that get_metadata() return an error. initialize() only
 /// initializes block devices, it does not write metadata.
 pub fn test_initialize(paths: &[&Path]) -> () {
-    assert!(paths.len() > 2);
+    assert!(paths.len() > 1);
 
-    let (paths1, paths2) = paths.split_at(2);
+    let (paths1, paths2) = paths.split_at(paths.len() / 2);
+    assert!(paths1.len() > 0 && paths2.len() > 0);
```
mulkieran: Better to get rid of this assert and corresponding asserts below. Othewrise, we're just checking correctness of Rust integer division, which we had better be able to rely on.

tests/util/setup_tests.rs
```diff
@@ -45,9 +45,10 @@ use libstratis::engine::strat_engine::StratEngine;
...
 
     let unique_devices = resolve_devices(paths1).unwrap();
```
mulkieran: A somewhat useful assert might be:

```assert_eq!(paths1.len(), unique_devices.len())```

but I wouldn't call it necessary.

File: tests/real_tests.rs [hunk 5]
```diff
@@ -75,114 +83,161 @@ fn get_devices(count: u8) -> Option<Vec<RealTestDev>> {
...
+/// in multiple invocations of the test, with differing numbers of block
+/// devices.
+fn test_with_spec<F>(limits: DeviceLimits, test: F) -> ()
     where F: Fn(&[&Path]) -> ()
 {
     init_logger();
-    let devices = get_devices(count).unwrap();
-    let device_paths: Vec<&Path> = devices.iter().map(|x| x.as_path()).collect();
-    test(&device_paths);
+    let runs = get_devices(limits).expect("couldn't get desired devices");
```
mulkieran: Please use ```unwrap()``` here.

mulkieran:
Generally speaking, the overall direction of this PR looks good.

agrover:
updated based on comments.

tests/real_tests.rs
```diff
@@ -75,114 +83,159 @@ fn get_devices(count: u8) -> Option<Vec<RealTestDev>> {
...
+fn test_with_spec<F>(limits: DeviceLimits, test: F) -> ()
     where F: Fn(&[&Path]) -> ()
 {
     init_logger();
-    let devices = get_devices(count).unwrap();
-    let device_paths: Vec<&Path> = devices.iter().map(|x| x.as_path()).collect();
-    test(&device_paths);
+    let runs = get_devices(limits).unwrap();
+    for run_paths in runs {
+        let devices: Vec<_> = run_paths.iter().map(|x| RealTestDev::new(x)).collect();
```
mulkieran: Can't you unify these two lines, w/ map function ```|x| RealTestDev::new(x).as_path()```?

tests/real_tests.rs
```diff
@@ -61,9 +62,16 @@ impl Drop for RealTestDev {
...
+#[allow(dead_code)]
+enum DeviceLimits {
+    Exactly(usize),
+    AtLeast(usize),
+    Range(usize, usize), // inclusive
+}
+
+/// Return one or more lists of devices to use, based upon the
+/// constraints. Returns None if constraints can't be met.
+fn get_devices(limits: DeviceLimits) -> Option<Vec<Vec<String>>> {
```
mulkieran: It would be nicer if ```get_devices()``` result type were ```Option<Vec<Vec<PathBuf>>>``` and ```RealTestDev::new()``` took a &Path as an argument, see: #554.

tests/real_tests.rs
```diff
@@ -75,114 +83,161 @@ fn get_devices(count: u8) -> Option<Vec<RealTestDev>> {
         .unwrap()
         .as_array()
         .unwrap();
-    if devpaths.len() < count as usize {
-        return None;
+
```
mulkieran: Having DeviceLimits be a single struct, as:

```
struct DeviceLimits {
  lower: usize, // inclusive
  upper: Option<usize> // exclusive 
}

// impl contains a new that ensures lower < upper if upper not None.
```

would work just as well. 

```DeviceLimits { lower: 9, upper: Some(10) }``` means Exactly(9)
```DeviceLimits { lower: 9, upper: None }``` means AtLeast(9)
```DeviceLimits { lower: 9, upper: Some(30) }``` means Range(9, 29)

Then,

```
let avail = devpaths.len();
if limits.lower > avail {
  return None;
}
let high_take = {
  if limits.lower == avail {
    None
  } else {
    match limits.upper {
      None => Some(avail),
      Some(upper) => {
        if limits.lower + 1 == upper {
          None
        } else {
          Some(min(limits.upper - 1, avail))
        }
      }
    }
  }
}
```
```low_take``` is just ```limits.lower``` and continue as before.

**Comment by mulkieran** [comment]
Ideally, a very similar thing should be done with the loopbacked testing infrastructure, i.e., the type of ```get_devices``` and ```test_with_spec``` could usefully change in the same manner.

File: tests/real_tests.rs (hunk 6 above)
Comment (agrover) [reply tests/real_tests.rs]: no this doesn't work because as_path returns a reference to something that is about to leave scope.

File: tests/real_tests.rs [hunk 9]
```diff
@@ -75,114 +83,161 @@ fn get_devices(count: u8) -> Option<Vec<RealTestDev>> {
...
+        }
+    };
+
+    let low_paths: Vec<String> = devpaths
+        .iter()
+        .take(low_take)
+        .map(|x| x.as_str().unwrap().to_owned())
+        .collect();
+
+    if maybe_high_take.is_none() {
```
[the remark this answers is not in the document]
agrover: ok

agrover:
Fixed an unwrap. Other comments are suggestions for further work. I recommend we merge this and then that opens it up for further work to be undertaken by anyone.

mulkieran:
Having DeviceLimits be a single struct, as:

```
struct DeviceLimits {
  lower: usize, // inclusive
  upper: Option<usize> // exclusive 
}

// impl contains a new that ensures lower < upper if upper not None.
```

would work just as well. 

```DeviceLimits { lower: 9, upper: Some(10) }``` means Exactly(9)
```DeviceLimits { lower: 9, upper: None }``` means AtLeast(9)
```DeviceLimits { lower: 9, upper: Some(30) }``` means Range(9, 29)

Then,

```
let avail = devpaths.len();
if limits.lower > avail {
  return None;
}
let high_take = {
  if limits.lower == avail {
    None
  } else {
    match limits.upper {
      None => Some(avail),
      Some(upper) => {
        if limits.lower + 1 == upper {
          None
        } else {
          Some(min(limits.upper - 1, avail))
        }
      }
    }
  }
}
```
```low_take``` is just ```limits.lower``` and continue as before.

The above isn't a suggestion for further work...it's a suggestion on how to take a thing that is written in what seems like an unnecessarily complicated and roundabout way and make it relatively simple and direct.

**Comment by agrover** [comment]
I actually implemented it that way and then switched to the enum approach, because it's easier to understand at the callsite. I see the redundancy in the implementation but to me that's less important. Any ways to improve the enum approach's implementation while keeping the enum?

**Comment by mulkieran** [comment]
FWIW, it's not easier for me to understand at call site (and it will be harder for many non-native speakers of English).

But if you want to keep mnemonic constructors you can convert to two-part representation at top of method as:

```
let (lower, upper) = match limits {
  Exactly(num) => (num, Some(num + 1)),
  AtLeast(num) => (num, None),
  Range(lower, upper) = {
    if lower > upper {
      // do error
    }
    (lower, Some(upper + 1))
  }
} 
```

then proceed more or less as above.

**Comment by agrover** [comment]
added commit to implement calc based on review suggestion.

File: tests/real_tests.rs [hunk 10]
```diff
@@ -61,9 +62,16 @@ impl Drop for RealTestDev {
     }
 }
 
-/// Set up count devices from configuration file.
-/// Wipe first MiB on each device.
-fn get_devices(count: u8) -> Option<Vec<RealTestDev>> {
+#[allow(dead_code)]
```
mulkieran: Please move the allow so that it is just above the Range constructor definition.

tests/real_tests.rs
```diff
@@ -75,114 +83,156 @@ fn get_devices(count: u8) -> Option<Vec<RealTestDev>> {
...
     where F: Fn(&[&Path]) -> ()
 {
     init_logger();
-    let devices = get_devices(count).unwrap();
-    let device_paths: Vec<&Path> = devices.iter().map(|x| x.as_path()).collect();
-    test(&device_paths);
+    let runs = get_devices(limits).unwrap();
+    for run_paths in runs {
+        let devices: Vec<_> = run_paths.iter().map(|x| RealTestDev::new(x)).collect();
+        let device_paths: Vec<_> = devices.iter().map(|x| x.as_path()).collect();
```
mulkieran: Can you squash these two lines?

mulkieran:
Please add an issue regarding assessing the use of similar changes to the loopbacked testing infrastructure.

Other issues can be handled in #554 .

tests/real_tests.rs
mulkieran: ```let device_paths``` and ```test```.

agrover: ok

tests/real_tests.rs
agrover: ok

agrover:
rebased and squashed. made suggested changes. opened issue #558.

tests/real_tests.rs
mulkieran: With recent changes, allow can be completely removed, so might as well do that.

tests/real_tests.rs
```diff
@@ -61,9 +62,16 @@ impl Drop for RealTestDev {
...
-/// Wipe first MiB on each device.
-fn get_devices(count: u8) -> Option<Vec<RealTestDev>> {
+#[allow(dead_code)]
+enum DeviceLimits {
+    Exactly(usize),
+    AtLeast(usize),
+    Range(usize, usize), // inclusive
+}
+
+/// Return one or more lists of devices to use, based upon the
```
mulkieran: "lists of devices" -> "lists of device nodes"

agrover:
fixed comment but left allow because removing it still causes a warning.

mulkieran:
Yeah, it turns out I made a mistake in my test. The allow should just be moved down so it is just above the Range constructor definition.

agrover:
ok, that's how it is right now.

PR #563, opened 2017-09-12
Implement D-Bus Blockdev API

agrover:
Big one here. Hard to test since I think API changes will break cli?

Trait changes:
* Blockdev: Implement needed accessor methods
* Pool: Add methods to get a blockdev based on uuid to Pool trait
* Pool: Add save_state() since BlockDev::set_user_id() can't do this
        itself, DBus code has to.
* Pool: Change add_blockdevs() to return dev UUID instead of paths, since
        paths can be obtained by getting the blockdev from the pool and
        then calling devnode().
* Engine: Change create_pool() to just return pool UUID, same reason.

sim/strat blockdev implementation:
* Change blockdev collections from Vec to HashMap to make lookups by
  uuid more natural.
* sim: Add fields as needed to SimDev
* strat: Add fields as needed to StratBlockDev and BlockDevSave
* strat: Get location and disk_id for each blockdev from saved state, if
         available, in get_blockdevs().

D-Bus:
* Make return signature changes to CreatePool and AddDevs: object paths
  instead of devnodes.
* Create blockdev dbus objects.

general:
* Add BlockDevState enum
* Update tests for changes

### DIFF
--- a/src/dbus_api/api.rs
+++ b/src/dbus_api/api.rs
@@ -71,13 +71,14 @@ fn create_pool(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
                                          MessageItem::Array(vec![], "o".into())]);
             let pool = get_mut_pool!(engine; pool_uuid; default_return; return_message);
 
-            let mut bd_object_paths = Vec::new();
-            for dev_uuid in pool.blockdevs().iter().map(|bd| bd.uuid()) {
-                bd_object_paths
-                    .push(MessageItem::ObjectPath(create_dbus_blockdev(dbus_context,
-                                                                       pool_object_path.clone(),
-                                                                       dev_uuid)));
-            }
+            let bd_object_paths = pool.blockdevs()
+                .iter()
+                .map(|bd| {
+                         MessageItem::ObjectPath(create_dbus_blockdev(dbus_context,
+                                                                      pool_object_path.clone(),
+                                                                      bd.uuid()))
+                     })
+                .collect::<Vec<_>>();
 
             let return_path = MessageItem::ObjectPath(pool_object_path);
             let return_list = MessageItem::Array(bd_object_paths, "o".into());
--- a/src/dbus_api/api.rs
+++ b/src/dbus_api/api.rs
@@ -66,9 +66,9 @@ fn create_pool(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
         Ok(pool_uuid) => {
             let pool_object_path: dbus::Path =
                 create_dbus_pool(dbus_context, object_path.clone(), pool_uuid);
-            let return_sig = "(oao)";
-            let default_return = MessageItem::Array(vec![], return_sig.into());
-
+            let default_return =
+                MessageItem::Struct(vec![MessageItem::ObjectPath(default_object_path()),
+                                         MessageItem::Array(vec![], "o".into())]);
             let pool = get_mut_pool!(engine; pool_uuid; default_return; return_message);
 
             let mut bd_object_paths = Vec::new();
@@ -264,8 +264,9 @@ pub fn connect(engine: Rc<RefCell<Engine>>)
     let (tree, object_path) = get_base_tree(DbusContext::new(engine));
     let dbus_context = tree.get_data().clone();
 
-    // This should never panic as create_dbus_pool() and
-    // create_dbus_filesystem() do not borrow the engine.
+    // This should never panic as create_dbus_pool(),
+    // create_dbus_filesystem(), and create_dbus_blockdev() do not borrow the
+    // engine.
     for pool in local_engine.borrow().pools() {
         let pool_path = create_dbus_pool(&dbus_context, object_path.clone(), pool.uuid());
         for fs_uuid in pool.filesystems().iter().map(|f| f.uuid()) {
--- a/src/dbus_api/api.rs
+++ b/src/dbus_api/api.rs
@@ -32,6 +32,7 @@ use engine::{Engine, Redundancy};
 use stratis::VERSION;
 
 use super::filesystem::create_dbus_filesystem;
+use super::blockdev::create_dbus_blockdev;
 use super::pool::create_dbus_pool;
 use super::types::{DeferredAction, DbusContext, DbusErrorEnum, TData};
 use super::util::STRATIS_BASE_PATH;
@@ -56,27 +57,30 @@ fn create_pool(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
 
     let object_path = m.path.get_name();
     let dbus_context = m.tree.get_data();
-    let result = dbus_context
-        .engine
-        .borrow_mut()
-        .create_pool(name, &blockdevs, tuple_to_option(redundancy), force);
+    let mut engine = dbus_context.engine.borrow_mut();
+    let result = engine.create_pool(name, &blockdevs, tuple_to_option(redundancy), force);
 
     let return_message = message.method_return();
 
     let msg = match result {
-        Ok((uuid, devnodes)) => {
+        Ok(pool_uuid) => {
             let pool_object_path: dbus::Path =
-                create_dbus_pool(dbus_context, object_path.clone(), uuid);
-            let paths = devnodes
-                .iter()
-                .map(|d| {
-                         d.to_str()
-                             .expect("'d' originated in the 'devs' D-Bus argument")
-                             .into()
-                     });
-            let paths = paths.map(MessageItem::Str).collect();
+                create_dbus_pool(dbus_context, object_path.clone(), pool_uuid);
+            let return_sig = "(oao)";
+            let default_return = MessageItem::Array(vec![], return_sig.into());
+
+            let pool = get_mut_pool!(engine; pool_uuid; default_return; return_message);
+
+            let mut bd_object_paths = Vec::new();
+            for dev_uuid in pool.blockdevs().iter().map(|bd| bd.uuid()) {
+                bd_object_paths
+                    .push(MessageItem::ObjectPath(create_dbus_blockdev(&dbus_context,
+                                                                       pool_object_path.clone(),
+                                                                       dev_uuid)));
+            }
+
             let return_path = MessageItem::ObjectPath(pool_object_path);
-            let return_list = MessageItem::Array(paths, "s".into());
+            let return_list = MessageItem::Array(bd_object_paths, "o".into());
             let return_value = MessageItem::Struct(vec![return_path, return_list]);
             let (rc, rs) = ok_message_items();
             return_message.append3(return_value, rc, rs)
@@ -201,7 +205,7 @@ fn get_base_tree<'a>(dbus_context: DbusContext) -> (Tree<MTFn<TData>, TData>, db
         .in_arg(("redundancy", "(bq)"))
         .in_arg(("force", "b"))
         .in_arg(("devices", "as"))
-        .out_arg(("result", "(oas)"))
+        .out_arg(("result", "(oao)"))
         .out_arg(("return_code", "q"))
         .out_arg(("return_string", "s"));
 
@@ -266,6 +270,9 @@ pub fn connect(engine: Rc<RefCell<Engine>>)
         let pool_path = create_dbus_pool(&dbus_context, object_path.clone(), pool.uuid());
         for fs_uuid in pool.filesystems().iter().map(|f| f.uuid()) {
             create_dbus_filesystem(&dbus_context, pool_path.clone(), fs_uuid);
+        }
+        for dev_uuid in pool.blockdevs().iter().map(|bd| bd.uuid()) {
+            create_dbus_blockdev(&dbus_context, pool_path.clone(), dev_uuid);
         }
     }
 
--- a/src/dbus_api/api.rs
+++ b/src/dbus_api/api.rs
@@ -74,7 +74,7 @@ fn create_pool(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
             let mut bd_object_paths = Vec::new();
             for dev_uuid in pool.blockdevs().iter().map(|bd| bd.uuid()) {
                 bd_object_paths
-                    .push(MessageItem::ObjectPath(create_dbus_blockdev(&dbus_context,
+                    .push(MessageItem::ObjectPath(create_dbus_blockdev(dbus_context,
                                                                        pool_object_path.clone(),
                                                                        dev_uuid)));
             }
--- a/src/dbus_api/blockdev.rs
+++ b/src/dbus_api/blockdev.rs
@@ -37,7 +37,7 @@ pub fn create_dbus_blockdev<'a>(dbus_context: &DbusContext,
                                 -> dbus::Path<'a> {
     let f = Factory::new_fn();
 
-    let set_userid_method = f.method("SetUserId", (), set_user_id)
+    let set_userid_method = f.method("SetUserInfo", (), set_user_info)
         .in_arg(("id", "s"))
         .out_arg(("changed", "b"))
         .out_arg(("return_code", "q"))
@@ -48,15 +48,15 @@ pub fn create_dbus_blockdev<'a>(dbus_context: &DbusContext,
         .emits_changed(EmitsChangedSignal::Const)
         .on_get(get_blockdev_devnode);
 
-    let hardware_id_property = f.property::<&str, _>("HardwareId", ())
+    let hardware_info_property = f.property::<&str, _>("HardwareId", ())
         .access(Access::Read)
         .emits_changed(EmitsChangedSignal::Const)
-        .on_get(get_blockdev_hardware_id);
+        .on_get(get_blockdev_hardware_info);
 
-    let user_id_property = f.property::<&str, _>("UserId", ())
+    let user_info_property = f.property::<&str, _>("UserId", ())
         .access(Access::Read)
-        .emits_changed(EmitsChangedSignal::Const)
-        .on_get(get_blockdev_user_id);
+        .emits_changed(EmitsChangedSignal::False)
+        .on_get(get_blockdev_user_info);
 
     let initialization_time_property = f.property::<&str, _>("InitializationTime", ())
         .access(Access::Read)
@@ -68,7 +68,7 @@ pub fn create_dbus_blockdev<'a>(dbus_context: &DbusContext,
         .emits_changed(EmitsChangedSignal::False)
         .on_get(get_blockdev_physical_size);
 
-    let state_property = f.property::<&str, _>("State", ())
+    let state_property = f.property::<u16, _>("State", ())
         .access(Access::Read)
         .emits_changed(EmitsChangedSignal::False)
         .on_get(get_blockdev_state);
@@ -94,12 +94,12 @@ pub fn create_dbus_blockdev<'a>(dbus_context: &DbusContext,
         .add(f.interface(interface_name, ())
                  .add_m(set_userid_method)
                  .add_p(devnode_property)
-                 .add_p(hardware_id_property)
+                 .add_p(hardware_info_property)
                  .add_p(initialization_time_property)
                  .add_p(total_physical_size_property)
                  .add_p(pool_property)
                  .add_p(state_property)
-                 .add_p(user_id_property)
+                 .add_p(user_info_property)
                  .add_p(uuid_property));
 
     let path = object_path.get_name().to_owned();
@@ -107,7 +107,7 @@ pub fn create_dbus_blockdev<'a>(dbus_context: &DbusContext,
     path
 }
 
-fn set_user_id(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
+fn set_user_info(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
     let message: &Message = m.msg;
     let mut iter = message.iter_init();
 
@@ -138,12 +138,12 @@ fn set_user_id(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
                                                            &blockdev_uuid))
                             })?;
 
-        let id_changed = blockdev.user_id() != &Some(new_id.to_owned());
+        let id_changed = blockdev.user_info() != Some(new_id);
 
         blockdev
-            .set_user_id(Some(new_id))
+            .set_user_info(Some(new_id))
             .map_err(|err| {
-                         MethodErr::failed(&format!("set_user_id failed for object path {}: {}",
+                         MethodErr::failed(&format!("set_user_info failed for object path {}: {}",
                                                     object_path,
                                                     err))
                      })?;
@@ -225,30 +225,20 @@ fn get_blockdev_devnode(i: &mut IterAppend,
                           |p| Ok(MessageItem::Str(format!("{}", p.devnode().display()))))
 }
 
-fn get_blockdev_hardware_id(i: &mut IterAppend,
-                            p: &PropInfo<MTFn<TData>, TData>)
-                            -> Result<(), MethodErr> {
-    fn get_hardware_id(blockdev: &BlockDev) -> Result<MessageItem, MethodErr> {
-        match *blockdev.hardware_id() {
-            Some(ref id) => Ok(MessageItem::Str(id.to_owned())),
-            None => Ok(MessageItem::Str("".to_owned())),
-        }
-    }
-
-    get_blockdev_property(i, p, get_hardware_id)
+fn get_blockdev_hardware_info(i: &mut IterAppend,
+                              p: &PropInfo<MTFn<TData>, TData>)
+                              -> Result<(), MethodErr> {
+    get_blockdev_property(i,
+                          p,
+                          |p| Ok(MessageItem::Str(p.hardware_info().unwrap_or("").to_owned())))
 }
 
-fn get_blockdev_user_id(i: &mut IterAppend,
-                        p: &PropInfo<MTFn<TData>, TData>)
-                        -> Result<(), MethodErr> {
-    fn get_user_id(blockdev: &BlockDev) -> Result<MessageItem, MethodErr> {
-        match *blockdev.user_id() {
-            Some(ref id) => Ok(MessageItem::Str(id.to_owned())),
-            None => Ok(MessageItem::Str("".to_owned())),
-        }
-    }
-
-    get_blockdev_property(i, p, get_user_id)
+fn get_blockdev_user_info(i: &mut IterAppend,
+                          p: &PropInfo<MTFn<TData>, TData>)
+                          -> Result<(), MethodErr> {
+    get_blockdev_property(i,
+                          p,
+                          |p| Ok(MessageItem::Str(p.user_info().unwrap_or("").to_owned())))
 }
 
 fn get_blockdev_initialization_time(i: &mut IterAppend,
@@ -272,13 +262,13 @@ fn get_blockdev_state(i: &mut IterAppend,
                       -> Result<(), MethodErr> {
     fn get_state(blockdev: &BlockDev) -> Result<MessageItem, MethodErr> {
         let state = match blockdev.state() {
-            BlockDevState::Missing => "Missing".into(),
-            BlockDevState::Bad => "Bad".into(),
-            BlockDevState::Spare => "Spare".into(),
-            BlockDevState::NotInUse => "Not-in-use".into(),
-            BlockDevState::InUse => "In-use".into(),
+            BlockDevState::Missing => 0,
+            BlockDevState::Bad => 1,
+            BlockDevState::Spare => 2,
+            BlockDevState::NotInUse => 3,
+            BlockDevState::InUse => 4,
         };
-        Ok(MessageItem::Str(state))
+        Ok(MessageItem::UInt16(state))
     }
 
     get_blockdev_property(i, p, get_state)
--- a/src/dbus_api/blockdev.rs
+++ b/src/dbus_api/blockdev.rs
@@ -58,7 +58,7 @@ pub fn create_dbus_blockdev<'a>(dbus_context: &DbusContext,
         .emits_changed(EmitsChangedSignal::False)
         .on_get(get_blockdev_user_info);
 
-    let initialization_time_property = f.property::<&str, _>("InitializationTime", ())
+    let initialization_time_property = f.property::<u64, _>("InitializationTime", ())
         .access(Access::Read)
         .emits_changed(EmitsChangedSignal::Const)
         .on_get(get_blockdev_initialization_time);
@@ -111,7 +111,10 @@ fn set_user_info(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
     let message: &Message = m.msg;
     let mut iter = message.iter_init();
 
-    let new_id: &str = get_next_arg(&mut iter, 0)?;
+    let new_id: Option<&str> = match get_next_arg(&mut iter, 0)? {
+        "" => None,
+        val => Some(val),
+    };
 
     let dbus_context = m.tree.get_data();
     let object_path = m.path.get_name();
@@ -138,26 +141,21 @@ fn set_user_info(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
                                                            &blockdev_uuid))
                             })?;
 
-        let id_changed = blockdev.user_info() != Some(new_id);
+        let id_changed = blockdev.user_info() != new_id;
+        blockdev.set_user_info(new_id);
 
-        blockdev
-            .set_user_info(Some(new_id))
-            .map_err(|err| {
-                         MethodErr::failed(&format!("set_user_info failed for object path {}: {}",
-                                                    object_path,
-                                                    err))
-                     })?;
         id_changed
     };
 
     // FIXME: engine should decide to save state, not this function
-    pool.save_state()
-        .map_err(|err| {
-                     MethodErr::failed(&format!("Could not save state for object path {}: {}",
-                                                object_path,
-                                                err))
-                 })?;
-
+    if id_changed {
+        pool.save_state()
+            .map_err(|err| {
+                         MethodErr::failed(&format!("Could not save state for object path {}: {}",
+                                                    object_path,
+                                                    err))
+                     })?;
+    }
 
     let (rc, rs) = ok_message_items();
     let msg = return_message.append3(MessageItem::Bool(id_changed), rc, rs);
@@ -247,7 +245,7 @@ fn get_blockdev_initialization_time(i: &mut IterAppend,
                                     -> Result<(), MethodErr> {
     get_blockdev_property(i,
                           p,
-                          |p| Ok(MessageItem::Str(p.initialization_time().to_rfc3339())))
+                          |p| Ok(MessageItem::UInt64(p.initialization_time().timestamp() as u64)))
 }
 
 fn get_blockdev_physical_size(i: &mut IterAppend,
--- a/src/dbus_api/blockdev.rs
+++ b/src/dbus_api/blockdev.rs
@@ -1,0 +1,283 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use dbus;
+use dbus::Message;
+use dbus::MessageItem;
+use dbus::arg::IterAppend;
+use dbus::tree::Access;
+use dbus::tree::EmitsChangedSignal;
+use dbus::tree::Factory;
+use dbus::tree::MTFn;
+use dbus::tree::MethodErr;
+use dbus::tree::MethodInfo;
+use dbus::tree::MethodResult;
+use dbus::tree::PropInfo;
+
+use uuid::Uuid;
+
+use super::super::engine::BlockDev;
+use super::super::engine::types::BlockDevState;
+
+use super::types::{DbusContext, DbusErrorEnum, OPContext, TData};
+
+use super::util::STRATIS_BASE_PATH;
+use super::util::STRATIS_BASE_SERVICE;
+use super::util::code_to_message_items;
+use super::util::get_next_arg;
+use super::util::get_parent;
+use super::util::get_uuid;
+use super::util::ok_message_items;
+
+
+pub fn create_dbus_blockdev<'a>(dbus_context: &DbusContext,
+                                parent: dbus::Path<'static>,
+                                uuid: Uuid)
+                                -> dbus::Path<'a> {
+    let f = Factory::new_fn();
+
+    let set_userid_method = f.method("SetUserId", (), set_user_id)
+        .in_arg(("id", "s"))
+        .out_arg(("changed", "b"))
+        .out_arg(("return_code", "q"))
+        .out_arg(("return_string", "s"));
+
+    let devnode_property = f.property::<&str, _>("Devnode", ())
+        .access(Access::Read)
+        .emits_changed(EmitsChangedSignal::Const)
+        .on_get(get_blockdev_devnode);
+
+    let hardware_id_property = f.property::<&str, _>("HardwareId", ())
+        .access(Access::Read)
+        .emits_changed(EmitsChangedSignal::Const)
+        .on_get(get_blockdev_hardware_id);
+
+    let user_id_property = f.property::<&str, _>("UserId", ())
+        .access(Access::Read)
+        .emits_changed(EmitsChangedSignal::Const)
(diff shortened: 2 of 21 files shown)


mulkieran:
Suggest moving "Rename BlockDev to StratBlockDev" to separate PR for faster merging.

mulkieran:
"Add a "blockdevs" method..." commit looks good to me.

trgill:
Looks good to me... we might want to consider adding some tests to Jenkins that use busctl and the CLI.

src/engine/strat_engine/blockdev.rs
```diff
@@ -7,33 +7,44 @@
...
 
 
 #[derive(Debug)]
-pub struct BlockDev {
+pub struct StratBlockDev {
     dev: Device,
     pub devnode: PathBuf,
     bda: BDA,
     used: RangeAllocator,
+    pub location: Option<String>,
```
mulkieran: These can and should be private.

mulkieran: ```location``` is probably a bad name for this field and ```user_id``` might not be so good for accessor method. How about ```user_supplied_string```... that's descriptive.

src/engine/strat_engine/blockdev.rs
```diff
@@ -7,33 +7,44 @@
...
+    pub disk_id: Option<String>,
```
mulkieran: ```disk_id``` -> ```hardware_id```. Accessor method name is more precise.

src/engine/engine.rs
```diff
@@ -20,13 +20,23 @@ pub trait HasName: Debug {
...
 }
 
+pub trait BlockDev: HasUuid {
+    /// path of the device node
+    fn devnode(&self) -> PathBuf;
+
+    fn user_id(&self) -> &Option<String>;
+    fn set_user_id(&mut self, location: Option<&str>) -> EngineResult<()>;
+    fn hardware_id(&self) -> &Option<String>;
+    fn initialization_time(&self) -> String;
```
mulkieran: I'ld rather this be a TimeStamp...let D-Bus API layer convert it to its preferred format.

src/engine/sim_engine/blockdev.rs
```diff
@@ -16,16 +25,60 @@ use super::randomization::Randomizer;
...
+    fn set_user_id(&mut self, location: Option<&str>) -> EngineResult<()> {
+        self.location = location.map(|x| x.to_owned());
+        Ok(())
+    }
+
+    fn hardware_id(&self) -> &Option<String> {
+        &self.disk_id
+    }
+
+    fn initialization_time(&self) -> String {
```
mulkieran: See comment on trait about this method.

src/engine/strat_engine/blockdev.rs
```diff
@@ -112,10 +123,57 @@ impl BlockDev {
...
+    fn set_user_id(&mut self, location: Option<&str>) -> EngineResult<()> {
+        self.location = location.map(|x| x.to_owned());
+        Ok(())
+    }
+
+    fn hardware_id(&self) -> &Option<String> {
+        &self.disk_id
+    }
+
+    fn initialization_time(&self) -> String {
```
mulkieran: See comment on trait about this method.

src/engine/strat_engine/blockdev.rs
```diff
@@ -112,10 +123,57 @@ impl BlockDev {
...
+        Utc.timestamp(self.initialization_time() as i64, 0)
+            .to_rfc3339()
+    }
+
+    fn total_size(&self) -> Sectors {
+        self.bda.dev_size()
+    }
+
+    fn state(&self) -> BlockDevState {
+        BlockDevState::InUse
```
mulkieran: This can't be right. Doesn't it need a TODO or explanation of some sort?

src/engine/strat_engine/blockdev.rs
```diff
@@ -112,10 +123,57 @@ impl BlockDev {
...
+        &self.disk_id
+    }
+
+    fn initialization_time(&self) -> String {
+        Utc.timestamp(self.initialization_time() as i64, 0)
+            .to_rfc3339()
+    }
+
+    fn total_size(&self) -> Sectors {
+        self.bda.dev_size()
```
mulkieran: Probably you want ```current_capacity()``` here.

mulkieran:
P.S. In StratBlockDev implementation you have ended up w/ two ```initialization_time()``` methods. That is almost certainly one too many.

src/engine/strat_engine/blockdev.rs
agrover: Well, we (meaning..me) has let a little inconsistency arise between the design doc and the DBus API doc. Design doc calls them `location` and `disk_id`, API doc calls them `UserId` and `HardwareId`. I'm using the design doc terms here. Would you be in favor of adopting the API doc's terms in the design doc? Or some other terms. Consistency is good. Naming is hard. 🙁

src/engine/strat_engine/blockdev.rs
agrover: ok

src/engine/engine.rs
agrover: You mean a `DateTime<Utc>`? Makes sense.

src/engine/strat_engine/blockdev.rs
agrover: ok

src/engine/strat_engine/blockdev.rs
agrover: ok

agrover:
Rebased.

* change initialization_time to return a `DateTime<Utc>`
* Add some comments to BlockDev trait
* Add TODO in strat_engine::BlockDev::state()
* Remove extra initialization_time fn
* Make location and disk_id fields not pub
* fix `total_size` to call `current_capacity`

Not yet fixed:
* settling on universal names for location and disk_id.

src/dbus_api/blockdev.rs
```diff
@@ -0,0 +1,285 @@
...
+
+fn get_blockdev_physical_size(i: &mut IterAppend,
+                              p: &PropInfo<MTFn<TData>, TData>)
+                              -> Result<(), MethodErr> {
+    get_blockdev_property(i,
+                          p,
+                          |p| Ok(MessageItem::Str(format!("{}", *p.total_size()))))
+}
+
+fn get_blockdev_state(i: &mut IterAppend,
```
mulkieran: This looks like it might be done better if done in the same way as the error values and redundancy values, i.e., codes are exported as part of the Manager API and the state itself is exported as a number. The most obvious example is Redundancy.

Defining BlockDevState and exporting codes could be put in a separate PR, for quicker merging.

agrover:
rebased.

src/engine/engine.rs
```diff
@@ -26,8 +27,26 @@ pub trait Filesystem: HasName + HasUuid {
 }
 
 pub trait BlockDev: HasUuid {
-    /// path of the device node
+    /// Get the path of the device node for this device.
```
mulkieran: Reiterating concern about ```user_id``` name. It does not seem to be an id, properly speaking.

src/engine/engine.rs
```diff
@@ -26,8 +27,26 @@ pub trait Filesystem: HasName + HasUuid {
...
     fn devnode(&self) -> PathBuf;
+
+    /// Get the user-settable string associated with this blockdev.
+    fn user_id(&self) -> &Option<String>;
+
+    /// Set the user-settable string associated with this blockdev.
+    fn set_user_id(&mut self, user_id: Option<&str>) -> EngineResult<()>;
+
+    /// Get the hardware ID for this blockdev.
+    fn hardware_id(&self) -> &Option<String>;
```
mulkieran: Should there be multiple hardware ids, should this have type ```Vec<String>```, or something richer?

src/engine/engine.rs
```diff
@@ -44,7 +63,7 @@ pub trait Pool: HasName + HasUuid {
     /// Returns a list of device nodes corresponding to devices actually added.
     /// Returns an error if a blockdev can not be added because it is owned
     /// or there was an error while reading or writing a blockdev.
-    fn add_blockdevs(&mut self, paths: &[&Path], force: bool) -> EngineResult<Vec<PathBuf>>;
+    fn add_blockdevs(&mut self, paths: &[&Path], force: bool) -> EngineResult<Vec<DevUuid>>;
```
mulkieran: This seems like a good change...header comment needs updating.

src/engine/engine.rs
```diff
@@ -109,7 +137,7 @@ pub trait Engine: Debug {
                    blockdev_paths: &[&Path],
                    redundancy: Option<u16>,
                    force: bool)
-                   -> EngineResult<(PoolUuid, Vec<PathBuf>)>;
+                   -> EngineResult<PoolUuid>;
```
mulkieran: This change seems good...header comment needs updating.

src/engine/strat_engine/blockdev.rs
```diff
@@ -26,19 +26,25 @@ pub struct StratBlockDev {
     pub devnode: PathBuf,
     bda: BDA,
     used: RangeAllocator,
+    location: Option<String>,
+    disk_id: Option<String>,
```
mulkieran: Reiterating concern re naming of attributes. ```disk_id``` is actually ```hardware_id``` while ```location``` is actually ```user_id````? Hard to know what is what.

src/engine/engine.rs
```diff
@@ -26,8 +27,26 @@ pub trait Filesystem: HasName + HasUuid {
...
 
 pub trait BlockDev: HasUuid {
-    /// path of the device node
+    /// Get the path of the device node for this device.
     fn devnode(&self) -> PathBuf;
+
+    /// Get the user-settable string associated with this blockdev.
+    fn user_id(&self) -> &Option<String>;
+
+    /// Set the user-settable string associated with this blockdev.
```
mulkieran: I don't understand why the only argument is an Option type.

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -66,14 +66,17 @@ pub fn map_to_dm(bsegs: &[BlkDevSegment]) -> Vec<Segment> {
 
 #[derive(Debug)]
 pub struct BlockDevMgr {
-    block_devs: Vec<StratBlockDev>,
+    block_devs: HashMap<DevUuid, StratBlockDev>,
```
mulkieran: This is almost certainly a bad idea. Better to stick w/ Vec until we get a purpose built structure.

src/engine/strat_engine/blockdev.rs
```diff
@@ -129,10 +135,40 @@ impl BlockDev for StratBlockDev {
...
+    fn hardware_id(&self) -> &Option<String> {
+        &self.disk_id
+    }
+
+    fn initialization_time(&self) -> DateTime<Utc> {
+        Utc.timestamp(self.bda.initialization_time() as i64, 0)
+    }
+
+    fn total_size(&self) -> Sectors {
+        self.current_capacity()
```
mulkieran: Whoops! I think what this really ought to be is:

```
self.avail_range().1
```

src/dbus_api/blockdev.rs
agrover: I think this comment is superseded by the discussions we've had more recently about removing the other exported codes, and relying on the API spec.

src/engine/engine.rs
agrover: Trying to make it possible to unset the value.

src/engine/engine.rs
agrover: trying to initially keep it simple. I'm not sure how far we want to go for this feature, given it is not a critical one.

src/engine/strat_engine/blockdevmgr.rs
agrover: HashMap seems to be a step in the right direction. What other container would be better? `Table` is meant for things needing lookups by multiple values, but blockdevs don't have names, just uuids.

agrover:
Rebased.
* Fix a comment
* fix `total_size` implementation

Remaining issues:
* should blockdev states be exposed as strings or enumerated values
* settling on universal names for location and disk_id.
* container used for blockdevs in blockdevmgr

src/engine/strat_engine/blockdevmgr.rs
mulkieran: They have device numbers and device nodes as well.

src/dbus_api/blockdev.rs
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    get_blockdev_property(i, p, get_hardware_id)
+}
+
+fn get_blockdev_user_id(i: &mut IterAppend,
+                        p: &PropInfo<MTFn<TData>, TData>)
+                        -> Result<(), MethodErr> {
+    fn get_user_id(blockdev: &BlockDev) -> Result<MessageItem, MethodErr> {
+        match *blockdev.user_id() {
```
mulkieran: You could be so much more rustic here, something like:

```Ok(MessageItem::Str(*blockdev.user_id().or("").map(|x| x.to_owned())))```

File: src/dbus_api/blockdev.rs [hunk 18]
```diff
@@ -0,0 +1,285 @@
...
+    get_blockdev_property(i,
+                          p,
+                          |p| Ok(MessageItem::Str(format!("{}", p.devnode().display()))))
+}
+
+fn get_blockdev_hardware_id(i: &mut IterAppend,
+                            p: &PropInfo<MTFn<TData>, TData>)
+                            -> Result<(), MethodErr> {
+    fn get_hardware_id(blockdev: &BlockDev) -> Result<MessageItem, MethodErr> {
+        match *blockdev.hardware_id() {
```
mulkieran: Same comment as below.

src/dbus_api/blockdev.rs
mulkieran: You could be so much more rustic here, something like:

```Ok(MessageItem::Str(*blockdev.user_id().or("").map(|x| x.to_owned())))```

File: src/dbus_api/blockdev.rs [hunk 19]
```diff
@@ -0,0 +1,285 @@
...
+        .access(Access::Read)
+        .emits_changed(EmitsChangedSignal::Const)
+        .on_get(get_blockdev_initialization_time);
+
+    let total_physical_size_property = f.property::<&str, _>("TotalPhysicalSize", ())
+        .access(Access::Read)
+        .emits_changed(EmitsChangedSignal::False)
+        .on_get(get_blockdev_physical_size);
+
+    let state_property = f.property::<&str, _>("State", ())
```
mulkieran: I think it would be nicer if this was numeric.

src/dbus_api/blockdev.rs
```diff
@@ -0,0 +1,285 @@
...
+        .on_get(get_blockdev_devnode);
+
+    let hardware_id_property = f.property::<&str, _>("HardwareId", ())
+        .access(Access::Read)
+        .emits_changed(EmitsChangedSignal::Const)
+        .on_get(get_blockdev_hardware_id);
+
+    let user_id_property = f.property::<&str, _>("UserId", ())
+        .access(Access::Read)
+        .emits_changed(EmitsChangedSignal::Const)
```
mulkieran: Should be ```EmitsChangedSignal::False``` because it can change.

src/dbus_api/blockdev.rs
```diff
@@ -0,0 +1,285 @@
...
+                      -> Result<(), MethodErr> {
+    fn get_state(blockdev: &BlockDev) -> Result<MessageItem, MethodErr> {
+        let state = match blockdev.state() {
+            BlockDevState::Missing => "Missing".into(),
+            BlockDevState::Bad => "Bad".into(),
+            BlockDevState::Spare => "Spare".into(),
+            BlockDevState::NotInUse => "Not-in-use".into(),
+            BlockDevState::InUse => "In-use".into(),
+        };
+        Ok(MessageItem::Str(state))
```
mulkieran: Similar to above, would be nice if this were a number, not a string.

src/dbus_api/blockdev.rs
```diff
@@ -0,0 +1,285 @@
...
+            .set_user_id(Some(new_id))
+            .map_err(|err| {
+                         MethodErr::failed(&format!("set_user_id failed for object path {}: {}",
+                                                    object_path,
+                                                    err))
+                     })?;
+        id_changed
+    };
+
+    pool.save_state()
```
mulkieran: I think the engine should be doing the deciding about whether or not the pool needs to write out metadata.

src/engine/engine.rs
agrover: ok. let's try `user_info` and `hardware_info`.

src/engine/engine.rs
agrover: ok

src/engine/engine.rs
agrover: ok

src/engine/strat_engine/blockdevmgr.rs
agrover: we're not doing lookups on those at this point.

src/engine/strat_engine/blockdev.rs
agrover: ok

src/dbus_api/blockdev.rs
agrover: ok

src/dbus_api/blockdev.rs
agrover: this is a much bigger issue, e.g. see #583. I'm thinking I should leave this and we can replace it when we have something better in place.

src/dbus_api/blockdev.rs
agrover: cool. This also led to changing `user/hardware_info()` to return `Option<&str>` instead of `&Option<String>`, which is more convenient all around.

src/dbus_api/blockdev.rs
agrover: ok. I'll need to update the api doc.

agrover:
rebased.

* rename user_id and hardware_id to user_info and hardware_info
* Make blockdev state an integer instead of a string
* Fix some outdated comments
* Make user_info() and hardware_info() return Option<&str> instead of &Option<String>

remaining issues:

none?

agrover:
rebased.

src/dbus_api/blockdev.rs
```diff
@@ -0,0 +1,275 @@
...
+        .access(Access::Read)
+        .emits_changed(EmitsChangedSignal::Const)
+        .on_get(get_blockdev_devnode);
+
+    let hardware_info_property = f.property::<&str, _>("HardwareId", ())
+        .access(Access::Read)
+        .emits_changed(EmitsChangedSignal::Const)
+        .on_get(get_blockdev_hardware_info);
+
+    let user_info_property = f.property::<&str, _>("UserId", ())
```
mulkieran: ```UserNote```, ```UserInfo``` ?

src/dbus_api/blockdev.rs
mulkieran: Ok. Please leave a FIXME to mark the spot.

src/dbus_api/blockdev.rs
mulkieran: That's correct. The method looks good to me.

src/dbus_api/api.rs
```diff
@@ -56,27 +57,31 @@ fn create_pool(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
...
-                .iter()
-                .map(|d| {
-                         d.to_str()
-                             .expect("'d' originated in the 'devs' D-Bus argument")
-                             .into()
-                     });
-            let paths = paths.map(MessageItem::Str).collect();
+                create_dbus_pool(dbus_context, object_path.clone(), pool_uuid);
+            let return_sig = "(oao)";
+            let default_return = MessageItem::Array(vec![], return_sig.into());
```
mulkieran: I don't think this is correct default return. It's an array with items with type ```(oao)```. It should be more like:
```MessageItem::Struct(vec![MessageItem::Array(vec![], ...```.

File: src/dbus_api/pool.rs [hunk 25]
```diff
@@ -171,25 +172,31 @@ fn add_devs(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
...
-    let msg = match pool.add_blockdevs(&blockdevs, force) {
-        Ok(devnodes) => {
-            let paths = devnodes
+    let result = pool.add_blockdevs(&blockdevs, force);
+    let msg = match result {
+        Ok(ref infos) => {
+            let mut return_value = Vec::new();
+            for uuid in infos {
+                let bd_object_path: dbus::Path =
+                    create_dbus_blockdev(dbus_context, object_path.clone(), *uuid);
```
mulkieran: Might as well be consistent w/ above, wrap ```create_dbus_blockdev...``` w/ ```MessageItem::ObjectPath``` call as done in api.rs and drop line that starts at 185.

src/dbus_api/api.rs
```diff
@@ -270,6 +275,9 @@ pub fn connect(engine: Rc<RefCell<Engine>>)
         for fs_uuid in pool.filesystems().iter().map(|f| *f.uuid()) {
             create_dbus_filesystem(&dbus_context, pool_path.clone(), fs_uuid);
         }
+        for dev_uuid in pool.blockdevs().iter().map(|bd| *bd.uuid()) {
```
mulkieran: Might as well update comment above as it is necessary to hold for ```create_dbus_blockdev``` as well as others.

src/dbus_api/blockdev.rs
agrover: oops, yes.

src/dbus_api/pool.rs
agrover: yep, ok.

src/dbus_api/api.rs
agrover: ok

src/dbus_api/api.rs
agrover: ok I think I got this right, please take a look.

agrover:
* Fix default return construction in dbus create_pool()
* Fix comment in connect()
* Change property names to UserInfo and HardwareInfo
* Add FIXME for saving state
* Streamline code in add_devs

src/dbus_api/blockdev.rs
```diff
@@ -0,0 +1,276 @@
...
+                 .add_p(state_property)
+                 .add_p(user_info_property)
+                 .add_p(uuid_property));
+
+    let path = object_path.get_name().to_owned();
+    dbus_context.actions.borrow_mut().push_add(object_path);
+    path
+}
+
+fn set_user_info(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
```
mulkieran: How could this method ever allow setting user info to None. Is that desired?

src/dbus_api/blockdev.rs
```diff
@@ -0,0 +1,276 @@
...
+            .map_err(|err| {
+                         MethodErr::failed(&format!("set_user_info failed for object path {}: {}",
+                                                    object_path,
+                                                    err))
+                     })?;
+        id_changed
+    };
+
+    // FIXME: engine should decide to save state, not this function
+    pool.save_state()
```
mulkieran: I know this has a FIXME, but I still think that saving state should be conditional on something actually having changed.

src/dbus_api/blockdev.rs
```diff
@@ -0,0 +1,276 @@
...
+        let blockdev =
+            pool.get_mut_blockdev(blockdev_uuid)
+                .ok_or_else(|| {
+                                MethodErr::failed(&format!("no blockdev with uuid {}",
+                                                           &blockdev_uuid))
+                            })?;
+
+        let id_changed = blockdev.user_info() != Some(new_id);
+
+        blockdev
```
mulkieran: I think that, perhaps, this should be conditional on equality of new info w/ current info.

src/dbus_api/blockdev.rs
```diff
@@ -0,0 +1,276 @@
...
+
+fn get_blockdev_user_info(i: &mut IterAppend,
+                          p: &PropInfo<MTFn<TData>, TData>)
+                          -> Result<(), MethodErr> {
+    get_blockdev_property(i,
+                          p,
+                          |p| Ok(MessageItem::Str(p.user_info().unwrap_or("").to_owned())))
+}
+
+fn get_blockdev_initialization_time(i: &mut IterAppend,
```
mulkieran: Maybe initialization time should actually just be a number, and client should do any formatting.

src/dbus_api/api.rs
```diff
@@ -56,27 +57,31 @@ fn create_pool(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
...
-                     });
-            let paths = paths.map(MessageItem::Str).collect();
+                create_dbus_pool(dbus_context, object_path.clone(), pool_uuid);
+            let default_return =
+                MessageItem::Struct(vec![MessageItem::ObjectPath(default_object_path()),
+                                         MessageItem::Array(vec![], "o".into())]);
+            let pool_uuid_ref = &pool_uuid;
+            let pool = get_mut_pool!(engine; pool_uuid_ref; default_return; return_message);
+
+            let mut bd_object_paths = Vec::new();
```
mulkieran: Apparently, you can substitute a map() for this for loop, like you do down below.

src/dbus_api/blockdev.rs
agrover: Well I think so. The user may attach user info to a device, and then later want to remove what they added. Of course they could set it to an empty string but that's not ideal. Is there a way we can make it possible to unset the info using this method? Or do we unfortunately need an `clear_user_info` method as well?

src/dbus_api/blockdev.rs
agrover: oh, ok.

src/dbus_api/blockdev.rs
agrover: I'll put it in the conditional block.

src/dbus_api/blockdev.rs
agrover: Hmm, in `get_blockdev_user_info` we convert None to Some(""). This suggests another alternative would be doing the opposite in `set`. I guess my main goal is to avoid having `user_info = ""` in the on-disk json.

src/dbus_api/blockdev.rs
agrover: seconds since the epoch? Fine with me.

src/dbus_api/api.rs
agrover: ok will do.

agrover:
rebased.

* Change BlockDev::set_user_info() to never fail
* Change passing "" to set_user_info to be converted to a None when calling BlockDev::set_user_info()
* Make calling pool.save_state() conditional on if anything changed
* Change InitializationTime property to be seconds since epoch instead of string
* Use collect() to build object path list in create_pool()

src/dbus_api/blockdev.rs
```diff
@@ -0,0 +1,274 @@
...
+    let id_changed = {
+        let blockdev =
+            pool.get_mut_blockdev(blockdev_uuid)
+                .ok_or_else(|| {
+                                MethodErr::failed(&format!("no blockdev with uuid {}",
+                                                           &blockdev_uuid))
+                            })?;
+
+        let id_changed = blockdev.user_info() != new_id;
+        blockdev.set_user_info(new_id);
```
mulkieran: Still dunno why setting, even if user info not changed.

src/engine/sim_engine/blockdev.rs
```diff
@@ -19,13 +22,40 @@ use super::randomization::Randomizer;
...
+
+    fn set_user_info(&mut self, user_info: Option<&str>) -> () {
+        self.user_info = user_info.map(|x| x.to_owned());
+    }
+
+    fn hardware_info(&self) -> Option<&str> {
+        self.hardware_info.as_ref().map(|x| &**x)
+    }
+
+    fn initialization_time(&self) -> DateTime<Utc> {
```
mulkieran: Probably ```initialization_time``` result should just be u64. Otherwise, it seems like there's some pointless round-tripping to take a date time, change into a timestamp, and then retrieve it and turn it back into a date time.

src/dbus_api/blockdev.rs
agrover: it's harmless since the operation is idempotent. Avoiding conditionals makes code simpler.

src/engine/sim_engine/blockdev.rs
agrover: I think the type is useful to understand what it is when reading the code, and converting to a standard format internally ensures we don't have hidden dependencies between our on-disk format and our API.

src/engine/strat_engine/setup.rs
```diff
@@ -198,7 +198,18 @@ pub fn get_blockdevs(pool_uuid: PoolUuid,
                     RangeAllocator::new(actual_size,
                                         segment_table.get(bda.dev_uuid()).unwrap_or(&vec![]))?;
 
-                blockdevs.push(StratBlockDev::new(*device, devnode.to_owned(), bda, allocator));
+                let (location, disk_id) = pool_save
+                    .block_devs
+                    .get(bda.dev_uuid())
+                    .map(|bd_save| (bd_save.user_info.clone(), bd_save.hardware_info.clone()))
```
mulkieran: This should be an error, not an ```unwrap_or()```, because it would mean there is no entry for the blockdev in the pools metadata.

agrover:
fixed get_blockdevs.

src/engine/strat_engine/blockdev.rs
```diff
@@ -129,10 +135,39 @@ impl BlockDev for StratBlockDev {
...
+        // TODO: Implement states for blockdevs
+        BlockDevState::InUse
+    }
 }
 
 impl Recordable<BlockDevSave> for StratBlockDev {
     fn record(&self) -> BlockDevSave {
-        BlockDevSave { devnode: self.devnode.clone() }
+        BlockDevSave {
+            dev: Some(self.devnode.clone()),
```
mulkieran: If we really might not have a devnode to record, then it seems like BlockDev's devnode field should be an Option type, like the other two fields that get recorded.

src/engine/strat_engine/setup.rs
```diff
@@ -198,7 +198,21 @@ pub fn get_blockdevs(pool_uuid: PoolUuid,
                     RangeAllocator::new(actual_size,
                                         segment_table.get(bda.dev_uuid()).unwrap_or(&vec![]))?;
 
-                blockdevs.push(StratBlockDev::new(*device, devnode.to_owned(), bda, allocator));
+                let bd_save = pool_save
```
mulkieran: This is the only place where we have ready access to the blockdev's former devnode, if different from current. Is that worth recording? It's pretty evanescent, because once the pool metadata has been written, access to the prior devnode will be lost for good.

src/engine/strat_engine/blockdev.rs
```diff
@@ -129,10 +135,39 @@ impl BlockDev for StratBlockDev {
...
+    fn set_user_info(&mut self, user_info: Option<&str>) -> () {
+        self.user_info = user_info.map(|x| x.to_owned());
+    }
+
+    fn hardware_info(&self) -> Option<&str> {
+        self.hardware_info.as_ref().map(|x| &**x)
+    }
+
+    fn initialization_time(&self) -> DateTime<Utc> {
+        Utc.timestamp(self.bda.initialization_time() as i64, 0)
```
mulkieran: Please add:

``` // This cast will result in an incorrect, negative value starting in the year 292,277,026,596. ```

Comment (agrover) [reply src/engine/strat_engine/blockdev.rs]: ok, but I'm also adding a smiley!!!

File: src/engine/strat_engine/blockdev.rs (hunk 35 above)
Comment (agrover) [reply src/engine/strat_engine/blockdev.rs]: I think we will always have a devnode in StratBlockDev. I would flip it around and ask: why have devnode in BlockDevSave? It seems like it might be handy to know (maybe just informationally, to say "can't activate pool b/c blockdev [uuid] last seen at [devnode] is missing") but if it's not in the json then it shouldn't be an error because it's not used for anything critical, and maybe in the future we want to leave it out.

So this might be a spec issue worth further discussion around the value of it in the json, but as the spec is now I think this code is correct.

File: src/engine/strat_engine/blockdevmgr.rs [hunk 38]
```diff
@@ -91,7 +94,7 @@ impl BlockDevMgr {
     pub fn uuid_to_devno(&self) -> Box<Fn(&DevUuid) -> Option<Device>> {
         let uuid_map: HashMap<DevUuid, Device> = self.block_devs
             .iter()
-            .map(|bd| (*bd.uuid(), *bd.device()))
+            .map(|(_, bd)| (*bd.uuid(), *bd.device()))
```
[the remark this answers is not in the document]
agrover: agree.

src/engine/strat_engine/setup.rs
agrover: As mentioned in another comment, I think it's only interesting from an error-reporting perspective. (Or maybe, if we find two blockdevs with the same dev_uuid, we tie-break using former devnode? bleh.) I'm not sure former devnode is very useful, but I'm not sure it isn't, either.

agrover:
rebased.

* Add comment and change map in uuid_to_devno()

src/engine/strat_engine/blockdev.rs
mulkieran: I've filed an issue.

src/engine/sim_engine/blockdev.rs
```diff
@@ -19,13 +22,40 @@ use super::randomization::Randomizer;
 pub struct SimDev {
     pub devnode: PathBuf,
     rdm: Rc<RefCell<Randomizer>>,
-    uuid: Uuid,
+    pub uuid: Uuid,
```
mulkieran: This surely doesn't need to be public.

agrover: ok

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -204,24 +212,34 @@ impl BlockDevMgr {
     /// Get references to managed blockdevs.
     pub fn blockdevs(&self) -> Vec<&BlockDev> {
         self.block_devs
-            .iter()
+            .values()
             .map(|bd| bd as &BlockDev)
             .collect()
     }
 
+    pub fn get_blockdev_by_uuid(&self, uuid: &DevUuid) -> Option<&BlockDev> {
```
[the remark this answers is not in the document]
agrover: ok

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -204,24 +212,34 @@ impl BlockDevMgr {
...
+        self.block_devs.get(uuid).map(|bd| bd as &BlockDev)
+    }
+
+    pub fn get_mut_blockdev_by_uuid(&mut self, uuid: &DevUuid) -> Option<&mut BlockDev> {
```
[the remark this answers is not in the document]
agrover: ok

src/engine/strat_engine/serde_structs.rs
```diff
@@ -39,7 +39,12 @@ pub struct PoolSave {
 
 #[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct BlockDevSave {
-    pub devnode: PathBuf,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub dev: Option<PathBuf>,
```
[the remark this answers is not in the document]
agrover: hmm, ok

src/engine/engine.rs
```diff
@@ -94,22 +107,36 @@ pub trait Pool: HasName + HasUuid {
...
     fn blockdevs(&self) -> Vec<&BlockDev>;
+
+    /// Get the blockdev in this pool with this UUID.
+    fn get_blockdev(&self, uuid: &DevUuid) -> Option<&BlockDev>;
+
+    /// Get the mutable filesystem in this pool with this UUID.
+    fn get_mut_blockdev(&mut self, uuid: &DevUuid) -> Option<&mut BlockDev>;
+
+    /// Save the state of the pool.
+    fn save_state(&mut self) -> EngineResult<()>;
```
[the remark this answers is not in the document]
agrover: ok. I've opened #614 and reference it in the fixme.

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -335,7 +353,7 @@ pub fn initialize(pool_uuid: PoolUuid,
             let allocator = RangeAllocator::new(bda.dev_size(), &[(Sectors(0), bda.size())])
                 .expect("bda.size() < bda.dev_size() and single range");
 
-            bds.push(StratBlockDev::new(dev, devnode.to_owned(), bda, allocator));
+            bds.push(StratBlockDev::new(dev, devnode.to_owned(), bda, allocator, None, None));
```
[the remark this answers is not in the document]
agrover: FIXMEs are for broken or temporary things, right? I think this is just unimplemented. I've opened #615.

agrover:
* Fix up new code to use UUIDs by value
* Un-pub members of SimDev
* Rename BlockDevSave::dev to devnode
* Extend comment for BlockDev::initialization_time()

src/engine/engine.rs
```diff
@@ -26,8 +27,27 @@ pub trait Filesystem: HasName + HasUuid {
...
 
 pub trait BlockDev: HasUuid {
-    /// path of the device node
+    /// Get the path of the device node for this device.
     fn devnode(&self) -> PathBuf;
+
+    /// Get the user-settable string associated with this blockdev.
+    fn user_info(&self) -> Option<&str>;
+
+    /// Set the user-settable string associated with this blockdev.
```
mulkieran: Have this return a boolean if value was changed.

src/dbus_api/blockdev.rs
```diff
@@ -0,0 +1,274 @@
...
+    let pool = get_mut_pool!(engine; pool_uuid; default_return; return_message);
+
+    let id_changed = {
+        let blockdev = pool.get_mut_blockdev(blockdev_data.uuid)
+            .ok_or_else(|| {
+                            MethodErr::failed(&format!("no blockdev with uuid {}",
+                                                       blockdev_data.uuid))
+                        })?;
+
+        let id_changed = blockdev.user_info() != new_id;
```
mulkieran: If ```BlockDev::set_user_info``` is changed as suggested, then these three lines can just be:

```blockdev.set_user_info(new_id)```

File: src/dbus_api/blockdev.rs [hunk 47]
```diff
@@ -0,0 +1,274 @@
...
+        engine
+            .get_pool(pool_uuid)
+            .ok_or_else(|| {
+                            MethodErr::failed(&format!("no pool corresponding to uuid {}",
+                                                       &pool_uuid))
+                        })?;
+    let blockdev =
+        pool.get_blockdev(blockdev_data.uuid)
+            .ok_or_else(|| {
+                            MethodErr::failed(&format!("noblockdev with uuid {}",
```
mulkieran: Typo, needs a space.

agrover:
* change BlockDev::set_user_info to return bool
* Fix typo in error message
* Add TODO to get and use hardware info

src/engine/sim_engine/blockdev.rs
```diff
@@ -37,8 +37,13 @@ impl BlockDev for SimDev {
         self.user_info.as_ref().map(|x| &**x)
     }
 
-    fn set_user_info(&mut self, user_info: Option<&str>) -> () {
-        self.user_info = user_info.map(|x| x.to_owned());
+    fn set_user_info(&mut self, user_info: Option<&str>) -> bool {
```
mulkieran: Might as well macroize this and its twin.

mulkieran:
Depends on #619.

src/engine/sim_engine/blockdev.rs
agrover: I'd rather not, this is less than the amount of code I think warrants a macro.

mulkieran: It makes sense to make things that are shared by both the engines into macros, regardless of size. That way we know that they are shared. That is the main point of `src/engine/macros.rs`.

agrover: ok that makes sense to me, will do.

agrover:
rebase
* macroize set_user_info
* fix a clippy error

agrover:
@trgill review?

agrover:
👏 👏 👏 👏 👏 👏 👏 👏  much thanks!!!

agrover:
This was undone manually b/c I accidentally merged before squashing. #622 is corrected (squashed) equivalent.

Issue #309, opened 2017-04-18
Excellent Logging


agrover:
We have basic logging now but it makes sense at some point to devote additional thought to how to make our logging excellent. This is because logs are the primary way we have to understand failures that occur on our user's systems, and we want to be able to fix these issues without requiring them to reproduce them, since this may be very hard to do and an additional burden to place on them.

We'd like
* Minimal (no?) logging while operating normally
* Comprehensive logging and view of system state when something goes wrong, that the customer can send in, possibly via sosreport or similar.
* Minimal runtime space/cpu overhead
* A policy for instrumenting the code consistently with log output statements

mulkieran:
I'm still hoping for an agreement to leave logging statements out of commits in both devicemapper and stratisd until such time as we figure this out.

mulkieran:
One thing worth logging is a bad thing that we don't return an error for. This would happen if we could proceed with our task, even when we encounter an error. A good example is when reading metadata from a bunch of devices, the read of metadata from a device may fail. That is an error, but we don't want the method to fail, because it can try another device for the metadata. But the fact that the read failed for some device is definitely worth a warning.

agrover:
idea: we don't want to spew all the time, but it would be nice if something went wrong, we had a "black box" that recorded info that would help us figure out what happened.

mulkieran:
Another point is that logs entries in themselves are timestamped and have a syntactic location. Errors generated w/ error chain give an incomplete view of the stack. Together, these might be quite informative.

agrover:
current plan for an initial implementation:
* create a buffered logging implementation (keep env_logger when --debug)
* Log periodic complete state dumps to the logger using Debug trait
* Log other relevant info that is not captured by state changes
* on toplevel error, panic, or other failure event, trigger the logger. Requires using `std::panic::set_hook`.

mulkieran:
Possibly useful: https://crates.io/crates/loggy.

PR #568, opened 2017-09-14
MDV: Do not keep MDV mounted

### LINKED ISSUE #358: MDVs are mounted all the time
Right now we mount the MDV for each pool under `/dev/stratis`. We really could avoid this, if we wanted to -- only mounting it when we needed to read/write from it.

The only issue would be we'd probably want to do some sort of lazy unmounting, to prevent bad behavior when lots was going on.

agrover:
Instead of mounting the MDV when MetadataVol is created and unmounting
in teardown, only mount in each method for as long as necessary.

Add a `with_mount` function that takes a closure, and executes it with the
MDV mounted, and then unmounts afterwards.

Remove the check for EBUSY, since this should never happen.

fixes #358

### DIFF
--- a/src/engine/strat_engine/mdv.rs
+++ b/src/engine/strat_engine/mdv.rs
@@ -10,7 +10,7 @@ use std::fs::{create_dir, OpenOptions, read_dir, remove_file, rename};
 use std::io::ErrorKind;
 use std::io::prelude::*;
 use std::os::unix::io::AsRawFd;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use nix;
 use nix::mount::{MsFlags, mount, umount};
@@ -39,6 +39,45 @@ pub struct MetadataVol {
     mount_pt: PathBuf,
 }
 
+/// A helper struct that borrows the MetadataVol and ensures that the MDV is
+/// mounted as long as it is borrowed, and then unmounted.
+#[derive(Debug)]
+struct MountedMDV<'a> {
+    mdv: &'a MetadataVol,
+}
+
+impl<'a> MountedMDV<'a> {
+    /// Borrow the MDV and ensure it's mounted.
+    fn mount(mdv: &MetadataVol) -> EngineResult<MountedMDV> {
+        match mount(Some(&mdv.dev.devnode()),
+                    &mdv.mount_pt,
+                    Some("xfs"),
+                    MsFlags::empty(),
+                    None as Option<&str>) {
+            Err(nix::Error::Sys(nix::Errno::EBUSY)) => {
+                // The device is already mounted at the specified mountpoint
+                Ok(())
+            }
+            Err(err) => Err(err),
+            Ok(_) => Ok(()),
+        }?;
+
+        Ok(MountedMDV { mdv })
+    }
+
+    fn mount_pt(&self) -> &Path {
+        &self.mdv.mount_pt
+    }
+}
+
+impl<'a> Drop for MountedMDV<'a> {
+    fn drop(&mut self) {
+        if let Err(err) = umount(&self.mdv.mount_pt) {
+            warn!("Could not unmount MDV: {}", err)
+        }
+    }
+}
+
 impl MetadataVol {
     /// Initialize a new Metadata Volume.
     pub fn initialize(pool_uuid: &PoolUuid, dev: LinearDev) -> EngineResult<MetadataVol> {
@@ -63,44 +102,36 @@ impl MetadataVol {
             }
         }
 
-        if let Err(err) = mount(Some(&dev.devnode()),
-                                &mount_pt,
-                                Some("xfs"),
-                                MsFlags::empty(),
-                                None as Option<&str>) {
-            match err {
-                nix::Error::Sys(nix::Errno::EBUSY) => {
-                    // EBUSY is the error returned in this context when the
-                    // device is already mounted at the specified mountpoint.
+        let mdv = MetadataVol { dev, mount_pt };
+
+        {
+            let mount = MountedMDV::mount(&mdv)?;
+
+            if let Err(err) = create_dir(mount.mount_pt().join(FILESYSTEM_DIR)) {
+                if err.kind() != ErrorKind::AlreadyExists {
+                    return Err(From::from(err));
                 }
-                _ => return Err(From::from(err)),
             }
-        }
 
-        if let Err(err) = create_dir(&mount_pt.join(FILESYSTEM_DIR)) {
-            if err.kind() != ErrorKind::AlreadyExists {
-                return Err(From::from(err));
-            }
-        }
+            for dir_e in read_dir(mount.mount_pt().join(FILESYSTEM_DIR))? {
+                let dir_e = dir_e?;
 
-        for dir_e in read_dir(mount_pt.join(FILESYSTEM_DIR))? {
-            let dir_e = dir_e?;
-
-            // Clean up any lingering .temp files. These should only
-            // exist if there was a crash during save_fs().
-            if dir_e.path().ends_with(".temp") {
-                match remove_file(dir_e.path()) {
-                    Err(err) => {
-                        debug!("could not remove file {:?}: {}",
-                               dir_e.path(),
-                               err.description())
+                // Clean up any lingering .temp files. These should only
+                // exist if there was a crash during save_fs().
+                if dir_e.path().ends_with(".temp") {
+                    match remove_file(dir_e.path()) {
+                        Err(err) => {
+                            debug!("could not remove file {:?}: {}",
+                                   dir_e.path(),
+                                   err.description())
+                        }
+                        Ok(_) => debug!("Cleaning up temp file {:?}", dir_e.path()),
                     }
-                    Ok(_) => debug!("Cleaning up temp file {:?}", dir_e.path()),
                 }
             }
         }
 
-        Ok(MetadataVol { dev, mount_pt })
+        Ok(mdv)
     }
 
     /// Save info on a new filesystem to persistent storage, or update
@@ -116,6 +147,8 @@ impl MetadataVol {
             .with_extension("json");
 
         let temp_path = path.clone().with_extension("temp");
+
+        let _mount = MountedMDV::mount(self)?;
 
         // Braces to ensure f is closed before renaming
         {
@@ -141,6 +174,9 @@ impl MetadataVol {
             .join(FILESYSTEM_DIR)
             .join(fs_uuid.simple().to_string())
             .with_extension("json");
+
+        let _mount = MountedMDV::mount(self)?;
+
         if let Err(err) = remove_file(fs_path) {
             if err.kind() != ErrorKind::NotFound {
                 return Err(From::from(err));
@@ -154,7 +190,9 @@ impl MetadataVol {
     pub fn filesystems(&self) -> EngineResult<Vec<FilesystemSave>> {
         let mut filesystems = Vec::new();
 
-        for dir_e in read_dir(self.mount_pt.join(FILESYSTEM_DIR))? {
+        let mount = MountedMDV::mount(self)?;
+
+        for dir_e in read_dir(mount.mount_pt().join(FILESYSTEM_DIR))? {
             let dir_e = dir_e?;
 
             if dir_e.path().ends_with(".temp") {
@@ -173,7 +211,6 @@ impl MetadataVol {
 
     /// Tear down a Metadata Volume.
     pub fn teardown(self, dm: &DM) -> EngineResult<()> {
-        umount(&self.mount_pt)?;
         self.dev.teardown(dm)?;
 
         Ok(())
--- a/src/engine/strat_engine/mdv.rs
+++ b/src/engine/strat_engine/mdv.rs
@@ -83,6 +83,23 @@ impl MetadataVol {
             }
         }
 
+        for dir_e in read_dir(mount_pt.join(FILESYSTEM_DIR))? {
+            let dir_e = dir_e?;
+
+            // Clean up any lingering .temp files. These should only
+            // exist if there was a crash during save_fs().
+            if dir_e.path().ends_with(".temp") {
+                match remove_file(dir_e.path()) {
+                    Err(err) => {
+                        debug!("could not remove file {:?}: {}",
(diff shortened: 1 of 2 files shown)


agrover:
Yeah I see what you're saying. Updated.

src/engine/strat_engine/mdv.rs
```diff
@@ -124,33 +143,41 @@ impl MetadataVol {
...
-
-            // Clean up any lingering .temp files. These should only
-            // exist if there was a crash during save_fs().
-            if dir_e.path().ends_with(".temp") {
-                match remove_file(dir_e.path()) {
-                    Err(err) => {
-                        debug!("could not remove file {:?}: {}",
-                               dir_e.path(),
-                               err.description())
+        with_mount(&self.dev.devnode(), &self.mount_pt, || {
```
mulkieran: With current code, fs will be mounted every 10 seconds if nothing is happening, just to do housekeeping here. Does that seem too often?

src/engine/strat_engine/mdv.rs
```diff
@@ -159,26 +186,30 @@ impl MetadataVol {
...
+
+            }
+            Ok(())
+        })?;
 
         Ok(filesystems)
     }
 
     /// Tear down a Metadata Volume.
     pub fn teardown(self, dm: &DM) -> EngineResult<()> {
```
mulkieran: Should the ```umount``` here be kept? Does fs need to be unmounted for teardown to succeed?

src/engine/strat_engine/mdv.rs
```diff
@@ -33,6 +33,32 @@ const DEV_PATH: &'static str = "/dev/stratis";
 
 const FILESYSTEM_DIR: &'static str = "filesystems";
 
+fn with_mount<F, O>(dev: &Path, mount_point: &Path, op: F) -> EngineResult<O>
```
mulkieran: Needs a header comment.

src/engine/strat_engine/mdv.rs
```diff
@@ -33,6 +33,32 @@ const DEV_PATH: &'static str = "/dev/stratis";
...
+                // device is already mounted at the specified mountpoint.
+            }
+            _ => return Err(From::from(err)),
+        }
+    };
+
+    let res = op().map_err(|err| err.into());
+
+    umount(mount_point).unwrap_or_else(|err| {
+                                           error!("Could not unmount {:?}: {}", mount_point, err)
```
mulkieran: Failure to unmount may be an error for unmount, but it is definitely not an error for stratisd.

agrover: `warn!()`?

src/engine/strat_engine/mdv.rs
agrover: That `check` code is rather pointless. How about if I move the temp file removal to `setup` and we can just scrap `check`.

src/engine/strat_engine/mdv.rs
agrover: The assumption is that the MDV is unmounted between calls, so it should already be unmounted when teardown is called.

src/engine/strat_engine/mdv.rs
agrover: ok

agrover:
rebased. Made changes as recommended. Added "MDV: Remove check()" commit.

src/engine/strat_engine/mdv.rs
mulkieran: It might not be. The related PR talks about an optimization to sometimes not unmount if busy, and a previous unmount could have failed.

src/engine/strat_engine/mdv.rs
```diff
@@ -33,6 +33,31 @@ const DEV_PATH: &'static str = "/dev/stratis";
 
 const FILESYSTEM_DIR: &'static str = "filesystems";
 
+/// Mount the MDV, execute a method while it is mounted, then unmount the MDV.
+fn with_mount<F, O>(dev: &Path, mount_point: &Path, op: F) -> EngineResult<O>
+    where F: FnOnce() -> EngineResult<O>
```
mulkieran: Probably worthwhile to parameterize on the result type.

src/engine/strat_engine/mdv.rs
```diff
@@ -33,6 +33,31 @@ const DEV_PATH: &'static str = "/dev/stratis";
 
 const FILESYSTEM_DIR: &'static str = "filesystems";
 
+/// Mount the MDV, execute a method while it is mounted, then unmount the MDV.
```
mulkieran: Maybe better.

```
/// Executes op on the mounted MDV returning op's result.
/// Returns an error if the MDV can not be mounted at the specified mountpoint.
/// Does not return an error if unmount fails; failure to unmount does not prevent the specified operation succeeding.
```

src/engine/strat_engine/mdv.rs
```diff
@@ -33,6 +33,31 @@ const DEV_PATH: &'static str = "/dev/stratis";
...
+        match err {
+            nix::Error::Sys(nix::Errno::EBUSY) => {
+                // EBUSY is the error returned in this context when the
+                // device is already mounted at the specified mountpoint.
+            }
+            _ => return Err(From::from(err)),
+        }
+    };
+
+    let res = op().map_err(|err| err.into());
```
mulkieran: If parameterized on result type, maybe this can be just ```let res = op();```.

mulkieran: Actually, can't this just be ```let res = op();``` regardless of parameterization?

agrover:
rebased.

new approach: Mutably borrow the MDV in creating a `MountedMDV`, and in doing so ensure it it mounted as long as the borrow is alive, and then that it is unmounted when the borrow ends. Add an `is_mounted` flag to avoid unneeded calls to `mount` and `unmount`.

src/engine/strat_engine/mdv.rs
```diff
@@ -63,35 +108,48 @@ impl MetadataVol {
...
                 }
-                _ => return Err(From::from(err)),
             }
-        }
 
-        if let Err(err) = create_dir(&mount_pt.join(FILESYSTEM_DIR)) {
-            if err.kind() != ErrorKind::AlreadyExists {
-                return Err(From::from(err));
+            for dir_e in read_dir(mount.mount_pt().join(FILESYSTEM_DIR))? {
+                let dir_e = dir_e?;
```
mulkieran: This should get its own function, like:

(The comments are just an explanation of the reasoning).
```
/// Remove temp files from the designated directory.
/// Fails if the directory can not be read.
/// Persists if an individual directory entry can not be read due to an intermittent IO error.
/// Returns a list of error resulting from remove operations. Ideally, this list is empty.
// This informational result can be upgraded or altered as necessary.
fn remove_temp_files(dir) -> EngineResult<Vec<std::io::Error>> {
  let mut remaining = Vec::new();
  for path in read_dir(dir)?
    // If there's an intermittent IO error on one dir_entry it probably doesn't make sense to fail
    // Presumably it's ok to miss a file that may not be a ".temp" file anyway.
    .filter_map(|e| e.ok())
    // Get the path, pronto, it is what we care about
    .map(|e| e.path())
    // Be interested only in the .temp files
    .filter(|p| p.endswith(".temp")) {
      remove_file(path).unwrap_or_else(|e| remaining.push(e))
  } 
  Ok(remaining)
}
```

src/engine/strat_engine/mdv.rs
```diff
@@ -37,6 +37,51 @@ const FILESYSTEM_DIR: &'static str = "filesystems";
...
+                Ok(_) => Ok(()),
+            }?;
+
+            mdv.is_mounted = true;
+        }
+
+        Ok(MountedMDV { mdv })
+    }
+
+    fn mount_pt(&self) -> &Path {
```
mulkieran: I'm not sure this method serves any purpose.

src/engine/strat_engine/mdv.rs
```diff
@@ -63,35 +108,48 @@ impl MetadataVol {
...
-                                MsFlags::empty(),
-                                None as Option<&str>) {
-            match err {
-                nix::Error::Sys(nix::Errno::EBUSY) => {
-                    // EBUSY is the error returned in this context when the
-                    // device is already mounted at the specified mountpoint.
+        let mut mdv = MetadataVol {
+            dev,
+            mount_pt,
+            is_mounted: false,
```
mulkieran: I know we're using this in the code, but it's not clear to me that we're using it for anything useful.

src/engine/strat_engine/mdv.rs
```diff
@@ -37,6 +37,51 @@ const FILESYSTEM_DIR: &'static str = "filesystems";
...
+/// mounted as long as it is borrowed, and then unmounted.
+#[derive(Debug)]
+struct MountedMDV<'a> {
+    mdv: &'a mut MetadataVol,
+}
+
+impl<'a> MountedMDV<'a> {
+    /// Borrow the MDV and ensure it's mounted.
+    fn mount(mdv: &mut MetadataVol) -> EngineResult<MountedMDV> {
+        if !mdv.is_mounted {
```
mulkieran: We think that it might be possible for the volume to be mounted when we think it is not, but we do not allow for the possibility that the volume is unmounted when we think it is. Why is that?

My preference would be to omit ```is_mounted``` and just mount the thing here.

agrover: Ok.

src/engine/strat_engine/mdv.rs
agrover: once MountedMDV is created, the MetadataVol is no longer accessible, but some methods still need access to the mount point path.

src/engine/strat_engine/mdv.rs
```diff
@@ -37,6 +37,51 @@ const FILESYSTEM_DIR: &'static str = "filesystems";
...
+
+    fn mount_pt(&self) -> &Path {
+        &self.mdv.mount_pt
+    }
+}
+
+impl<'a> Drop for MountedMDV<'a> {
+    fn drop(&mut self) {
+        match umount(&self.mdv.mount_pt) {
+            Err(err) => warn!("Could not unmount MDV: {}", err),
```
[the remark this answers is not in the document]
agrover: If unmount fails, we can panic (which I don't think is appropriate), or print a message, but we can't return an error, and we shouldn't just silently let it go, because if unmount is failing then we need to know to find out why. We have #309 to improve this en masse, and have guidelines in the style guide, but until then what should we do here?

src/engine/strat_engine/mdv.rs
agrover: ok, removed..

src/engine/strat_engine/mdv.rs
```diff
@@ -63,35 +108,48 @@ impl MetadataVol {
...
+                    return Err(From::from(err));
                 }
-                _ => return Err(From::from(err)),
             }
-        }
 
-        if let Err(err) = create_dir(&mount_pt.join(FILESYSTEM_DIR)) {
-            if err.kind() != ErrorKind::AlreadyExists {
-                return Err(From::from(err));
+            for dir_e in read_dir(mount.mount_pt().join(FILESYSTEM_DIR))? {
```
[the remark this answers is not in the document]
agrover: there's a slight possibility of .temp files on the MDV, and we don't want them to hang around forever.

src/engine/strat_engine/mdv.rs
agrover: Maybe, but lets limit this PR to changes related to mounting and open another PR for that.

agrover:
rebased. Deleted `is_mounted` flag from `MetadataVol`.

src/engine/strat_engine/mdv.rs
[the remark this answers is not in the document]
mulkieran: That's not a compelling reason to risk failure of the whole method. Why not just delete this chunk and file an issue about it?

src/engine/strat_engine/mdv.rs
```diff
@@ -133,33 +186,13 @@ impl MetadataVol {
...
-                }
-            }
-        }
-
-        Ok(())
-    }
-
     /// Get list of filesystems stored on the MDV.
-    pub fn filesystems(&self) -> EngineResult<Vec<FilesystemSave>> {
+    pub fn filesystems(&mut self) -> EngineResult<Vec<FilesystemSave>> {
```
mulkieran: I think you can remove a bunch of muts, of which this one.

agrover:
hey cool. zap `mut`s.

src/engine/strat_engine/mdv.rs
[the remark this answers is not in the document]
agrover: Hey you're right but just as a matter of process I'd prefer if we left the code in this PR and then opened another issue/PR to fix it, rather than removing it here and then opening an issue to re-add it. This keeps this PR from having anything unrelated to MDV mount/unmount in it.

src/engine/strat_engine/mdv.rs
mulkieran: Ok. I've opened an issue and assigned it to myself: https://github.com/stratis-storage/stratisd/issues/579.

src/engine/strat_engine/mdv.rs
```diff
@@ -39,6 +39,45 @@ pub struct MetadataVol {
...
+impl<'a> MountedMDV<'a> {
+    /// Borrow the MDV and ensure it's mounted.
+    fn mount(mdv: &MetadataVol) -> EngineResult<MountedMDV> {
+        match mount(Some(&mdv.dev.devnode()),
+                    &mdv.mount_pt,
+                    Some("xfs"),
+                    MsFlags::empty(),
+                    None as Option<&str>) {
+            Err(nix::Error::Sys(nix::Errno::EBUSY)) => {
+                // We thought we weren't mounted but we actually were
```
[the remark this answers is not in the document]
agrover: ok

src/engine/strat_engine/mdv.rs
agrover: I think it's conceptually simpler to borrow the entire struct so I'm leaving this as is.

agrover:
updated comment. Please merge.

PR #573, opened 2017-09-20
Add a FIXME marker

mulkieran:
It would be wrong for Stratis to crash in this context.


mulkieran:
@trgill review, please?

PR #585, opened 2017-09-29
Update to latest nix and specify version

agrover:
Update library versions by running "cargo update". No issues, except that
nix's Error type now supports description().

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15,7 +15,7 @@ dependencies = [
  "libc 0.2.30 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "loopdev 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "nix 0.8.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "nix 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "quickcheck 0.4.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


mulkieran:
I think maybe we should consider whether it might be important for our required version to lag whatever is being packaged for Fedora. Not all Fedora packages stay totally up-to-date w/ package releases.

mulkieran:
@cuviper Do you have any thoughts or knowledge of best practices for package updating?

cuviper:
The short answer is that Fedora doesn't package any Rust crates yet.  Do what you like! 😉 
(There are actually a few on rawhide now, but this is my short answer, so wave those away...)

Longer term, it's harder to predict how we'll handle this.  Nix is the only semver-breaking change in this PR, so that's the only one I would really hesitate about updating in a Fedora release.  It's not really kosher that you have `nix = "0"` for your dependency, because there are probably breaking changes between any `0.x` and `0.y` -- `nix-0.8` to `nix-0.9` in this case.

The other updates here are all semver-compatible (in theory), so should be fair game.  When we do get full packaging in place, it's likely that we won't hold firm to every project's specific `Cargo.lock`, only their looser `Cargo.toml` requirements, so make sure that's right.

I'd say you should do your upstream development with the latest dependencies, and that's probably what you'll find in Rawhide packages too.  Older Fedora may or may not update those, possibly (hopefully) with compat packages in place.  If they don't update, you'll either have to not update Stratis either, or else patch it to work with older deps.  Enterprise distros amplify this dilemma, as usual...

mulkieran:
I like the nix change fine, but I'm worried that by updating in this way we will fail to notice when we need to update Cargo.toml, see: #588 .

agrover:
@mulkieran review/merge?

PR #586, opened 2017-10-01
Add more descriptive errors to test_linear_device

jsalzbergedu:
More descriptive errors on test_linear_device. Please let me know if this is a bad idea or if I can do anything to make it better.

### DIFF
--- a/tests/util/dm_tests.rs
+++ b/tests/util/dm_tests.rs
@@ -31,12 +31,12 @@ use libstratis::engine::strat_engine::metadata::MIN_MDA_SECTORS;
 /// Verify that the sum of the lengths of the available range of the
 /// blockdevs in the linear device equals the size of the linear device.
 pub fn test_linear_device(paths: &[&Path]) -> () {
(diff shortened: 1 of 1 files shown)


trgill:
Can one of the admins verify this patch?

mulkieran:
Thanks for the patch!

We have a rule, perhaps not yet in our style guide, the we use expect only for unwraps() that we believe that we can prove logically will always succeed.

An odd drawback of expects() is, that while you may gain an explanation you can lose other information, see discussion: https://github.com/rust-lang-nursery/rust-clippy/issues/1921.

What we're working on, which we hope will help, is adding the use of the error_chain module to both devicemapper and stratisd. Here's the PR in devicemapper: https://github.com/stratis-storage/devicemapper-rs/pull/163. Then,  when an unwrap fails, you'll be able to view the error, as now, the stack trace associated w/ that error, the error that caused the error that your viewing, the stack trace associated with that error, and so on.

PR #594, opened 2017-10-05
 Update devnode_to_devno() to return None for NotFound errors

trgill:
Update devnode_to_devno() to return None for NotFound errors

This error can happen if /dev/ has a sybolic link to a file that doesn't
exist or a file has gone away between reading the contents of /dev/ and
when we check the meta data.

### DIFF
--- a/src/engine/strat_engine/device.rs
+++ b/src/engine/strat_engine/device.rs
@@ -6,7 +6,7 @@
 
 use std::collections::HashMap;
 use std::fs::File;
-use std::io::{BufWriter, Seek, Write, SeekFrom};
+use std::io::{BufWriter, ErrorKind, Seek, SeekFrom, Write};
 use std::fs::OpenOptions;
 use std::os::linux::fs::MetadataExt;
 use std::os::unix::prelude::AsRawFd;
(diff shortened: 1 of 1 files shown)


mulkieran:
If /dev/ has a symbolic link to a file that doesn't exist, ```devnode_to_devno``` should return ```Ok(None)```.

trgill:
I'm not sure I understand your comment.

You are requesting I change devnode_to_devno() to return Ok(None) for a symbolic link to a file that doesn't exist?  Currently it returns an error because path.metadata() returns an error.

mulkieran:
Yes. If there _is_ no device number for the device node, returning None is the right thing to do.

trgill:
Ok.  Do you agree with the change to find_all() to continue in the case of an error from devnode_to_devno()?  I'm not sure if there are other cases.

agrover:
👍

mulkieran:
We should try to hold the line in ```find_all()``` itself. We don't want to fail to notice the fact that we have failed to get a device that might actually belong to stratisd.

trgill:
@mulkieran ok.  I've made the changes I think you've suggested.

PR #595, opened 2017-10-05
Have pool.check() return a Result

mulkieran:
Fixes a bug introduced in #537.

### DIFF
--- a/src/engine/macros.rs
+++ b/src/engine/macros.rs
@@ -91,7 +91,8 @@ macro_rules! rename_pool_pre {
 macro_rules! check_engine {
     ( $s:ident ) => {
         for pool in &mut $s.pools {
-            pool.check();
+            // FIXME: It is not really correct to ignore result of pool.check().
+            let _ = pool.check();
         }
     }
 }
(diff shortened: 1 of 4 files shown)


mulkieran:
With this change, suddenly all tests succeed.

mulkieran:
@agrover review?

agrover:
Isn't this masking that the pool check didn't succeed? Do we know why the check is failing?

mulkieran:
We have no evidence that the check was ever failing. ```<expr>.unwrap_or(error!("message"))``` prints "message" regardless of the value of ```<expr>```.

agrover:
I am seeing that message intermittently.

mulkieran:
It happens exactly when ```pool.check()``` runs and completes. So, if you're occasionally not seeing it, that must mean that pool.check() is not being run on occasion.

https://play.rust-lang.org/?gist=147aef4052459cdfd8a179c05025189c&version=stable

mulkieran:
rebased.

agrover:
oh yeah...`unwrap_or` always evaluates what's in the parens, I should've used `unwrap_or_else`!

mulkieran:
rebased.

PR #596, opened 2017-10-05
suspend running loopback tests on Travis

trgill:
Suspend the loopback tests until we either get access to newer xfsprogs
on the Travis VM or are able to run inside a container.

Stratis requires xfsprogs 4.2.0 or greater.  The -U <uuid> feature is
used to keep the Stratis UUID for the filesystem in sync with the UUID
used by XFS.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,9 +18,6 @@ matrix:
           env: TASK=docs
         - rust: stable
           env: TASK=test
-        - rust: stable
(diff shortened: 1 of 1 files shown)


mulkieran:
Please be specific in the commit about what version of xfsprogs we need and why, for future reference. Otherwise, looks good.

agrover:
👍

PR #598, opened 2017-10-05
Dm events

agrover:
pending dm-rs changes. WIP.

### DIFF
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -85,20 +85,51 @@ fn run() -> StratisResult<()> {
 
     let (dbus_conn, mut tree, dbus_context) = libstratis::dbus_api::connect(Rc::clone(&engine))?;
 
-    // Get a list of fds to poll for
-    let mut fds: Vec<_> = dbus_conn
-        .watch_fds()
-        .iter()
-        .map(|w| w.to_pollfd())
-        .collect();
+    let mut fds = Vec::new();
+
+    let mut eventable = engine.borrow_mut().get_eventable()?;
+    let engine_fds_end_idx = match eventable {
+        Some(ref mut evt) => {
+            fds.push(libc::pollfd {
+                         fd: evt.get_pollable_fd(),
+                         revents: 0,
+                         events: libc::POLLIN,
+                     });
+            1
+        }
+        None => 0,
+    };
+
+    // Don't timeout if eventable lets us be event-driven
+    let poll_timeout = if engine_fds_end_idx != 0 { -1 } else { 10000 };
 
     loop {
-        // Poll them with a 10 s timeout
-        let r = unsafe { libc::poll(fds.as_mut_ptr(), fds.len() as libc::c_ulong, 10000) };
-        assert!(r >= 0);
+        // Unconditionally call evented() if engine has no eventable.
+        // This looks like a bad idea, but the only engine that has
+        // no eventable is the sim engine, and for that engine, evented()
+        // is essentially a no-op.
+        if engine_fds_end_idx == 0 {
+            engine.borrow_mut().evented()?;
+        }
 
-        // And handle incoming events
-        for pfd in fds.iter().filter(|pfd| pfd.revents != 0) {
+        // Handle engine fd, if there is one.
+        // If engine has no eventable, there is no engine fd; in that case
+        // the body of the loop is executed zero times.
+        for pfd in fds[..engine_fds_end_idx]
+                .iter_mut()
+                .filter(|pfd| pfd.revents != 0) {
+            pfd.revents = 0;
+            eventable
+                .as_mut()
+                .expect("index < engine_fds_end_idx <=> index == 0 <=> eventable.is_some()")
+                .clear_event()?;
+            engine.borrow_mut().evented()?;
+        }
+
+        // Iterate through D-Bus file descriptors
+        for pfd in fds[engine_fds_end_idx..]
+                .iter()
+                .filter(|pfd| pfd.revents != 0) {
             for item in dbus_conn.watch_handle(pfd.fd, WatchEvent::from_revents(pfd.revents)) {
                 if let Err(r) = libstratis::dbus_api::handle(&dbus_conn,
                                                              &item,
@@ -109,8 +140,13 @@ fn run() -> StratisResult<()> {
             }
         }
 
-        // Ask the engine to check its pools
-        engine.borrow_mut().check()
+        // Refresh list of dbus fds to poll for every time. This can change as
+        // D-Bus clients come and go.
+        fds.truncate(engine_fds_end_idx);
+        fds.extend(dbus_conn.watch_fds().iter().map(|w| w.to_pollfd()));
+
+        let r = unsafe { libc::poll(fds.as_mut_ptr(), fds.len() as libc::c_ulong, poll_timeout) };
+        assert!(r >= 0);
     }
 }
 
--- a/src/engine/engine.rs
+++ b/src/engine/engine.rs
@@ -3,6 +3,7 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 use std::fmt::Debug;
+use std::os::unix::io::RawFd;
 use std::path::{Path, PathBuf};
 
 use chrono::{DateTime, Utc};
@@ -172,4 +173,22 @@ pub trait Engine: Debug {
 
     /// Get all pools belonging to this engine.
     fn pools(&self) -> Vec<&Pool>;
+
+    /// If the engine would like to include an event in the message loop, it
+    /// may return an Eventable from this method.
+    fn get_eventable(&mut self) -> EngineResult<Option<Box<Eventable>>>;
+
+    /// Notify the engine that an event has occurred on the Eventable.
+    fn evented(&mut self) -> EngineResult<()>;
+}
+
+/// Allows an Engine to include a fd in the event loop. See
+/// Engine::get_eventable() and Engine::evented().
+pub trait Eventable {
+    /// Get fd the engine would like to monitor for activity
+    fn get_pollable_fd(&mut self) -> RawFd;
+
+    /// Assuming level-triggered semantics, clear the event that caused the
+    /// Eventable to trigger.
+    fn clear_event(&mut self) -> EngineResult<()>;
 }
--- a/src/engine/strat_engine/engine.rs
+++ b/src/engine/strat_engine/engine.rs
@@ -2,11 +2,13 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+use std::collections::HashMap;
+use std::os::unix::io::{AsRawFd, RawFd};
 use std::path::Path;
 
-use devicemapper::DM;
+use devicemapper::{DM, DmNameBuf};
 
-use super::super::engine::{Engine, HasName, HasUuid, Pool};
+use super::super::engine::{Engine, Eventable, HasName, HasUuid, Pool};
 use super::super::errors::{EngineError, EngineResult, ErrorEnum};
 use super::super::structures::Table;
 use super::super::types::{DevUuid, PoolUuid, Redundancy, RenameAction};
@@ -14,6 +16,8 @@ use super::super::types::{DevUuid, PoolUuid, Redundancy, RenameAction};
 use super::cleanup::teardown_pools;
 use super::pool::StratPool;
 use super::setup::find_all;
+
+const REQUIRED_DM_MINOR_VERSION: u32 = 37;
 
 #[derive(Debug, PartialEq, Eq)]
 pub enum DevOwnership {
@@ -25,6 +29,22 @@ pub enum DevOwnership {
 #[derive(Debug)]
 pub struct StratEngine {
     pools: Table<StratPool>,
+    // Maps name of DM devices we are watching to the most recent event number
+    // we've handled for each
+    watched_dev_last_event_nrs: HashMap<DmNameBuf, u32>,
+}
+
+impl Eventable for DM {
+    /// Get file we'd like to have monitored for activity
+    fn get_pollable_fd(&mut self) -> RawFd {
+        self.file().as_raw_fd()
+    }
+
+    fn clear_event(&mut self) -> EngineResult<()> {
+        self.arm_poll()?;
+
+        Ok(())
+    }
 }
 
 impl StratEngine {
@@ -32,9 +52,18 @@ impl StratEngine {
     /// 1. Verify the existance of Stratis /dev directory.
     /// 2. Setup all the pools belonging to the engine.
     ///
+    /// Returns an error if the kernel doesn't support required DM features.
     /// Returns an error if there was an error reading device nodes.
     /// Returns an error if there was an error setting up any of the pools.
     pub fn initialize() -> EngineResult<StratEngine> {
+        let minor_dm_version = DM::new()?.version()?.1;
+        if minor_dm_version < REQUIRED_DM_MINOR_VERSION {
+            let err_msg = format!("Requires DM minor version {} but kernel only supports {}",
+                                  REQUIRED_DM_MINOR_VERSION,
+                                  minor_dm_version);
+            return Err(EngineError::Engine(ErrorEnum::Error, err_msg));
+        }
+
         let pools = find_all()?;
 
         let mut table = Table::default();
@@ -50,7 +79,10 @@ impl StratEngine {
             }
         }
 
-        Ok(StratEngine { pools: table })
+        Ok(StratEngine {
+               pools: table,
+               watched_dev_last_event_nrs: HashMap::new(),
+           })
     }
 
     /// Teardown Stratis, preparatory to a shutdown.
@@ -121,6 +153,32 @@ impl Engine for StratEngine {
 
     fn pools(&self) -> Vec<&Pool> {
         self.pools.into_iter().map(|x| x as &Pool).collect()
+    }
+
+    fn get_eventable(&mut self) -> EngineResult<Option<Box<Eventable>>> {
+        Ok(Some(Box::new(DM::new()?)))
+    }
+
+    fn evented(&mut self) -> EngineResult<()> {
+        let device_list: HashMap<_, _> = DM::new()?
+            .list_devices()?
+            .into_iter()
+            .map(|(dm_name, _, event_nr)| {
+                     (dm_name, event_nr.expect("Supported DM versions always provide a value"))
+                 })
+            .collect();
+
+        for pool in &mut self.pools {
+            for dm_name in pool.get_eventing_dev_names() {
+                if device_list.get(&dm_name) > self.watched_dev_last_event_nrs.get(&dm_name) {
+                    pool.event_on(&dm_name)?;
+                }
+            }
+        }
+
+        self.watched_dev_last_event_nrs = device_list;
+
+        Ok(())
     }
 }
 
--- a/src/engine/strat_engine/pool.rs
+++ b/src/engine/strat_engine/pool.rs
@@ -11,7 +11,7 @@ use std::vec::Vec;
 use serde_json;
 use uuid::Uuid;
 
-use devicemapper::{Device, DM, Sectors};
+use devicemapper::{Device, DM, DmName, DmNameBuf, Sectors, ThinPoolDev};
 
 use super::super::engine::{Filesystem, BlockDev, HasName, HasUuid, Pool};
 use super::super::errors::{EngineError, EngineResult, ErrorEnum};
@@ -116,6 +116,11 @@ impl StratPool {
 
     pub fn has_filesystems(&self) -> bool {
         self.thin_pool.has_filesystems()
+    }
+
+    /// The names of DM devices belonging to this pool that may generate events
+    pub fn get_eventing_dev_names(&self) -> Vec<DmNameBuf> {
+        self.thin_pool.get_eventing_dev_names()
     }
 }
 
--- a/src/engine/strat_engine/pool.rs
+++ b/src/engine/strat_engine/pool.rs
@@ -11,7 +11,7 @@ use std::vec::Vec;
 use serde_json;
 use uuid::Uuid;
 
-use devicemapper::{Device, DM, DmName, DmNameBuf, Sectors, ThinPoolDev};
+use devicemapper::{Device, DM, DmName, DmNameBuf, Sectors};
 
 use super::super::engine::{Filesystem, BlockDev, HasName, HasUuid, Pool};
 use super::super::errors::{EngineError, EngineResult, ErrorEnum};
@@ -121,6 +121,12 @@ impl StratPool {
     /// The names of DM devices belonging to this pool that may generate events
     pub fn get_eventing_dev_names(&self) -> Vec<DmNameBuf> {
         self.thin_pool.get_eventing_dev_names()
+    }
+
+    pub fn event_on(&mut self, _dm_name: &DmName) -> EngineResult<()> {
+        // TODO: Just check the device that evented. Currently checks
+        // everything.
+        self.check()
     }
 }
 
--- a/src/engine/strat_engine/thinpool.rs
+++ b/src/engine/strat_engine/thinpool.rs
@@ -10,8 +10,9 @@ use std::process::Command;
 use uuid::Uuid;
 
 use devicemapper as dm;
-use devicemapper::{DM, DataBlocks, DmDevice, DmName, IEC, LinearDev, MetaBlocks, Sectors, Segment,
-                   ThinDev, ThinDevId, ThinPoolDev, ThinPoolStatusSummary, device_exists};
+use devicemapper::{DM, DataBlocks, DmDevice, DmName, DmNameBuf, IEC, LinearDev, MetaBlocks,
+                   Sectors, Segment, ThinDev, ThinDevId, ThinPoolDev, ThinPoolStatusSummary,
+                   device_exists};
 
 use super::super::engine::{Filesystem, HasName};
 use super::super::errors::{EngineError, EngineResult, ErrorEnum};
@@ -182,8 +183,6 @@ impl ThinPool {
                                         &format_flex_name(pool_uuid, FlexRole::ThinData),
                                         None,
                                         &map_to_dm(&data_segments))?;
-
-
 
         let thinpool_dev = ThinPoolDev::setup(dm,
                                               &thinpool_name,
@@ -539,6 +538,15 @@ impl ThinPool {
             self.filesystems.insert(filesystem);
             Ok(RenameAction::Renamed)
         }
+    }
+
+    /// The names of DM devices belonging to this pool that may generate events
+    pub fn get_eventing_dev_names(&self) -> Vec<DmNameBuf> {
+        vec![format_flex_name(self.pool_uuid, FlexRole::ThinMeta),
+             format_flex_name(self.pool_uuid, FlexRole::ThinData),
+             format_flex_name(self.pool_uuid, FlexRole::MetadataVolume),
+             format_thinpool_name(self.pool_uuid, ThinPoolRole::Pool)]
+
     }
 }
 
--- a/src/engine/sim_engine/engine.rs
+++ b/src/engine/sim_engine/engine.rs
@@ -9,7 +9,7 @@ use std::iter::FromIterator;
 use std::path::Path;
 use std::rc::Rc;
 
-use super::super::engine::{Engine, HasName, HasUuid, Pool};
+use super::super::engine::{Engine, Eventable, HasName, HasUuid, Pool};
 use super::super::errors::{EngineError, EngineResult, ErrorEnum};
 use super::super::structures::Table;
 use super::super::types::{PoolUuid, Redundancy, RenameAction};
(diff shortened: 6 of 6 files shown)


agrover:
`event_on()` only placeholder code.

agrover:
Rebased.
* Added some missing trait method docstrings

src/bin/stratisd.rs
```diff
@@ -109,8 +133,13 @@ fn run() -> StratisResult<()> {
             }
         }
 
-        // Ask the engine to check its pools
-        engine.borrow_mut().check()
+        // Refresh list of dbus fds to poll for every time. This can change as
+        // D-Bus clients come and go.
+        fds.truncate(dbus_fd_start_idx);
+        fds.extend(dbus_conn.watch_fds().iter().map(|w| w.to_pollfd()));
```
mulkieran: In this new arrangement, the first time through the loop there are no d-bus events to be interested in, but there were before. Is that intentional?

src/bin/stratisd.rs
```diff
@@ -109,8 +133,13 @@ fn run() -> StratisResult<()> {
...
+
+        let r = unsafe { libc::poll(fds.as_mut_ptr(), fds.len() as libc::c_ulong, poll_timeout) };
```
mulkieran: Why does whether or not engine has an eventable determine timeout for all interesting file descriptors?

src/bin/stratisd.rs
```diff
@@ -109,8 +133,13 @@ fn run() -> StratisResult<()> {
             }
         }
 
-        // Ask the engine to check its pools
-        engine.borrow_mut().check()
+        // Refresh list of dbus fds to poll for every time. This can change as
+        // D-Bus clients come and go.
+        fds.truncate(dbus_fd_start_idx);
```
mulkieran: Just set up dbus fds here (rather than adding to uber-list of fds.

src/engine/strat_engine/engine.rs
```diff
@@ -122,6 +141,42 @@ impl Engine for StratEngine {
     fn pools(&self) -> Vec<&Pool> {
         self.pools.into_iter().map(|x| x as &Pool).collect()
     }
+
+    fn get_eventable(&mut self) -> EngineResult<Option<Box<Eventable>>> {
```
mulkieran: I'm not sure that Box in the actual return type is a good idea.

src/engine/strat_engine/engine.rs
```diff
@@ -25,6 +27,20 @@ pub enum DevOwnership {
 #[derive(Debug)]
 pub struct StratEngine {
     pools: Table<StratPool>,
+    watched_devs: HashMap<DmNameBuf, u32>, // name, event_nr
```
mulkieran: Needs a comment on new field, something like:

```// Maps name of DM device to most recent event number associated w/ the device```

File: src/engine/strat_engine/engine.rs [hunk 6]
```diff
@@ -122,6 +141,42 @@ impl Engine for StratEngine {
...
+                    if new_event_nr > *prev_event_nr {
+                        // Event! We know the dm name, but don't know which
+                        // pool
+                        for pool in &mut self.pools {
+                            if pool.get_eventables().contains(dm_name) {
+                                pool.event_on(dm_name)?;
+                                break;
+                            }
+                        }
+                        *prev_event_nr = new_event_nr;
```
mulkieran: Given that ```watched_devs```is reset just below, I'm not sure that this assignment achieves anything.

src/engine/strat_engine/pool.rs
```diff
@@ -117,6 +117,19 @@ impl StratPool {
     pub fn has_filesystems(&self) -> bool {
         self.thin_pool.has_filesystems()
     }
+
+    pub fn get_eventables(&self) -> Vec<DmNameBuf> {
+        self.thin_pool.get_eventables()
+    }
+
+    pub fn event_on(&mut self, dm_name: &DmName) -> EngineResult<()> {
+        println!("event on {}", dm_name);
```
mulkieran: Expecting this to go away, of course.

src/engine/strat_engine/engine.rs
```diff
@@ -122,6 +141,42 @@ impl Engine for StratEngine {
...
+                    }
+                }
+            }
+        }
+
+        self.watched_devs = device_list
```
mulkieran: I don't thing these devs are really watched. Maybe ```last_observed_event_nrs``` would be a better name.

src/engine/strat_engine/engine.rs
```diff
@@ -122,6 +141,42 @@ impl Engine for StratEngine {
...
+
+    fn get_eventable(&mut self) -> EngineResult<Option<Box<Eventable>>> {
+        Ok(Some(Box::new(DM::new()?)))
+    }
+
+    fn evented(&mut self) -> EngineResult<()> {
+        let device_list = DM::new()?.list_devices()?;
+
+        for &(ref dm_name, _, new_event_nr) in &device_list {
+            if let Some(new_event_nr) = new_event_nr {
```
mulkieran: I think you can drop both these if's. Instead,

```if new_event_nr > self.watched_devs.get_mut(dm_name) {```

should work out correctly.

File: src/engine/strat_engine/engine.rs [hunk 10]
```diff
@@ -122,6 +141,42 @@ impl Engine for StratEngine {
...
+                if let Some(prev_event_nr) = self.watched_devs.get_mut(dm_name) {
+                    if new_event_nr > *prev_event_nr {
+                        // Event! We know the dm name, but don't know which
+                        // pool
+                        for pool in &mut self.pools {
```
mulkieran: Might as well use iterator ```find()``` method here.Or possibly ```position``` and then ```get_mut()``` if found.

src/engine/engine.rs
```diff
@@ -172,4 +173,20 @@ pub trait Engine: Debug {
 
     /// Get all pools belonging to this engine.
     fn pools(&self) -> Vec<&Pool>;
+
+    /// If the engine would like to include an event in the message loop, it
+    /// may return an Eventable from this method.
+    fn get_eventable(&mut self) -> EngineResult<Option<Box<Eventable>>>;
```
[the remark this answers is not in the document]
agrover: We can do this when such a thing becomes necessary.

src/engine/engine.rs
```diff
@@ -172,4 +173,20 @@ pub trait Engine: Debug {
...
+
+    /// Notify the engine that an event has occurred on the Eventable.
```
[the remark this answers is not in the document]
agrover: That's all we need for StratEngine right now. It's really hard to know what we'll need in the future, and easy enough to add in the future as well, so this is just doing what's required for today's situation where we only have Sim and StratEngine.

src/engine/strat_engine/thinpool.rs
```diff
@@ -543,6 +542,14 @@ impl ThinPool {
             Ok(RenameAction::Renamed)
         }
     }
+
+    pub fn get_eventables(&self) -> Vec<DmNameBuf> {
```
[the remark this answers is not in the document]
agrover: I don't see how to avoid making this Vec without making things more complex.

src/bin/stratisd.rs
```diff
@@ -85,20 +85,44 @@ fn run() -> StratisResult<()> {
...
-    // Get a list of fds to poll for
-    let mut fds: Vec<_> = dbus_conn
-        .watch_fds()
-        .iter()
-        .map(|w| w.to_pollfd())
-        .collect();
+    let mut fds = Vec::new();
+
+    let mut eventable = engine.borrow_mut().get_eventable()?;
+    let poll_timeout = match eventable {
```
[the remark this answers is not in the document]
agrover: You have to eventually make one list of `struct pollfd` because that's what `libc::poll` takes, see `man 2 poll`. And then that's also where you look after it returns to tell which (if any) of the fds have had an event. I'm not sure doing as you describe simplifies things -- so we now have an `Option<libc::pollfd>` but you still have to copy it into a Vec with the dbus fds, one after the other, and then check the array to dispatch events. How can this be made simpler? There seems to be a lot of irreducible ugliness here (which is what things like tokio is supposed to improve, and maybe we should use that in the future, but it has its own set of complexities that IMHO make this simpler right now.)

src/bin/stratisd.rs
```diff
@@ -85,20 +85,44 @@ fn run() -> StratisResult<()> {
...
-        // Poll them with a 10 s timeout
-        let r = unsafe { libc::poll(fds.as_mut_ptr(), fds.len() as libc::c_ulong, 10000) };
-        assert!(r >= 0);
+        // Handle all Eventable events before polling. If an event arrives
+        // after calling clear_event(), we will pop out of the poll
+        // immediately to call evented() again. This guarantees no events are
+        // missed but may result in an unneeded call to evented().
+        match eventable {
+            Some(ref mut evt) => {
+                if fds[0].revents != 0 {
```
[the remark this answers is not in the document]
agrover: We can't do this because it all needs to be in the uber-list.

src/bin/stratisd.rs
agrover: We can't do this because it all needs to be in the uber-list.

src/bin/stratisd.rs
agrover: The loop is structured to avoid duplicating code. There are things that need to happen before `poll()` is called the first time (setup), and then there are things (checking, setup) every time `poll()` wakes up. That ends up putting the call to `poll()` at the bottom of the loop, so the first time through the "checking" part is moot, we're just doing the "setup" part... and then in subsequent loops both checking and setup are done.

src/bin/stratisd.rs
agrover: Because if the engine doesn't support events then it needs to poll -- periodically have a chance to check itself to see if it needs to do something. (We were doing this before for StratEngine to call `check()`, which would e.g. extend the flex data device if needed, etc..)

src/engine/strat_engine/engine.rs
agrover: ok

src/engine/strat_engine/engine.rs
agrover: We need to return a trait object here, so I think this is needed.

src/engine/strat_engine/engine.rs
agrover: oh yeah! ok.

src/engine/strat_engine/engine.rs
agrover: yes, ok

src/engine/strat_engine/pool.rs
agrover: yup.

src/engine/strat_engine/engine.rs
agrover: I think there was a theoretical concern about event_nr wrapping?

src/engine/strat_engine/engine.rs
agrover: ok will try.

src/bin/stratisd.rs
mulkieran: I should be more specific...most of those fd's have nothing to do with the engine's eventables. Why is their poll-timeout dependent on the existance of the eventable when it wasn't before?

src/engine/strat_engine/engine.rs
mulkieran: Actually, I think there may be a bug in your current formulation. I think that if ```watched_devs``` lacks an entry for ```dm_name``` but there was a corresponding ```event_nr``` for the name in the device list then you had better take the action. But in your current code, you don't.

src/bin/stratisd.rs
agrover: I don't follow -- there's just one timeout for all the fds we're polling. The other fds, the D-Bus fds, are already event-driven so if any of them have activity, we return from poll() immediately and handle them. Once we can avoid polling by making the engine check() calls event-driven, then there's no need to "fake" a DM event every 10 seconds, so the timeout serves no purpose any more.

src/engine/strat_engine/engine.rs
agrover: If watched_devs lacks an entry, then an event happened on a DM device we don't care about. DM::list_devices() returns all DM devices on the system. If e.g. an LVM configured DM device has an event we will still end up in this function and should take no action.

src/engine/strat_engine/engine.rs
agrover: On second thought, they are. This is a list of only the dm names we care about, and the last seen event numbers for each. I'll try to name it something better, and we can see if it's better.

src/engine/strat_engine/engine.rs
agrover: got this working with find(), but it's ugly:
```rust
if let Some(pool) = (&mut self.pools)
                               .into_iter()
                               .find(|ref mut pool| {
                                         pool.get_eventing_dev_names().contains(dm_name)
                                     }) {
                            pool.event_on(dm_name)?;
                        }
```

I think working on #492 might help, not sure. Maybe we can go with what's there for now and then circle back.

agrover:
rebased.
* Added 2 squashable commits that rename & add comment for clarity

src/engine/strat_engine/engine.rs
mulkieran: I don't think that's quite right.

Last time we called ```evented()```, a particular device might not have existed and ```watched_dev_last_event_nrs``` would not contain a key for that device. But this time we call ```evented()``` the device does exist, so it is contained in ```list_devices()``` result but not in ```watched_dev_last_event_nrs```. I think we need to do something in that case.

src/bin/stratisd.rs
mulkieran:
```
if let Some(ref mut evt) = eventable {
  fds.push(...)
}
let dbus_fd_start_idx = fds.len();
```

src/bin/stratisd.rs
mulkieran: instead of ```poll_timeout```,

```if dbus_fd_start_idx == 0 { 10000 } else { -1 }```

File: src/engine/strat_engine/thinpool.rs [hunk 16]
```diff
@@ -543,6 +542,14 @@ impl ThinPool {
             Ok(RenameAction::Renamed)
         }
     }
+
+    pub fn get_eventing_dev_names(&self) -> Vec<DmNameBuf> {
```
mulkieran: I guess method is probably fine. Suggested header comment:

```
/// The names of DM devices belonging to this pool that may generate events
```

?

src/engine/strat_engine/pool.rs
```diff
@@ -117,6 +117,19 @@ impl StratPool {
     pub fn has_filesystems(&self) -> bool {
         self.thin_pool.has_filesystems()
     }
+
+    pub fn get_eventing_dev_names(&self) -> Vec<DmNameBuf> {
```
mulkieran: This method probably also needs a header comment like the one below.

agrover:
rebased.
* Add comments to get_eventing_dev_names functions
* Redo code around poll_timeout and dbus_fd_start_idx to make it clearer
      their relative importance

src/bin/stratisd.rs
```diff
@@ -85,20 +85,45 @@ fn run() -> StratisResult<()> {
...
+        None => 0,
+    };
+
+    let poll_timeout = if dbus_fd_start_idx == 0 { 10000 } else { -1 };
 
     loop {
-        // Poll them with a 10 s timeout
-        let r = unsafe { libc::poll(fds.as_mut_ptr(), fds.len() as libc::c_ulong, 10000) };
-        assert!(r >= 0);
+        // Handle all Eventable events before polling. If an event arrives
```
mulkieran: You're only handling one or zero Eventable events in this code. It doesn't generalize.

agrover: Ok I'll clarify. We can have multiple DM events that result in a single Eventable, but at this level that is not relevant.

agrover:
> Why do we need to return a trait object? Is it due to unknown size?

Yes. We know we have (a reference to) a struct that implements Eventable, but we don't know anything else about the struct, including its size.

agrover:
* fix comment about Eventable

src/bin/stratisd.rs
```diff
@@ -109,8 +134,13 @@ fn run() -> StratisResult<()> {
...
 
-        // Ask the engine to check its pools
-        engine.borrow_mut().check()
+        // Refresh list of dbus fds to poll for every time. This can change as
+        // D-Bus clients come and go.
+        fds.truncate(dbus_fd_start_idx);
+        fds.extend(dbus_conn.watch_fds().iter().map(|w| w.to_pollfd()));
+
+        let r = unsafe { libc::poll(fds.as_mut_ptr(), fds.len() as libc::c_ulong, poll_timeout) };
+        assert!(r >= 0);
```
mulkieran: I think ```assert!``` is probably not the write way to check ```poll()``` return code.

src/bin/stratisd.rs
```diff
@@ -85,20 +85,45 @@ fn run() -> StratisResult<()> {
...
+        None => 0,
+    };
+
+    let poll_timeout = if dbus_fd_start_idx == 0 { 10000 } else { -1 };
 
     loop {
-        // Poll them with a 10 s timeout
-        let r = unsafe { libc::poll(fds.as_mut_ptr(), fds.len() as libc::c_ulong, 10000) };
-        assert!(r >= 0);
+        // Handle the Eventable (if there is one) before polling. If an event
```
mulkieran: This code still seems very funky, but I feel as if I understand it.

src/engine/engine.rs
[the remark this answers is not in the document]
mulkieran: I honestly think it might make the funky code up above easier. But, I'm happy to leave it for now.

src/engine/engine.rs
```diff
@@ -172,4 +173,20 @@ pub trait Engine: Debug {
...
+    fn evented(&mut self) -> EngineResult<()>;
+}
+
+pub trait Eventable {
```
mulkieran: Please document the purpose of the trait.

src/bin/stratisd.rs
agrover: We really shouldn't be using `libc::poll` at all. I've opened #655, which will include removing that assert.

src/bin/stratisd.rs
agrover: it is funky. At some point we may want to look at mio or tokio crates to abstract the messiness of the event loop. I've opened #656 to track.

agrover:
squashed and rebased.

src/engine/strat_engine/engine.rs
```diff
@@ -122,6 +143,41 @@ impl Engine for StratEngine {
...
     }
+
+    fn get_eventable(&mut self) -> EngineResult<Option<Box<Eventable>>> {
+        Ok(Some(Box::new(DM::new()?)))
+    }
+
+    fn evented(&mut self) -> EngineResult<()> {
+        let device_list = DM::new()?.list_devices()?;
+
+        for &(ref dm_name, _, new_event_nr) in &device_list {
```
mulkieran: Either we'll be using a newer version of devicemapper, i which case every single one of these ```new_event_nr``` values will be ```Some``` or we'll not, in which case they will all be ```None```. In the second case, no pool will ever be "evented on".

We should definitely document what our stance is on that situation.

agrover:
> I thin nix::poll will have the same meaning in its return value, though, so it might still be worth handling here.

libc::poll returns a c_int, but nix::poll returns a `Result<c_int>`, so it will be different in that the -1 that libc::poll returns to indicate error an error is checked-for by Nix, and results in an `Err` value. I think with nix::poll we can just `?` it, since we don't happen to care about the value (the number of fds with events) as long as there was no error.

mulkieran:
> > I thin nix::poll will have the same meaning in its return value, though, so it might still be worth handling here.

> libc::poll returns a c_int, but nix::poll returns a `Result<c_int>`, so it will be different in that the -1 that libc::poll returns to indicate error an error is checked-for by Nix, and results in an `Err` value. I think with nix::poll we can just `?` it, since we don't happen to care about the value (the number of fds with events) as long as there was no error.

Ok!

src/engine/strat_engine/engine.rs
agrover: Yeah this is a problem. Or at least a weirdness. Maybe if `None` we should instead *always* call event_on() for all watched devs?

mulkieran: I think that if devicemapper version is early so that no event nrs available, then the only meaningful thing is to call ```event_on()``` for all pools.

I think that in that same case, there will be no watched devs.

agrover:
* Added minimum DM version check to StratEngine and simplified `evented()` implementation now that we can assume list_devices() always returns an event_nr.

src/engine/strat_engine/engine.rs
```diff
@@ -35,6 +55,14 @@ impl StratEngine {
     /// Returns an error if there was an error reading device nodes.
     /// Returns an error if there was an error setting up any of the pools.
     pub fn initialize() -> EngineResult<StratEngine> {
+        let minor_dm_version = DM::new()?.version()?.1;
```
mulkieran: We might want to document the reason for this requirement in some docs.

src/engine/strat_engine/engine.rs
```diff
@@ -122,6 +153,40 @@ impl Engine for StratEngine {
...
+                            break;
+                        }
+                    }
+                }
+            }
+        }
+
+        self.watched_dev_last_event_nrs = device_list
+            .into_iter()
+            .filter_map(|(dm_name, _, event_nr)| match event_nr {
```
mulkieran: There's no need to filter here anymore, since ```even_nr``` can not be ```None```.

agrover: ok

src/engine/strat_engine/engine.rs
agrover: opened https://github.com/stratis-storage/stratis-docs/issues/72

agrover:
rebased.
* Don't need `filter_map`, just use `map`
* Add a comment line to StratEngine::initialize()

src/engine/strat_engine/engine.rs
```diff
@@ -122,6 +154,39 @@ impl Engine for StratEngine {
...
+
+    fn get_eventable(&mut self) -> EngineResult<Option<Box<Eventable>>> {
+        Ok(Some(Box::new(DM::new()?)))
+    }
+
+    fn evented(&mut self) -> EngineResult<()> {
+        let device_list = DM::new()?.list_devices()?;
+
+        for &(ref dm_name, _, new_event_nr) in &device_list {
+            let new_event_nr = new_event_nr.expect("Supported DM versions always provide a value");
```
mulkieran: Now that we know that ```new_event_nr``` is always ```Some``` this is simpler to think about.

I still think that

```
if new_event_nr > self.watched_dev_last_event_nrs.get(dm_name) {
  ...
}
```
is the correct thing to do.

There are only two possiblites.

1. ```prev_event_nr``` is ```None```. In that case, the device has shown up since the last time ```evented()``` was called. So there may be something to do about it, and we should handle it.
2. ```prev_event_nr``` matches ```Some```. In that case, there will be a comparsion between the values of the two numbers and if ```new_event_nr``` is greater than ```prev_event_nr``` something will be done.

src/engine/strat_engine/engine.rs
```diff
@@ -122,6 +154,39 @@ impl Engine for StratEngine {
...
+            if let Some(prev_event_nr) = self.watched_dev_last_event_nrs.get(dm_name) {
+                if new_event_nr > *prev_event_nr {
+                    // Event! We know the dm name, but don't know which
+                    // pool
+                    for pool in &mut self.pools {
```
mulkieran: Was it impossible to use ```find``` here?

src/engine/strat_engine/pool.rs
```diff
@@ -117,6 +117,19 @@ impl StratPool {
...
         self.thin_pool.has_filesystems()
     }
+
+    /// The names of DM devices belonging to this pool that may generate events
+    pub fn get_eventing_dev_names(&self) -> Vec<DmNameBuf> {
+        self.thin_pool.get_eventing_dev_names()
+    }
+
+    pub fn event_on(&mut self, _dm_name: &DmName) -> EngineResult<()> {
+        // We could be smart and just check the thing that evented, but for
```
mulkieran: Could you turn this into a ```TODO``` or a ```FIXME```, please?

src/engine/strat_engine/pool.rs
```diff
@@ -117,6 +117,19 @@ impl StratPool {
...
+        // now check everything
+        self.check()?;
```
mulkieran: Single line, ```self.check()``` would be more aesthetic, but this is not important.

src/engine/strat_engine/engine.rs
agrover: Doesn't compile.
```
   --> src/engine/strat_engine/engine.rs:167:31
    |
167 |             if new_event_nr > self.watched_dev_last_event_nrs.get(dm_name) {
    |                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected u32, found enum `std::option::Option`
    |
    = note: expected type `u32`
               found type `std::option::Option<&u32>`
```

src/engine/strat_engine/engine.rs
agrover: no, but it looks like this:
```
if let Some(pool) =
    (&mut self.pools)
        .into_iter()
        .find(|pool| pool.get_eventing_dev_names().contains(dm_name)) {
    pool.event_on(dm_name)?;
}
```

which doesn't seem much better than what's there now.

src/engine/strat_engine/pool.rs
agrover: ok

src/engine/strat_engine/pool.rs
agrover: sure, why not

agrover:
* made style changes to event_on()

src/engine/strat_engine/engine.rs
mulkieran: Yeah, you need to lose the ```expect()``` call as well. Then they are both ```Option``` type and hopefully ```Option``` has enough cleverness that ```>``` will still work.

mulkieran: But in any case, if you do nothing, I think you have  a bug.

src/bin/stratisd.rs
```diff
@@ -85,20 +85,45 @@ fn run() -> StratisResult<()> {
...
+        // arrives after calling clear_event(), we will pop out of the poll
```
mulkieran: most of this comment is obscure to me. "If an event arrives after calling clear_event()``` is the unclearest to me. I'm not sure we do guarantee that no events are missed. We could miss 100 and then detect the 101st, AFAICT.

agrover: Can you give an example of scenario where you think we would miss an event? Then I can walk through it.

mulkieran: Well I'm assuming a one-to-one and onto relationship between events and event nrs. Suppose you call list devices. You look at all devices and the highest event_nr is n. Later you call list_devices again. The device which previously had event_nr n now has event_nr n + 2. All other devices had the same event nr they had previously. I don't think that is impossible, but that means that you missed event corresponding to event nr n + 1.

agrover: Yes, that is certainly possible. That's working as intended. The goal is to get woken up when an event happens, and then know which DM device(s) evented (once or more) times since the last time we handled events. The assumption is we'd get the current status of the device and take action based on the status. If the status shows multiple things needing attention with the device then we should resolve all of them, rather than getting a separate event for each issue.

agrover:
rebased.
* Rewrite evented() to handle if we don't know the last seen event for a watched dev

src/engine/strat_engine/engine.rs
```diff
@@ -122,6 +154,41 @@ impl Engine for StratEngine {
...
+            .list_devices()?
+            .into_iter()
+            .map(|(dm_name, _, event_nr)| {
+                     (dm_name, event_nr.expect("Supported DM versions always provide a value"))
+                 })
+            .collect();
+
+        for pool in &mut self.pools {
+            for dm_name in pool.get_eventing_dev_names() {
+                if let Some(new_event_nr) = device_list.get(&dm_name) {
```
mulkieran:
```
if device_list.get(&dm_name) > self.watched_dev_last_event_nrs.get(&dm_name) {
  pool.even_on(&dm_name)?;
}
```

should still take careof lines 173 - 185, AFAICT.

agrover:
ooh that's slick!

Made that change, rebased, and squashed.

mulkieran:
I prefer the arrangement shown here: https://github.com/mulkieran/stratisd/commit/9cefb5f9efe4b9bae26ad9198755ac1c1646ea08. It gives a uniformity of structure as now the code works over the list. Also, it makes the relation between the idx and existance of eventable more explicit.

mulkieran:
Actually, I prefer the whole approach in: https://github.com/mulkieran/stratisd/commits/agrover-dm-events.

mulkieran:
Also, I'm not convinced that ```poll_timeout``` conditional is not reversed. Please verify that its correct and please add an explanatory comment.

mulkieran:
I'm gonna try one more thing...

mulkieran:
Probably the best available: https://github.com/mulkieran/stratisd/tree/agrover-dm-events-2.

agrover:
Do you want to open a separate PR so I can review your latest changes?

agrover:
rebased.

PR #603, opened 2017-10-11
update get_mount_point() to return EngineResult<Option<PathBuf>>

trgill:
The change allows the caller to distinguish between the filesytem being
unmounted and an error determining if it is mounted.


trgill:
@agrover @mulkieran please review - this commit was split from the snapshot PR.

PR #604, opened 2017-10-11
Simplify test-loop and test-real targets

mulkieran:
Sure beats alphabetizing the individual tests.


mulkieran:
Guess it only works if they're part of unit test framework.

PR #606, opened 2017-10-13
Master devicemapper 0.11.0 next

### LINKED ISSUE #550: The problem with ThinPool::extend_data
It works like this:

1. Call devicemapper methods to extend the underlying thinpool, actually the thin pool's data device with the given segments. The problem here is that it's not actually extension, the last data segment of the device is coalesced w/ the first data segment of the incoming devices if possible.
2. Run the identical algorithm as is used in ```devicemapper-rs::LinearDev::extend()``` in order to make the segments that the thinpool stores match the segments on the linear device.
3. Return.

Step (2) is really brittle, since it relies for correctness on algorithms in two distinct libraries to be identical.

Proposal 1: Do not coalesce in either place. This is simple, but it is also brittle. Someone might re-add cleverness to devicemapper, causing a breakage in stratisd.

Proposal 2: Calculate the segments, coalescing as desired, in stratisd. _Set_ all the segments in the underlying devicemapper linear device. This needs a new function in LinearDev.

I prefer proposal (2) because ultimately it is less brittle and it gives all clients of devicemapper greater flexibility in how they arrange their segments.

mulkieran:
Resolves #550.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,10 +9,10 @@ dependencies = [
  "crc 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "custom_derive 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "dbus 0.5.3 (registry+https://github.com/rust-lang/crates.io-index)",
- "devicemapper 0.11.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "devicemapper 0.11.0 (git+https://github.com/stratis-storage/devicemapper-rs)",
  "enum_derive 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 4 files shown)


mulkieran:
rebased.

mulkieran:
rebased.

mulkieran:
test this please

mulkieran:
rebased and updates.

mulkieran:
@agrover @trgill review? Let's do a new devicemapper release. We have bug fixes and API changes, and nothing really in the queue.

mulkieran:
The only thing that's failing is clippy, so we should consider this a success. @agrover @trgill review?

agrover:
LGTM 👍

mulkieran:
Depends on #619, since clippy actually requires some code changes.

mulkieran:
Rebased, newest version of devicemapper master branch.

Issue #491, opened 2017-07-27
Investigate improvements into mnt crate


agrover:
We are using the `mnt` crate to parse /proc/mounts to get information we need to link mount points up with the Stratis thin devices that they may be on. This library may not be handling all cases, e.g. if multiple filesystems mounted on the same mount point. We should make sure it can meet our needs, possibly by contributing improvements as needed.

tasleson:
@agrover

> e.g. if multiple filesystems mounted on the same mount point.

Did you mean to say multiple mount points for the same filesystem?

agrover:
I think both are possible!

tasleson:
Yes both are possible, I guess I don't know of a use case for multiple FS mounted to the same mount point as only that last mounted FS is visible, but if it can be done it should be handled in the API :-)

agrover:
https://en.wikipedia.org/wiki/Union_mount#Linux I think e.g. Docker uses UnionFS

agrover:
BTW we're not actually even using `mnt` crate right now, we're parsing `df` output!!!

tasleson:
Oh, yes.  I forgot about union and overlay FS et. al.

tasleson:
Objective achieved with `mnt 0.3.1` and PR #625

PR #625, opened 2017-10-23
Use mnt crate

tasleson:
This PR should address https://github.com/stratis-storage/stratisd/issues/491

Using updated `0.3.1` mnt crate we have what we need to replace the fork & exec of `df`.

### DIFF
--- a/src/engine/strat_engine/filesystem.rs
+++ b/src/engine/strat_engine/filesystem.rs
@@ -3,11 +3,11 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 use std::path::{Path, PathBuf};
-use std::process::Command;
 
 use devicemapper::{Bytes, DmDevice, DmName, DM, IEC, SECTOR_SIZE, Sectors, ThinDev, ThinDevId,
                    ThinStatus, ThinPoolDev};
 
+use mnt::{MountParam, MountIter};
 use nix::sys::statvfs::statvfs;
 use nix::sys::statvfs::vfs::Statvfs;
 use nix::mount::{mount, MsFlags, umount};
@@ -149,20 +149,35 @@ impl StratFilesystem {
         current_size
     }
 
-    /// Get the mount_point for this filesystem
-    /// TODO Replace this code with something less brittle
+    /// Get one (non-deterministic in the presence of errors) of the mount_point(s) for the file
+    /// system that is contained on the block device referred to as self.devnode(), i.e. the device
+    /// node, while ignoring parse errors as long as at least one mount point is found.
     pub fn get_mount_point(&self) -> EngineResult<Option<PathBuf>> {
-        let output = Command::new("df")
-            .arg("--output=target")
-            .arg(&self.devnode())
-            .output()?;
-        if output.status.success() {
-            let output_str = String::from_utf8_lossy(&output.stdout);
-            return Ok(output_str.lines().last().map(PathBuf::from));
+        let device_node = self.devnode();
+        let search = device_node.to_str().ok_or_else(|| EngineError::Engine(ErrorEnum::Error,
+                                    format!("Unable to represent devnode as string {:?}", *self)))?;
+
+        let m_iter = MountIter::new_from_proc()
+            .map_err(|e| {
+                         EngineError::Engine(ErrorEnum::Error,
+                                             format!("Error reading /proc/mounts {:?}", e))
+                     })?;
+
+        let mut last_error: Option<String> = None;
+        for mp in m_iter {
+            match mp {
+                Ok(mount) => {
+                    if mount.contains(&MountParam::Spec(search)) {
+                        return Ok(Some(mount.file));
+                    }
+                }
+                Err(e) => {
+                    last_error = Some(format!("Error during parsing {:?} {:?}", *self, e));
+                }
+            }
         }
-        Err(EngineError::Engine(ErrorEnum::Error,
-                                format!("Failed to get filesystem mountpoint for {:?}",
-                                        self.devnode())))
+
+        last_error.map_or(Ok(None), |e| Err(EngineError::Engine(ErrorEnum::Error, e)))
     }
 
     /// Tear down the filesystem.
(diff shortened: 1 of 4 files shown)


tasleson:
The mnt crate is causing an issue, need to investigate.  Thanks @trgill

tasleson:
OK, my initial try at this was a massive fail, the `get_mount` from mnt crate does not work when you pass it a device node, not sure how I missed that.  I've added a new functionality to the library which does add what we need and submitted a PR: https://github.com/stemjail/mnt-rs/pull/1

tasleson:
Re-based and updated to leverage proposed: https://github.com/stemjail/mnt-rs/commits/feature-search-pr1

src/engine/strat_engine/filesystem.rs
```diff
@@ -150,19 +150,19 @@ impl StratFilesystem {
...
-            .output()?;
-        if output.status.success() {
-            let output_str = String::from_utf8_lossy(&output.stdout);
-            return Ok(output_str.lines().last().map(PathBuf::from));
+
+        let m_iter = MountIter::new_from_proc().expect("We expect to be able to open /proc/mounts");
+        match m_iter
+                  .map(|m| m.ok().unwrap())
+                  .find(|m| {
+                            m.contains(&MountParam::Spec(&String::from(self.devnode()
```
l0kod: you can remove the `String::from` now

tasleson:
@trgill Tests running for 20+ hours?

tasleson:
Re-based and using upstream `0.3.1`.

src/engine/strat_engine/filesystem.rs
```diff
@@ -150,19 +150,38 @@ impl StratFilesystem {
...
-        if output.status.success() {
-            let output_str = String::from_utf8_lossy(&output.stdout);
-            return Ok(output_str.lines().last().map(PathBuf::from));
+        if let Some(search) = self.devnode().to_str() {
+            match MountIter::new_from_proc() {
+                Ok(m_iter) => {
+                    for mp in m_iter {
+                        match mp {
+                            Ok(mount) => {
+                                if mount.contains(&MountParam::Spec(&search)) {
```
l0kod: …but here you will get the first match

src/engine/strat_engine/filesystem.rs
```diff
@@ -150,19 +150,38 @@ impl StratFilesystem {
...
     /// Get the mount_point for this filesystem
-    /// TODO Replace this code with something less brittle
     pub fn get_mount_point(&self) -> EngineResult<Option<PathBuf>> {
-        let output = Command::new("df")
-            .arg("--output=target")
-            .arg(&self.devnode())
-            .output()?;
-        if output.status.success() {
-            let output_str = String::from_utf8_lossy(&output.stdout);
-            return Ok(output_str.lines().last().map(PathBuf::from));
```
l0kod: I don't know if that matters but here you were getting the last match…

src/engine/strat_engine/filesystem.rs
```diff
@@ -150,19 +150,38 @@ impl StratFilesystem {
...
+        if let Some(search) = self.devnode().to_str() {
```
mulkieran: Generally you try to use the functions on Option, so, instead of the let-if, you would do:

```
let search = self.devnode().to_str().ok_or_else(|| EngineError::Engine(ErrorEnum::Error, format!("Unable to represent devnode as string...")))?;
let m_iter = MountIter::new_from_proc().map_err(|_| EngineError::Engine(ErrorEnum::Error, "Error opening /proc/mounts..."))?;
for mp in m_iter {
  // probably more to be done here....
}
```

src/engine/strat_engine/filesystem.rs
```diff
@@ -150,19 +150,38 @@ impl StratFilesystem {
...
+            match MountIter::new_from_proc() {
+                Ok(m_iter) => {
+                    for mp in m_iter {
+                        match mp {
+                            Ok(mount) => {
+                                if mount.contains(&MountParam::Spec(search)) {
```
mulkieran: I guess there couldn't possibly be two matches for search?

tasleson: There could be `N` matching items for search, but all we are trying to do is find one so we can call `xfs` expand passing it as one of the arguments.

mulkieran: If any are equally good, then it is better not to be eager about returning an error if there was a parse failure on any entry.

src/engine/strat_engine/filesystem.rs
```diff
@@ -150,19 +150,38 @@ impl StratFilesystem {
     }
 
     /// Get the mount_point for this filesystem
```
mulkieran: If there are multiple possible mountpoints and you are looking for any one of them, then the definite article is a poor choice here. I think the header could stand to be updated a bit.

tasleson:
@mulkieran Please take another peek, thanks!

src/engine/strat_engine/filesystem.rs
tasleson: No, it shouldn't matter.  Thanks for the thorough review!

src/engine/strat_engine/filesystem.rs
```diff
@@ -149,20 +149,39 @@ impl StratFilesystem {
...
+        let search = device_node.to_str().ok_or_else(|| EngineError::Engine(ErrorEnum::Error,
+                                    format!("Unable to represent devnode as string {:?}", *self)))?;
+
+        let m_iter = MountIter::new_from_proc()
+            .map_err(|e| {
+                         EngineError::Engine(ErrorEnum::Error,
+                                             format!("Error opening /proc/mounts {:?}", e))
+                     })?;
+
+        for mp in m_iter {
```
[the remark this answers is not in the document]
mulkieran: You could probably simplify by making type of ```last_error``` ```Option<String>``` since the only thing that might vary is the message.

tasleson:
@mulkieran All suggestions completed, please take another peek, thanks!

src/engine/strat_engine/filesystem.rs
```diff
@@ -145,24 +145,40 @@ impl StratFilesystem {
     /// Return an extend size for the thindev under the filesystem
     /// TODO: returning the current size will double the space provisoned to
     /// the thin device.  We should determine if this is a reasonable value.
+    ///
```
mulkieran: Accidental extra line introduced.

tasleson: @mulkieran  Ugh, not sure how that got in there, thanks!  PR spun again.

agrover:
no add'l issues, 👍, merging as last reviewer. Thanks!

Issue #626, opened 2017-10-24
Blockdev DBus objectpaths not removed when pool is removed


agrover:
We should automatically remove these when the pool is removed. The hard part is determining their object paths from `destroy_pool`. We can get to the Pool's Blockdevs but they don't remember their objectpaths. If we could iterate the Tree then we could remove all Blockdevs whose Pool property refers to the pool being destroyed, but the tree is not iterable.

I'm currently thinking if we add a `children: Vec<ObjectPath>` (or whatever type) to the pool's context then that would let us do it.

mulkieran:
Related: #644.

mulkieran:
We believe that this was fixed by #660.

PR #628, opened 2017-10-24
Fix pool destroy for blockdevs

agrover:
We need to remove blockdev objectpaths from DBus when the pool is destroyed. Unfortunately, this is hard for the reasons stated in #626.

### DIFF
--- a/src/dbus_api/blockdev.rs
+++ b/src/dbus_api/blockdev.rs
@@ -20,7 +20,7 @@ use uuid::Uuid;
 use super::super::engine::BlockDev;
 use super::super::engine::types::BlockDevState;
 
-use super::types::{DbusContext, DbusErrorEnum, OPContext, TData};
+use super::types::{DbusContext, DbusErrorEnum, OPContext, OPType, TData};
 
 use super::util::STRATIS_BASE_PATH;
 use super::util::STRATIS_BASE_SERVICE;
@@ -89,7 +89,8 @@ pub fn create_dbus_blockdev<'a>(dbus_context: &DbusContext,
 
     let interface_name = format!("{}.{}", STRATIS_BASE_SERVICE, "blockdev");
 
-    let object_path = f.object_path(object_name, Some(OPContext::new(parent, uuid)))
+    let object_path = f.object_path(object_name,
+                                    Some(OPContext::new(parent, uuid, OPType::Blockdev)))
         .introspectable()
         .add(f.interface(interface_name, ())
                  .add_m(set_userid_method)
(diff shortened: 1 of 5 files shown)


agrover:
Hmm, well we're handling filesystems differently -- if any exist, you can't even remove the pool.

If the larger issue is that OPContext has fields that are invalid for different objectpaths, maybe there's a better way. Let me take another stab.

mulkieran:
> Hmm, well we're handling filesystems differently -- if any exist, you can't even remove the pool.

This is certainly true. But a pool may develop other sorts of children, somehow.

> If the larger issue is that OPContext has fields that are invalid for different objectpaths, maybe there's a better way. Let me take another stab.

I think we really want to be consistent with our parent/child relationships. But you are right that some auxiliary data structure does seem to be required in the OPContext.

src/dbus_api/blockdev.rs
```diff
@@ -31,10 +32,10 @@ use super::util::get_uuid;
 use super::util::ok_message_items;
 
 
-pub fn create_dbus_blockdev<'a>(dbus_context: &DbusContext,
-                                parent: dbus::Path<'static>,
-                                uuid: Uuid)
-                                -> dbus::Path<'a> {
+pub fn create_dbus_blockdev(dbus_context: &DbusContext,
```
mulkieran: Rather than pulling the code out of ```create_dbus_blockdev``` I think you're better off putting more in. Is it possible to add a parameter that refers to the list that needs to be updated here?

agrover:
Here's an additional commit that adds that auxiliary data structure.

I'll try as you suggest and put more stuff in `create_dbus_blockdev` and friends, see how that works out.

I think there may be more to your initial comment I'm not addressing but I'm not quite getting what you're saying. If so, can you restate?

agrover:
* fix clippy
* Remove commit that changed create_dbus_{filesystem,blockdev} to do less
* Changed create_dbus_blockdev to also add Path to the Pool.children hashset

agrover:
@mulkieran review?

mulkieran:
Is it possible to call ```get_manage_objects``` method internally in stratisd in order to find all things that refer to thing being removed and remove them? If possible, that might be a much better approach.

agrover:
Like I mentioned in #626 there are a couple different options for fixing this, and using GetManagedObjects is another one. (FWIW I don't think using the D-Bus interface in implementing the D-Bus interface is a good idea. I personally prefer making `Tree` iterable and inspecting that, but that requires an upstream change.) As with all PRs, no code is forever, and maybe this code will be superseded, but right now this code solves the issue and is better than the status quo.

agrover:
Based on our conversation, I will investigate upstream dbus changes that could make this easier to implement cleanly.

agrover:
new dbus-rs features obsolete this PR, closing.

PR #629, opened 2017-10-24
Reduce spurious test output

### LINKED ISSUE #605: Would be nice to find a way to make xfs_growfs a bit more quiet
It's a bit of an inconvenience in testing output and might be inconvenient in the real world. But I don't think it's really urgent.

tasleson:
xfs command line tool was dumping output that was ending up in the
test output, changed how we run them to remove output.  We may want
to do this for all of the command line tools we exec.

Resolves #605


mulkieran:
The code looks fine to me. ```mkfs.xfs``` is certainly a candidate for this change as well. Perhaps those in other files as well.

agrover:
👍 MALV (merging as last reviewer) Thanks @tasleson !

PR #630, opened 2017-10-25
Do more clean up in unit tests by leveraging the drop trait more

tasleson:
Add code in the drop trait so that when it runs we:

* Remove all mounted file systems that were created from the unit test
* ~Remove all the dm tables using `dmsetup remove_all --force`~
* Remove only those dm devices which following the stratis naming convention

~Tested with numerous unit tests that panic and we do appear to clean up the system.  However, we may end up causing a kernel oops in the process, ref: https://bugzilla.redhat.com/show_bug.cgi?id=1506287~

When we tear down the dm devices in the correct order we don't oops the kernel.

~Note: This change build up on PR https://github.com/stratis-storage/stratisd/pull/625 as it uses the mnt crate too.~
Note: Using standard released mount crate  and I've removed the changes for https://github.com/stratis-storage/stratisd/pull/625

### DIFF
--- a/src/engine/strat_engine/tests/util.rs
+++ b/src/engine/strat_engine/tests/util.rs
@@ -1,0 +1,51 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use std::path::PathBuf;
+
+use devicemapper::{DM, DmFlags, DevId};
+
+use nix::mount::umount;
+use mnt::get_submounts;
+
+/// Attempt to remove all device mapper devices which match the stratis naming convention.
+/// FIXME: Current implementation complicated by https://bugzilla.redhat.com/show_bug.cgi?id=1506287
+fn dm_stratis_devices_remove() {
+
+    let dm = DM::new().unwrap();
+
+    loop {
+        let mut progress_made = false;
+        for d in dm.list_devices()
+                .unwrap()
+                .iter()
+                .filter(|d| format!("{}", d.0.as_ref()).starts_with("stratis-1")) {
+            progress_made |= dm.device_remove(&DevId::Name(&d.0), DmFlags::empty())
+                .is_ok();
+        }
+
+        if !progress_made {
+            break;
+        }
+    }
+}
+
+/// Try and un-mount any filesystems that have the name stratis in the mount point.
+fn stratis_filesystems_unmount() {
+    for m in get_submounts(&PathBuf::from("/"))
+            .unwrap()
+            .iter()
+            .filter(|m| m.file.to_str().map_or(false, |s| s.contains("stratis"))) {
+        umount(&m.file).unwrap();
+    }
+}
+
+/// When a unit test panics we can leave the system in an inconsistent state.  This function
+/// tries to clean up by un-mounting any mounted file systems which contain the string
+/// "stratis_testing" and then it tries to remove any device mapper tables which are also stratis
+/// created.
+pub fn clean_up() -> () {
+    stratis_filesystems_unmount();
+    dm_stratis_devices_remove();
+}
--- a/src/engine/strat_engine/tests/loopbacked.rs
+++ b/src/engine/strat_engine/tests/loopbacked.rs
@@ -14,6 +14,7 @@ use self::loopdev::{LoopControl, LoopDevice};
 
 use super::logger::init_logger;
 use super::tempdir::TempDir;
+use super::util::clean_up;
 
 use super::super::device::wipe_sectors;
 
@@ -34,6 +35,7 @@ impl LoopTestDev {
     /// Create a new loopbacked device.
     /// Create its backing store of 1 GiB wiping the first 1 MiB.
     pub fn new(lc: &LoopControl, path: &Path) -> LoopTestDev {
+        clean_up();
         let mut f = OpenOptions::new()
             .read(true)
             .write(true)
@@ -71,6 +73,7 @@ impl LoopTestDev {
 
 impl Drop for LoopTestDev {
     fn drop(&mut self) {
+        clean_up();
         self.detach()
     }
 }
--- a/src/engine/strat_engine/tests/mod.rs
+++ b/src/engine/strat_engine/tests/mod.rs
@@ -7,5 +7,6 @@ extern crate log;
 pub extern crate tempdir;
 
 mod logger;
+mod util;
 pub mod loopbacked;
 pub mod real;
--- a/src/engine/strat_engine/tests/real.rs
+++ b/src/engine/strat_engine/tests/real.rs
@@ -13,6 +13,7 @@ use serde_json::{Value, from_reader};
 use devicemapper::{Bytes, IEC, Sectors};
 
 use super::logger::init_logger;
+use super::util::clean_up;
 
 use super::super::device::wipe_sectors;
 
@@ -24,6 +25,7 @@ impl RealTestDev {
     /// Construct a new test device for the given path.
     /// Wipe initial MiB to clear metadata.
     pub fn new(path: &Path) -> RealTestDev {
+        clean_up();
         wipe_sectors(path, Sectors(0), Bytes(IEC::Mi).sectors()).unwrap();
         RealTestDev { path: PathBuf::from(path) }
     }
@@ -36,6 +38,7 @@ impl RealTestDev {
 
 impl Drop for RealTestDev {
     fn drop(&mut self) {
+        clean_up();
         wipe_sectors(&self.path, Sectors(0), Bytes(IEC::Mi).sectors()).unwrap();
     }
 }


trgill:
Doesn't Fedora install over LVM/DM devices by default?  My test VM has:

# dmsetup ls
fedora-swap	(253:1)
fedora-root	(253:0)

FYI - I used this script to cleanup our Jenkins test systems after failures - it is a bit ugly...

umount `df | awk '{print $6}' | fgrep stratis`
dmsetup remove `dmsetup ls | awk '{print $1}' | fgrep thin-fs`
dmsetup remove `dmsetup ls | awk '{print $1}' | fgrep stratis_test_filesystem`
dmsetup remove `dmsetup ls | awk '{print $1}' | fgrep mdv`
dmsetup remove `dmsetup ls | awk '{print $1}' | fgrep pool`
dmsetup remove `dmsetup ls | awk '{print $1}' | fgrep flex-thinmeta`
dmsetup remove `dmsetup ls | awk '{print $1}' | fgrep flex-thindata`

tasleson:
@trgill It does, I guess I need to be more specific in the interaction with `dm` so we don't blitz everything.  However, I thought we were instructing people to **not** utilize lvm & dm if they want to hack on stratis.

mulkieran:
If you're using a text editor you can still run cargo fmt and get the benefit.

```make fmt``` should tell you if there is anything wrong, but I usually just use ```PATH=${HOME}/.cargo/bin:${PATH} cargo fmt``` to do the change (having run ```make fmt``` at some time in the past to install the correct version).

File: src/engine/strat_engine/filesystem.rs [hunk 1]
```diff
@@ -145,19 +145,22 @@ impl StratFilesystem {
...
-    /// TODO Replace this code with something less brittle
     pub fn get_mount_point(&self) -> EngineResult<Option<PathBuf>> {
-        let output = Command::new("df")
-            .arg("--output=target")
-            .arg(&self.devnode())
-            .output()?;
-        if output.status.success() {
-            let output_str = String::from_utf8_lossy(&output.stdout);
-            return Ok(output_str.lines().last().map(PathBuf::from));
+        match get_mount_search(&Search::Spec(String::from(self.devnode().to_str().unwrap_or("")))) {
```
mulkieran: I can't even begin to comment on this part until mnt PR settles down a bit.

src/engine/strat_engine/tests/util.rs
```diff
@@ -0,0 +1,51 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use std::env;
+use std::path::PathBuf;
+use std::process::Command;
+
+use nix::mount::{umount};
+use mnt::{get_submounts};
+
+pub fn dm_remove_all() -> () {
```
mulkieran: I like header comments that specify what a function is supposed to accomplish.

src/engine/strat_engine/tests/util.rs
```diff
@@ -0,0 +1,51 @@
...
+                     .expect("failed to execute process dmsetup");
+
+    if !output.status.success() {
+        eprintln!("dmsetup status: {}", output.status);
+        eprintln!("dmsetup stdout: {}", String::from_utf8_lossy(&output.stdout));
+        eprintln!("dmsetup stderr: {}", String::from_utf8_lossy(&output.stderr));
+    }
+}
+
+pub fn clean_up() -> () {
```
mulkieran: Header comment here also would be good.

src/engine/strat_engine/tests/util.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    // To allow debugging, especially when running individual test cases lets offer an option
+    // to not clean up
+    if env::var("STRATIS_TEST_SKIP_CLEAN_UP").is_ok() {
+        eprintln!("Skipping clean up as requested");
+        return;
+    }
+
+    // Do our best to clean up after a unit test
```
mulkieran: If you had a precise header comment you could remove this comment, and state precisely what "cleanup" means in the header comment.

src/engine/strat_engine/tests/util.rs
```diff
@@ -0,0 +1,51 @@
...
+    if !output.status.success() {
+        eprintln!("dmsetup status: {}", output.status);
+        eprintln!("dmsetup stdout: {}", String::from_utf8_lossy(&output.stdout));
+        eprintln!("dmsetup stderr: {}", String::from_utf8_lossy(&output.stderr));
+    }
+}
+
+pub fn clean_up() -> () {
+
+    // To allow debugging, especially when running individual test cases lets offer an option
```
mulkieran: Unless you're using this already in your own work, please leave it out.

src/engine/strat_engine/tests/util.rs
mulkieran: If it isn't necessary to make this public you could leave it private.

trgill:
@tasleson yeah, you are right about not using LVM/dm, but not sure we make that as clear as we should if we are going to remove all DM devices as test cleanup.  FYI - some of the tests mount filesystems - I don't think the "remove all" will work if it is mounted.

tasleson:
@trgill That's why the `chean_up` function removes all mounted stratis file systems before removing all the dm tables.  I'll re-work this PR to selectively remove only stratis dm tables, so we don't bork someones system.  However, in reality they really shouldn't be using dm for production data when stratis is in development and mucking with them.

tasleson:
@trgill @mulkieran  I've re-worked this change incorporating suggestions from both of you.

tasleson:
@trgill @mulkieran OK I've made some more changes, I missed some from my last push.

I wanted to use `lazy_static` for the regular expression but was unable to get the crate to work and using the `regex` crate was being difficult too.  Additionally I was unable to make the `clean_up` function not public for it to work for me.

I'm going to go back and re-read the chapter on crates as obviously I'm missing something on how I should be structuring things.

tasleson:
@mulkieran Looks like `clippy` is having problems :confused:

mulkieran:
@tasleson #636.

tasleson:
@mulkieran OK, I've done as you suggested, please take a peek, thanks!

src/engine/strat_engine/tests/util.rs
```diff
@@ -0,0 +1,70 @@
...
+        }
+        Err(e) => eprintln!("Error on DM::new() {:?}", e),
+    }
+}
+
+/// When a unit test panics we can leave the system in an inconsistent state.  This function
+/// tries to clean up by un-mounting any mounted file systems which contain the string
+/// "stratis_testing" and then it removes any device mapper tables which are also stratis created.
+pub fn clean_up() -> () {
+    match get_submounts(&PathBuf::from("/")) {
```
mulkieran: You could make this block into a separate function, like ```unmount_stratis_filesystems```.

You will want to check that all filesystems made by tests are actually so conveniently named. I don't think I've wriiten too many filesystem tests, but my name choices tend to be completely random.

If a stratis filesystem can't be unmounted isn't the next step, getting rid of the devicemapper devices, at least partially doomed?

src/engine/strat_engine/tests/util.rs
```diff
@@ -0,0 +1,70 @@
...
+use devicemapper::{DM, DmFlags, DevId};
+
+use nix::mount::umount;
+use mnt::get_submounts;
+
+/// Remove all device mapper devices which have a name which matches the naming convention for
+/// a stratis created device.
+fn dm_stratis_devices_remove() {
+    match DM::new() {
+        Ok(dm) => {
```
mulkieran: While we never unwrap in source code, we unwrap everywhere both in the test code and in the test infrastructure. We could reconsider the policy, but we shouldn't do it piecemeal. So, for instance, ```DM::new()``` should definitely just be unwrapped.

src/engine/strat_engine/tests/util.rs
```diff
@@ -0,0 +1,70 @@
...
+            // Lets assume that nesting won't go more than 10 deep...
+            // DM::table_deps returns a devices parent, not child which is what we would ideally
+            // want to systematically remove the devices in the correct order.
+            for _ in 0..10 {
+                let mut remain = 0;
```
mulkieran: ```let mut remaining = false;```

src/engine/strat_engine/tests/util.rs
```diff
@@ -0,0 +1,70 @@
...
+
+                match dm.list_devices() {
+                    Ok(devices) => {
+                        for d in devices {
```
mulkieran:
```
remaining = devices.iter().map(|d| d.0).filter(|d| format!(d.as_ref()).starts_with("...")).map(|d| dm.device_remove(...)).any(|r| r.is_err());
```

src/engine/strat_engine/tests/util.rs
```diff
@@ -0,0 +1,70 @@
...
+                            }
+                        }
+                    }
+                    Err(e) => {
+                        eprintln!("Error on list_devices() {:?}", e);
+                        break;
+                    }
+                }
+
+                if remain == 0 {
```
mulkieran:
```
if remaining {
    break;
}
```

src/engine/strat_engine/tests/util.rs
```diff
@@ -0,0 +1,61 @@
...
+use nix::mount::umount;
+use mnt::get_submounts;
+
+/// Remove all device mapper devices which have a name which matches the naming convention for
+/// a stratis created device.
+fn dm_stratis_devices_remove() {
+
+    let dm = DM::new().unwrap();
+
+    // Lets assume that nesting won't go more than 10 deep...
```
mulkieran: You could drop this comment and put the truer facts in the header. Something like:

```
/// iteratively remove device mapper devices which have a name which matches the naming convention for a stratis-created device.
/// This method does at most ten passes, removing all devices which are not depended on by another device at each pass.
```

src/engine/strat_engine/tests/util.rs
```diff
@@ -0,0 +1,61 @@
...
+        }
+    }
+}
+
+/// Try and un-mount any filesystems that have the name stratis in the mount point.
+fn stratis_filesystems_unmount() {
+    match get_submounts(&PathBuf::from("/")) {
+        Ok(mp) => {
+            for m in mp.iter()
+                    .filter(|m| m.file.to_str().unwrap_or("").contains("stratis")) {
```
mulkieran: ```m.file.to_str().map_or(false, |s| s.contains("stratis"))``` is somewhat better Rust.

src/engine/strat_engine/tests/util.rs
tasleson: IMHO the header documentation should state what it does, not how it does it.  When we get a fix for the kernel oops bug, then there will be no more for loop, just a single pass across the devices that match.

src/engine/strat_engine/tests/util.rs
tasleson: Done

tasleson:
@mulkieran @trgill @agrover I added `eprintln!` to the code that tries to clean up the mess left by a panic which occurs during the unit test.  We really just want to do the best we can.  In testing we appear to clean up nicely without any of the error statements firing.  However, IMHO we should like to know if the cleanup code isn't working without causing a failure there from causing unit test to fail, but maybe they should?

src/engine/strat_engine/tests/util.rs
mulkieran: I agree with the sentiment. But unfortunately the header comment says cleans up _all_ devices and "cleans up some devices" seems a bit slapdash.

What about keep cleans up all, with a FIXME indicating that it will really and truly clean up all when it is able to call ```DM::remove_all()``` ?

src/engine/strat_engine/tests/util.rs
```diff
@@ -0,0 +1,60 @@
...
+use nix::mount::umount;
+use mnt::get_submounts;
+
+/// Remove all device mapper devices which have a name which matches the naming convention for
+/// a stratis created device.
+fn dm_stratis_devices_remove() {
+
+    let dm = DM::new().unwrap();
+
+    // TODO: Remove this for loop when we can utilize the DM_DEFERRED_REMOVE flag and not cause
```
mulkieran: Change TODO inline to FIXME in header as:

```
FIXME: When remove_devices() can be called with DM_DEFERRED_REMOVE flag without causing a kernel oops
then this method will really remove _all_ devices.
```
You could also include a link to the bz.

mulkieran:
I vote we change those eprintln's to unwraps. We can always reconsider.

tasleson:
@mulkieran My concern with changing the `eprintln!` to unwrap, is that do we really want to panic in the middle of cleaning up a panic?  The `clean_up` function is running before and after the unit test, to try and ensure that we are in a good consistent state before we run and that we leave it in a good state when done too.

tasleson:
@mulkieran Re-worked dm device removal, should handle most cases better and I updated the comment.  Still not sure about `unwrap` vs. `eprintln!`

tasleson:
@mulkieran I've removed the `eprintln!` and everything has been converted to `unwrap`.

mulkieran:
Last reviewer, merging.

Issue #627, opened 2017-10-24
`stratis pool list` only works second time with freshly started daemon


agrover:
There seems to be some sort of startup-related thing that is causing a first request to get all objects properly exported on DBus.

1. start stratisd
1. create a pool (if one doesn't exist)
1. restart stratisd
1. `stratis pool list` 1st time shows nothing
1. `stratis pool list` 2nd time shows the existing pool

PR #638, opened 2017-10-30
dbus api: Handle deferred at start up

tasleson:
When api::connect is executed the code iterates over the engine processing
all the objects.  It makes calls to create_dbus_pool, create_dbus_filesystem,
and create dbus_block_dev.  The problem is all of these functions add the
changes to the "action" queue.  Thus when then the first client does any
method call it will appear the system is empty until after we process
the deferred event queue after handling the initial request.

This change moves the deferred processing into a separate function and
calls it in api::connect before returning, so that the tree has all the
queued objects applied to it.

Resolves: #627

### DIFF
--- a/src/dbus_api/api.rs
+++ b/src/dbus_api/api.rs
@@ -34,7 +34,7 @@ use stratis::VERSION;
 use super::filesystem::create_dbus_filesystem;
 use super::blockdev::create_dbus_blockdev;
 use super::pool::create_dbus_pool;
-use super::types::{DeferredAction, DbusContext, DbusErrorEnum, TData};
+use super::types::{ActionQueue, DeferredAction, DbusContext, DbusErrorEnum, TData};
 use super::util::STRATIS_BASE_PATH;
 use super::util::STRATIS_BASE_SERVICE;
 use super::util::code_to_message_items;
@@ -262,7 +262,7 @@ pub fn connect(engine: Rc<RefCell<Engine>>)
 
     let local_engine = Rc::clone(&engine);
 
-    let (tree, object_path) = get_base_tree(DbusContext::new(engine));
+    let (mut tree, object_path) = get_base_tree(DbusContext::new(engine));
     let dbus_context = tree.get_data().clone();
 
     // This should never panic as create_dbus_pool(),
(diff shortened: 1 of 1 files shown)


src/dbus_api/api.rs
```diff
@@ -282,9 +282,31 @@ pub fn connect(engine: Rc<RefCell<Engine>>)
 
     c.register_name(STRATIS_BASE_SERVICE, NameFlag::ReplaceExisting as u32)?;
 
+    process_deferred_actions(&c, &mut tree, &dbus_context)?;
+
     Ok((c, tree, dbus_context))
 }
 
+fn process_deferred_actions(c: &Connection,
```
mulkieran: Please put header comment on this method.

src/dbus_api/api.rs
```diff
@@ -299,19 +321,8 @@ pub fn handle(c: &Connection,
...
-                    c.register_object_path(path.get_name())?;
-                    tree.insert(path);
-                }
-                DeferredAction::Remove(path) => {
-                    c.unregister_object_path(&path);
-                    tree.remove(&path);
-                }
-            }
-        }
+        // Take all adds/removes from tree and handle them
```
mulkieran: Please drop this comment.

src/dbus_api/api.rs
mulkieran: Are you able to pass only the actions, not the whole context.?

tasleson: I tried, but after spending way too much time I gave up.

src/dbus_api/api.rs
tasleson: Done

src/dbus_api/api.rs
tasleson: Done

agrover:
@tasleson I pushed a commit to pass actions instead of entire DbusContext, please take a look.

tasleson:
@agrover LGTM, Thanks!

agrover:
ok cool. If you'd be so kind as to squash it into the previous commit, we may be ready for final review.

tasleson:
@agrover Commit squashed and I re-based, so hopefully this will pass CI with a new updated clippy.

agrover:
👍 , merging even though clippy is broken. Thanks!

PR #640, opened 2017-11-01
Make process_deferred_actions take a simple ActionQueue argument

mulkieran:
The advantages of this are that process_deferred_actions has a cleaner
interface, and the borrow_mut()'s are all in the places that they occur
rather than hidden in process_deferred_actions. This makes it a bit easier
to track them and to make sure that they do not conflict with other
borrows.


agrover:
👍

mulkieran:
This is only a smidge short of trivial, and it's acked, so I'll go ahead and merge.

Issue #591, opened 2017-10-04
Remove constants from Manager interface once stratis-cli v0.0.4 is released


mulkieran:
And then add them to the D-Bus API doc.

PR #641, opened 2017-11-02
Remove obsoleted constants

mulkieran:
Instead of publishing the constants dynamically we publish them statically,
in the D-Bus API reference.

Resolves: #591.


mulkieran:
Once this is done, we should update the D-Bus API Reference.

PR #649, opened 2017-11-13
Fixes and alterations to some tests

mulkieran:
Somewhat miscellaneous fixes to tests:

* Use assert_eq, assert_neq macros where possible in tests.
* Use umount2 and MNT_DETACH in test infrastructure to support loopbacked tests slightly better.
* Omit the code that is there just for cleanup in tests. We expect that the cleanup code in the test infrastructure should do all the cleaning up that is necessary.

### DIFF
--- a/src/engine/strat_engine/blockdevmgr.rs
+++ b/src/engine/strat_engine/blockdevmgr.rs
@@ -538,24 +538,24 @@ mod tests {
         BlockDevMgr::initialize(uuid1, paths1, MIN_MDA_SECTORS, false).unwrap();
 
         let pools = find_all().unwrap();
-        assert!(pools.len() == 1);
+        assert_eq!(pools.len(), 1);
         assert!(pools.contains_key(&uuid1));
         let devices = pools.get(&uuid1).expect("pools.contains_key() was true");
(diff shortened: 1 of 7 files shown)


mulkieran:
@agrover @trgill @tasleson review, please?

trgill:
I was thinking the cleanup code in tests was a good practice.  But, if others feel this is better - it is fine with me.

tasleson:
> I was thinking the cleanup code in tests was a good practice. But, if others feel this is better - it is fine with me.

I agree with this as it gives us additional opportunities to test the code to tear down devices etc.  The code we added to ensure the test environment was consistent exercises totally different code paths.  I suppose one could make the argument that those tear down functions have there own unit tests and that exercising them here is redundant.

mulkieran:
> > I was thinking the cleanup code in tests was a good practice. But, if others feel this is better - it is fine with me.

> I agree with this as it gives us additional opportunities to test the code to tear down devices etc.  The code we added to ensure the test environment was consistent exercises totally different code paths.  I suppose one could make the argument that those tear down functions have their own unit tests and that exercising them here is redundant.

With this change we still continue to exercise the teardown functions as part of individual tests

PR #651, opened 2017-11-14
Do not use unwrap in clean_up

tasleson:
~When we are trying to clean up the test environment after a panic if
we panic again we no long get anything of use from the unit test.  The
process aborts and we get no back trace or anything which we can
leverage to figure out what happened.~

Updated:

    Before we execute the unit test we will do a sanity check on the
    test environment which will panic and end that unit test if bad.  If
    the environment is good we will run the test capturing the result
    and then clean up capturing its result.  We will then ensure that
    the test and the post clean ups were successful as well.  This
    approach removes the potential of a panic during a panic and allowing
    us to know when we encounter an issue.

### DIFF
--- a/src/engine/strat_engine/tests/util.rs
+++ b/src/engine/strat_engine/tests/util.rs
@@ -4,48 +4,90 @@
 
 use std::path::PathBuf;
 
-use devicemapper::{DM, DmFlags, DevId};
+use devicemapper::{DM, DmFlags, DmResult, DevId};
 
 use nix::mount::{MNT_DETACH, umount2};
 use mnt::get_submounts;
 
-/// Attempt to remove all device mapper devices which match the stratis naming convention.
-/// FIXME: Current implementation complicated by https://bugzilla.redhat.com/show_bug.cgi?id=1506287
-fn dm_stratis_devices_remove() {
 
-    let dm = DM::new().unwrap();
+mod cleanup_errors {
+    use mnt;
+    use nix;
 
-    loop {
-        let mut progress_made = false;
-        for d in dm.list_devices()
-                .unwrap()
-                .iter()
-                .filter(|d| format!("{}", d.0.as_ref()).starts_with("stratis-1")) {
-            progress_made |= dm.device_remove(&DevId::Name(&d.0), DmFlags::empty())
-                .is_ok();
-        }
-
-        if !progress_made {
-            break;
+    error_chain!{
+        foreign_links {
+            Mnt(mnt::ParseError);
+            Nix(nix::Error);
         }
     }
 }
 
-/// Try and un-mount any filesystems that have the name stratis in the mount point.
-fn stratis_filesystems_unmount() {
-    for m in get_submounts(&PathBuf::from("/"))
-            .unwrap()
-            .iter()
-            .filter(|m| m.file.to_str().map_or(false, |s| s.contains("stratis"))) {
-        umount2(&m.file, MNT_DETACH).unwrap();
+use self::cleanup_errors::{Error, Result, ResultExt};
+
+
+/// Attempt to remove all device mapper devices which match the stratis naming convention.
+/// FIXME: Current implementation complicated by https://bugzilla.redhat.com/show_bug.cgi?id=1506287
+fn dm_stratis_devices_remove() -> Result<()> {
+
+    /// One iteration of removing devicemapper devices
+    fn one_iteration(dm: &DM) -> DmResult<(bool, Vec<String>)> {
+        let mut progress_made = false;
+        let mut remain = Vec::new();
+
+        for d in dm.list_devices()?
+                .iter()
+                .filter(|d| format!("{}", d.0.as_ref()).starts_with("stratis-1")) {
+
+            match dm.device_remove(&DevId::Name(&d.0), DmFlags::empty()) {
+                Ok(_) => progress_made = true,
+                Err(_) => remain.push(format!("{}", d.0.as_ref())),
+            }
+        }
+        Ok((progress_made, remain))
     }
+
+    let dm = DM::new().chain_err(|| "Unable to initialize DM")?;
+
+    loop {
+        let (progress_made, remain) = one_iteration(&dm)
+            .map_err(|e| {
+                Error::with_chain(e,
+                                  "Error while attempting to remove stratis device mapper devices")}
+                )?;
+
+        if !progress_made {
+            if remain.len() != 0 {
+                bail!("We were unable to remove all stratis device mapper devices {:?}",
+                      remain);
+            }
+            break;
+        }
+    }
+
+    Ok(())
+}
+
+/// Try and un-mount any filesystems that have the name stratis in the mount point, returning
+/// immediately on the first one we are unable to unmount.
+fn stratis_filesystems_unmount() -> Result<()> {
+    || -> Result<()> {
+        let mounts = get_submounts(&PathBuf::from("/"))?;
+        for m in mounts
+                .iter()
+                .filter(|m| m.file.to_str().map_or(false, |s| s.contains("stratis"))) {
+            umount2(&m.file, MNT_DETACH)?;
+        }
+        Ok(())
+    }()
+            .map_err(|e| Error::with_chain(e, "unable to unmount all stratis filesystems"))
 }
 
 /// When a unit test panics we can leave the system in an inconsistent state.  This function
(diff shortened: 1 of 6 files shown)


mulkieran:
I think the following might be a better solution.

Have the functions that actually perform the tests return a result type. For example, ```test_pool_rename``` which now returns nothing, should return a, for example, TestResult. It should be possible to easily define this with enough From implementations to make changing the tests themselves fairly simple. Much would simply involve replacing the string ```.unwrap()``` with a ```?```. In the cases where the ```unwrap```s unwrap an ```Option``` type there would be an extra step.

Have the ```test_with_spec``` methods, loopbacked and real, capture the result of running the test, whether error or ```Ok(())```.

Run the ```clean()``` function in ```test_with_spec```. Ideally, the clean methods and perhaps some other methods can return their own TestInfraError, possibly constructed via error-chain for the stack trace benefits that can be derived therefrom. If the ```clean()``` method returns an error, print it out using Debug format string.

Unwrap the result previously obtained from the actual test.

tasleson:
@mulkieran  It's not difficult to remove the `unwraps` and propagate the error up as you suggest with a `?`, but what are you proposing for the `assert*` usage in the unit tests themselves, the `test` cannot capture the panic...

Well it could if we could get something like the following to work:
``` rust
 let panic_result = panic::catch_unwind(|| {
        test_result = test(&devices.iter().map(|x| x.as_path()).collect::<Vec<_>>());
 });
```

I saw this [before](https://medium.com/@ericdreichert/test-setup-and-teardown-in-rust-without-a-framework-ba32d97aa5ab) and thought we could adopt something similar, but I was unable to get it to compile for our use case.

mulkieran:
Maybe ```panic_unwind``` is the right choice. Here's something that compiles: https://github.com/mulkieran/stratisd/tree/master-panic-unwind. Not in any way complete.

tasleson:
@mulkieran OK, I've utilized what you had with a couple of small changes, how does this look?

tasleson:
@mulkieran Error chain utilized, `clean_up` returning a result, please take a peek, thanks!

src/lib.rs
```diff
@@ -23,9 +23,14 @@ extern crate serde_json;
...
 extern crate log;
 
+#[cfg(test)]
+#[macro_use]
+extern crate error_chain;
+
 #[cfg(test)]
 extern crate quickcheck;
 
+
```
mulkieran: Stray line?

src/engine/strat_engine/tests/util.rs
```diff
@@ -9,43 +9,72 @@ use devicemapper::{DM, DmFlags, DevId};
...
         }
 
         if !progress_made {
+            if remain.len() != 0 {
+                bail!("We were unable to remove all stratis device mapper devices {:?}",
+                      remain);
+            }
             break;
+        } else {
+            remain.clear();
```
mulkieran: Think this qualifies as premature optimization. Why not just initialize the vector in the loop?

src/engine/strat_engine/tests/util.rs
```diff
@@ -9,43 +9,72 @@ use devicemapper::{DM, DmFlags, DevId};
...
     }
+
+    Ok(())
 }
 
 /// Try and un-mount any filesystems that have the name stratis in the mount point.
-fn stratis_filesystems_unmount() {
-    for m in get_submounts(&PathBuf::from("/"))
-            .unwrap()
+fn stratis_filesystems_unmount() -> Result<()> {
```
mulkieran: For header comment, please add something like: ```Exit immediately on failure to unmount a filesystem.```

src/engine/strat_engine/tests/util.rs
```diff
@@ -9,43 +9,72 @@ use devicemapper::{DM, DmFlags, DevId};
...
+
+    let mounts = get_submounts(&PathBuf::from("/"))
```
mulkieran: For the entire body of this method, I would consider doing something more like:

```
|| {
  for m in get_submounts(&PathBuf::from("/"))?.iter().filter(...) {
    umount2(...)?; 
  }
}().map_err(|e| Error::with_chain(e, "Failed to unmount all mounted filesystems belonging to stratis")
```

or a fancier version that was able to report the unmounted filesystems.

The benefit is that you're able to chain onto the other errors a higher level error which describes the way in which this method, the one in which you're doing the chaining, failed. I would consider trying something similar in the more complicate method above, as well, but there's a lot more going on there, so harder to explain.

tasleson:
@mulkieran @agrover OK, part of me really appreciates suggestions like, do this instead:

``` rust
|| {
        for m in get_submounts(&PathBuf::from("/"))?.iter().filter(|m| m.file.to_str().map_or(false, |s| s.contains("stratis"))) {
            umount2(&m.file, MNT_DETACH)?;
        }
    }().map_err(|e| Error::with_chain(e, "Failed to unmount all stratis test mounted filesystems"));
```
even though at the moment I can't get it to compile, as it increases my knowledge of the language.

```
error[E0618]: expected function, found `_`
  --> src/engine/strat_engine/tests/util.rs:58:8
   |
58 |       || {
   |  ________^
59 | |         for m in get_submounts(&PathBuf::from("/"))?.iter().filter(|m| m.file.to_str().map_or(false, |s| s.contains("stratis"))) {
60 | |             umount2(&m.file, MNT_DETACH)?;
61 | |         }
62 | |     }().map_err(|e| Error::with_chain(e, "Failed to unmount all stratis test mounted filesystems"));
   | |_______^

```
In this case I understand that the value of the {} after the closure is not something you can call as a function, but I'm not sure how this expression is suppose to manifest into a result in which we can call `map_err` on.  It looks like some attempt at a try/catch block.

However, at the same time when I compare what I had
```rust
    let mounts = get_submounts(&PathBuf::from("/"))
        .chain_err(|| "Unable to retrieve mount points")?;

    for m in mounts
            .iter()
            .filter(|m| m.file.to_str().map_or(false, |s| s.contains("stratis"))) {
        umount2(&m.file, MNT_DETACH)
            .chain_err(|| format!("Unable to umount {:?}", m.file))?;
    }
```
to what is suggested, I can't help feeling that we are trying so very hard to utilize the full syntactic sugar of the language just because we can, not because it's needed.  Maybe this all steams from the fact that I'm still very much learning the language, but what are we gaining with one version over the other if functionally they achieve the same result?  One of them is more approachable to a larger audience IMHO.

tasleson:
I've update the PR to incorporate some of the changes.  I also fixed an `unwrap` that I missed.  After trying for way too long I was unable to make the suggested changes to `stratis_filesystems_unmount` and it remains as it was.  Please provide a working suggestion or amend the commit.

mulkieran:
@tasleson, I was trying to suggest something like this: https://github.com/mulkieran/stratisd/tree/tasleson-no_unwrap_in_cleanup.

It basically takes the orginal code of the function, and sticks it in an anonymous closure which it then immediately invokes. That gives a Result, and you call ```map_err()``` on that.

I had forgotten the need to add foreign links, to unify the result type of the closure. It seems like neither of mnt or nix use ```error_crate```, so I had to use foreign links instead of links.

We make anonymous closures and pass them to other functions all the time in Rust. The only twist here is making an anonymous closure and immediately invoking it. But the formatting should make it pretty easy to see what's happening.

tasleson:
@mulkieran Thanks for the working sample, I've cherry-picked your change and added it to this PR.  I got close to what you have provided here, I just stumbled on handling the anonymous function returning differing result types.  I tried to `Box` the result, but that yielded some ugly `Sized` compiler errors.

tasleson:
Re-based to pickup latest clippy.

mulkieran:
I think the step here is worth taking: https://github.com/mulkieran/stratisd/tree/tasleson-no_unwrap_in_cleanup.

But after that, probably consider this as complete as it is going to get. Further improvements would induce a lot of ugliness (which would be fixed better by error-chaining up stratisd as appropriate).

tasleson:
@mulkieran Added 849ac2c069ce00c0826ccf5ced0d911e6f619d3b to PR.

tasleson:
@mulkieran Should be good to go!

mulkieran:
@tasleson Code looks as good as is probably possible. Can you squash them into fewer commits? My commit messages were pretty casual, and if possible those shouldn't appear in final version.

tasleson:
@mulkieran Flattened into 1, I tried, but I couldn't take it any further :-)

PR #652, opened 2017-11-14
Fix loop back tests

tasleson:
The tests do not work when run concurrently.

With this correction my loop back tests run with no errors on my local VM.


agrover:
👍

mulkieran:
I'm merging, because all Jenkins tests actually succeeded.

PR #660, opened 2017-11-17
Fix pool destroy for blockdevs (take 2)

agrover:
This is another, cleaner way of achieving what #628 was trying to do.

### DIFF
--- a/src/dbus_api/api.rs
+++ b/src/dbus_api/api.rs
@@ -108,7 +108,7 @@ fn destroy_pool(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
             dbus_context
                 .actions
                 .borrow_mut()
-                .push_remove(object_path);
+                .push_remove(&object_path, m.tree);
             return_message.append3(action, msg_code_ok(), msg_string_ok())
         }
         Err(err) => {
--- a/src/dbus_api/types.rs
+++ b/src/dbus_api/types.rs
@@ -8,7 +8,7 @@ use std::convert::From;
 use std::rc::Rc;
 
 use dbus::Path;
-use dbus::tree::{DataType, MTFn, ObjectPath};
+use dbus::tree::{DataType, MTFn, ObjectPath, Tree};
 
 use uuid::Uuid;
 
@@ -127,9 +127,24 @@ impl ActionQueue {
         self.queue.push_back(DeferredAction::Add(object_path))
     }
 
-    /// Push a Remove action onto the back of the queue.
-    pub fn push_remove(&mut self, object_path: Path<'static>) {
-        self.queue.push_back(DeferredAction::Remove(object_path))
+    /// Push Remove actions for a path and its immediate descendants. Not
+    /// recursive, since no multi-level parent-child relationships currently
+    /// exist.
+    // Note: Path x is a child of path y if x's context's parent field is y.
+    pub fn push_remove(&mut self, item: &Path<'static>, tree: &Tree<MTFn<TData>, TData>) -> () {
+        for opath in tree.iter()
+                .filter(|opath| {
+                            opath
+                                .get_data()
+                                .as_ref()
+                                .map_or(false, |op_cxt| op_cxt.parent == *item)
+                        }) {
+            self.queue
+                .push_back(DeferredAction::Remove(opath.get_name().clone()))
+        }
+
+        self.queue
+            .push_back(DeferredAction::Remove(item.clone()))
     }
 
     /// Drain the queue.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -146,11 +146,11 @@ dependencies = [
 
 [[package]]
 name = "dbus"
-version = "0.6.0"
+version = "0.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "libc 0.2.31 (registry+https://github.com/rust-lang/crates.io-index)",
- "libdbus-sys 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libdbus-sys 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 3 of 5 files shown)


src/dbus_api/api.rs
```diff
@@ -105,6 +105,23 @@ fn destroy_pool(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
 
     let msg = match dbus_context.engine.borrow_mut().destroy_pool(pool_uuid) {
         Ok(action) => {
+            let bd_iface_name = format!("{}.{}", STRATIS_BASE_SERVICE, "blockdev");
```
mulkieran: I would generalize this in a number of ways.

* Forget about the interface. That's putting engine brains into the D-Bus API. Just search on the parent path. Note that everything has an OPContext, so you don't have to worry about what type it is.
* Consider searching recursively, just to be sure that you remove everything.
* Abstract this operation, of removing all things that have a parent that is the thing that you're removing into a separate function.
* Consider wrapping the ```push_remove``` action in some macro or something, such that the clean up of dependent objects happens automatically.

agrover:
waiting on working on this until we get a released dbus-rs version (0.6.1?) that has the iterators we're using.

trgill:
when this is merged - we should re-enable the Destroy2TestCase in stratisd-client-dbus

agrover:
removed temp commit and rebased since dbus 0.6.1 is out, but haven't tackled review suggestions yet.

agrover:
> Forget about the interface. That's putting engine brains into the D-Bus API. Just search on the parent path. Note that everything has an OPContext, so you don't have to worry about what type it is.

Not quite universal because Manager has a None OpContext.

> Consider searching recursively, just to be sure that you remove everything.

Ok. Well, what I did was add a new ActionQueue method that removes recursively. 

> Abstract this operation, of removing all things that have a parent that is the thing that you're removing into a separate function.

Ok.

> Consider wrapping the push_remove action in some macro or something, such that the clean up of dependent objects happens automatically.

Added `push_remove_recursive`.

src/dbus_api/types.rs
```diff
@@ -132,6 +132,23 @@ impl ActionQueue {
         self.queue.push_back(DeferredAction::Remove(object_path))
     }
 
+    /// Push Remove actions for a path and all paths that reference it via
+    /// the OpContext.parent field.
+    pub fn push_remove_recursive(&mut self,
+                                 item: &Path<'static>,
+                                 tree: &Tree<MTFn<TData>, TData>)
+                                 -> () {
+        for opath in tree.iter() {
```
mulkieran: Try instead:

```
let internal = |root: &Path<'static>| -> () {
    for opath in tree.iter().filter(|opath| opath.get_data().as_ref().map_or(false, |op_ctx| op_ctx.parent == *item) {
        internal(opath.get_name())
    }
    self.push_remove(root.clone());
};

internal(item);
```

The advantage is that you no longer appear to be recursing on the tree, which is nice, because that is not what is happening.

You may have to use a move, though.

src/dbus_api/types.rs
```diff
@@ -132,6 +132,23 @@ impl ActionQueue {
         self.queue.push_back(DeferredAction::Remove(object_path))
     }
 
+    /// Push Remove actions for a path and all paths that reference it via
```
mulkieran: Might as well get rid of original ```push_remove()``` and reimplement it as ```push_remove_recursive()```. There will never be any good reason to remove an object path and not to remove the object paths that it depends on that I can think of. Because if you do, it's just about certain that the next call will be a failure.

src/dbus_api/types.rs
agrover: I have no objections to this suggestion, but am having some trouble getting it to work (i.e. compile). Any chance you could take a stab?

agrover:
Rebase.
* Fold recursive stuff into `push_remove` method instead of creating another method.

src/dbus_api/types.rs
```diff
@@ -127,9 +127,20 @@ impl ActionQueue {
         self.queue.push_back(DeferredAction::Add(object_path))
     }
 
-    /// Push a Remove action onto the back of the queue.
-    pub fn push_remove(&mut self, object_path: Path<'static>) {
-        self.queue.push_back(DeferredAction::Remove(object_path))
+    /// Push Remove actions for a path and all paths that reference it via
+    /// the OpContext.parent field.
+    pub fn push_remove(&mut self, item: &Path<'static>, tree: &Tree<MTFn<TData>, TData>) -> () {
+        for opath in tree.iter() {
```
mulkieran: It is preferable to make the recursive call only on the value the algorithm actually recurs on, which in this case is the object path.

Also, it's probably worth making a comment on the algorithmic complexity. At first glance, it looks quadratic, and assuming a large number of object paths, that wouldn't be so great.

agrover:
rebase.

src/dbus_api/types.rs
agrover: I have no objections to this suggestion, but am having some trouble getting it to work (i.e. compile). Any chance you could take a stab?

mulkieran: For the first part, recursive closure are forbidden in Rust: http://smallcultfollowing.com/babysteps/blog/2013/04/30/the-case-of-the-recurring-closure/. That's a real drawback.

mulkieran: Alas, it looks like any attempt might shortly be outlawed. But I'ld still prefer something more like this https://github.com/mulkieran/stratisd/tree/agrover-fix-pool-destroy-for-blockdevs for the following reasons:
1. It makes it explicit that this is the transitive closure of the child relation. That's what the recursion is trying to achieve.
2. It uses filter instead of an if, which I just think is clearer.

Or, if it is not necessary to be recursive, and we think it won't be for the foreseeable future, just don't use recursion.

agrover: Yeah, `filter` is good here. Ok, will respin based on your version.

agrover:
update based on review to use `filter` instead of conditional.

PR #642, opened 2017-11-03
Master devicemapper 0.13 next

mulkieran:

PR #665, opened 2017-11-21
remove custom_derive and enum_derive dependencies

ignatenkobrain:
Signed-off-by: Igor Gnatenko <ignatenko@redhat.com>


mulkieran:
I think we should wait on this until the new devicemapper version is updated. Once stratisd depends on that version, then custom_derive will disappear from Cargo.lock.

mulkieran:
Ok. Merging this, doing new devicemapper release, rebasing and updating #642, and then merging that. Here goes.

Issue #615, opened 2017-10-19
Get hardware info for blockdevs


agrover:
We should be getting hardware info when adding blockdevs to a pool. Check out https://github.com/agrover/sg3-rs for some initial steps on a crate to get SCSI hw info.

agrover:
update: probably it's easier to get this by querying libudev.

tasleson:
Fixed

PR #668, opened 2017-11-27
Fetch HW WWN for blockdev

tasleson:
ref: https://github.com/stratis-storage/stratisd/issues/615

### DIFF
--- a/src/engine/strat_engine/blockdevmgr.rs
+++ b/src/engine/strat_engine/blockdevmgr.rs
@@ -25,6 +25,7 @@ use super::engine::DevOwnership;
 use super::metadata::{BDA, MIN_MDA_SECTORS, StaticHeader, validate_mda_size};
 use super::range_alloc::RangeAllocator;
 use super::serde_structs::{BlockDevSave, Recordable};
+use super::util::hw_lookup;
 
 const MIN_DEV_SIZE: Bytes = Bytes(IEC::Gi);
 const MAX_NUM_TO_WRITE: usize = 10;
@@ -364,8 +365,11 @@ fn initialize(pool_uuid: PoolUuid,
             let allocator = RangeAllocator::new(bda.dev_size(), &[(Sectors(0), bda.size())])
                 .expect("bda.size() < bda.dev_size() and single range");
 
-            // TODO: support getting hw info and passing in here. See #615
-            bds.push(StratBlockDev::new(dev, devnode.to_owned(), bda, allocator, None, None));
+            let hw_id = match hw_lookup(devnode) {
+                Ok(id) => id,
+                Err(_) => None,  // TODO: Log this failure so that it can be addressed.
+            };
+            bds.push(StratBlockDev::new(dev, devnode.to_owned(), bda, allocator, None, hw_id));
         } else {
             // TODO: check the return values and update state machine on failure
             let _ = BDA::wipe(&mut f);
--- a/src/engine/strat_engine/util.rs
+++ b/src/engine/strat_engine/util.rs
@@ -3,6 +3,7 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 /// Utilities to support Stratis.
+extern crate libudev;
 
 use std::path::Path;
 use std::process::Command;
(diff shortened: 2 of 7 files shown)


src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -364,8 +365,11 @@ fn initialize(pool_uuid: PoolUuid,
             let allocator = RangeAllocator::new(bda.dev_size(), &[(Sectors(0), bda.size())])
                 .expect("bda.size() < bda.dev_size() and single range");
 
-            // TODO: support getting hw info and passing in here. See #615
-            bds.push(StratBlockDev::new(dev, devnode.to_owned(), bda, allocator, None, None));
+            let hw_id = match hw_lookup(devnode) {
```
mulkieran: Yes. This is exactly the place where we would like to log because we ought not to fail, as the value is optional, but we would like to know about the problem. Feel free to put a TODO here, probably right next to the error.

src/engine/strat_engine/util.rs
```diff
@@ -72,3 +79,32 @@ pub fn set_uuid(devnode: &Path, uuid: Uuid) -> EngineResult<()> {
         Err(EngineError::Engine(ErrorEnum::Error, err_msg))
     }
 }
+
+/// Lookup the WWN from the udev db using the device node eg. /dev/sda
+pub fn hw_lookup(dev_node_search: &Path) -> strat_engine_util::Result<Option<String>> {
```
mulkieran: Given that we have no idea what error system we'll be using, please don't use error_chain.

Try instead adding a new error constructor for Udev errors to EngineError. Implement From for this constructor and it should be auto-promoted by the ```?``` operator. Transform ```chain_err(...)?``` to ```?``` everywhere and it may all compile.

src/engine/strat_engine/util.rs
```diff
@@ -72,3 +79,32 @@ pub fn set_uuid(devnode: &Path, uuid: Uuid) -> EngineResult<()> {
...
+    enumerator
+        .match_property("DEVTYPE", "disk")
+        .chain_err(|| "Unable to match udev property")?;
+
+    for device in enumerator
+            .scan_devices()
+            .chain_err(|| "Unable to scan udev db")? {
+        if dev_node_search == device.devnode().chain_err(|| "device.devnode error")? {
+            if let Some(wwid) = device.property_value("ID_WWN") {
+                return Ok(Some(String::from(wwid.to_str()
```
mulkieran: I don't know why you need ```to_str``` here.

tasleson: The property value is returned as an `OsString` just trying to get it into a `String` format.

tasleson:
@mulkieran Change re-worked to incorporate your changes, please take peek, thanks!

src/engine/strat_engine/util.rs
```diff
@@ -54,3 +55,21 @@ pub fn set_uuid(devnode: &Path, uuid: Uuid) -> EngineResult<()> {
...
+/// Lookup the WWN from the udev db using the device node eg. /dev/sda
+pub fn hw_lookup(dev_node_search: &Path) -> EngineResult<Option<String>> {
+    let context = libudev::Context::new()?;
+    let mut enumerator = libudev::Enumerator::new(&context)?;
+    enumerator.match_subsystem("block")?;
+    enumerator.match_property("DEVTYPE", "disk")?;
+
+    let result = enumerator
+        .scan_devices()?
+        .find(|x| dev_node_search == x.devnode().unwrap_or(Path::new("")))
```
mulkieran: ```|x| x.devnode().map_or(false, |d| dev_node_search == d)``` is a more Rustic choice for closure.

src/engine/strat_engine/util.rs
```diff
@@ -54,3 +55,21 @@ pub fn set_uuid(devnode: &Path, uuid: Uuid) -> EngineResult<()> {
...
 }
+
+/// Lookup the WWN from the udev db using the device node eg. /dev/sda
+pub fn hw_lookup(dev_node_search: &Path) -> EngineResult<Option<String>> {
+    let context = libudev::Context::new()?;
+    let mut enumerator = libudev::Enumerator::new(&context)?;
+    enumerator.match_subsystem("block")?;
+    enumerator.match_property("DEVTYPE", "disk")?;
+
+    let result = enumerator
```
mulkieran: Probably unnecessary to assign to ```result``` and then immediately return it.

src/engine/strat_engine/util.rs
```diff
@@ -54,3 +55,21 @@ pub fn set_uuid(devnode: &Path, uuid: Uuid) -> EngineResult<()> {
...
+        .scan_devices()?
+        .find(|x| dev_node_search == x.devnode().unwrap_or(Path::new("")))
+        .map_or(Ok(None), |dev| {
+            Ok(dev.property_value("ID_WWN")
+                   .map_or(None, |r| r.to_str().map_or(None, |x| Some(String::from(x)))))
```
mulkieran: The unlikely failure of ```to_str()``` to return a value should probably be treated as an error.

tasleson:
@mulkieran @agrover @trgill The code in this PR compiles, but clippy flags an issue, however if you do as it suggests you get the following compiler error:

```
   Compiling libstratis v0.1.2 (file:///home/tasleson/projects/stratis/stratisd)
error[E0597]: `context` does not live long enough
  --> src/engine/strat_engine/util.rs:73:1
   |
62 |     let mut enumerator = libudev::Enumerator::new(&context)?;
   |                                                    ------- borrow occurs here
...
73 | }
   | ^ `context` dropped here while still borrowed
   |
   = note: values in a scope are dropped in the opposite order they are created

error[E0597]: `enumerator` does not live long enough
  --> src/engine/strat_engine/util.rs:73:1
   |
66 |     enumerator
   |     ---------- borrow occurs here
...
73 | }
   | ^ `enumerator` dropped here while still borrowed
   |
   = note: values in a scope are dropped in the opposite order they are created

error: aborting due to 2 previous errors

error: Could not compile `libstratis`.
```
I encountered this before I pushed the current change, but was unable to find a better solution.  So I'm up for suggestions on how to make the compiler and clippy happy.

src/engine/strat_engine/util.rs
tasleson: Code will not compile it, get 2 borrow errors.

mulkieran: Has bug been filed w/ clippy yet? If not, probably worth it.

mulkieran: _ Perhaps_ instead of ```String::from(x)```, ```x.to_owned()``` would avoid error. It might ease the borrow checker's difficulties. But I wouldn't bet on it.

tasleson: Tried it, same error...

tasleson:
@mulkieran @trgill Thanks for your help!  This should be ready to go!

Issue #670, opened 2017-11-28
Capture stdout & stderr for all command line tools


tasleson:
When calls to some of the command line tools fail we can still get output dumped to `stdout` and `stderr`, lets handle all cases.

agrover:
fixed

PR #671, opened 2017-11-28
Use common function for CLI calls

tasleson:
Create a common function which captures stdout & stderr so that we
never see it get spewed out to the journal unless we want to have it
in a logged error message.  This also has a nice side effect of
removing copy & pasted code.

Note: This is a follow on to changes introduced with 94dbe87.  Corrects https://github.com/stratis-storage/stratisd/issues/670


agrover:
👍 , merging. Thanks!

Issue #676, opened 2017-12-04
Update udev-rs to new upstream


tasleson:
It appears `https://github.com/Smithay/udev-rs` is the new upsteam.  Change the dependency and update the code.

tasleson:
Not making the switch just yet...

PR #677, opened 2017-12-04
Update udev library

tasleson:
The library appears to have a new upstream maintainer:

https://github.com/Smithay/udev-rs


tasleson:
Addresses https://github.com/stratis-storage/stratisd/issues/676

mulkieran:
Not exactly. The new version is because the maintainer of the previous version was fairly unresponsive and the maintainer of the new version was tired of waiting around. So this requires a bit of thought as to whether to jump ship or not.

tasleson:
@mulkieran Kind of trivial to swap back and forth, so I'm not going to lose too much sleep if this is needed.

tasleson:
Updated PR to do less change to make it easier to swap back and forth.

tasleson:
Lets leave this be for now.

PR #678, opened 2017-12-05
Rework Table and remove HasName/HasUuid traits

agrover:
Change Table in two ways:
1. Implement using two HashMaps instead of Vec and two HashMaps. This seems simpler, although uuid and name are no longer exactly equal when doing lookups.
1. Externalize keys from the stored struct. This should make mutable iteration better. But, this does cause some nontrivial ripples in our code that uses Table (most of it).

Right now I'm looking for general feedback on the acceptability of this approach in review-prelim. If this is a good direction, then we can proceed with detailed review in review-final.

Please read individual commit logs for specifics. Thanks!

### DIFF
--- a/src/engine/structures.rs
+++ b/src/engine/structures.rs
@@ -199,19 +199,33 @@ impl<T> Table<T> {
     }
 
     /// Inserts an item for given uuid and name.
-    /// Possibly returns the item displaced.
-    pub fn insert(&mut self, name: Name, uuid: Uuid, item: T) -> Option<(Name, Uuid, T)> {
+    /// Possibly returns the items displaced.
+    /// If two items are displaced, the one displaced by matching name is
+    /// returned first.
+    pub fn insert(&mut self, name: Name, uuid: Uuid, item: T) -> Option<Vec<(Name, Uuid, T)>> {
         match self.name_to_uuid.insert(name.clone(), uuid) {
             Some(old_uuid) => {
                 // (existing name, _)
-                match self.items.insert(uuid, (name, item)) {
+                match self.items.insert(uuid, (name.clone(), item)) {
                     // (existing name, existing uuid)
-                    Some((old_name, old_item)) => Some((old_name, uuid, old_item)),
+                    Some((old_name, old_item)) => {
+                        // double eviction
+                        if old_name != name {
+                            assert_ne!(old_uuid, uuid);
+                            let (_, item2) = self.items.remove(&old_uuid).expect("should be there");
+                            self.name_to_uuid
+                                .remove(&old_name)
+                                .expect("should be there");
+                            Some(vec![(old_name, uuid, old_item), (name, old_uuid, item2)])
+                        } else {
+                            Some(vec![(old_name, uuid, old_item)])
+                        }
+                    }
                     // (existing name, new uuid)
                     None => {
                         let (old_name, old_item) =
                             self.items.remove(&old_uuid).expect("should be there");
-                        Some((old_name, old_uuid, old_item))
+                        Some(vec![(old_name, old_uuid, old_item)])
                     }
                 }
             }
@@ -221,7 +235,7 @@ impl<T> Table<T> {
                     let old_uuid = self.name_to_uuid
                         .remove(&old_name)
                         .expect("should be there");
-                    Some((old_name, old_uuid, old_item))
+                    Some(vec![(old_name, old_uuid, old_item)])
                 } else {
                     // (new name, new uuid)
                     None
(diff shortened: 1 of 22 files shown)


agrover:
review @mulkieran @trgill @tasleson ?

agrover:
rebased.

agrover:
rebased.

tasleson:
Correct use order:
Note: Using ASCII value sort order
```
src/engine/structures.rs 5 [HashMap, hash_map]
src/engine/strat_engine/cleanup.rs 7 [EngineError, EngineResult, ErrorEnum]
src/engine/sim_engine/engine.rs 19 [Name, PoolUuid, Redundancy, RenameAction]
```

src/engine/structures.rs
```diff
@@ -433,22 +419,22 @@ mod tests {
...
         // The table contains the new item and has no memory of the old.
         assert!(t.contains_uuid(uuid));
         assert!(t.contains_name(name2));
         assert!(!t.contains_name(name));
-        assert_eq!(t.get_by_uuid(uuid).unwrap().stuff, thing_key2);
-        assert_eq!(t.get_by_name(&name2).unwrap().stuff, thing_key2);
-        assert_eq!(t.len(), 1);
+        assert!(t.get_by_uuid(uuid).unwrap().1.stuff == thing_key2);
+        assert!(t.get_by_name(&name2).unwrap().1.stuff == thing_key2);
+        assert!(t.len() == 1);
```
tasleson: Please use `assert_eq` for all of the asserts you are doing which do an equality comparison, which appears to be many of the lines you edited in this commit.

tasleson:
Is there a way to simplify the name change so that we don't have to do `&*foo.name()` everywhere we need to use the name?

src/engine/structures.rs
```diff
@@ -102,117 +166,112 @@ impl<T: HasName + HasUuid> Table<T> {
...
-
-            // Remove the item we want to remove and also the uuid mapping
-            let item = self.items.swap_remove(index);
-            self.name_map.remove(&*item.name());
-
-            // Remove the uuid again, in case there is only one item.
-            self.uuid_map.remove(&uuid);
-
+        if let Some((_, item)) = self.items.remove(&uuid) {
+            let name = self.name_to_uuid
```
tasleson: Maybe I'm missing something here, but when you remove the item you are getting the name and item right?  As you already know the name can't you just skip the `.ter().find ...` and go straight at `self.name_to_uuid.remove(&*name)`

agrover: You're right! Nice catch!

src/engine/structures.rs
agrover: Not that I can see. Things must've gotten flipped during some rebasing session.

src/engine/structures.rs
```diff
@@ -2,62 +2,121 @@
...
+pub struct Iter<'a, T: 'a> {
+    items: hash_map::Iter<'a, Uuid, (Name, T)>,
+}
 
-    fn into_iter(self) -> IterMut<'a, T> {
-        self.items.iter_mut()
+impl<'a, T> Iterator for Iter<'a, T> {
+    type Item = (&'a Name, &'a Uuid, &'a T);
+
+    #[inline]
```
mulkieran: Could you leave ```inline``` hints out, unless you're absolutely convinced that they are desirable? We don't do this anywhere else.

src/engine/structures.rs
```diff
@@ -66,117 +125,108 @@ impl<T: HasName + HasUuid> Table<T> {
...
-
-        self.items.push(item);
-
-        match (name_item, uuid_item) {
-            (None, None) => vec![],
-            (None, Some(uuid_item)) => vec![uuid_item],
-            (Some(name_item), None) => vec![name_item],
-            (Some(name_item), Some(uuid_item)) => vec![name_item, uuid_item],
+    /// Possibly returns the item displaced.
+    pub fn insert(&mut self, name: Name, uuid: Uuid, item: T) -> Option<(Name, Uuid, T)> {
```
mulkieran: I don't see any way that this can be right. If the name and the uuid are both already in the table, and they don't belong to the same item, then in order to insert the new item, and maintain the invariant that each uuid maps to its name and each name maps to its uuid then two items must be ejected.

In this specific implementation, you are ejecting both, but only returning one.

agrover: ok

src/engine/structures.rs
agrover: This is what other container implementations do, so I'm gonna leave it.

agrover:
Hi, I've pulled in all the commits from that PR, they look good.

PR #680, opened 2017-12-06
meta expand

trgill:
Extend the meta device in a pool when the usage exceeds the META_LOWATER.
Add tests for the extend meta code.

### DIFF
--- a/src/engine/strat_engine/thinpool/thinpool.rs
+++ b/src/engine/strat_engine/thinpool/thinpool.rs
@@ -40,6 +40,37 @@ pub const INITIAL_DATA_SIZE: DataBlocks = DataBlocks(768);
 const INITIAL_MDV_SIZE: Sectors = Sectors(32 * IEC::Ki); // 16 MiB
 
 
+pub struct ThinPoolSizeParams {
+    meta_size: MetaBlocks,
+    data_size: DataBlocks,
+    mdv_size: Sectors,
+}
+
+impl ThinPoolSizeParams {
+    /// The number of Sectors in the MetaBlocks.
+    pub fn meta_size(&self) -> Sectors {
+        self.meta_size.sectors()
+    }
+    /// The number of Sectors in the DataBlocks.
+    pub fn data_size(&self) -> Sectors {
+        *self.data_size * DATA_BLOCK_SIZE
+    }
+    /// MDV size
+    pub fn mdv_size(&self) -> Sectors {
+        self.mdv_size
+    }
+}
+
+impl Default for ThinPoolSizeParams {
+    fn default() -> ThinPoolSizeParams {
+        ThinPoolSizeParams {
+            meta_size: INITIAL_META_SIZE,
+            data_size: INITIAL_DATA_SIZE,
+            mdv_size: INITIAL_MDV_SIZE,
+        }
+    }
+}
+
 /// A ThinPool struct contains the thinpool itself, the spare
 /// segments for its metadata device, and the filesystems and filesystem
 /// metadata associated with it.
@@ -60,15 +91,16 @@ impl ThinPool {
     /// Make a new thin pool.
     pub fn new(pool_uuid: PoolUuid,
                dm: &DM,
+               thin_pool_size: &ThinPoolSizeParams,
                data_block_size: Sectors,
                low_water_mark: DataBlocks,
                block_mgr: &mut BlockDevMgr)
                -> EngineResult<ThinPool> {
         let mut segments_list =
-            match block_mgr.alloc_space(&[ThinPool::initial_metadata_size(),
-                                          ThinPool::initial_metadata_size(),
-                                          ThinPool::initial_data_size(),
-                                          ThinPool::initial_mdv_size()]) {
+            match block_mgr.alloc_space(&[thin_pool_size.meta_size(),
+                                          thin_pool_size.meta_size(),
+                                          thin_pool_size.data_size(),
+                                          thin_pool_size.mdv_size()]) {
                 Some(sl) => sl,
                 None => {
                     let err_msg = "Could not allocate sufficient space for thinpool devices.";
@@ -93,9 +125,7 @@ impl ThinPool {
                                         &format_flex_name(pool_uuid, FlexRole::ThinMeta),
                                         None,
                                         &map_to_dm(&meta_segments))?;
-        wipe_sectors(&meta_dev.devnode(),
-                     Sectors(0),
-                     ThinPool::initial_metadata_size())?;
+        wipe_sectors(&meta_dev.devnode(), Sectors(0), thin_pool_size.meta_size())?;
 
         let data_dev = LinearDev::setup(dm,
                                         &format_flex_name(pool_uuid, FlexRole::ThinData),
@@ -243,20 +273,6 @@ impl ThinPool {
            })
     }
 
-    /// Initial size for a pool's meta data device.
-    fn initial_metadata_size() -> Sectors {
-        INITIAL_META_SIZE.sectors()
-    }
-
-    /// Initial size for a pool's data device.
-    fn initial_data_size() -> Sectors {
-        *INITIAL_DATA_SIZE * DATA_BLOCK_SIZE
-    }
-
-    /// Initial size for a pool's filesystem metadata volume.
-    fn initial_mdv_size() -> Sectors {
-        INITIAL_MDV_SIZE
-    }
 
     /// Run status checks and take actions on the thinpool and its components.
     pub fn check(&mut self, dm: &DM, bd_mgr: &mut BlockDevMgr) -> EngineResult<()> {
@@ -654,8 +670,13 @@ mod tests {
         let pool_uuid = Uuid::new_v4();
         let dm = DM::new().unwrap();
         let mut mgr = BlockDevMgr::initialize(pool_uuid, paths, MIN_MDA_SECTORS, false).unwrap();
-        let mut pool = ThinPool::new(pool_uuid, &dm, DATA_BLOCK_SIZE, DATA_LOWATER, &mut mgr)
-            .unwrap();
+        let mut pool = ThinPool::new(pool_uuid,
+                                     &dm,
+                                     &ThinPoolSizeParams::default(),
+                                     DATA_BLOCK_SIZE,
+                                     DATA_LOWATER,
+                                     &mut mgr)
+                .unwrap();
 
         let fs_uuid = pool.create_filesystem("stratis_test_filesystem", &dm, None)
             .unwrap();
@@ -738,8 +759,13 @@ mod tests {
         let pool_uuid = Uuid::new_v4();
         let dm = DM::new().unwrap();
         let mut mgr = BlockDevMgr::initialize(pool_uuid, paths, MIN_MDA_SECTORS, false).unwrap();
-        let mut pool = ThinPool::new(pool_uuid, &dm, DATA_BLOCK_SIZE, DATA_LOWATER, &mut mgr)
-            .unwrap();
+        let mut pool = ThinPool::new(pool_uuid,
+                                     &dm,
+                                     &ThinPoolSizeParams::default(),
+                                     DATA_BLOCK_SIZE,
+                                     DATA_LOWATER,
+                                     &mut mgr)
+                .unwrap();
 
         let fs_uuid = pool.create_filesystem(&name1, &dm, None).unwrap();
 
@@ -777,8 +803,13 @@ mod tests {
         let pool_uuid = Uuid::new_v4();
         let dm = DM::new().unwrap();
         let mut mgr = BlockDevMgr::initialize(pool_uuid, paths, MIN_MDA_SECTORS, false).unwrap();
-        let mut pool = ThinPool::new(pool_uuid, &dm, DATA_BLOCK_SIZE, DATA_LOWATER, &mut mgr)
-            .unwrap();
+        let mut pool = ThinPool::new(pool_uuid,
+                                     &dm,
+                                     &ThinPoolSizeParams::default(),
+                                     DATA_BLOCK_SIZE,
+                                     DATA_LOWATER,
+                                     &mut mgr)
+                .unwrap();
 
         let fs_uuid = pool.create_filesystem("fsname", &dm, None).unwrap();
 
@@ -828,8 +859,13 @@ mod tests {
         let pool_uuid = Uuid::new_v4();
         let dm = DM::new().unwrap();
         let mut mgr = BlockDevMgr::initialize(pool_uuid, paths, MIN_MDA_SECTORS, false).unwrap();
-        let mut pool = ThinPool::new(pool_uuid, &dm, DATA_BLOCK_SIZE, DATA_LOWATER, &mut mgr)
-            .unwrap();
+        let mut pool = ThinPool::new(pool_uuid,
+                                     &dm,
+                                     &ThinPoolSizeParams::default(),
+                                     DATA_BLOCK_SIZE,
+                                     DATA_LOWATER,
+                                     &mut mgr)
+                .unwrap();
         let fs_name = "stratis_test_filesystem";
         let fs_uuid = pool.create_filesystem(&fs_name, &dm, None).unwrap();
         let thin_id = pool.get_filesystem_by_uuid(fs_uuid).unwrap().thin_id();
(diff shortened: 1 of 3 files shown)


src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -935,6 +935,50 @@ mod tests {
...
+                                          DATA_BLOCK_SIZE,
+                                          Some(MetaBlocks(16)),
+                                          DATA_LOWATER,
+                                          &mut mgr)
+                .unwrap();
+
+        match thin_pool.status(&dm).unwrap() {
+            dm::ThinPoolStatus::Working(ref status) => {
+                let usage = &status.usage;
+                assert_eq!(usage.total_meta.0, MetaBlocks(16).0);
```
mulkieran: ```assert_eq!(usage.total_meta, MetaBlock(16))``` ought to work since both are meta blocks.

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -935,6 +935,50 @@ mod tests {
...
+            }
+            dm::ThinPoolStatus::Fail => panic!("thin_pool.status() failed"),
+        }
+        // The meta device is smaller than META_LOWATER, so it should be expanded
+        // in the thin_pool.check() call.
+        let _ = thin_pool.check(&dm, &mut mgr);
```
mulkieran: Might as well unwrap() here.

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -935,6 +935,50 @@ mod tests {
...
+        match thin_pool.status(&dm).unwrap() {
+            dm::ThinPoolStatus::Working(ref status) => {
+                let usage = &status.usage;
+                // validate that the meta has been expanded.
+                assert!(usage.total_meta.0 > MetaBlocks(16).0);
```
mulkieran: ```assert!(usage.total_meta > MetaBlocks(16));```

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -261,11 +261,14 @@ impl ThinPool {
         INITIAL_MDV_SIZE
     }
 
+    pub fn status(&mut self, dm: &DM) -> EngineResult<dm::ThinPoolStatus> {
```
mulkieran: I don't think this method is needed for testing.

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -283,16 +284,20 @@ impl ThinPool {
                     }
                 }
 
-                let usage = &status.usage;
-                if usage.used_meta > usage.total_meta - META_LOWATER {
-                    // TODO: Extend meta device
+                if status.usage.used_meta >
```
mulkieran: This is minor, but I liked stowing ```status.usage``` in ```usage```. Reduces line length a little.

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -867,6 +935,50 @@ mod tests {
...
+
+    #[test]
+    pub fn real_test_meta_expand() {
+        real::test_with_spec(real::DeviceLimits::Range(2, 3), test_meta_expand);
+    }
+
+    fn test_meta_expand(paths: &[&Path]) -> () {
+        let pool_uuid = Uuid::new_v4();
+        let dm = DM::new().unwrap();
+        let mut mgr = BlockDevMgr::initialize(pool_uuid, paths, MIN_MDA_SECTORS, false).unwrap();
```
mulkieran: Right here, you could be passing paths which are not the original device paths, but are actually a linear device or devices that is only a little bit bigger than the initial size of the pool. Then it is still necessary to figure out how to expand the metadata faster than the data. Possibly you could do that by making lots and lots of filesystems of negligible size.

trgill: How will providing a smaller amount of storage to the pool do any good?  Doing that won't change the initial size of the meta or the data devices.

mulkieran: You're correct. I had the problem backwards, unfortunately.

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -277,16 +284,20 @@ impl ThinPool {
...
+                   cmp::max(status.usage.total_meta, META_LOWATER) - META_LOWATER {
+                    match self.extend_thinpool_meta(dm, status.usage.total_meta, bd_mgr) {
+                        #![allow(single_match)]
+                        Ok(_) => {}
+                        Err(_) => {} // TODO: Take pool offline?
+                    }
                 }
 
-                if usage.used_data > usage.total_data - DATA_LOWATER {
+                if status.usage.used_data > status.usage.total_data - DATA_LOWATER {
```
mulkieran: Do the same trick as above here, to ensure no overflow on subtraction.

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -351,45 +362,47 @@ impl ThinPool {
...
+    /// Return the number of MetaBlocks added.
+    fn extend_thinpool_meta(&mut self,
+                            dm: &DM,
+                            extend_size: MetaBlocks,
+                            bd_mgr: &mut BlockDevMgr)
+                            -> EngineResult<MetaBlocks> {
+        if let Some(mut new_meta_regions) = bd_mgr.alloc_space(&[extend_size.sectors()]) {
+            self.extend_meta(dm,
+                             &new_meta_regions
+                                  .pop()
```
mulkieran: I doubt you need to pop here, just use ```first()```.   ```new_meta_regions``` can then be immutable. Might as well fix up ```extend_thinpool``` which does same thing in separate commit.

mulkieran:
Regarding the question of setting metadata region size...looking over commit message and associated comments on the values (there are none), it looks to me like these initial values were somewhat arbitrarily chosen. Which means that they have less significance than I had thought formerly. So it seems to me like the better choice would be to make all three sizes non-optional for ```ThinPool::new()```, perhaps dreaming up a struct to keep them better organized. Then they can be passed as arguments on all calls to ThinPool::new(). With a struct, you could define default such that it held the default values for all these initial values, which would make the calls more convenient.

src/engine/strat_engine/thinpool/thinpool.rs
trgill: Why not?  I'm using it to verify the meta total allocated space is increased.  Isn't it needed to validate the meta was expanded?

src/engine/strat_engine/thinpool/thinpool.rs
trgill: Ok - I'll put it back.

src/engine/strat_engine/thinpool/thinpool.rs
trgill: ok

src/engine/strat_engine/thinpool/thinpool.rs
trgill: ok

src/engine/strat_engine/thinpool/thinpool.rs
trgill: ok

src/engine/strat_engine/thinpool/thinpool.rs
mulkieran: The only test that you add is in this file, so you could just as well use ```thin_pool.thin_pool.status(&dm)``` in your test. There is no need for making anything public.

trgill: I see.  thanks.

trgill:
Ok.  That all looks good to me.

The meta_size should be MetaBlocks, right?  (not Sectors) 

data_size => DataBlocks too...

mulkieran:
Yes. ```data_size``` and ```meta_size``` would almost certainly be better using their more specific types, rather than Sectors.

src/engine/strat_engine/thinpool/thinpool.rs
trgill: ok

src/engine/strat_engine/thinpool/thinpool.rs
trgill: ok - I broke both out into commits.

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -40,6 +40,37 @@ pub const INITIAL_DATA_SIZE: DataBlocks = DataBlocks(768);
...
+}
+
+impl ThinPoolSizeParams {
+    /// The number of Sectors in the MetaBlocks.
+    pub fn meta_size(&self) -> Sectors {
+        self.meta_size.sectors()
+    }
+    /// The number of Sectors in the DataBlocks.
+    pub fn data_size(&self) -> Sectors {
+        self.data_size.0 * DATA_BLOCK_SIZE
```
mulkieran: Please dereference here. Ideally, the field would be private outside the module in which its defined. But due to some compiler constraints, which we put up with so we can define Sectors constants, it is not. Eventuallly, we hope that these constraints will go away, and when they do, the compiler will complain about accessing the .0 field out side the module. Meanwhile, we just have to pretend.

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -295,14 +296,18 @@ impl ThinPool {
...
+                        Err(_) => {} // TODO: Take pool offline?
+                    }
                 }
 
                 if usage.used_data > usage.total_data - DATA_LOWATER {
                     // Request expansion of physical space allocated to the pool
                     // TODO: we just request that the space be doubled here.
                     // A more sophisticated approach might be in order.
-                    match self.extend_thinpool(dm, usage.total_data, bd_mgr) {
+                    match self.extend_thinpool(dm, status.usage.total_data, bd_mgr) {
```
mulkieran: Just use ```usage```, not ```status.usage```.

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -946,6 +946,54 @@ mod tests {
...
+    }
+
+    fn test_meta_expand(paths: &[&Path]) -> () {
+        let pool_uuid = Uuid::new_v4();
+        let dm = DM::new().unwrap();
+        let mut mgr = BlockDevMgr::initialize(pool_uuid, paths, MIN_MDA_SECTORS, false).unwrap();
+        // Create a ThinPool with a very small meta device.
+        let mut thin_pool = ThinPool::new(pool_uuid,
+                                          &dm,
+                                          &ThinPoolSizeParams {
```
mulkieran: Here you can use strange Default syntax:
```
&ThinPoolSizeParams {
    meta_size: MetaBlocks(16),
    ...Default::default(),
}
```

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -278,15 +295,20 @@ impl ThinPool {
...
+                if usage.used_meta > cmp::max(usage.total_meta, META_LOWATER) - META_LOWATER {
+                    match self.extend_thinpool_meta(dm, usage.total_meta, bd_mgr) {
+                        #![allow(single_match)]
+                        Ok(_) => {}
+                        Err(_) => {} // TODO: Take pool offline?
+                    }
                 }
 
-                if usage.used_data > usage.total_data - DATA_LOWATER {
+                if status.usage.used_data >
```
mulkieran: Just use ```usage```, not ```status.usage``` here, too.

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -867,6 +946,54 @@ mod tests {
...
+        // This test requires more than 1 GiB.
+        loopbacked::test_with_spec(loopbacked::DeviceLimits::Range(2, 3), test_meta_expand);
+    }
+
+    #[test]
+    pub fn real_test_meta_expand() {
+        real::test_with_spec(real::DeviceLimits::Range(2, 3), test_meta_expand);
+    }
+
+    fn test_meta_expand(paths: &[&Path]) -> () {
```
mulkieran: Needs a header comment.

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -867,6 +946,54 @@ mod tests {
...
+        match thin_pool.thin_pool.status(&dm).unwrap() {
+            dm::ThinPoolStatus::Working(ref status) => {
+                let usage = &status.usage;
+                assert_eq!(usage.total_meta, MetaBlocks(16));
+            }
+            dm::ThinPoolStatus::Fail => panic!("thin_pool.status() failed"),
+        }
+        // The meta device is smaller than META_LOWATER, so it should be expanded
+        // in the thin_pool.check() call.
+        let _ = thin_pool.check(&dm, &mut mgr);
```
mulkieran: unwrap the value rather than ignoring it.

src/engine/strat_engine/thinpool/thinpool.rs
mulkieran: Got this syntax from https://stackoverflow.com/questions/19650265/is-there-a-faster-shorter-way-to-initialize-variables-in-a-rust-struct, but I think I originally ran across it in programming Rust book.

src/engine/strat_engine/thinpool/thinpool.rs
trgill: Your comment is about self.data_size.0?  I thought the rule was .0 if inside implementation * if outside?

mulkieran: I'm not sure what you're saying there. But, pretend it's a regular struct w/ a private field. You could access that private field within the module in which that struct was defined, which for Sectors, is the types module, but you could not access that field outside the types module. If you follow this rule, then you should not access the field here.

trgill: ok

src/engine/strat_engine/thinpool/thinpool.rs
trgill: ok

src/engine/strat_engine/thinpool/thinpool.rs
trgill: yup

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -867,6 +945,56 @@ mod tests {
...
+        match thin_pool.thin_pool.status(&dm).unwrap() {
+            dm::ThinPoolStatus::Working(ref status) => {
+                let usage = &status.usage;
+                assert_eq!(usage.total_meta, MetaBlocks(16));
+            }
+            dm::ThinPoolStatus::Fail => panic!("thin_pool.status() failed"),
+        }
+        // The meta device is smaller than META_LOWATER, so it should be expanded
+        // in the thin_pool.check() call.
+        let _ = thin_pool.check(&dm, &mut mgr).unwrap();
```
mulkieran: Just unwrap()...leave out the ```let _``` and it should still compile.

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -867,6 +945,56 @@ mod tests {
...
+    /// a meta device smaller than the META_LOWATER.
+    fn test_meta_expand(paths: &[&Path]) -> () {
+        let pool_uuid = Uuid::new_v4();
+        let dm = DM::new().unwrap();
+        let mut mgr = BlockDevMgr::initialize(pool_uuid, paths, MIN_MDA_SECTORS, false).unwrap();
+        // Create a ThinPool with a very small meta device.
+        let mut thin_pool = ThinPool::new(pool_uuid,
+                                          &dm,
+                                          &ThinPoolSizeParams {
+                                               meta_size: MetaBlocks(16),
```
mulkieran: ```MetaBlocks(16)``` probably deservies a name in this method, since it gets used four times to mean the same thing.

mulkieran:
@agrover review?

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -295,14 +296,18 @@ impl ThinPool {
 
                 let usage = &status.usage;
                 if usage.used_meta > usage.total_meta - META_LOWATER {
-                    // TODO: Extend meta device
+                    match self.extend_thinpool_meta(dm, usage.total_meta, bd_mgr) {
```
agrover: By passing `usage.total_meta` we're setting the meta extension policy to be "double the size". I think it might not be super obvious this is where that policy is being enacted, so maybe add a comment explicitly stating this, or even a line `let meta_extend_size = usage.total_meta` and then pass `meta_extend_size` could make it clear this is intentional.

trgill: ok - sounds good

Issue #498, opened 2017-08-02
Create symlinks to filesystems under /dev/stratis/<poolname>/<fsname>


agrover:
Create symlinks to filesystems when they are created in a pool. These symlinks should also be updated for pool and fs renames.

agrover:
The problem this solves is that after creating a filesystem in a pool, there is no easy way for the user to mount it. The user needs to use the DM-named `/dev/dm-X`, but only after finding out which one corresponds with their newly-created filesystem

Furthermore, the dm-X for a given device can change across reboots for various reasons. We need more stable names, like what LVM provides.

mulkieran:
We should be very careful how we do this, lest we end up in a situation that nobody knows how to fix. See, for example, https://bugzilla.redhat.com/show_bug.cgi?id=1263441.

agrover:
I don't think that particular BZ is applicable to us. But, I've been looking at udev code to try and understand possible negative interactions between it, and us maintaining symlinks as described.

(Q: Why not leverage udev rules to create & maintain these symlinks for us? A: Wouldn't the need to set up symlinks for filesystems including root in the initrd preclude using udev for this?)

trgill:
This PR will be helpful for automated testing.  Currently there is no way to script creating a filesystem and mounting it.   The "name" doesn't correspond to /dev/mapper/<> node.

PR #682, opened 2017-12-06
Make symlinks

agrover:
This PR implements maintenance of /dev/stratis, as mentioned in #498. The big question I have is where to put the code this PR adds -- right now it's in stratis/stratis.rs for lack of a better place.

### DIFF
--- a/src/engine/strat_engine/devlinks.rs
+++ b/src/engine/strat_engine/devlinks.rs
@@ -1,0 +1,110 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use std::collections::HashSet;
+use std::fs;
+use std::io::ErrorKind;
+use std::os::unix::fs::symlink;
+use std::path::{Path, PathBuf};
+use std::str;
+
+use super::super::errors::EngineResult;
+
+use engine::Pool;
+
+pub const DEV_PATH: &str = "/dev/stratis";
+
+/// Set up directories and symlinks under /dev/stratis based on current
+/// config. Clear out any directory or file that doesn't correspond to a pool
+/// or filesystem.
+// Don't just remove and recreate everything in case there are processes
+// (e.g. user shells) with the current working directory within the tree.
+pub fn setup_devlinks<'a, I: Iterator<Item = &'a Pool>>(pools: I) -> EngineResult<()> {
+    if let Err(err) = fs::create_dir(DEV_PATH) {
+        if err.kind() != ErrorKind::AlreadyExists {
+            return Err(From::from(err));
+        }
+    }
+
+    let mut existing_dirs = fs::read_dir(DEV_PATH)?
+        .map(|dir_e| dir_e.and_then(|d| Ok(d.file_name().into_string().expect("Unix is utf-8"))))
+        .collect::<Result<HashSet<_>, _>>()?;
+
+    for pool in pools {
+        if !existing_dirs.remove(pool.name()) {
+            pool_added(pool.name())?;
+        }
+
+        let pool_path: PathBuf = vec![DEV_PATH, pool.name()].iter().collect();
+
+        let mut existing_files = fs::read_dir(pool_path)?
+            .map(|dir_e| {
+                     dir_e.and_then(|d| Ok(d.file_name().into_string().expect("Unix is utf-8")))
+                 })
+            .collect::<Result<HashSet<_>, _>>()?;
+
+        for fs in pool.filesystems() {
+            filesystem_added(pool.name(), fs.name(), &fs.devnode())?;
+            existing_files.remove(fs.name());
+        }
+
+        for leftover in existing_files {
+            filesystem_removed(pool.name(), &leftover)?;
+        }
+    }
+
+    for leftover in existing_dirs {
+        pool_removed(str::from_utf8(leftover.as_bytes()).expect("is valid utf8"))?
+    }
+
+    Ok(())
+}
+
+/// Create a directory when a pool is added.
+pub fn pool_added(pool: &str) -> EngineResult<()> {
+    let p: PathBuf = vec![DEV_PATH, pool].iter().collect();
+    fs::create_dir(&p)?;
+    Ok(())
+}
+
+/// Remove the directory and its contents when the pool is removed.
+pub fn pool_removed(pool: &str) -> EngineResult<()> {
+    let p: PathBuf = vec![DEV_PATH, pool].iter().collect();
+    fs::remove_dir_all(&p)?;
+    Ok(())
+}
+
+/// Rename the directory to match the pool's new name.
+pub fn pool_renamed(old_name: &str, new_name: &str) -> EngineResult<()> {
+    let old: PathBuf = vec![DEV_PATH, old_name].iter().collect();
+    let new: PathBuf = vec![DEV_PATH, new_name].iter().collect();
+    fs::rename(&old, &new)?;
+    Ok(())
+}
+
+/// Create a symlink to the new filesystem's block device within its pool's
+/// directory.
+pub fn filesystem_added(pool_name: &str, fs_name: &str, devnode: &Path) -> EngineResult<()> {
+    let p: PathBuf = vec![DEV_PATH, pool_name, fs_name].iter().collect();
+
+    // Remove existing and recreate to ensure it points to the correct devnode
+    let _ = fs::remove_file(&p);
+    symlink(devnode, &p)?;
+    Ok(())
+}
+
+/// Remove the symlink when the filesystem is destroyed.
+pub fn filesystem_removed(pool_name: &str, fs_name: &str) -> EngineResult<()> {
+    let p: PathBuf = vec![DEV_PATH, pool_name, fs_name].iter().collect();
+    fs::remove_file(&p)?;
+    Ok(())
+}
+
+/// Rename the symlink to track the filesystem's new name.
+pub fn filesystem_renamed(pool_name: &str, old_name: &str, new_name: &str) -> EngineResult<()> {
+    let old: PathBuf = vec![DEV_PATH, pool_name, old_name].iter().collect();
+    let new: PathBuf = vec![DEV_PATH, pool_name, new_name].iter().collect();
+    fs::rename(&old, &new)?;
+    Ok(())
+}
--- a/src/engine/strat_engine/engine.rs
+++ b/src/engine/strat_engine/engine.rs
@@ -14,6 +14,7 @@ use super::super::structures::Table;
 use super::super::types::{DevUuid, PoolUuid, Redundancy, RenameAction};
 
 use super::cleanup::teardown_pools;
+use super::devlinks;
 use super::pool::StratPool;
 use super::setup::find_all;
 
@@ -79,10 +80,14 @@ impl StratEngine {
             }
         }
 
-        Ok(StratEngine {
-               pools: table,
-               watched_dev_last_event_nrs: HashMap::new(),
-           })
+        let engine = StratEngine {
+            pools: table,
+            watched_dev_last_event_nrs: HashMap::new(),
+        };
+
+        devlinks::setup_devlinks(engine.pools().into_iter())?;
+
+        Ok(engine)
     }
 
     /// Teardown Stratis, preparatory to a shutdown.
@@ -113,12 +118,29 @@ impl Engine for StratEngine {
         let pool = StratPool::initialize(name, &dm, blockdev_paths, redundancy, force)?;
 
         let uuid = pool.uuid();
+        devlinks::pool_added(pool.name())?;
         self.pools.insert(pool);
+
         Ok(uuid)
     }
 
     fn destroy_pool(&mut self, uuid: PoolUuid) -> EngineResult<bool> {
-        destroy_pool!{self; uuid}
+        if let Some(pool) = self.pools.get_by_uuid(uuid) {
+            if pool.has_filesystems() {
+                return Err(EngineError::Engine(ErrorEnum::Busy,
+                                               "filesystems remaining on pool".into()));
+            };
+        } else {
+            return Ok(false);
+        }
+
+        let pool = self.pools
+            .remove_by_uuid(uuid)
+            .expect("Must succeed since self.pools.get_by_uuid() returned a value");
+        let pool_name = pool.name().to_owned();
+        pool.destroy()?;
+        devlinks::pool_removed(&pool_name)?;
+        Ok(true)
     }
 
     fn rename_pool(&mut self, uuid: PoolUuid, new_name: &str) -> EngineResult<RenameAction> {
@@ -135,6 +157,7 @@ impl Engine for StratEngine {
             Err(err)
         } else {
             self.pools.insert(pool);
+            devlinks::pool_renamed(&old_name, new_name)?;
             Ok(RenameAction::Renamed)
         }
     }
--- a/src/engine/macros.rs
+++ b/src/engine/macros.rs
@@ -14,23 +14,6 @@ macro_rules! calculate_redundancy {
     }
 }
 
-macro_rules! destroy_pool {
-    ( $s:ident; $uuid: ident) => {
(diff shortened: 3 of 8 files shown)


src/engine/strat_engine/mdv.rs
```diff
@@ -73,7 +79,9 @@ impl<'a> MountedMDV<'a> {
 impl<'a> Drop for MountedMDV<'a> {
     fn drop(&mut self) {
         if let Err(err) = umount(&self.mdv.mount_pt) {
-            warn!("Could not unmount MDV: {}", err)
+            warn!("Could not unmount MDV: {}", err);
```
mulkieran: The drop method needs a comment explaining why this is happening. Not a header comment, just a regular one.

agrover: you mean, why we're trying to umount, or why the umount might have failed?

agrover: in any case, broke out this commit into its own PR, #689, so let's continue over there.

agrover:
rebased and moved a commit to its own PR.

agrover:
rebased.

agrover:
rebased

mulkieran:
Tests don't compile.

agrover:
Fixed breakage in test code.

src/stratis/stratis.rs
```diff
@@ -2,5 +2,103 @@
...
 
 pub const VERSION: &str = env!("CARGO_PKG_VERSION");
+
+pub const DEV_PATH: &str = "/dev/stratis";
+
+/// Set up directories and symlinks under /dev/stratis based on current
+/// config. Clear out other stuff.
+pub fn setup_devlinks(engine: &Rc<RefCell<Engine>>) -> StratisResult<()> {
+    let mut existing_dirs = HashSet::new();
+    for dir_e in fs::read_dir(DEV_PATH)? {
```
trgill: The read_dir() causes a failure on one of the Jenkins machines because /dev/stratis doesn't exist.  I thought something was wrong with the test machine - so I looked at it.  Just an FYI.

agrover: thanks for noticing these issues!

mulkieran:
I _think_ the two commits are separate and could be put in separate PRs.

agrover:
@mulkieran yes. I've opened #703.

agrover:
Moved /dev/stratis creation out of stratengine code into setup_devlinks to address the issue @trgill found. Created an issue for the case that /dev/stratis is a regular file, since I think this is separate.

mulkieran:
@tasleson review?

src/stratis/stratis.rs
```diff
@@ -2,5 +2,110 @@
...
+/// config. Clear out other stuff.
+pub fn setup_devlinks(engine: &Rc<RefCell<Engine>>) -> StratisResult<()> {
+    if let Err(err) = fs::create_dir(DEV_PATH) {
+        if err.kind() != ErrorKind::AlreadyExists {
+            return Err(From::from(err));
+        }
+    }
+
+    let mut existing_dirs = HashSet::new();
+    for dir_e in fs::read_dir(DEV_PATH)? {
```
mulkieran: Minor: I bet there's an opportunity for a map and a collect here.

mulkieran:
The Sim Engine doesn't effect the world at all. Therefore, I don't think the management of these symlinks needs to be in this higher level. It could go at the strat engine level.

agrover:
> What is the justification for putting calls to the new methods you defined into the D-Bus layer?

I thought it was the better of two bad choices between the API layer and the engine. (We don't have a spot for method-specific code that isn't engine- or api-specific, right now.)

This was driven by a preconception that we needed this also to be there when SimEngine was used. But if you don't think so, then ok, let's put it in StratEngine! (If we ever get another "real" engine implementation, we may need to revisit.)

src/engine/strat_engine/engine.rs
```diff
@@ -113,12 +118,29 @@ impl Engine for StratEngine {
...
+            };
+        } else {
+            return Ok(false);
+        }
+
+        let pool = self.pools
+            .remove_by_uuid(uuid)
+            .expect("Must succeed since $s.pool.get_by_uuid() returned a value");
+        let pool_name = pool.name().to_owned();
+        pool.destroy()?;
```
tasleson: Would it be better to remove the the devlink before you remove the actual pool?  Trying to think which is better, leaving a symlink for a pool that failed to be removed or missing a symlink for a pool that still exists?

I keep thinking it should be done in reverse order of creation.

agrover:
rebased and squashed. @mulkieran review?

mulkieran:
```stratisd_client_dbus``` failure is spurious.

File: src/engine/strat_engine/devlinks.rs [hunk 5]
```diff
@@ -0,0 +1,118 @@
...
+use std::str;
+
+use super::super::errors::EngineResult;
+
+use engine::Engine;
+
+pub const DEV_PATH: &str = "/dev/stratis";
+
+/// Set up directories and symlinks under /dev/stratis based on current
+/// config. Clear out other stuff.
```
mulkieran: Maybe you want to be precise about meaning of "other stuff".

src/engine/strat_engine/devlinks.rs
```diff
@@ -0,0 +1,118 @@
...
+    if let Err(err) = fs::create_dir(DEV_PATH) {
+        if err.kind() != ErrorKind::AlreadyExists {
+            return Err(From::from(err));
+        }
+    }
+
+    let mut existing_dirs: HashSet<_> = fs::read_dir(DEV_PATH)?
+        .map(|dir_e| {
+                 dir_e
+                     .expect("direntry is valid")
```
mulkieran: This expect() is mistaken. Here's the docs:

```This io::Result will be an Err if there's some sort of intermittent IO error during iteration.```.

File: src/engine/strat_engine/devlinks.rs [hunk 7]
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    for leftover in existing_dirs {
+        pool_removed(str::from_utf8(leftover.as_bytes()).expect("is valid utf8"))
+    }
+
+    Ok(())
+}
+
+pub fn pool_added(pool: &str) -> () {
```
mulkieran: Why not just fail on all of these?

src/engine/strat_engine/devlinks.rs
```diff
@@ -0,0 +1,118 @@
...
+                     .expect("direntry is valid")
+                     .file_name()
+                     .into_string()
+                     .expect("Unix is utf8")
+             })
+        .collect();
+
+    for pool in engine.pools() {
+        if !existing_dirs.contains(pool.name()) {
+            pool_added(pool.name());
```
mulkieran: I would suggest that if you can't add the thing, you should just fail. Your method claims to do X. If it can't do X, and this is part of doing X, then returning an error is the correct thing to do, because it didn't do X.

src/engine/strat_engine/devlinks.rs
```diff
@@ -0,0 +1,118 @@
...
+        }
+        existing_dirs.remove(pool.name());
```
mulkieran: You can combine the remove and the ```if``` above by using the return value of ```remove()```.

src/engine/strat_engine/devlinks.rs
```diff
@@ -0,0 +1,118 @@
...
+
+        let pool_path: PathBuf = vec![DEV_PATH, pool.name()].iter().collect();
+
+        let mut existing_files: HashSet<_> = fs::read_dir(pool_path)?
+            .map(|dir_e| {
+                     dir_e
+                         .expect("direntry is valid")
```
mulkieran: Same as previous expect comment.

src/engine/strat_engine/devlinks.rs
```diff
@@ -0,0 +1,118 @@
...
+                         .into_string()
+                         .expect("Unix is utf8")
+                 })
+            .collect();
+
+        for fs in pool.filesystems() {
+            if !existing_files.contains(fs.name()) {
+                filesystem_added(pool.name(), fs.name(), &fs.devnode());
+            }
+            existing_files.remove(fs.name());
```
mulkieran: Same remove idiom can be used here as well.

src/engine/strat_engine/devlinks.rs
```diff
@@ -0,0 +1,118 @@
...
+                         .expect("direntry is valid")
+                         .file_name()
+                         .into_string()
+                         .expect("Unix is utf8")
+                 })
+            .collect();
+
+        for fs in pool.filesystems() {
+            if !existing_files.contains(fs.name()) {
+                filesystem_added(pool.name(), fs.name(), &fs.devnode());
```
mulkieran: Better to return an error.

src/engine/strat_engine/devlinks.rs
```diff
@@ -0,0 +1,118 @@
...
+            existing_files.remove(fs.name());
+        }
+
+        for leftover in existing_files {
+            filesystem_removed(pool.name(), &leftover)
+        }
+    }
+
+    for leftover in existing_dirs {
+        pool_removed(str::from_utf8(leftover.as_bytes()).expect("is valid utf8"))
```
mulkieran: Better to return an error.

src/engine/strat_engine/engine.rs
```diff
@@ -113,12 +118,29 @@ impl Engine for StratEngine {
...
         let uuid = pool.uuid();
+        devlinks::pool_added(pool.name());
         self.pools.insert(pool);
+
         Ok(uuid)
     }
 
     fn destroy_pool(&mut self, uuid: PoolUuid) -> EngineResult<bool> {
-        destroy_pool!{self; uuid}
+        if let Some(pool) = self.pools.get_by_uuid(uuid) {
```
mulkieran: I think you can still use the macro, almost unmodified, so it makes sense to do so.

src/engine/strat_engine/engine.rs
```diff
@@ -113,12 +118,29 @@ impl Engine for StratEngine {
...
+        } else {
+            return Ok(false);
+        }
+
+        let pool = self.pools
+            .remove_by_uuid(uuid)
+            .expect("Must succeed since self.pool.get_by_uuid() returned a value");
+        let pool_name = pool.name().to_owned();
+        pool.destroy()?;
+        devlinks::pool_removed(&pool_name);
```
mulkieran: Just fail here. Pool only partially removed, bad.

src/engine/strat_engine/devlinks.rs
```diff
@@ -0,0 +1,118 @@
...
+                                                     old.display(),
+                                                     new.display())
+                                          });
+}
+
+pub fn filesystem_added(pool_name: &str, fs_name: &str, devnode: &Path) -> () {
+    let p: PathBuf = vec![DEV_PATH, pool_name, fs_name].iter().collect();
+
+    // Remove existing and recreate to ensure it points to the correct devnode
+    // Failing to remove existing is ok.
```
mulkieran: Why is it ok?

src/engine/strat_engine/devlinks.rs
```diff
@@ -0,0 +1,118 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use std::collections::HashSet;
+use std::fs;
+use std::io::ErrorKind;
+use std::os::unix::fs::symlink;
```
mulkieran: In theory we're running on Linux. Why is unix specific method OK?

agrover: Linux is Unix (-like)

src/engine/strat_engine/devlinks.rs
agrover: ok

src/engine/strat_engine/devlinks.rs
agrover: ok. let's bail if this ever happens, I guess.

src/engine/strat_engine/devlinks.rs
agrover: ok. changing everything to return a result instead of `()`.

src/engine/strat_engine/devlinks.rs
agrover: 👍

src/engine/strat_engine/devlinks.rs
agrover: Went a step further and removed the conditional entirely. This will remove/add all fs symlinks, and should protect against if a symlink exists but points to the wrong dev, which the current code doesn't fixup.

src/engine/strat_engine/devlinks.rs
agrover: It's not. Basically the entire PR was operating under a "best effort" approach, whereas now it won't.

src/engine/strat_engine/engine.rs
agrover: I tried but didn't have any luck without causing other ugly ripples.

agrover:
rebased and updated based on comments.

src/engine/strat_engine/devlinks.rs
```diff
@@ -0,0 +1,102 @@
...
+use std::str;
+
+use super::super::errors::EngineResult;
+
+use engine::Engine;
+
+pub const DEV_PATH: &str = "/dev/stratis";
+
+/// Set up directories and symlinks under /dev/stratis based on current
+/// config. Clear out anything that doesn't belong.
```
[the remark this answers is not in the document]
agrover: ok sounds good

src/engine/strat_engine/devlinks.rs
```diff
@@ -0,0 +1,102 @@
...
+// Don't just remove and recreate everything in case there are processes
+// (e.g. user shells) with the current working directory within the tree.
+pub fn setup_devlinks(engine: Option<&Engine>) -> EngineResult<()> {
```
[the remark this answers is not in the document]
agrover: I actually tried this but couldn't figure out how to express what the test code needs succinctly, an `iterator<item=&Pool>` that immediately yields `None`. Any ideas?

src/engine/strat_engine/devlinks.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    for leftover in existing_dirs {
+        pool_removed(str::from_utf8(leftover.as_bytes()).expect("is valid utf8"))?
+    }
+
+    Ok(())
+}
+
+pub fn pool_added(pool: &str) -> EngineResult<()> {
```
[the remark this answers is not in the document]
agrover: ok will do.

src/engine/strat_engine/devlinks.rs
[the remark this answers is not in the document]
mulkieran: Make an empty vec and call ```.iter()```. I bet you'll need to stick in a type annotation at invocation site to help out the checker; due to a puzzling limitation in Rust type-inference.

agrover:
updated based on comments.

src/engine/strat_engine/engine.rs
```diff
@@ -79,10 +80,14 @@ impl StratEngine {
...
-        Ok(StratEngine {
-               pools: table,
-               watched_dev_last_event_nrs: HashMap::new(),
-           })
+        let engine = StratEngine {
+            pools: table,
+            watched_dev_last_event_nrs: HashMap::new(),
+        };
+
+        devlinks::setup_devlinks(engine.pools().into_iter())?;
```
mulkieran: So long as ```Table::into_iter()``` doesn't consume all the pools this is good. It looks like it doesn't, but I'm not quite sure.

agrover: (sorry for delay, fell off my radar!)

`Engine::pools()` returns a `Vec<&Pool>` by value, so this vec of trait objects is the thing being consumed rather than the pools themselves.

(`Engine::pools()` also calls into_iter(), but this is ok because Table::into_iter() acts like .iter() right now. See #678 or #509.)

trgill: @mulkieran this ok with you?

Issue #656, opened 2017-11-15
Use a library to manage the event loop


agrover:
We're doing it ourselves right now. We should look into Mio (still lowish level) or Tokio (higher-level) if the event loop continues to grow more complex.

mulkieran:
This is getting less optional all the time, so I took off the enhancement label.

agrover:
This assumes an implementation that @tasleson may choose to pick an alternative solution, so may become invalid, we'll see.

agrover:
I don't know about mio, but I think recent dbus versions have added tokio support, we could look into that.

mulkieran:
There's a separate dbus-tokio crate, now, but I think it is still a work-in-progress.

agrover:
Not sure this is worth doing any more, since we have done more work to improve our event loop code, and other solutions are really more network-focused than what we need. Closing.

PR #690, opened 2017-12-17
Handle udev add events

tasleson:
Block devices can be delayed at boot or be systematically added after the system is up and running.  Handle these udev events by evaluating block devices when they are added, to see if they have
a valid stratis signature.  If they do we will attempt to assemble it into a complete pool.

Preliminary manual testing shows that this change appears to work, but it needs quite a bit of scrutiny and a bunch of added unit tests before it can be considered complete.  Making it available so people can see the direction I'm taking with it, to see if it's appropriate.  I'm trying to utilize as much of the existing code as possible.

### DIFF
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -2,6 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+extern crate devicemapper;
 extern crate libstratis;
 #[macro_use]
 extern crate log;
@@ -10,15 +11,18 @@ extern crate clap;
 #[cfg(feature="dbus_enabled")]
 extern crate dbus;
 extern crate libc;
+extern crate libudev;
 
 #[cfg(test)]
 extern crate quickcheck;
 
+use std::cell::RefCell;
 use std::env;
 use std::error::Error;
-use std::rc::Rc;
-use std::cell::RefCell;
+use std::os::unix::io::AsRawFd;
+use std::path::PathBuf;
 use std::process::exit;
+use std::rc::Rc;
 
 use clap::{App, Arg};
 use env_logger::LogBuilder;
@@ -26,6 +30,8 @@ use log::{LogLevelFilter, SetLoggerError};
 
 #[cfg(feature="dbus_enabled")]
 use dbus::WatchEvent;
+
+use devicemapper::Device;
 
 use libstratis::engine::{Engine, SimEngine, StratEngine};
 use libstratis::stratis::{StratisResult, StratisError, VERSION};
@@ -53,6 +59,25 @@ fn initialize_log(debug: bool) -> Result<(), SetLoggerError> {
     builder.init()
 }
 
+/// Given a udev event check to see if it's an add and if it is return the device node and
+/// devicemapper::Device.
+fn handle_udev_add(event: &libudev::Event) -> Option<(PathBuf, Device)> {
+    if event.event_type() == libudev::EventType::Add {
+        let device = event.device();
+        return device
+                   .devnode()
+                   .and_then(|devnode| {
+                                 device
+                                     .devnum()
+                                     .and_then(|devnum| {
+                                                   Some((PathBuf::from(devnode),
+                                                         Device::from(devnum)))
+                                               })
+                             });
+    }
+    None
+}
+
 fn run() -> StratisResult<()> {
 
     let matches = App::new("stratis")
@@ -69,6 +94,18 @@ fn run() -> StratisResult<()> {
     initialize_log(matches.is_present("debug"))
         .expect("This is the first and only invocation of this method; it must succeed.");
 
+    // We must setup a udev listener before we initialize the
+    // engine. It is possible that a device may appear after the engine has read the
+    // /dev directory but before it has completed initialization. Unless the udev
+    // event has been recorded, the engine will be unaware of the existence of the
+    // device.
+    // This is especially important since stratisd is required to be able to run
+    // during early boot.
+    let context = libudev::Context::new()?;
+    let mut monitor = libudev::Monitor::new(&context)?;
+    monitor.match_subsystem_devtype("block", "disk")?;
+    let mut udev = monitor.listen()?;
+
     let engine: Rc<RefCell<Engine>> = {
         if matches.is_present("sim") {
             info!("Using SimEngine");
@@ -79,54 +116,109 @@ fn run() -> StratisResult<()> {
         }
     };
 
-    #[cfg(feature="dbus_enabled")]
-    let (dbus_conn, mut tree, dbus_context) = libstratis::dbus_api::connect(Rc::clone(&engine))?;
+    /*
+    The file descriptor array indexes are laid out in the following:
+
+    0   == Always udev fd index
+    1   == engine index if eventable
+    1/2 == Start of dbus client file descriptor(s), 1 if engine is not eventable, else 2
+    */
+    const FD_INDEX_UDEV: usize = 0;
+    const FD_INDEX_ENGINE: usize = 1;
+
 
     let mut fds = Vec::new();
 
+    fds.push(libc::pollfd {
+                 fd: udev.as_raw_fd(),
+                 revents: 0,
+                 events: libc::POLLIN,
+             });
+
     let mut eventable = engine.borrow_mut().get_eventable()?;
-    let engine_fds_end_idx = match eventable {
+
+    // The variable _dbus_client_index_start is only used when dbus support is compiled in, thus
+    // we denote the value as not needed to compile when dbus support is not included.
+    let (engine_eventable, poll_timeout, _dbus_client_index_start) = match eventable {
         Some(ref mut evt) => {
             fds.push(libc::pollfd {
                          fd: evt.get_pollable_fd(),
                          revents: 0,
                          events: libc::POLLIN,
                      });
-            1
+
+            // Don't timeout if eventable, we are event driven
+            (true, -1, FD_INDEX_ENGINE + 1)
         }
-        None => 0,
+
+        // We periodically need to timeout as we are not event driven
+        None => (false, 10000, FD_INDEX_ENGINE),
     };
 
-    // Don't timeout if eventable lets us be event-driven
-    let poll_timeout = if engine_fds_end_idx != 0 { -1 } else { 10000 };
+    #[cfg(feature="dbus_enabled")]
+    let (mut dbus_conn, mut tree, base_object_path, mut dbus_context) =
+        libstratis::dbus_api::connect(Rc::clone(&engine))?;
 
     loop {
-        // Unconditionally call evented() if engine has no eventable.
-        // This looks like a bad idea, but the only engine that has
-        // no eventable is the sim engine, and for that engine, evented()
-        // is essentially a no-op.
-        if engine_fds_end_idx == 0 {
-            engine.borrow_mut().evented()?;
+        // Process any udev block events
+        if fds[FD_INDEX_UDEV].revents != 0 {
+            loop {
+                match udev.receive_event() {
+                    Some(event) => {
+                        if let Some((devnode, device)) = handle_udev_add(&event) {
+
+                            // If block evaluate returns an error we are going to ignore it as
+                            // there is nothing we can do for a device we are getting errors with.
+                            let pool_uuid = engine
+                                .borrow_mut()
+                                .block_evaluate(devnode, device)
+                                .unwrap_or(None);
+
+                            // We need to pretend that we aren't using the variable _pool_uuid so
+                            // that we can conditionally compile out the register_pool when dbus
+                            // is not enabled.
+                            if let Some(_pool_uuid) = pool_uuid {
+                                #[cfg(feature="dbus_enabled")]
+                                libstratis::dbus_api::register_pool(&mut dbus_conn,
+                                                                    Rc::clone(&engine),
+                                                                    &mut dbus_context,
+                                                                    &mut tree,
+                                                                    _pool_uuid,
+                                                                    &base_object_path)?;
+                            }
+                        }
+                    }
+                    None => {
+                        break;
+                    }
+                };
+            }
         }
 
-        // Handle engine fd, if there is one.
-        // If engine has no eventable, there is no engine fd; in that case
-        // the body of the loop is executed zero times.
-        for pfd in fds[..engine_fds_end_idx]
-                .iter_mut()
-                .filter(|pfd| pfd.revents != 0) {
-            pfd.revents = 0;
-            eventable
-                .as_mut()
-                .expect("index < engine_fds_end_idx <=> index == 0 <=> eventable.is_some()")
-                .clear_event()?;
+        // Handle engine events, if the engine is eventable
+        if engine_eventable {
+            if fds[FD_INDEX_ENGINE].revents != 0 {
+                fds[FD_INDEX_ENGINE].revents = 0;
+
+                eventable
+                    .as_mut()
+                    .expect("eventable.is_some()")
+                    .clear_event()?;
+
+                engine.borrow_mut().evented()?;
+            }
+        } else {
+            // Unconditionally call evented() if engine has no eventable.
+            // This looks like a bad idea, but the only engine that has
+            // no eventable is the sim engine, and for that engine, evented()
+            // is essentially a no-op.
             engine.borrow_mut().evented()?;
         }
 
         // Iterate through D-Bus file descriptors (if enabled)
         #[cfg(feature="dbus_enabled")]
         {
-            for pfd in fds[engine_fds_end_idx..]
+            for pfd in fds[_dbus_client_index_start..]
                     .iter()
                     .filter(|pfd| pfd.revents != 0) {
                 for item in dbus_conn.watch_handle(pfd.fd, WatchEvent::from_revents(pfd.revents)) {
@@ -141,7 +233,7 @@ fn run() -> StratisResult<()> {
 
             // Refresh list of dbus fds to poll for every time. This can change as
             // D-Bus clients come and go.
-            fds.truncate(engine_fds_end_idx);
+            fds.truncate(_dbus_client_index_start);
 
             fds.extend(dbus_conn.watch_fds().iter().map(|w| w.to_pollfd()));
         }
--- a/src/dbus_api/api.rs
+++ b/src/dbus_api/api.rs
@@ -24,8 +24,9 @@ use dbus::tree::MethodInfo;
 use dbus::tree::PropInfo;
 use dbus::tree::Tree;
 use dbus::ConnectionItem;
+use uuid::Uuid;
 
-use engine::Engine;
+use engine::{Engine, Pool};
 use stratis::VERSION;
 
 use super::filesystem::create_dbus_filesystem;
@@ -194,9 +195,22 @@ fn get_base_tree<'a>(dbus_context: DbusContext) -> (Tree<MTFn<TData>, TData>, db
     (base_tree.add(obj_path), path)
 }
 
+/// Given an Pool, create all the needed dbus objects to represent it.
+fn register_pool_dbus(dbus_context: &DbusContext, pool: &Pool, object_path: &dbus::Path<'static>) {
+    let pool_path = create_dbus_pool(dbus_context, object_path.clone(), pool.uuid());
+    for fs_uuid in pool.filesystems().iter().map(|f| f.uuid()) {
+        create_dbus_filesystem(dbus_context, pool_path.clone(), fs_uuid);
+    }
+    for dev_uuid in pool.blockdevs().iter().map(|bd| bd.uuid()) {
+        create_dbus_blockdev(dbus_context, pool_path.clone(), dev_uuid);
+    }
+}
+
+/// Connect a stratis engine to dbus.
 #[allow(type_complexity)]
-pub fn connect(engine: Rc<RefCell<Engine>>)
-               -> Result<(Connection, Tree<MTFn<TData>, TData>, DbusContext), dbus::Error> {
+pub fn connect<'a>
+    (engine: Rc<RefCell<Engine>>)
+     -> Result<(Connection, Tree<MTFn<TData>, TData>, dbus::Path<'a>, DbusContext), dbus::Error> {
     let c = Connection::get_private(BusType::System)?;
 
     let local_engine = Rc::clone(&engine);
@@ -204,26 +218,33 @@ pub fn connect(engine: Rc<RefCell<Engine>>)
     let (mut tree, object_path) = get_base_tree(DbusContext::new(engine));
     let dbus_context = tree.get_data().clone();
 
-    // This should never panic as create_dbus_pool(),
-    // create_dbus_filesystem(), and create_dbus_blockdev() do not borrow the
-    // engine.
+    // This should never panic as register_pool_dbus does not borrow the engine.
     for pool in local_engine.borrow().pools() {
-        let pool_path = create_dbus_pool(&dbus_context, object_path.clone(), pool.uuid());
-        for fs_uuid in pool.filesystems().iter().map(|f| f.uuid()) {
-            create_dbus_filesystem(&dbus_context, pool_path.clone(), fs_uuid);
-        }
-        for dev_uuid in pool.blockdevs().iter().map(|bd| bd.uuid()) {
-            create_dbus_blockdev(&dbus_context, pool_path.clone(), dev_uuid);
-        }
+        register_pool_dbus(&dbus_context, pool, &object_path);
     }
 
     tree.set_registered(&c, true)?;
-
     c.register_name(STRATIS_BASE_SERVICE, NameFlag::ReplaceExisting as u32)?;
 
     process_deferred_actions(&c, &mut tree, &mut dbus_context.actions.borrow_mut())?;
 
-    Ok((c, tree, dbus_context))
+    Ok((c, tree, object_path, dbus_context))
+}
+
+/// Given the UUID of a pool, register all the pertinent information with dbus.
+pub fn register_pool(c: &Connection,
+                     engine: Rc<RefCell<Engine>>,
+                     dbus_context: &DbusContext,
+                     tree: &mut Tree<MTFn<TData>, TData>,
+                     pool_uuid: Uuid,
+                     object_path: &dbus::Path<'static>)
+                     -> Result<(), dbus::Error> {
+
+    if let Some(pool) = engine.borrow().get_pool(pool_uuid) {
+        register_pool_dbus(dbus_context, pool, object_path);
+        return process_deferred_actions(c, tree, &mut dbus_context.actions.borrow_mut());
+    }
+    Ok(())
 }
 
 /// Update the dbus tree with deferred adds and removes.
--- a/src/engine/engine.rs
+++ b/src/engine/engine.rs
@@ -9,7 +9,7 @@ use std::path::{Path, PathBuf};
 use chrono::{DateTime, Utc};
 use uuid::Uuid;
 
-use devicemapper::Sectors;
+use devicemapper::{Sectors, Device};
 
 use super::errors::EngineResult;
 use super::types::{BlockDevState, FilesystemUuid, PoolUuid, DevUuid, RenameAction};
@@ -146,6 +146,14 @@ pub trait Engine: Debug {
                    redundancy: Option<u16>,
                    force: bool)
                    -> EngineResult<PoolUuid>;
+
+    /// Evaluate a device node & devicemapper::Device to see if it's a valid
+    /// stratis device.  If all the devices are present in the pool and the pool isn't already
+    /// up and running, it will get setup and the pool uuid will be returned.
+    fn block_evaluate(&mut self,
+                      dev_node: PathBuf,
+                      device: Device)
+                      -> EngineResult<Option<PoolUuid>>;
 
     /// Destroy a pool.
     /// Ensures that the pool of the given UUID is absent on completion.
--- a/src/engine/sim_engine/engine.rs
+++ b/src/engine/sim_engine/engine.rs
@@ -2,12 +2,16 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+extern crate libc;
+
 use std::cell::RefCell;
 use std::collections::HashSet;
 use std::collections::hash_map::RandomState;
 use std::iter::FromIterator;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::rc::Rc;
+
+use devicemapper::Device;
 
 use super::super::engine::{Engine, Eventable, HasName, HasUuid, Pool};
 use super::super::errors::{EngineError, EngineResult, ErrorEnum};
@@ -56,6 +60,15 @@ impl Engine for SimEngine {
         self.pools.insert(pool);
 
         Ok(uuid)
+    }
+
+    fn block_evaluate(&mut self,
+                      dev_node: PathBuf,
+                      device: Device)
+                      -> EngineResult<Option<PoolUuid>> {
+        assert_ne!(dev_node, PathBuf::from("/"));
+        assert_ne!(libc::dev_t::from(device), 0);
+        Ok(None)
     }
 
     fn destroy_pool(&mut self, uuid: PoolUuid) -> EngineResult<bool> {
--- a/src/engine/strat_engine/engine.rs
+++ b/src/engine/strat_engine/engine.rs
@@ -2,11 +2,13 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+extern crate libc;
+
 use std::collections::HashMap;
 use std::os::unix::io::{AsRawFd, RawFd};
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
-use devicemapper::{DM, DmNameBuf};
+use devicemapper::{DM, Device, DmNameBuf};
 
 use super::super::engine::{Engine, Eventable, HasName, HasUuid, Pool};
 use super::super::errors::{EngineError, EngineResult, ErrorEnum};
@@ -16,7 +18,7 @@ use super::super::types::{DevUuid, PoolUuid, Redundancy, RenameAction};
 use super::cleanup::teardown_pools;
 use super::devlinks;
 use super::pool::StratPool;
-use super::setup::find_all;
+use super::setup::{find_all, is_stratis_device};
 
 const REQUIRED_DM_MINOR_VERSION: u32 = 37;
 
@@ -30,6 +32,11 @@ pub enum DevOwnership {
 #[derive(Debug)]
 pub struct StratEngine {
     pools: Table<StratPool>,
+
+    // Map of stratis devices that have been found but one or more stratis block devices are missing
+    // which prevents the associated pools from being setup.
+    incomplete_pools: HashMap<PoolUuid, HashMap<Device, PathBuf>>,
+
     // Maps name of DM devices we are watching to the most recent event number
     // we've handled for each
     watched_dev_last_event_nrs: HashMap<DmNameBuf, u32>,
@@ -52,10 +59,13 @@ impl StratEngine {
     /// Setup a StratEngine.
     /// 1. Verify the existence of Stratis /dev directory.
     /// 2. Setup all the pools belonging to the engine.
+    ///    a. Places any devices which belong to a pool, but are not complete
+    ///       in the incomplete pools data structure.
     ///
     /// Returns an error if the kernel doesn't support required DM features.
     /// Returns an error if there was an error reading device nodes.
-    /// Returns an error if there was an error setting up any of the pools.
+    /// Returns an error and tears down all the pools if we find a pool with a
+    /// duplicate name found.
     pub fn initialize() -> EngineResult<StratEngine> {
         let dm = DM::new()?;
         let minor_dm_version = dm.version()?.1;
@@ -69,20 +79,29 @@ impl StratEngine {
         let pools = find_all()?;
 
         let mut table = Table::default();
+        let mut incomplete_pools = HashMap::new();
+
         for (pool_uuid, devices) in &pools {
-            let evicted = table.insert(StratPool::setup(&dm, *pool_uuid, devices)?);
-            if !evicted.is_empty() {
-
-                // TODO: update state machine on failure.
-                let _ = teardown_pools(table.empty());
-
-                let err_msg = "found two pools with the same id or name";
-                return Err(EngineError::Engine(ErrorEnum::Invalid, err_msg.into()));
+            match StratPool::setup(&dm, *pool_uuid, devices) {
+                Ok(pool) => {
+                    let evicted = table.insert(pool);
+                    if !evicted.is_empty() {
+                        // TODO: update state machine on failure.
+                        let _ = teardown_pools(table.empty());
+                        let err_msg = "found two pools with the same name";
+                        return Err(EngineError::Engine(ErrorEnum::Invalid, err_msg.into()));
+                    }
+                }
+                Err(_) => {
+                    // Unable to setup pool, it might become complete later.
+                    incomplete_pools.insert(*pool_uuid, devices.clone());
+                }
             }
         }
 
         let engine = StratEngine {
             pools: table,
+            incomplete_pools,
             watched_dev_last_event_nrs: HashMap::new(),
         };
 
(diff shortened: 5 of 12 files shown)


src/bin/stratisd.rs
```diff
@@ -83,10 +105,23 @@ fn run() -> StratisResult<()> {
         }
     };
 
-    let (dbus_conn, mut tree, dbus_context) = libstratis::dbus_api::connect(Rc::clone(&engine))?;
+    let (mut dbus_conn, mut tree, base_object_path, mut dbus_context) =
```
mulkieran: It was pretty messy when ```fds``` was shared by only two different types of file descriptors requiring special treatment. If it's going to be shared by 3, which it probably has to be, then this whole setup should get some improvement.

src/bin/stratisd.rs
```diff
@@ -46,6 +49,25 @@ fn write_or_panic(err: StratisError) -> () {
...
 
+/// Given a udev event check to see if it's an add and if it is return the device node and
+/// device number.
+fn handle_udev_add(event: &libudev::Event) -> Option<(PathBuf, u64)> {
+    match event.event_type() {
+        libudev::EventType::Add => {
+            info!("Hot plug add {:?}{:?}",
+                  event.device().devnum(),
+                  event.device().devnode());
+            if let Some(devnode) = event.device().devnode() {
```
mulkieran: Probably a true Rustacean would go with:

```
let device = event.device();
device.devnode().and_then(|devnode| device.devnum().and_then(|devnum| (PathBuf::from(devnode), devnum)))
```
or something similar as whole body of function.

src/engine/strat_engine/setup.rs
```diff
@@ -26,6 +26,67 @@ use super::range_alloc::RangeAllocator;
...
+    //
+    // 2. ENOMEDIUM: The device has no medium. An example of this case is an
+    // empty optical drive.
+    //
+    // Note that it is better to be conservative and return with an
+    // error in any case where failure to read the device could result
+    // in bad data for Stratis. Additional exceptions may be added,
+    // but only with a complete justification.
+    if f.is_err() {
+        let err = f.unwrap_err();
```
mulkieran: Please use a separate commit to change ```unwrap_err()``` to ```expect_err("f.is_err()")```. We use expect function when something can be proved to be true. We probably didn't know that ```expect_err()``` existed when this code was written, hence the oversight.

src/engine/strat_engine/setup.rs
```diff
@@ -26,6 +26,67 @@ use super::range_alloc::RangeAllocator;
...
+            }
+        }
+    }
+
+    let mut f = f.expect("unreachable if f is err");
+    if let DevOwnership::Ours(pool_uuid, _) = StaticHeader::determine_ownership(&mut f)? {
+        // No value should ever be ejected, because duplicate device nodes
+        // are filtered out above. Therefore, the return value of insert()
+        // might as well be ignored.
+        debug!("STRATIS DEVICE {:?} {:?}", pool_uuid, devnode);
```
mulkieran: I expect this will go away in finished copy.

src/engine/strat_engine/setup.rs
```diff
@@ -26,6 +26,67 @@ use super::range_alloc::RangeAllocator;
...
+
+        let _ = pool_map
+            .entry(pool_uuid)
+            .or_insert_with(HashMap::new)
+            .insert(Device::from(devno), devnode);
```
mulkieran: It was probably cowardly to have a ```let _``` instead of an ```expect()``` here, see comment above. But now, the circumstance on which the comment relies are no longer true, so the code needs some modfication. Given these changes, I would _not_ pass the HashMap, instead I'ld just return the relevant values.

src/bin/stratisd.rs
agrover: #656 is for this.

src/engine/strat_engine/setup.rs
tasleson: If I return the relevant values I will need to duplicate the same code to insert them into the hash map, is that really better?  Yes the comment is wrong now.

mulkieran: I'm not sure what you mean by duplicate. Won't the insertion code be in just one place?

tasleson: There are two different code paths which call `is_stratis_device`.  If each of them needs the return values placed into a hashmap there will be duplicate insertion code for each to do so.

tasleson:
Squashed and rebased

tasleson:
@mulkieran I believe I've taken into account all your suggestions for this PR

src/bin/stratisd.rs
```diff
@@ -35,6 +38,29 @@ fn print_err(err: StratisError) -> () {
     eprintln!("{}", err.description());
 }
 
+/// Given a udev event check to see if it's an add and if it is return the device node and
+/// device number.
+fn handle_udev_add(event: &libudev::Event) -> Option<(PathBuf, u64)> {
```
agrover: Recommend converting device.devnum() to a devicemapper::Device asap, and then returning that instead of `u64`. (Device has a from(u64, aka libc::dev_t) method.) Change `block_evaluate` to take a `Device`, too.

src/stratis/errors.rs
```diff
@@ -82,3 +87,10 @@ impl From<EngineError> for StratisError {
         StratisError::Engine(err)
     }
 }
+
+/// Allow ability to convert from libudev error to stratis error
```
mulkieran: Might as well skip the comment, as it is a very common idiom and none of the others are commented.

src/bin/stratisd.rs
```diff
@@ -35,6 +38,29 @@ fn print_err(err: StratisError) -> () {
...
+    match event.event_type() {
+        libudev::EventType::Add => {
+            let device = event.device();
+            info!("Hot plug add {:?}{:?}", device.devnum(), device.devnode());
```
mulkieran: Remove in final version.

src/bin/stratisd.rs
```diff
@@ -35,6 +38,29 @@ fn print_err(err: StratisError) -> () {
     eprintln!("{}", err.description());
 }
 
+/// Given a udev event check to see if it's an add and if it is return the device node and
+/// device number.
+fn handle_udev_add(event: &libudev::Event) -> Option<(PathBuf, u64)> {
+    match event.event_type() {
```
mulkieran: Instead of a full match, since you are only interested in Add event, should do:

```
if event.event_type() == libudev::EventType::Add {
    // do stuff
}
```

src/bin/stratisd.rs
```diff
@@ -85,20 +124,45 @@ fn run() -> StratisResult<()> {
...
                          events: libc::POLLIN,
                      });
-            1
+            2
         }
-        None => 0,
+        None => 1,
     };
 
     // Don't timeout if eventable lets us be event-driven
```
mulkieran: The comment and the code don't match here. ```engine_fds_end_idx``` is not equal to 0 always.
So this line is equivalent to:

```let poll_timeout = -1;```

Is that what is wanted?

Comment (tasleson) [reply src/bin/stratisd.rs]: You found a bug in my change, this should be:
`let poll_timeout = if engine_fds_end_idx != 1 { -1 } else { 10000 };`

**Comment by tasleson** [comment]
@agrover @mulkieran Updated code based on feedback, thanks! 

if everything looks OK I will squash and then merge.

**Comment by agrover** [comment]
Thanks. One last change is that Device is `Copy` so it doesn't need to be passed by reference into `block_evaluate`.

File: src/bin/stratisd.rs [hunk 11]
```diff
@@ -19,6 +21,8 @@ use std::error::Error;
 use std::rc::Rc;
 use std::cell::RefCell;
 use std::process::exit;
+use std::os::unix::io::AsRawFd;
```
mulkieran: Please alphabetize your imports. In theory, we alphabetize them everywhere within sub categories of standard lib, external crate, our crates , imports within stratisd. This one clearly escaped, but no need to exacerbate the problem.

agrover:
👍

src/bin/stratisd.rs
```diff
@@ -85,20 +123,45 @@ fn run() -> StratisResult<()> {
...
-    let poll_timeout = if engine_fds_end_idx != 0 { -1 } else { 10000 };
+    let poll_timeout = if engine_fds_end_idx != 1 { -1 } else { 10000 };
 
     loop {
+        // Process any udev block events
+        if fds[0].revents != 0 {
+            loop {
+                match udev.receive_event() {
+                    Some(event) => {
+                        if let Some(added) = handle_udev_add(&event) {
```
mulkieran: I would think you could destructure the tuple right here, as:

```
if let Some((devnode, device)) = handle_udev_add(&event) {
```

but maybe not.

src/bin/stratisd.rs
```diff
@@ -85,20 +123,45 @@ fn run() -> StratisResult<()> {
...
+                            let pool_uuid = engine.borrow_mut().block_evaluate(added.0, added.1);
+                            if let Some(_pool_uuid) = pool_uuid {
```
mulkieran: If you are not interested in ```pool_uuid``` just say:

```
if pool_uuid.is_some() {
  // do stuff unrelated to actual value
}
```

Or possibly 
```
if engine.borrow_mut().block_evaluate(...).is_some() {
  // do stuff unrelated to the value of the pool uuid returned by block_evaluate, but dependent on its existance
}
```

src/bin/stratisd.rs
```diff
@@ -85,20 +123,45 @@ fn run() -> StratisResult<()> {
...
+                                #[cfg(feature="dbus_enabled")]
```
mulkieran: I am surprised that our tests pass with this as this leaves an empty code block when ```dbus_enabled``` is false. But I guess the compiler must not mind.

src/dbus_api/api.rs
```diff
@@ -26,7 +26,9 @@ use dbus::tree::Tree;
 use dbus::ConnectionItem;
 
 use engine::Engine;
+use engine::Pool;
```
mulkieran: ```use engine::{Engine, Pool};``` was what we decided to go with a while ago.

src/dbus_api/api.rs
```diff
@@ -26,7 +26,9 @@ use dbus::tree::Tree;
...
 use stratis::VERSION;
+use uuid::Uuid;
```
mulkieran: ```uuid``` is an external crate, so import should go above stratis import.

src/bin/stratisd.rs
tasleson: The compiler did mind until I renamed `pool_uuid` to `_pool_uuid` instead.

src/dbus_api/api.rs
```diff
@@ -208,22 +222,32 @@ pub fn connect(engine: Rc<RefCell<Engine>>)
...
     // engine.
     for pool in local_engine.borrow().pools() {
-        let pool_path = create_dbus_pool(&dbus_context, object_path.clone(), pool.uuid());
-        for fs_uuid in pool.filesystems().iter().map(|f| f.uuid()) {
-            create_dbus_filesystem(&dbus_context, pool_path.clone(), fs_uuid);
-        }
-        for dev_uuid in pool.blockdevs().iter().map(|bd| bd.uuid()) {
-            create_dbus_blockdev(&dbus_context, pool_path.clone(), dev_uuid);
-        }
+        register_pool_dbus(&dbus_context, pool, &object_path);
```
mulkieran: Please adjust comment above this loop appropriately. Something like:

```
This should never panic as register_pool_dbus does not borrow the engine.
```

src/bin/stratisd.rs
mulkieran: Huh. So, you have to pretend that you're not using a variable in order to make coimpilation w.out ```dbus_enabled``` happy. That might be worth a comment; it's a pretty unusual situation.

src/engine/engine.rs
```diff
@@ -2,14 +2,16 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+extern crate devicemapper;
```
mulkieran: I'm surprised this is necessary, since it is externeded in ```src/lib.rs```. It can probably be removed.

src/bin/stratisd.rs
```diff
@@ -2,6 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+extern crate devicemapper;
```
mulkieran: You might be able to get rid of this ```extern```.

tasleson:
@mulkieran External crate usage re-worked as requested.

src/bin/stratisd.rs
```diff
@@ -85,20 +123,49 @@ fn run() -> StratisResult<()> {
...
+                };
+            }
+        }
+
         // Unconditionally call evented() if engine has no eventable.
         // This looks like a bad idea, but the only engine that has
         // no eventable is the sim engine, and for that engine, evented()
         // is essentially a no-op.
-        if engine_fds_end_idx == 0 {
+        if engine_fds_end_idx == 1 {
```
mulkieran: You need to make more changes in this code because of the use of indexes to mean really important stuff.

```
for pfd in fds[...engine_fds_end_idx]
```

is wrong now. In fact, looks like there needs to more changes in loop, as carefully thought out ```expect``` string is wrong as well.

I'm wondering if it would have been a little less bad if engine had come before udev. But if you're going to stick with this order, probably a constant called ```udev_fds_end_index``` would help.

src/dbus_api/api.rs
```diff
@@ -194,36 +195,56 @@ fn get_base_tree<'a>(dbus_context: DbusContext) -> (Tree<MTFn<TData>, TData>, db
...
+    }
+    for dev_uuid in pool.blockdevs().iter().map(|bd| bd.uuid()) {
+        create_dbus_blockdev(dbus_context, pool_path.clone(), dev_uuid);
+    }
+}
+
 #[allow(type_complexity)]
-pub fn connect(engine: Rc<RefCell<Engine>>)
-               -> Result<(Connection, Tree<MTFn<TData>, TData>, DbusContext), dbus::Error> {
+pub fn connect<'a>
```
mulkieran: It is time to comment this, now that it returns a 4-tuple.

tasleson:
@agrover @mulkieran  Please take a look at the event loop, I've **tried** to make it less error prone.

src/bin/stratisd.rs
```diff
@@ -72,54 +97,108 @@ fn run() -> StratisResult<()> {
...
-                .expect("index < engine_fds_end_idx <=> index == 0 <=> eventable.is_some()")
-                .clear_event()?;
+        // Handle engine events, if the engine is eventable
+        if engine_eventable {
+            if fds[FD_INDEX_ENGINE].revents != 0 {
+                fds[FD_INDEX_ENGINE].revents = 0;
+
+                eventable
+                    .as_mut()
+                    .expect("eventable.is_some() and engine fd.revents != 0")
```
mulkieran: Just "eventable.is_some()", I think. We use the string to explain the logical reason why ```expect()``` can never fail.

src/bin/stratisd.rs
```diff
@@ -72,54 +97,108 @@ fn run() -> StratisResult<()> {
...
+    0   == Always udev fd index
+    1   == engine index if eventable
+    1/2 == Start of dbus client file descriptor(s), 1 if engine is not eventable, else 2
+    */
+    const FD_INDEX_UDEV: usize = 0;
+    const FD_INDEX_ENGINE: usize = 1;
+
     #[cfg(feature="dbus_enabled")]
-    let (dbus_conn, mut tree, dbus_context) = libstratis::dbus_api::connect(Rc::clone(&engine))?;
+    let (mut dbus_conn, mut tree, base_object_path, mut dbus_context) =
```
mulkieran: I think it makes sense to move this line down, so that it is just above ```loop```.

src/engine/strat_engine/engine.rs
```diff
@@ -65,25 +73,41 @@ impl StratEngine {
...
+                                // TODO: update state machine on failure.
+                                let _ = teardown_pools(engine.pools.empty());
+                                let err_msg = "found two pools with the same id or name";
+                                return Err(EngineError::Engine(ErrorEnum::Invalid, err_msg.into()));
+                            }
+                        }
+                        Err(e) => {
+                            // Unable to setup pool, it might become complete later?
+                            engine.holding_pool.insert(*pool_uuid, devices.clone());
+                            warn!("Error while doing a pool::setup for pool uuid {} {:?}, \
```
mulkieran: Please skip the ```warn!```. It might be useful to shove the error into the ```incomplete_pools``` (```holding_pool```) map along with the rest.

src/engine/strat_engine/engine.rs
```diff
@@ -65,25 +73,41 @@ impl StratEngine {
...
-        let pools = find_all()?;
-
-        let mut table = Table::default();
-        for (pool_uuid, devices) in &pools {
-            let evicted = table.insert(StratPool::setup(*pool_uuid, devices)?);
-            if !evicted.is_empty() {
-
-                // TODO: update state machine on failure.
-                let _ = teardown_pools(table.empty());
+        let mut engine = StratEngine {
```
mulkieran: I don't think it's necessary to make the StratEngine struct until after you've processed the pools. The reason for this is that I don't think you would handle an event until after the pool is done initializing.

src/engine/strat_engine/engine.rs
```diff
@@ -65,25 +73,41 @@ impl StratEngine {
...
+                            if !evicted.is_empty() {
+                                // TODO: update state machine on failure.
+                                let _ = teardown_pools(engine.pools.empty());
+                                let err_msg = "found two pools with the same id or name";
+                                return Err(EngineError::Engine(ErrorEnum::Invalid, err_msg.into()));
+                            }
+                        }
+                        Err(e) => {
+                            // Unable to setup pool, it might become complete later?
+                            engine.holding_pool.insert(*pool_uuid, devices.clone());
```
mulkieran: Properly speaking, you should check for the unlikely chance that ```holding_pool``` (```incomplete_pools```) already had a bunch of devices w/ that pool UUID _or_ you could state that the last such found is good enough for your purposes, since one incomplete pool is just as good as another.

src/engine/strat_engine/engine.rs
tasleson: This comment is not entirely clear to me, but I've moved the creation to where it was before, I've put it closer to the way it was before.

It's entirely possible that when we start the daemon and do a device scan that we only find one device of a multi-device pool, thus we will be placing that device into the map of incomplete pools, this has nothing to do with event handing.  However, this discussion does remind me that the order is very important and that we just overlooked a race condition when bringing up the daemon.  We must register for udev events before we walk the device tree, otherwise a device could show up just after the scan but before we register for events, thus causing us to miss a device and fail to bring up a pool.  I will reorder stratisd.rs to account for this.

src/engine/strat_engine/engine.rs
tasleson: This isn't possible as this code path is only executed when we initialize the daemon, thus incomplete_pools will never have an existing entry with the pool uuid.

src/engine/strat_engine/engine.rs
```diff
@@ -65,25 +73,41 @@ impl StratEngine {
...
+        match find_all() {
+            Ok(pools) => {
+                for (pool_uuid, devices) in &pools {
+                    match StratPool::setup(*pool_uuid, devices) {
+                        Ok(pool) => {
+                            let evicted = engine.pools.insert(pool);
+                            if !evicted.is_empty() {
+                                // TODO: update state machine on failure.
+                                let _ = teardown_pools(engine.pools.empty());
+                                let err_msg = "found two pools with the same id or name";
```
mulkieran: This error message was kind of buggy. It should say, "found two pools with the same name". It is impossible to find two pools with the same UUID, because the pools are being read from ```find_all()``` which is a HashMap from pool UUIDs to stuff.

src/bin/stratisd.rs
```diff
@@ -62,6 +87,14 @@ fn run() -> StratisResult<()> {
         .init()
         .expect("This is the first and only initialization of the logger; it must succeed");
 
+    // Setup udev events for block devices, we need to do this before we call initialize, so that
```
mulkieran: Please be more precise about the bug that is avoided here. It's obvious that either you must set up the udev context before you set up the engine _or_ that this code should be placed after the engine setup code. The comment tells me that the ordering of these blocks is _not_ random, i.e., that the person who wrote the comments believed that there was actually a reason to set up the udev context before setting up the engine. This is somewhat helpful. But it still leaves me completely at a loss as to what might go wrong if the order were reversed. In order to understand the code I want to know this. I could experiment by reversing the order and seeing what happens, but that is time consuming. I'ld rather get a definite explanation right here.

src/dbus_api/api.rs
```diff
@@ -194,36 +195,57 @@ fn get_base_tree<'a>(dbus_context: DbusContext) -> (Tree<MTFn<TData>, TData>, db
...
 
     process_deferred_actions(&c, &mut tree, &mut dbus_context.actions.borrow_mut())?;
 
-    Ok((c, tree, dbus_context))
+    Ok((c, tree, object_path, dbus_context))
+}
+
+/// Given the UUID of a pool, register all the pertinent information with dbus.
+pub fn register_pool(c: &Connection,
+                     engine: &Rc<RefCell<Engine>>,
```
mulkieran: I don't think you should clone within the method. It helps to reason about the code if you clone at the call site. That way it's clearer how many pointers are roaming around.

src/bin/stratisd.rs
```diff
@@ -72,54 +105,103 @@ fn run() -> StratisResult<()> {
...
+                        if let Some((devnode, device)) = handle_udev_add(&event) {
+                            let pool_uuid = engine.borrow_mut().block_evaluate(devnode, device);
+
+                            // We need to pretend that we aren't using the variable _pool_uuid so
+                            // that we can conditionally compile out the register_pool when dbus
+                            // is not enabled.
+                            if let Some(_pool_uuid) = pool_uuid {
+                                #[cfg(feature="dbus_enabled")]
+                                libstratis::dbus_api::register_pool(&mut dbus_conn,
+                                                                    &Rc::clone(&engine),
```
mulkieran: The comment about the signature of the function ```register_pool``` is also a comment about this call-site. I think you should just clone here and not pass as a reference.

src/engine/strat_engine/engine.rs
```diff
@@ -2,11 +2,14 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+extern crate libc;
+
 use std::collections::HashMap;
 use std::os::unix::io::{AsRawFd, RawFd};
 use std::path::Path;
+use std::path::PathBuf;
```
mulkieran: ```use std::path::{Path, PathBuf};```

src/engine/strat_engine/engine.rs
```diff
@@ -2,11 +2,14 @@
...
 
-use devicemapper::{DM, DmNameBuf};
+use devicemapper::{DM, DmNameBuf, Device};
```
mulkieran: Alphabetize, ```Device``` should come before ```DmNameBuf```.

tasleson:
@mulkieran Re-based, squashed and changed requested completed :-)

src/engine/sim_engine/engine.rs
```diff
@@ -2,12 +2,17 @@
...
 
+extern crate libc;
+
 use std::cell::RefCell;
 use std::collections::HashSet;
 use std::collections::hash_map::RandomState;
 use std::iter::FromIterator;
 use std::path::Path;
 use std::rc::Rc;
+use std::path::PathBuf;
```
mulkieran: Move this import up to where ```Path``` is already imported.

tasleson:
Re-based, merge conflicts resolved.

src/engine/strat_engine/engine.rs
```diff
@@ -124,6 +140,49 @@ impl Engine for StratEngine {
...
+                    if rc.is_some() || duplicate {
+                        self.incomplete_pools.remove(&pool_uuid);
+                    }
+
+                    return rc;
+                }
+
+                None
+            }
+            _ => None,
```
mulkieran: I think ```block_evaluate()``` should return an error if ```is_stratis_device``` returns an error. It didn't succesfully evaluate the device for being a stratis device or not. You can choose to ignore the error at a higher level with the ```_```. So structure would just be:

```
match is stratis_device(...)? {
    Some(uuid) => {
      // all the interesting stuff
    },
    None => None
}
```

And return type would be ```Err<Option<PoolUuid>>```.

src/bin/stratisd.rs
```diff
@@ -79,54 +116,103 @@ fn run() -> StratisResult<()> {
...
-        // is essentially a no-op.
-        if engine_fds_end_idx == 0 {
-            engine.borrow_mut().evented()?;
+        // Process any udev block events
+        if fds[FD_INDEX_UDEV].revents != 0 {
+            loop {
+                match udev.receive_event() {
+                    Some(event) => {
+                        if let Some((devnode, device)) = handle_udev_add(&event) {
+                            let pool_uuid = engine.borrow_mut().block_evaluate(devnode, device);
```
mulkieran: To ignore the possible error:

```
engine.borrow_mut().block_evaluate(devnode, device).unwrap_or(None);
```

and maybe an explanation of why it is ok to ignore it. That explanation might be that, if it is not our device, obviously there is nothing to do. If it our device yet we missed it, continuing with execution is still better than failing, as the worst that can happen is that we'ld be stuck forever with an incomplete pool.

src/engine/strat_engine/setup.rs
```diff
@@ -26,6 +26,64 @@ use super::range_alloc::RangeAllocator;
 use super::serde_structs::PoolSave;
 
 
+/// Determine if devnode is a stratis device, if it is we will add  pool uuid and device
+/// information to pool_map and return pool uuid.
+pub fn is_stratis_device(pool_map: &mut HashMap<PoolUuid, HashMap<Device, PathBuf>>,
```
mulkieran: I would be much happier if this did not take ```pool_map``` as an argument. That does mean that you have to insert into a map in two places, but that is better than having a significant side-effect buried in the method.

src/engine/strat_engine/engine.rs
```diff
@@ -124,6 +140,49 @@ impl Engine for StratEngine {
...
+    fn block_evaluate(&mut self, dev_node: PathBuf, device: Device) -> Option<PoolUuid> {
+        let mut duplicate = false;
+
+        match is_stratis_device(&mut self.incomplete_pools,
+                                dev_node,
+                                libc::dev_t::from(device)) {
+            Ok(process_it) => {
+                let mut rc = None;
+                if let Some(pool_uuid) = process_it {
+                    if let Some(devices) = self.incomplete_pools.get(&pool_uuid) {
```
mulkieran: I think that ```self.incomplete_pools.get(&pool_uuid)``` will always return ```Some``` value. The reason for that is that ```is_stratis_device``` always inserts into the hashmap that is its first argument whenever it returns a uuid. This appears to me to be a definite bug.

This goes back to my other point, which is that ```is_stratis_device``` would be a lot handier if did not take the hashmap argument but left it to the user to decide what values to be inserting where.

src/engine/strat_engine/engine.rs
```diff
@@ -124,6 +140,51 @@ impl Engine for StratEngine {
         Ok(uuid)
     }
 
+    /// Evaluate a device node & devicemapper::Device to see if it's a valid
+    /// stratis device.  If all the devices are present in the pool and the pool isn't already
+    /// up and running, it will get setup and the pool uuid will be returned.
```
mulkieran: Please add

```
// Returns an error if the status of the block device can not be evaluated.
```

src/engine/strat_engine/engine.rs
```diff
@@ -124,6 +140,51 @@ impl Engine for StratEngine {
...
+                    .get(&pool_uuid)
+                    .expect("We previously inserted pool uuid!");
+
+                if let Ok(pool_setup) = StratPool::setup(pool_uuid, devices) {
+                    self.pools.insert(pool_setup);
+                    rc = Some(pool_uuid);
+                }
+            } else {
+                // We have this pool already so we shall ignore.
+                warn!("Pool {} is already known, ignoring!", pool_uuid);
```
mulkieran: My new official policy is that I will not object to debug output of this specific sort, i.e., a notification of a problem that cannot be better encapsulated in the return value, as we would want to log this stuff in some way. The log message serves as a marker of our future, better, intentions. You get to be the first beneficiary of this policy :-).

src/engine/strat_engine/engine.rs
```diff
@@ -124,6 +140,51 @@ impl Engine for StratEngine {
...
+    /// up and running, it will get setup and the pool uuid will be returned.
+    fn block_evaluate(&mut self,
+                      dev_node: PathBuf,
+                      device: Device)
+                      -> EngineResult<Option<PoolUuid>> {
+        let mut rc = None;
+        let mut duplicate = false;
+
+        if let Some(pool_uuid) = is_stratis_device(&dev_node)? {
+
```
mulkieran: Please remove use of ```rc``` and ```duplicate``` variables by restructuring like:

```
if let Some(pool_uuid) = ... {
  if self.pools.contains(uuid) {
    // warn
    // remove pool_uuid entry from incomplete pools
    // Ok(None)
  } else {
    // insert current device into incomplete pools
    // get the devices corresponding to pool_uuid from incomplete pools
    // let rc = if setup of new pool succeeds
    //        remove from incomplete pools
    //        insert into pools
    //        Some(pool_uuid)
    // else
    //        None
    // Ok(rc)
  }
  // no return value here, cause it's taken care of by the if block.
}
```

You have to call remove on the incomplete pools on two different paths, but you get to eliminate two varaibles with semantic meaning.
I think it is less confusing this way.

mulkieran: I realized 2 further things about this code while riding the bus.

1. You shouldn't remove the ```pool_uuid``` entry from ```incomplete_pools``` if it is in ```pools```. You can show that the engine always maintains the invariant that the keys of ```pools``` and the keys of ```incomplete_pools``` are disjoint sets. If the key is in one, then it can't be in the other.

If this ever turns out to be false, it would actually be nice to look at some data indicating where you went wrong. So, never any reason to remove from ```incomplete_pools``` in this situation.

2. There is a possibility that when you insert into ```pools``` a pool that is already there will be ejected. This is because ```pools``` actually has two unique ids, the pool name and the pool uuid. Here you ensure that there is no conflict with the uuids, but it is easy to come up with a sequence of events that result in a name conflict. I think that in that case the newly created pool should be held in a special place reserved for problematic pools, i.e, the existing pool should not be ejected.

src/engine/strat_engine/setup.rs
```diff
@@ -26,6 +26,54 @@ use super::range_alloc::RangeAllocator;
 use super::serde_structs::PoolSave;
 
 
+/// Determine if devnode is a stratis device, if it is we will add  pool uuid and device
+/// information to pool_map and return pool uuid.
+pub fn is_stratis_device(devnode: &PathBuf) -> EngineResult<Option<PoolUuid>> {
+
+    let f = OpenOptions::new().read(true).open(&devnode);
```
mulkieran: At this point this code really only has two branches, f err and f not, so might as well just turn the whole method into one big match expression. That way, you get to avoid the ```expect```s.

src/engine/strat_engine/setup.rs
```diff
@@ -49,48 +97,8 @@ pub fn find_all() -> EngineResult<HashMap<PoolUuid, HashMap<Device, PathBuf>>> {
...
-        }
-
-        let mut f = f.expect("unreachable if f is err");
-        if let DevOwnership::Ours(pool_uuid, _) = StaticHeader::determine_ownership(&mut f)? {
-            // No value should ever be ejected, because duplicate device nodes
-            // are filtered out above. Therefore, the return value of insert()
-            // might as well be ignored.
-            let _ = pool_map
+        if let Some(pool_uuid) = is_stratis_device(&devnode)? {
+            pool_map
```
mulkieran: For clarity, this block and the block above could become:

```
match devnode_to_devno(&devnode)? {
  None => continue,
  Some(devno) => {
    if devno_set.insert(devno) {
      is_stratis_device(&devnode).and_then(|pool_uuid| pool_map.entry(pool_uuid).or_insert_with(...).insert(...))
    } else {
      continue;
    }
  }
}
```

That gets rid of any need for the comment that was in the previous version, because it is now very clear what is happenng and the code that separated these two has now been moved into ```is_stratis_device```.

src/engine/strat_engine/engine.rs
tasleson: Will calling `setup` with a duplicate pool name, but different uuid cause `setup` to fail?  If not then we should restructure the code, because it's a bad situation where we can succeed at calling `setup` only to fail because we cannot store the pool in the table data structure.  IMHO I'm thinking that we should have one entry point which brings up a pool and inserts it into the correct data structure(s).

mulkieran: I'm generally against having methods that work by side-effects if that can be avoided. Also, minimizing the number of parameters is generally a good idea.

You say it's a bad situation, where a pool can be setup, but turn out to be in a name-duplicate. Why is that a problem, and how would the restructuring that you are suggesting solve or alleviate the problem? Note that this situation is also encountered in ```StratPool::initialize```, where the action taken on a name conflict is just to return an error. I suspect that, given the new more forgiving context of handling incomplete pools that this PR introduces, that may be the wrong action to take in that context.

mulkieran: I guess, more fundamentally, and regardless of the code, what ought to be done if:
1. You find some blockdevs which are a bad fit for the current world, because they contain the same name as some other blockdevs that are already in an established pool, but a different UUID?
2. What should be done if these odd blockdevs form a complete pool, or should the code stop before it even gets to fidning out if that is true or not?

tasleson: `StratPool::setup` has a side effect that it makes the pool active and available on the system.  I dislike that at the moment we have to call setup, go through that entire process to get the pool object so we can ask the question, what is your name (yes we could add a method for that)?  Then we realize that we have a pool by that name which violates a rule in the Engine, but we already made it public.

I've made a change to my PR so in this case we will `teardown` the newly found pool and leave it in the incomplete pools while logging a error until we figure out what we ideally want to do.  The user could recover by renaming the existing live pool and then generate a synthetic hot plug event on one of the other devices in the other inactive pool.  We would then dynamically bring the pool back up.

mulkieran: We haven't made the pool public as a consequence of calling ```setup()```. And the purpose of calling ```setup()``` is to set up the pool, we can't really get around that.

The problem is that the devnode itself doesn't contain the name of the pool it belongs to. That's in the pool's metadata which can only be discovered by reading the blockdevs belonging to the pool and finding the ones that contain the most recent metadata. That's why getting that information is in ```StratPool::setup()```. But, if it were necessary or desirable, it could probably be lifted out of ```StratPool::setup()``` since the whole activity is encapsulated in the ```get_metadata``` method.

tasleson: > We haven't made the pool public as a consequence of calling `setup()`

Are you sure?  It sure looks like after setup dm tables have been loaded and the associated block devices are now present in `/dev`

src/engine/sim_engine/engine.rs
mulkieran: This import should be combined w/ import of Path above as ```use std::path::{Path, PathBuf};```

src/engine/strat_engine/engine.rs
```diff
@@ -16,7 +18,7 @@ use super::super::types::{DevUuid, PoolUuid, Redundancy, RenameAction};
 use super::cleanup::teardown_pools;
 use super::devlinks;
 use super::pool::StratPool;
-use super::setup::find_all;
+use super::setup::{is_stratis_device, find_all};
```
mulkieran: alphabetize, "f" comes before "i".

src/engine/strat_engine/engine.rs
```diff
@@ -124,6 +140,58 @@ impl Engine for StratEngine {
...
+                    .entry(pool_uuid)
+                    .or_insert_with(HashMap::new)
+                    .insert(device, dev_node);
+
+                let devices = self.incomplete_pools
+                    .get(&pool_uuid)
+                    .expect("We previously inserted pool uuid!")
+                    .clone();
+
+                let rc = if let Ok(pool) = StratPool::setup(pool_uuid, &devices) {
```
mulkieran: I think we're going to discuss the problem of duplicate names in team meeting, so there's no reason to try to implement my suggestion before that.

However, I grant your previous point that setting up a device that you can't use just to tear it down again is not so great. My suggestion would be to pull the call to ```get_metadata``` out of ```StratPool::setup()``` and invoke it here. That way you can check whether there is a name conflict _before_ the pool is setup.

As far as I'm concerned, leaving the list of devs stuck in ```incomplete_pools``` is fine.

But I think those error messages are kind of pointless. It would be much better if the engine could do a health check on itself, and report a coordinated report on its incomplete pools, explaining the whole situation, after a reasonable time interval. It might be worth opening an issue at this point.

And finally, this behavior is inconsistent with what is done in ```StratEngine::initialize()``` where discovery of name conflict results in immediate error and termination of program. We should explain why that is in the comments, as there may be a perfectly good reason to behave differently, or else we should make the behavior consistent.

But, as I said, might as well wait until meeting before implementing any of this.

src/engine/strat_engine/engine.rs
mulkieran: I think this is just a being divided by a common language kind of thing. Of course, we have set up many dm devices, and I agree that that is not what we would wish to do.

src/engine/strat_engine/engine.rs
```diff
@@ -124,6 +140,58 @@ impl Engine for StratEngine {
...
+    /// up and running, it will get setup and the pool uuid will be returned.
+    ///
+    /// Returns an error if the status of the block device can not be evaluated.
+    fn block_evaluate(&mut self,
+                      dev_node: PathBuf,
+                      device: Device)
+                      -> EngineResult<Option<PoolUuid>> {
+        if let Some(pool_uuid) = is_stratis_device(&dev_node)? {
+            if self.pools.contains_uuid(pool_uuid) {
+                warn!("Pool {} is already known, ignoring!", pool_uuid);
```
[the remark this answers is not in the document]
tasleson: Do you believe the existing code (before this change) handles all possible error cases during the daemon starting and restarting which are outlined here?

src/engine/strat_engine/engine.rs
```diff
@@ -124,6 +140,58 @@ impl Engine for StratEngine {
...
+            } else {
+                self.incomplete_pools
+                    .entry(pool_uuid)
+                    .or_insert_with(HashMap::new)
+                    .insert(device, dev_node);
+
+                let devices = self.incomplete_pools
+                    .get(&pool_uuid)
+                    .expect("We previously inserted pool uuid!")
+                    .clone();
```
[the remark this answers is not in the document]
tasleson:
```
error[E0502]: cannot borrow `self.incomplete_pools` as mutable because it is also borrowed as immutable
   --> src/engine/strat_engine/engine.rs:171:25
    |
162 |                 let devices = self.incomplete_pools
    |                               --------------------- immutable borrow occurs here
...
171 |                         self.incomplete_pools.remove(&pool_uuid);
    |                         ^^^^^^^^^^^^^^^^^^^^^ mutable borrow occurs here
...
188 |             }
    |             - immutable borrow ends here
```

src/engine/strat_engine/engine.rs
tasleson: In this case the pool really isn't incomplete, it's all there we just have a name conflict.  If we had some reporting mechanism in place or engine check we could certainly not log a message, but we have none of that so logging a message is better than silently doing nothing IMHO.

IMHO exiting the daemon is a really bad thing to do and should be avoided.  Once we do that the customer has no way to interact with stratis, the CLI won't work, we cannot send dbus signals or generate any kind of alerts and the health of whatever might be up and running cannot be monitored.  The daemon really needs to be running.

src/engine/strat_engine/engine.rs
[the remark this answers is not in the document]
mulkieran: Before what change?

src/engine/strat_engine/engine.rs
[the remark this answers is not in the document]
mulkieran: Oh, I see. You might be better off removing the devices here, and reinserting them below if necessary. Beats a deep copy.

src/engine/strat_engine/engine.rs
mulkieran: > In this case the pool really isn't incomplete, it's all there we just have a name conflict. If we had some reporting mechanism in place or engine check we could certainly not log a message, but we have none of that so logging a message is better than silently doing nothing IMHO.

That is correct. I'm just maintaining that a reporting mechanism would be much better. Here you log if you have a name conflict, but of course you don't constantly log if you have an incomplete pool, because you expect that to change. But if the incomplete pool is never promoted to a complete pool, then that should be reported. But at this point we can't/don't. I'll file an issue.

src/engine/strat_engine/engine.rs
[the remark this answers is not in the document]
tasleson: My previous implementation didn't require a clone...

src/engine/strat_engine/engine.rs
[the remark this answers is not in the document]
tasleson: Clarification: 

Do you believe the existing code (what is currently in main branch) handles all possible error cases during the daemon starting and restarting which are outlined here?

src/engine/strat_engine/engine.rs
mulkieran: > IMHO exiting the daemon is a really bad thing to do and should be avoided. Once we do that the customer has no way to interact with stratis, the CLI won't work, we cannot send dbus signals or generate any kind of alerts and the health of whatever might be up and running cannot be monitored. The daemon really needs to be running.

So this is a reason to keep on going when, in the event loop, there is found to be a pool w/ a name conflict. That seems right to me. The question is, what about when the same situation arises in ```StratEngine::initialize```. Currently, the code exits. I'm saying that this _looks_ like an oversight, as I would expect the name conflict to be handled the same way in both places. If it is an oversight, than that code should be changed to be consistent. If it's deliberate, then a comment should be added to explain why the difference the way the same problem is handled between the two locations.

src/engine/strat_engine/engine.rs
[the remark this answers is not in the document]
mulkieran: The existing code lives in a very limited, simple world. The only devs it knows about are those it finds in /dev. All devs discovered are divided into disjoint sets according to their pool UUID. It exits on the only two possible ways to fail, failure to setup a pool, and a name conflict.

This PR creates a very much more complicated world, and as a consequence a whole lot more ways for stuff to go wrong. The problem here, of finding, after a pool has been created, another dev with the same pool UUID, can't even exist in the master branch, due to the division into disjoint sets of the entire available set of devs.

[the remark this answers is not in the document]
tasleson: I believe it can exist in the current branch, you only need to restart the daemon while a configuration is active and the hardware has changed.

tasleson:
@mulkieran I've updated the PR to add comments, remove clone etc.

tasleson:
Merge conflicts corrected, re-based and squished.

agrover:
@mulkieran 👍 ??

src/engine/strat_engine/engine.rs
```diff
@@ -125,6 +137,73 @@ impl Engine for StratEngine {
...
+                    // so we will ensure unique before we insert as we don't want to change the
+                    // existing state if we have a conflict.
+                    //
+                    // We will also _not_ exit the daemon as we do in initialize as we were
+                    // previously up and running for some duration of time.
+                    if !self.pools.contains_name(pool.name()) {
+                        self.pools.insert(pool);
+                        Some(pool_uuid)
+                    } else {
+                        let mut dev_paths = String::new();
```
mulkieran: Simplify to something like this, please.

```
let dev_paths = devices.values().map(|p| p.to_str().expect(...)).join(", ");
```

src/engine/strat_engine/engine.rs
```diff
@@ -125,6 +137,73 @@ impl Engine for StratEngine {
...
+    ///
+    /// Returns an error if the status of the block device can not be evaluated.
+    fn block_evaluate(&mut self,
+                      dev_node: PathBuf,
+                      device: Device)
+                      -> EngineResult<Option<PoolUuid>> {
+        let dm = DM::new()?;
+        if let Some(pool_uuid) = is_stratis_device(&dev_node)? {
+            if self.pools.contains_uuid(pool_uuid) {
+                // TODO: Handle the case where we have found a device for an already active pool.
```
mulkieran: Please open an issue for this TODO.

src/engine/strat_engine/engine.rs
```diff
@@ -171,6 +250,10 @@ impl Engine for StratEngine {
         get_mut_pool!(self; uuid)
     }
 
+    fn configure_simulator(&mut self, _denominator: u32) -> EngineResult<()> {
```
mulkieran: I think this move of ```configure_simulator``` belongs in a completely different PR, or is a rebasing mistake.

src/engine/strat_engine/engine.rs
```diff
@@ -69,20 +76,29 @@ impl StratEngine {
         let pools = find_all()?;
 
         let mut table = Table::default();
-        for (pool_uuid, devices) in &pools {
-            let evicted = table.insert(StratPool::setup(&dm, *pool_uuid, devices)?);
-            if !evicted.is_empty() {
-
-                // TODO: update state machine on failure.
-                let _ = teardown_pools(table.empty());
+        let mut incomplete_pools = HashMap::new();
```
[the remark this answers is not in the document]
tasleson: Done.

src/engine/strat_engine/engine.rs
tasleson: Done

src/engine/strat_engine/engine.rs
tasleson: Done

src/engine/strat_engine/engine.rs
tasleson: Rebase error, addressed with another rebase and merge conflict

mulkieran:
I expect that the idea is to squash the last few commits, so I added the do not merge until that's done, so nobody gets trigger happy.

tasleson:
PR squished

PR #692, opened 2017-12-19
Improve fs creation

agrover:
When creating more than one filesystem, create the first one, and then snapshot it to create the rest. This should be faster and more space-efficient.

### DIFF
--- a/src/dbus_api/pool.rs
+++ b/src/dbus_api/pool.rs
@@ -21,8 +21,6 @@ use dbus::tree::PropInfo;
 
 use uuid::Uuid;
 
-use devicemapper::Sectors;
-
 use engine::{Pool, RenameAction};
 
 use super::blockdev::create_dbus_blockdev;
@@ -52,10 +50,7 @@ fn create_filesystems(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
     let mut engine = dbus_context.engine.borrow_mut();
     let pool = get_mut_pool!(engine; pool_uuid; default_return; return_message);
 
-    let result =
-        pool.create_filesystems(&filesystems
-                                     .map(|x| (x, None))
-                                     .collect::<Vec<(&str, Option<Sectors>)>>());
+    let result = pool.create_filesystems(&filesystems.collect::<Vec<_>>());
 
     let msg = match result {
         Ok(ref infos) => {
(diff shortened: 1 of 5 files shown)


src/dbus_api/pool.rs
```diff
@@ -52,10 +50,7 @@ fn create_filesystems(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
     let mut engine = dbus_context.engine.borrow_mut();
     let pool = get_mut_pool!(engine; pool_uuid; default_return; return_message);
 
-    let result =
-        pool.create_filesystems(&filesystems
-                                     .map(|x| (x, None))
-                                     .collect::<Vec<(&str, Option<Sectors>)>>());
+    let result = pool.create_filesystems(&filesystems.collect::<Vec<_>>());
```
mulkieran: I guess it's too difficult to just revert the commits that introduced these changes?

src/engine/strat_engine/pool.rs
```diff
@@ -151,9 +151,16 @@ impl Pool for StratPool {
         // TODO: Roll back on filesystem initialization failure.
         let dm = DM::new()?;
         let mut result = Vec::new();
-        for name in names {
-            let fs_uuid = self.thin_pool.create_filesystem(name, &dm, None)?;
-            result.push((*name, fs_uuid));
+        let mut names_iter = names.into_iter();
```
mulkieran: Probably you want ```split_first``` here if you stick to this algorithm.

src/engine/strat_engine/pool.rs
```diff
@@ -151,9 +151,16 @@ impl Pool for StratPool {
...
+        if let Some(first_name) = names_iter.next() {
```
mulkieran: This whole approach seems a little circumscribed to me. Why can't each stratisd pool just keep around a secret, empty filesystem, which it snapshots whenever any request for a new filesystem comes in? Here this is an optimization which is not terribly useful if the user just requests one filesystem at a time.

How much faster is snapshotting than making a whole new filesystem, anyway?

src/engine/strat_engine/pool.rs
agrover: `names` is a HashSet not a slice so alas it's not available.

src/engine/strat_engine/pool.rs
agrover: Yes absolutely. This is a stepping stone to getting to that implementation, which has additional complexities and larger code changes that it would need.

> How much faster is snapshotting than making a whole new filesystem, anyway?

It's not! Very strange! See #694.

src/dbus_api/pool.rs
agrover: That was months ago. I'd be amazed if this is a straight reversion of that commit. I'll check, though!

agrover:
closing due to inactivity and/or not worth doing.

Issue #486, opened 2017-07-25
Might be possible to make functions used in setup(), i.e., ```find_all()```, ```get_metadata()```, ```get_blockdevs()``` more efficient as a class


mulkieran:
~~Probably ```find_all()``` should go to engine and the other two to pool. There are probably ways that they could be simplified.~~

~~This is not a task that should be attempted in the very near future.~~

As a result of cache tier work, these methods have been distributed and altered in a variety of ways. But they have also been made less public. This makes it more likely that they can be called only in contexts where more restrictive preconditions can be met. So this is a goal that is still worth working for.

agrover:
This implicitly includes the excessive opens/reads mentioned in #661.

mulkieran:
In PR #837 :

```find_all``` is used exclusively in ```StratEngine::initialize```.
```get_blockdevs``` is use exclusively in ```Backstore::setup```.
```get_metadata``` is used exclusively in ```setup::setup_pool```.
```is_stratis_device``` is used in ```setup::find_all``` and ```StratEngine::block_allocate```
```setup_pool``` is used in ```StratEngine``` in ```initialize``` and ```block_allocate```.

mulkieran:
blocked by libblkid work.

mulkieran:
These methods were completely reworked to support encryption, and are about as efficient as can be managed at this time. For example, previously the BDA was read multiple times, now it is always just read and stored once on any attempt to set up a pool. Closing.

PR #696, opened 2017-12-20
Add optional devicemapper cache to pool physical layer

mulkieran:
Related: #486.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -161,13 +161,11 @@ dependencies = [
 [[package]]
 name = "devicemapper"
 version = "0.15.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
+source = "git+https://github.com/mulkieran/devicemapper-rs?branch=master-cache#890af8ea2bb1d724bc774c0608cff439ebf14335"
 dependencies = [
  "bitflags 1.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "env_logger 0.4.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "error-chain 0.11.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.31 (registry+https://github.com/rust-lang/crates.io-index)",
- "log 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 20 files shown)


mulkieran:
rebased.

mulkieran:
rebased.

mulkieran:
@agrover review? Especially check the comments in store.rs, as they represent a development plan of sorts.

agrover:
store.rs looks like a good start. In future changes, I'm expecting to see a great deal of commonality between managing a group of data-tier blockdevs and a group of cache-tier blockdevs. For example, will Cache::new()'s cache param take a `Vec<StratBlockDev>` or similar, and will Cache have a second BlockDevMgr member for tracking them? Just curious if I'm envisioning where this is going correctly.

Next: Based on the assumption that for all the data tier blockdevs to be treated as the cachedev origin device, their usable space must be concatenated by way of a LinearDev, I wonder if it ends up making other things easier if we did this even in the no-cache-devices case.

Finally, "Store" is a pretty generic name. Is this more aptly called a "CachedBlockDevMgr" or something else a little more descriptive?

mulkieran:
Regarding name of Store...it is not a good name. But "CachedBlockDevMgr" isn't so great either, because the cache is optional. Are there any other suggestions, @tasleson, @trgill?

mulkieran:
> Next: Based on the assumption that for all the data tier blockdevs to be treated as the cachedev origin device, their usable space must be concatenated by way of a LinearDev, I wonder if it ends up making other things easier if we did this even in the no-cache-devices case.

Are you saying it might make sense to merge them all into a LinearDev and allocate from that?

agrover:
> What that might mean is that BlockDevMgr::alloc_space() is simply eliminated entirely.

Thinking about it some more, not quite, because it still needs to track the actual blockdevs the allocations are on for the on-disk metadata. The common code would be a little more complex overall vs status quo: (creating concatenated linear (+complex), allocating from linear (-complex), tracking linear-to-blockdevmgr ranges (+complex)) but it would make the non-cache config look like an edge condition of a with-cache config, rather than an entirely separate code path (--complex).

Anyway, my recommendation would be to get with-cache working, feeling free to break non-cache, and then supporting non-cache should fall out of that (different physical top layers, like you said.)

mulkieran:
I think we can abandon this in favor of #722.

PR #705, opened 2018-01-03
Master usage test

mulkieran:


mulkieran:
We failed on the assertion that the empty filesystem was less than 10 MiB. AFAICT, the only different between how we do this on devicemapper and how we do this on stratis is that we use the '-m uuid=', use a significantly larger data block size, and also update the mdv.

mulkieran:
If we don't increase the number of data blocks when we lower the data block size then we fail to make a filesystem at all, because we run out of space.

mulkieran:
Obsolete.

PR #706, opened 2018-01-05
Update dependencies...

ignatenkobrain:


mulkieran:
@ignatenkobrain fmt test is working again. It's failing 'cause it doesn't like your formatting.

src/engine/strat_engine/blockdevmgr.rs
```diff
@@ -203,7 +203,8 @@ impl BlockDevMgr {
 
         // TODO: consider making selection not entirely random, i.e, ensuring
         // distribution of metadata over different paths.
-        let saved = sample(&mut thread_rng(), candidates, MAX_NUM_TO_WRITE)
+        let saved = seq::sample_iter(&mut thread_rng(), candidates, MAX_NUM_TO_WRITE)
```
mulkieran: Semantics of ```sample_iter()``` differs from semantics of ```sample()```. Need something like:

```
let candidates = ...
  .values_mut()
  .filter(...).collect::<Vec<_>>();

let saved = seq::sample_slice(..., candidates, min(MAX_NUM_TO_WRITE, candidates.len()))
  .expect("requested sample size <= candidates.len()")
  ...
```

ignatenkobrain: does it?

```
pub fn sample<T, I, R>(rng: &mut R, iterable: I, amount: usize) -> Vec<T>
pub fn sample_iter<T, I, R>(rng: &mut R, iterable: I, amount: usize) -> Result<Vec<T>, Vec<T>> 
```

ignatenkobrain: Moreover, look at https://github.com/rust-lang-nursery/rand/blob/0.4.1/src/lib.rs#L955-L979

```rust
#[cfg(feature="std")]
#[inline(always)]
#[deprecated(since="0.4.0", note="renamed to seq::sample_iter")]
pub fn sample<T, I, R>(rng: &mut R, iterable: I, amount: usize) -> Vec<T>
    where I: IntoIterator<Item=T>,
          R: Rng,
{
    // the legacy sample didn't care whether amount was met
    seq::sample_iter(rng, iterable, amount)
        .unwrap_or_else(|e| e)
}
```

ignatenkobrain:
Btw, I don't know why it doesn't like formatting because IT IS ABSOLUTELY AS IT WANTS...

src/engine/strat_engine/blockdevmgr.rs
mulkieran: Semantics means behavior. You can tell by looking at the implementation of ```sample()``` that you pasted in above that the semantics differs. That ```unwrap_or_else()``` is necessary precisely because the semantics differs. But it's also a helpful hint. You could do the same in your code. Instead of my original suggestion:

```
let saved = seq::sample_slice(..., ..., ...)
  .unwrap_or_else(|e| e)
...
```

will do the trick.

ignatenkobrain: Fixed

agrover:
@ignatenkobrain perhaps rustfmt version differences are to blame? Are you using rustfmt 0.8.3? We pinned to that because of trouble with later versions.

(Of course, we're looking to stop pinning to 0.8.3, but as of right now we still are...)

mulkieran:
Looks like something wierd with Travis. Everything passes now. @agrover, ok to merge?

PR #711, opened 2018-01-08
add dbus service

ignatenkobrain:
Without it, stratisd is not activatable.

Signed-off-by: Igor Gnatenko <ignatenko@redhat.com>

### DIFF
--- a/org.storage.stratis1.service
+++ b/org.storage.stratis1.service
@@ -1,0 +1,4 @@
+[D-BUS Service]
+Name=org.storage.stratis1
+Exec=/usr/libexec/stratisd
+User=root


mulkieran:
@agrover, @tasleson review?

tasleson:
This looks good to me, only comment is on the binary path being hard-coded.  If we were using autotools I would know more about how to address this.

ignatenkobrain:
@tasleson why anyone on this planet would use autotools? :P 

But this file is mostly for reference & should work in 99% cases, so...

agrover:
I think we may need to do some followup work in this area but this is clearly needed now, so let's merge it.

PR #715, opened 2018-01-10
Extend mdv if needed (as part of save_fs, or elsewhere?)

agrover:
Doesn't compile because we need a mutable reference to the blockdevmgr to
allocate more space. Doable, but first let's agree on where this code
should go. Also, we probably don't need this right away, I just wanted to rebase.


agrover:
see #363

agrover:
Premature. Closing for now.

PR #723, opened 2018-01-23
Make dbus support a "feature" via "dbus_enabled"

tasleson:
We need the ability to run stratis daemon in initramfs.  In this
environment we don't have dbus support and need to have a binary
that doesn't require libdbus-1.so.* .

### DIFF
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -7,6 +7,7 @@ extern crate libstratis;
 extern crate log;
 extern crate env_logger;
 extern crate clap;
+#[cfg(feature="dbus_enabled")]
 extern crate dbus;
 extern crate term;
 extern crate libc;
@@ -24,6 +25,8 @@ use std::process::exit;
 use clap::{App, Arg};
 use log::LogLevelFilter;
 use env_logger::LogBuilder;
+
+#[cfg(feature="dbus_enabled")]
 use dbus::WatchEvent;
 
 use libstratis::engine::{Engine, SimEngine, StratEngine};
@@ -83,6 +86,7 @@ fn run() -> StratisResult<()> {
         }
     };
 
+    #[cfg(feature="dbus_enabled")]
     let (dbus_conn, mut tree, dbus_context) = libstratis::dbus_api::connect(Rc::clone(&engine))?;
 
     let mut fds = Vec::new();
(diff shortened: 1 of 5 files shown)


tasleson:
```
$ cargo build > /dev/null 2>&1 && ldd target/debug/stratisd
	linux-vdso.so.1 (0x00007ffc3b5e4000)
	libudev.so.1 => /lib64/libudev.so.1 (0x00007f1c7f036000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007f1c7e137000)
	librt.so.1 => /lib64/librt.so.1 (0x00007f1c7df2f000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f1c7dd10000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007f1c7daf9000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f1c7d724000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f1c7ee51000)
	libm.so.6 => /lib64/libm.so.6 (0x00007f1c7d40e000)
	libresolv.so.2 => /lib64/libresolv.so.2 (0x00007f1c7d1f4000)
	libselinux.so.1 => /lib64/libselinux.so.1 (0x00007f1c7cfcc000)
	libcap.so.2 => /lib64/libcap.so.2 (0x00007f1c7cdc7000)
	libpcre.so.1 => /lib64/libpcre.so.1 (0x00007f1c7cb54000)
```
```
$ cargo build --features dbus_enabled > /dev/null 2>&1 && ldd target/debug/stratisd
	linux-vdso.so.1 (0x00007ffcb0ba7000)
	libudev.so.1 => /lib64/libudev.so.1 (0x00007fd56a7be000)
	libdbus-1.so.3 => /lib64/libdbus-1.so.3 (0x00007fd5696de000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007fd5694da000)
	librt.so.1 => /lib64/librt.so.1 (0x00007fd5692d2000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007fd5690b3000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007fd568e9c000)
	libc.so.6 => /lib64/libc.so.6 (0x00007fd568ac7000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fd56a5d9000)
	libm.so.6 => /lib64/libm.so.6 (0x00007fd5687b1000)
	libresolv.so.2 => /lib64/libresolv.so.2 (0x00007fd568597000)
	libselinux.so.1 => /lib64/libselinux.so.1 (0x00007fd56836f000)
	libcap.so.2 => /lib64/libcap.so.2 (0x00007fd56816a000)
	libsystemd.so.0 => /lib64/libsystemd.so.0 (0x00007fd56a72f000)
	libpcre.so.1 => /lib64/libpcre.so.1 (0x00007fd567ef7000)
	liblzma.so.5 => /lib64/liblzma.so.5 (0x00007fd567cd1000)
	liblz4.so.1 => /lib64/liblz4.so.1 (0x00007fd567abc000)
	libgcrypt.so.20 => /lib64/libgcrypt.so.20 (0x00007fd5677ae000)
	libgpg-error.so.0 => /lib64/libgpg-error.so.0 (0x00007fd56759a000)
```

tasleson:
To build one binary that does both and leverage `dlopen` will be much more involved, but anything is possible given enough time and resources.

src/bin/stratisd.rs
```diff
@@ -143,6 +148,8 @@ fn run() -> StratisResult<()> {
         // Refresh list of dbus fds to poll for every time. This can change as
         // D-Bus clients come and go.
         fds.truncate(engine_fds_end_idx);
+
```
mulkieran: I guess that currently ```fds.truncate()``` becomes a no-op if dbus not enabled. But the comment just above it is now obviously out of place.

tasleson:
@mulkieran Re-worked, I think this is better.

agrover:
Very cool! Good progress. But this isn't exactly what we need. Yes we need a binary that doesn't link against D-Bus, but we want the binary ("stratisstart"? "stratis-init"?) to be built in parallel with the regular stratisd binary, not instead-of. We also want it to basically just call StratEngine::initialize() and then exit.

At least that's my current assumption for how all this would need to work 😃

agrover:
How does building without D-Bus support not cause stratisd-client-dbus tests to fail???

mulkieran:
@agrover @tasleson The explanation for ```stratisd_dbus_client``` test passing is that it uses the build target to build stratisd, and @tasleson's changes include adding the ```dbus_enabled``` flag to build target.

PR #726, opened 2018-01-25
Remove use of term crate

agrover:
Since 1.19 we can use eprintln!() to print to stderr. That leaves the only
benefit of term as printing the message in red, which is not worth it.


agrover:
rebased for merge conflicts.

mulkieran:
@trgill what do you think?

PR #728, opened 2018-01-26
Clarify how to allocate all available space on a blockdev

mulkieran:
Use None to mean give everything.

### DIFF
--- a/src/engine/strat_engine/blockdev.rs
+++ b/src/engine/strat_engine/blockdev.rs
@@ -91,8 +91,9 @@ impl StratBlockDev {
         self.bda.last_update_time()
     }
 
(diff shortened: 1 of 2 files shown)


mulkieran:
Cherry-picked from #722. Somewhat related, but not so closely that it can't be considered somewhat on its own merits.

mulkieran:
@agrover review?

agrover:
Using `None` to mean "all" feels weird. How about adding another fn called `request_all_space`?

mulkieran:
@agrover Here is the less code approach, will that work?

agrover:
Yes, composing available() and request() seems nice.

mulkieran:
Doesn't change any actual code, seems approved, merged.

Issue #631, opened 2017-10-26
Support cache tier


agrover:
1. List cache blockdevs in a pool

2. Add cache blockdevs

3. Remove cache blockdevs

4. Write-through caching only

mulkieran:
From #686 .

> @mulkieran: based upon [this article you found](https://access.redhat.com/articles/2106521) and our discussion, and [this thread](https://www.redhat.com/archives/linux-lvm/2014-August/msg00016.html), it sounds like the configuration being considered assumes caching the thinpool data device (as opposed to the thin devs directly, which is not supported.) This then presents two issues: 1) this leaves the thinpool meta dev -- which should be quite "hot" access-wise -- uncached. 2) Caching doesn't support the origin device (in this case the thinpool data dev) being resized underneath it. # 2 leads to the implication that we would need to wrap data dev extensions with tearing down and recreating the cache, hurting the cache's effectiveness, and that we would also suffer reduced performance because of # 1 unless we did something else (such as moving the thin meta dev to the cache device, as that thread suggests, but that has its own complexities).

> Nothing impossible with this, but it sounds less than ideal.

> Another thought I had that could resolve both these drawbacks is to cache the entire available space in the pool: create a linear dev that glues together the usable space in all the blockdevs and use that as the cache origin dev. Flex layer devs sit on top of the cache layer, and all of them including the thin meta dev benefit. Thin data dev extensions would no longer invalidate the cache, but adding (or removing) a blockdev would.

> Thoughts?

mulkieran:
Even if cache is right above blockdevs, it's still necessary to write metadata directly to the individual blockdevs, i.e., not through cache.

mulkieran:
But, when we allocate space, we need to make sure we're allocating space from the cache device, in some sense.

PR #730, opened 2018-01-29
The cache part of the cache tier work

mulkieran:
Resolves: #631.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -327,7 +327,6 @@ dependencies = [
  "crc 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "dbus 0.6.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "devicemapper 0.18.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "either 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "env_logger 0.4.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "error-chain 0.11.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.36 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -170,7 +170,7 @@ dependencies = [
 [[package]]
 name = "devicemapper"
 version = "0.18.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
+source = "git+https://github.com/stratis-storage/devicemapper-rs#b04c09c006c2c19cd1fea134c5fb5e5df785e14c"
 dependencies = [
  "bitflags 1.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "error-chain 0.11.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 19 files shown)


mulkieran:
rebased to bring in clippy update. There is no obvious reason reason for stratisd-dbus-client test failures: https://stratis-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/stratisd-client-dbus-tests/194/.

mulkieran:
It is expected that some of the stratisd-dbus-client tests will fail. This is because the method ```AddDevs``` has been renamed to ```AddBlockDevs```. I'll open a PR to demonstrate how that would be fixed up in a bit.

mulkieran:
@agrover review, please?

agrover:
It's better to think of cache devices being in a separate "Tier" rather than a different "Role" from data blockdevs. Let's save the word "role" for differentiating blockdevs within the same tier, (although this is theoretical as we don't do this differentiation yet, but an example of we might add some day is "raid spare" -- a blockdev requiring different treatment than the others, but within the same collection of blockdevs, the same tier.)

I think the semantics we want are that blockdevs in a tier can change roles in situ, but can only change tiers by being removed and re-added.

The above is probably more relevant to future PR work but I just wanted to share my thinking before changing the design doc. In this PR we're heading down the road to blockdevs from different tiers (and roles) exporting identical APIs -- the same set of properties and methods. I just want to confirm that that's the approach we want to take.

mulkieran:
rebased. Updated D-Bus tests. Expected to pass.

mulkieran:
It passed. We currently have no AddCacheDevs tests...

mulkieran:
Tests expected to pass.

mulkieran:
This is rebased off nix branch so that it can compile against devicemapper master.

mulkieran:
Just a helpful clippy error that time.

mulkieran:
rebased.

mulkieran:
rebased.

mulkieran:
@agrover review?

mulkieran:
There is no reason why cachedev's can't use the identical metadata layout as datadevs. Therefore, we can reuse our metadata module and just worry about the variable length metadata. Furthermore, I see no reason why any variable length metadata needs to be stored on the cachedevs. It makes just as much sense to store all information about cachedevs that is necessary to be stored in the variable length metadata that is written to the datadevs. And the metadata about the cachedevs need not be so ornate, since there isn't a complicated mapping from the upper layer to the devs. There is some info to store, though.

mulkieran:
clippy, a fix should be available soon.

mulkieran:
Jenkins failure is unusual: ": Some(Error { repr: Os { code: 84, message: "Invalid or incomplete multibyte or wide character" } "

agrover:
> There is no reason why cachedev's can't use the identical metadata layout as datadevs. Therefore, we can reuse our metadata module and just worry about the variable length metadata. Furthermore, I see no reason why any variable length metadata needs to be stored on the cachedevs. It makes just as much sense to store all information about cachedevs that is necessary to be stored in the variable length metadata that is written to the datadevs. And the metadata about the cachedevs need not be so ornate, since there isn't a complicated mapping from the upper layer to the devs. There is some info to store, though.

OK, sounds good.

mulkieran:
Fix some more clippy errors and examine a real bug. When setting up the Cache Device, its length doesn't match previous cache device. There are other problems, like that the device number doesn't match previous device number, but I think the error there is that we should not be checking the device number in the first place.

mulkieran:
Yay, devicemapper bug fix means test passes.

mulkieran:
Extending tests to check the cache status. More sanity, really.

mulkieran:
rebased.

mulkieran:
Jenkins tests failing no space left on device...seems unlikely.

mulkieran:
Did a soft reboot of fedora-server-1 to hopefully clear out Jenkins failures.

mulkieran:
build and docs Travis tests were stuck for some reason, everything else had passed.

mulkieran:
Hard reboot. Jenkins errors remain the same.

mulkieran:
Travis is doing fine, other than a clippy test that can't start.

mulkieran:
rebased.

mulkieran:
Superceded by #811.

Issue #201, opened 2017-01-12
Generating PropertiesChanged signals


agrover:
While it's not needed for stratis-cli, the DBus API must generate PropertiesChanged signals. We need to investigate dbus-rs's support for generating these -- whether it's some automatic thing the lib does on our behalf, or if not then what we need to do to generate them.

agrover:
From a brief look at the dbus-rs code, it appears that a property can automatically issue the signal when changed if set, by using the `emits_changed` builder method, but based upon advice we are avoiding writable properties, so this isn't helpful. We will likely need to generate signals ourselves when the value changes, either from being set by a `SetFoo` method or some other more indirect thing that causes the property value to change.

mulkieran:
No longer part of this milestone, still part of the final plan.

mulkieran:
We can take another look at this: https://github.com/diwic/dbus-rs/issues/66#issuecomment-328357598.

agrover:
depends on #838

agrover:
fixed by #1125

PR #731, opened 2018-01-29
Dbus propertychanged signals

agrover:
see #201. We want methods to be capable of emitting PropertyChanged signals when appropriate, as well as to generate signals outside the context of a D-Bus method call.

This is important because we want to make it very easy and clear for clients to monitor when a pool is approaching its capacity! They could do this by monitoring PropertyChanged events on `TotalPhysicalSize` and `TotalPhysicalUsed` and doing some math, or we will probably have a pool `State` property that would indicate trouble that could be monitored.

This PR will let us explore what properties we want to signal for, verify signals work, and develop CI tests for our signals.


mulkieran:
My only comment is that in due course we should have some explicit rationale for how we choose True, Invalidates, etc.

agrover:
Still working on this, will open a new PR when ready.

PR #742, opened 2018-02-06
Pass last_update_time as a parameter in BlockDevMgr::new

mulkieran:
Previously, it was set unconditionally.

If the parameter is not passed then the method is restricted in how and
when during execution it can be invoked. If we choose not to pass the
parameter, then we should document BlockDevMgr::new() a whole lot better.
It is easier to pass the parameter.


mulkieran:
@agrover review?

mulkieran:
This is more or less cherry-picked from #722. I think it would probably help if it went in ahead of time.

agrover:
👍

PR #744, opened 2018-02-07
Misc improvements

tasleson:
* Fix implementation order of traits
* Use initialization shorthand


tasleson:
I've updated `engine/strat_engine/engine.rs` for implementation order and added a typo fix.

PR #747, opened 2018-02-09
Do not forget thin_meta_dev_spare when getting segments for metadata

mulkieran:
This is a bug fix. Presumably tests were passing because we have difficulty
executing code paths where we need to use the spare.


mulkieran:
@agrover review? All stratisd tests passed, but the D-Bus tests are not working yet. So this is really ready to go.

PR #752, opened 2018-02-12
Update to new version of libstratis in Cargo.lock

mulkieran:


mulkieran:
@agrover review, please?

Note that this should have gone in w/ new version commit. On the other hand, all tests pass w/out it. I find this kind of confusing.

mulkieran:
Maybe this should actually be force pushed into master?

PR #749, opened 2018-02-09
Do not allocate blocks reserved for Stratis metadata

mulkieran:
This corrects a problem which, when an existing pool was set up, it was possible to allocate the space assigned to the metadata to the upper layer.

PR #753, opened 2018-02-12
Master range alloc test

mulkieran:
This is just a test, should fail.


mulkieran:
Failed exactly as hoped.

mulkieran:
No more need for this now that #749  is merged.

PR #755, opened 2018-02-12
Update stratisd-client-dbus introspection data to match daemon

mulkieran:
Now that this is actually part of stratisd it should always be kept up-to-date.


mulkieran:
This is just a cleanup step. It would be good to add a test for SnapshotFilesystem at some later date, but the real purpose is just to clear the decks so that cache PR doesn't have stray SnapshotFilesystem stuff in it.

mulkieran:
@agrover review?

PR #756, opened 2018-02-13
Correct use statements

tasleson:
Based on what's documented in https://stratis-storage.github.io/StratisStyleGuidelines.pdf only the third commit is really applicable (section 2), but I was dinged on items in the first two commits too.  We should add the changes from the first two commits to the style guide if people agree with them.


mulkieran:
@agrover review, please?

agrover:
👍 , merging as last reviewer

PR #758, opened 2018-02-14
Tasleson udev add test

mulkieran:


agrover:
#746 was merged so this is no longer needed.

PR #760, opened 2018-02-15
Agrover rework table4

mulkieran:
It passes all internal tests, most likely it will pass the rest.


agrover:
everything pulled into #678 so this is no longer needed.

Issue #773, opened 2018-02-19
Missing /dev/stratis causes failure to bring up pool(s)


tasleson:
If `/dev/stratis` is missing we will fail to read up the metadata.  I'm not sure what's changed, but I'm kind of surprised this hasn't shown up before.

agrover:
setup_devlinks should be creating /dev/stratis if it doesn't exist. setup_devlinks is called by StratEngine::initialize, which is called by stratisd::run. So yeah, either /dev/stratis should be there or stratisd should fail to start, as far as I can see.

tasleson:
> setup_devlinks should be creating /dev/stratis if it doesn't exist.

This is done **after** we try to assemble the pools which require `/dev/stratis` to exist.

tasleson:
Unit test: https://github.com/stratis-storage/stratisd/pull/777

tasleson:
Someone please push a PR which create the directory in the appropriate place.  I really don't care where or how, just that it's messing up my forward progress with other stuff.

PR #774, opened 2018-02-19
Correct /dev/stratis missing

tasleson:
If `/dev/stratis` is missing we partially bring a pool up.

Another idea we could do is replace `create_dir` with `create_dir_all`.

Corrects: https://github.com/stratis-storage/stratisd/issues/773

### DIFF
--- a/src/engine/strat_engine/engine.rs
+++ b/src/engine/strat_engine/engine.rs
@@ -92,7 +92,10 @@ impl StratEngine {
                         return Err(EngineError::Engine(ErrorEnum::Invalid, err_msg.into()));
                     }
                 }
-                Err(_) => {
+                Err(e) => {
+                    // TODO: Add ability to distinguish expected vs. unexpected errors
+                    // (expected error would be caused by missing block device(s).
+                    warn!("Pool {} failed to setup, reason ({:?}", pool_uuid, e);
                     // Unable to setup pool, it might become complete later.
                     incomplete_pools.insert(*pool_uuid, devices.clone());
                 }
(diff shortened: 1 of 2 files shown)


src/engine/strat_engine/engine.rs
```diff
@@ -76,6 +80,12 @@ impl StratEngine {
             return Err(EngineError::Engine(ErrorEnum::Error, err_msg));
         }
 
+        if let Err(err) = fs::create_dir(devlinks::DEV_PATH) {
```
agrover: If we're doing it here, we don't need to do it in setup_devlinks. (not that it hurts, it's just duplicate.) So you could remove it there. 

A step further would be to add another little function to devlinks.rs that does this and that you call from here, if you think it's worth it to keep all the devlink stuff in one spot. Hmm, I guess we can always do that later if the code becomes more significant.

src/engine/strat_engine/engine.rs
```diff
@@ -165,40 +178,45 @@ impl Engine for StratEngine {
...
+                        //
+                        // We will also _not_ exit the daemon as we do in initialize as we were
+                        // previously up and running for some duration of time.
+                        if !self.pools.contains_name(&pool_name) {
+                            self.pools.insert(pool_name, pool_uuid, pool);
+                            Some(pool_uuid)
+                        } else {
+                            let dev_paths = devices
+                                .values()
+                                .map(|p| p.to_str().expect("Expecting valid device path!"))
```
mulkieran: While you're changing all this code around, could you leave out the exclamation mark? We don't use it anywhere else in our ```expect```s. They are supposed to be statements explaining why the developer was confident when they wrote it that the code was logically incapable of failing. If you're not confident, then you'll need to do something more sophisticated to make it possible for the ```let-``` statement to evaluate to an error.

src/engine/strat_engine/engine.rs
```diff
@@ -165,40 +178,45 @@ impl Engine for StratEngine {
...
-                               pool_uuid,
-                               dev_paths);
-                        if let Err(e) = pool.teardown() {
-                            error!("Error while tearing down pool with duplicate name! {:?}!",
-                                   e);
+                                   pool_name,
+                                   pool_uuid,
+                                   dev_paths);
+                            if let Err(e) = pool.teardown() {
+                                error!("Error while tearing down pool with duplicate name! {:?}!",
```
mulkieran: There is weird punctuation here, too!

src/engine/strat_engine/engine.rs
```diff
@@ -92,7 +102,10 @@ impl StratEngine {
                         return Err(EngineError::Engine(ErrorEnum::Invalid, err_msg.into()));
                     }
                 }
-                Err(_) => {
+                Err(e) => {
+                    // TODO: Add ability to distinguish expected vs. unexpected errors
+                    // (expected error would be caused by missing block device(s).
+                    warn!("Pool {} failed to setup, reason ({:?}", pool_uuid, e);
```
mulkieran: There is wierd punctuation here, too.

tasleson:
Many of these changes have been addressed with other PR now.

src/engine/strat_engine/engine.rs
agrover: Huh? We need it set up, we should set it up in one place, not two, not zero.

PR #777, opened 2018-02-20
Extend basic engine initialization test

mulkieran:
To verify that engine initialization succeeds when /dev/stratis is missing.

Signed-off-by: mulhern <amulhern@redhat.com>

PR #778, opened 2018-02-20
Move definition of DEV_PATH to a more central location than devlinks file

mulkieran:


mulkieran:
I think this should go in before #777 content and #774 content. Otherwise this PR will have to fix up the other PRs, which seems wrong.

@agrover review?

tasleson:
Agreed & merged.

PR #781, opened 2018-02-21
Add another possiblity to the CacheRoles

mulkieran:
We need four names, including for the the actual cache device.

### DIFF
--- a/src/engine/strat_engine/backstore/backstore.rs
+++ b/src/engine/strat_engine/backstore/backstore.rs
@@ -70,7 +70,7 @@ impl DataTier {
 
         let mut ld = LinearDev::setup(dm,
                                       &format_backstore_name(block_mgr.pool_uuid(),
-                                                             CacheRole::Origin),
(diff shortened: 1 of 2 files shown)


mulkieran:
@agrover review? It's a question of whether you're happy with the naming scheme, really. We need four names, one for the cache, and three for the three sub-devices of the cache. I plan on using Device for the whole cache, but if you prefer something else it is very easy to change now.

agrover:
Yeah, names are tough, especially when there seem to be only three names for four different things.

`Documentation/device-mapper/cache.txt` refers to the first three as sub-devices. Could we append the first three enum variation names with "Sub" or "Subdev", and then the uber-device could then be free to be "Cache"? If that's not appealing, how about naming the last one "MainDevice"? If THAT is not appealing, at least we can add docstrings describing the first three as subdevices to the enum definition?

mulkieran:
@agrover is this satisfactory?

agrover:
:+1:

PR #784, opened 2018-02-22
Omit pool UUID field from ThinPool and BlockDevMgr

mulkieran:


mulkieran:
@agrover review, please? The key thing is that you agree that it is the correct thing to do. Other than that, once you have removed a field the compiler just tells you what to do and you do it until you are done.

tasleson:
IMHO it seems totally against OOA/OOP by having a pool object that doesn't even know what it's own name and UUID is.  Objects should have data and behavior.  When I have a pool and I want to create a FS on it, my first instinct isn't that I should need to call `create_filesystems` and pass the pool name and pool uuid to it when I'm calling a method on the pool itself.

I guess I just don't understand the overall design goals here...

agrover:
@mulkieran I'm good with the param reorderings so dm is first, if you want to split that out I think that can go in.

Let's take it a little slower with the other part. I know this flows from #678 but I think we need to mull this over a bit more. As you say, this is more of a consistency-of-implementation issue rather than an actual source of bugs, for the moment.

@tasleson I hear you but Rust isn't all-in on OOP. (e.g. no inheritance). But this started as more of a container-related issue, which might also manifest in other languages, where an object storing its own identifying info clashes with that info being used as a key for the object in a container. `Table` has two keys so it's twice as bad but the same basic issue. How do other projects or languages deal with this? Once 0.5 is released I hope we can study a little and figure out the best approach to take.

mulkieran:
rebased.

mulkieran:
Of course, every blockdev knows its pool UUID, via its BDA.

mulkieran:
See https://github.com/stratis-storage/stratisd/pull/730 "Have Backstore get pool uuid from its own DataTier" commit for motivation for this PR.

mulkieran:
Spurious Travis failure.

mulkieran:
rebased. Actually redone because the rebase had > 50 conflicts, but mostly the same (except that there was a bunch of new code added in between and UUID comes before name in this version).

mulkieran:
Blocked by #866, because I'ld like to be able to view the changes using UML diagrams, too.

mulkieran:
BlockDevMgr and ThinPool are just components of StratPool. They should not have a field standing for their pool UUID, regardless of whether StratPool stores one.

The UUID and the name are keys for looking up the pools, so they should not be viewed as intrinsic qualities of the pools. That's the argument for not storing them in the struct. I think it is a pretty good one.

Individual StratBlockDevs know their pool UUID, because that is stored in their BDA. That is fine, that UUID does not represent the pool they belong to, per se, but just the data that has been written on them.

mulkieran:
I"m moving this back to in progress, because I believe these changes relate to the FIXME in dbus_api:

```
// FIXME: engine should decide to save state, not this function
```

The basic problem is that ```set_user_info``` should not be a blockdev method, because that way there is no way to trigger a metadata update. Alternatively, globalness should happen, and operations that require a metadata update should make sure that is communicated to some central thing that does it. In any case, that's more API changes.

mulkieran:
rebased.

mulkieran:
I'm putting this back up for review. AFAIC, the pool's UUID should not be stored both in the data structure the engine uses to track its pools _and_ in the pool itself.

agrover:
I'm on board with this.

tasleson:
If we ripped out dbus and replaced it with varlink, where we simply had a vector of pool objects, where would the UUID reside?

mulkieran:
Who knows what we would do with varlink?. But if we have a pool name, we can always look up the pool UUID via the engine, and vice-versa.

mulkieran:
rebased.

mulkieran:
I failed to fix up some tests last time I rebased. This should be fixed.

mulkieran:
Doing one last rebase...

mulkieran:
No longer in review-prelim.

agrover:
Once the conflicts are resolved I'd like to go ahead and get this in.

mulkieran:
rebased. I'll merge if it passes CI.

mulkieran:
Missed a test.

PR #785, opened 2018-02-22
Set Dm UUID fields when creating devices

### LINKED ISSUE #782: Should set UUID on DM devices
Right now we set DM name, but haven't set DM uuid, since uuid is kind of less a differentiated uuid field and more like a "second name" field. It sounds like for expectations of other code such as blkid, it would be good to set uuid. Setting uuid to be the same as what we're setting DM name may be the easiest thing to do.

cc @tasleson @mulkieran

agrover:
Change dm name generation functions in dmnames.rs to also generate
DmUuidBufs.

Change calls to DM setup functions to specify UUID instead of None.

The only spot that still needs work is where we call ThinDev::snapshot,
because it needs to additionally take an Option<&DmUuid>.

see https://github.com/stratis-storage/devicemapper-rs/pull/270

fixes #782

### DIFF
--- a/src/engine/strat_engine/backstore/backstore.rs
+++ b/src/engine/strat_engine/backstore/backstore.rs
@@ -15,7 +15,7 @@ use super::super::super::engine::BlockDev;
 use super::super::super::errors::{EngineError, EngineResult, ErrorEnum};
 use super::super::super::types::{DevUuid, PoolUuid};
 
-use super::super::dmnames::{CacheRole, format_backstore_name};
+use super::super::dmnames::{CacheRole, format_backstore_ids};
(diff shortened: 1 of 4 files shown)


agrover:
@mulkieran sure, see https://github.com/tasleson/utils-linux/pull/2#issuecomment-367471065 .

mulkieran:
@agrover Ok, I read the link. You're correct about the really locking down, but I don't have enough context to understand the full issue from the link or its targets.

agrover:
@mulkieran pulled in your commit, thanks!

Still not quite ready, since we need a new cut of dm-rs for add'l `snapshot` param.

PR #787, opened 2018-02-23
Master blockdevs 2

mulkieran:


mulkieran:
@agrover review?

agrover:
I think that for 1.0 we're going to want to recognize extra space on a blockdev, but just not use it until after the user explicitly tells us to. This PR goes all the way towards pretending the extra space doesn't exist. Which is fine, but just be aware we're going to need to reintroduce the idea that actual capacity and recorded (usable) capacity may differ, in the future. (see the just-opened #792).

:+1:

PR #789, opened 2018-02-23
Add systemd service file

tasleson:
As we ultimately need Stratis up and running @ boot we need this.
First pass as there may be some things that need to be changed or
added.


agrover:
:+1:

PR #798, opened 2018-02-26
Hoist creating DM context out of StratPool::teardown() to world

mulkieran:
Mark StratEngine::teardown() to make it clear that, at least now,
it is used only for testing.


mulkieran:
@agrover review, please?

mulkieran:
I think this counts as ready.

mulkieran:
Did an adjustement to get around a clippy error.

I tried:

```Ok(<expression that might return DmError>?)```

Clippy doesn't like it because it looks like I'm making a mistake, because if there is not an error, the result is just ```()```. So, why do I get the result, when I know what it will be if the expression does not evalaute to an error? My solution: tack a ```map_err``` onto the end and strip off the ```Ok```. Hopefully this will work.

mulkieran:
Blocks #730.

Issue #763, opened 2018-02-15
Limit creating filesystems to 1 at a time


agrover:
Due to other issues, creating multiple filesystems can take so much time that the D-Bus method call can time out, which should never happen.

for 0.5, return an error if more than one filesystem is attempted to be created using CreateFilesystems, to prevent D-Bus timeout.

We should be able to address the underlying issue in the 1.0 timeframe, and then remove this restriction.

agrover:
resolved by #800

PR #800, opened 2018-02-28
Only allow one filesystem to be created at a time.

trgill:

### DIFF
--- a/src/dbus_api/pool.rs
+++ b/src/dbus_api/pool.rs
@@ -36,6 +36,11 @@ fn create_filesystems(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
     let return_message = message.method_return();
     let default_return: Vec<(dbus::Path, &str)> = Vec::new();
 
+    if filesystems.count() > 1 {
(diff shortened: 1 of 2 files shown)


mulkieran:
This is a general comment: have we agreed in principle that these restrictions on operations will be implemented solely in stratisd, i.e., the CLI will not be touched?

trgill:
Yes, that was my take on issue #763 

With this change the CLI has the following output: 

# stratis filesystem create p1 fs1 fs2
Execution failed: 5: creation of multiple filesystems currently not supported - create one filesystem at a time

I'm not finished - need to make sure the error code is correct. Also need to improve the wording.

mulkieran:
Now that we have D-Bus tests in stratisd, probably a good idea to add a test there, as well.

trgill:
@mulkieran any suggestions re: what to do with the tests that no longer pass?

I'm thinking I tag them to be skipped for now.  Re-working them to pass with a temporary change seems bad.

Agree?

mulkieran:
@trgill Yes, skipping seems right to me.

agrover:
lookin good. I set to review-final. Merge?

PR #802, opened 2018-02-28
SeekFrom::Start() takes bytes as a parameter (not Sectors)

trgill:
See: https://doc.rust-lang.org/std/io/enum.SeekFrom.html

Note: Currently only called with an offset of 0.


trgill:
@mulkieran review please?

PR #805, opened 2018-03-02
Trgill create one filesystem

mulkieran:


trgill:
Ok.  Thanks.

nth() takes a mutable self - to use it I need to make filesystems mutable.  I squashed the commit in the original PR.  (not sure that is the correct process here)

agrover:
Please just use count. It's clearer and perf doesn't matter here.

mulkieran:
@trgill ```nth()``` is starting to look scarier than ```count()```, because it requires mutable access on rustc 1.24.1 but not on 1.24.0 (my version). So, I'ld say go back to ```count()```.

agrover:
parent PR merged, closing

PR #810, opened 2018-03-05
When wiping metadev use LinearDev::size() method to specify amount

mulkieran:
The intention is that the whole device be wiped. This is clearer, must
be correct, and brings the whole thing into line with projected similar
action when constructing cache meta-device.


mulkieran:
@trgill review, please?

This is nearly but not quite trivial. It blocks #730, because I wish to get this change committed prior to the commit involving the related work on the cachedev.

PR #814, opened 2018-03-06
Add 'private' to uuid for internal device mapper devices

tasleson:
As requested by upstream `util-linux` lets make the uuid simpler for comparison.  I'm following the suggestion by Andy during our meeting.

I've also made the corresponding change to `util-linux` for our review before posting upstream on mailing list, ref. https://github.com/tasleson/utils-linux/pull/2

### DIFF
--- a/src/engine/strat_engine/dmnames.rs
+++ b/src/engine/strat_engine/dmnames.rs
@@ -89,20 +89,21 @@ impl Display for CacheRole {
 ///
 /// Prerequisite: len(format!("{}", FORMAT_VERSION)
 ///             + len("stratis")                         7
+///             + len("private")                         7
 ///             + len("flex")                            4
-///             + num_dashes                             4
+///             + num_dashes                             5
 ///             + len(pool uuid)                         32
 ///             + max(len(FlexRole))                     13
 ///             < 128 (129 for UUID)
 ///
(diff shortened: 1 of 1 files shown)


agrover:
I think life is easier if we keep name and uuid in sync. So, change the name as well as the uuid to include "private".

src/engine/strat_engine/dmnames.rs
```diff
@@ -88,12 +88,17 @@ impl Display for CacheRole {
...
+    (DmNameBuf::new(format!("stratis-{}-{}-flex-{}",
+                            FORMAT_VERSION,
+                            pool_uuid.simple().to_string(),
+                            role))
+             .expect("FORMAT_VERSION display length < 72"),
+     DmUuidBuf::new(format!("stratis-{}-private-{}-flex-{}",
+                            FORMAT_VERSION,
+                            pool_uuid.simple().to_string(),
+                            role))
+             .expect("FORMAT_VERSION display length < 81"))
```
mulkieran: Because you put more stuff into the UUID the FORMAT_VERSION for the UUID is constrained to be shorter than for the name. Length needs to be less than 63, not less than 81. Header should be updated appropriately, with the lower bound, i.e, 63 in this case. Others need to be changed as appropriate.

tasleson:
@agrover Changed both uuid & name to incorporate `private`
@mulkieran Re-based and updated following your previous change

PR #816, opened 2018-03-07
Formalize prerequisites about length of FORMAT_VERSION

mulkieran:
So that the values are more likely to remain correct as the values on
which they depend change.


mulkieran:
rust tried to compile my comments, because they looked so like code, but only when running tests. This should work better.

Issue #695, opened 2017-12-19
get_mount_point failing because devnode is not listed in /proc/mounts


agrover:
Instead of /proc/mounts showing "/dev/dm-9" as the device, it is showing "/dev/mapper/stratis-1-d13578ce75854ce59bb906a7f3fc31c2-thin-fs-fc6d2af9621d43f784c38ae32dcd5e0f". As a result, `get_mount_point` returns None, and the snapshot code doesn't think it has to do the mount/unmount trick to clear the log before setting a new uuid, causing mounting the snapshot to fail.

mulkieran:
So, who's wrong?

agrover:
I don't know. "/dev/mapper/stratis..." is a symlink to /dev/dm-9, so why it is being shown in /proc/mounts instead of dm-9 I have NO idea. So I'd really like to know why!!

But since it's existing behavior and in the kernel it's unlikely to change 😦 . We may need to extend `mnt` crate to look at /proc/self/mounts, which includes major:minor.

cc @tasleson

agrover:
this requires updates to mnt-rs, which I have made in a PR here: https://github.com/stemjail/mnt-rs/pull/4 so we can use major/minor to correlate with a mount point.

agrover:
fixed by #817 for now, and will address in next release with #820.

PR #817, opened 2018-03-07
Temporarily do not use get_mount_point in fs snapshot code

agrover:
It doesn't currently work, because its implementation via mnt-rs crate only
looks at /proc/mounts not /proc/self/mountinfo.

It should be safe to unconditionally do this, both when the origin fs is
mounted and when it is unmounted.


agrover:
review @trgill, that this is an ok thing to do as a short-term workaround?

agrover:
temp solution to #695

trgill:
yeah, I think this is probably the best approach

PR #825, opened 2018-03-08
Correct BDA layout

tasleson:
The code was incorrectly placing the second copy BDA @ sector 11 instead of
sector 9.  This was because:

f.write_all(&zeroed[SECTOR_SIZE * 7..])?;

actually writes 9 sectors (sectors 7..15).  This coupled with the first
2 sector writes places the second BDA copy at sector 11.  This was
found because libblkid was failing to zero the second copy signature.


agrover:
very interesting. Nice catch!

:+1: to merge when CI is done.

Issue #827, opened 2018-03-08
Include 32-bit (i686) builds in CI


agrover:
0.5.0 doesn't build cleanly on i686, which only became apparent when packaging failed. It would be better to include i686 builds as part of our CI, so these things are corrected earlier. This could be done on a native rustc, or I've just become aware that cross-compilation might also work without too much effort.

agrover:
I can't get this to work. Somebody else want to try? See #849 for as far as I got.

agrover:
Alternatively, it might be easier to get this going on Jenkins, as @trgill pointed out!

mati865:
I see it's assigned already but there is no activity.
Would you mind if I hijacked this issue and fixed Travis?

trgill:
Hi @mati865,

Thanks for having a look.  

I added #875 to build 32 bit on our Jenkins based CI.  If there is a problem - it should fail.  We invoke it via:

TARGET=i686-unknown-linux-gnu make build

I was thinking Jenkins is good enough - but maybe other folks would like Travis fixed?  @mulkieran ?

mulkieran:
If there was a good way to do it on Travis that would be great.

PR #828, opened 2018-03-08
Fix fs_usage() for variable-size return values

agrover:
The nix::sys::statvfs::Statfvs object returned from Nix's statvfs()
function has methods that change return value depending on the
CPU architecture width. In order to work on 32 bit systems, we need to cast
these return values to u64, since our data types are not defined using
arch-specific sizes.

### DIFF
--- a/src/engine/strat_engine/thinpool/filesystem.rs
+++ b/src/engine/strat_engine/thinpool/filesystem.rs
@@ -221,10 +221,8 @@ impl Filesystem for StratFilesystem {
 pub fn fs_usage(mount_point: &Path) -> EngineResult<(Bytes, Bytes)> {
     let stat = statvfs(mount_point)?;
 
-    // stat.block_size() is type c_ulong, which is 32 bits on some archs.
-    // Upcast to u64.
-    let f_bsize = stat.block_size() as u64;
-
-    let blocks = stat.blocks();
-    Ok((Bytes(f_bsize * blocks), Bytes(f_bsize * (blocks - stat.blocks_free()))))
+    // Upcast all arch-dependent variable width values to u64
+    let (block_size, blocks, blocks_free) =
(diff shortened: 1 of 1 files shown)


src/engine/strat_engine/thinpool/filesystem.rs
```diff
@@ -225,6 +225,6 @@ pub fn fs_usage(mount_point: &Path) -> EngineResult<(Bytes, Bytes)> {
     // Upcast to u64.
     let f_bsize = stat.block_size() as u64;
 
-    let blocks = stat.blocks();
-    Ok((Bytes(f_bsize * blocks), Bytes(f_bsize * (blocks - stat.blocks_free()))))
+    let blocks = stat.blocks() as u64;
```
mulkieran: I think we've just decided to upcast everything we get from statvfs to u64, just to be safe.

What about making that very clear, as:

```
// Upcast all statvfs() result fields to u64 (on some architectures some are narrower types)
let (block_size, blocks, blocks_free) = (stat.block_size() as u64, stat.blocks() as u64, stat.blocks_free() as u64);
Ok((Bytes(block_size * blocks), Bytes(block_size * (blocks - blocks_free))))
```

agrover: I don't think this rule would be the right way to go. I'd rather just have 32bit CI builds to catch this, see #827.

mulkieran: I don't understand what you are saying.

agrover: > I think we've just decided to upcast everything we get from statvfs to u64, just to be safe.

Maybe I misunderstood you to begin with... forget it. Let me try again.

That code suggestion is fine, I'm going to tweak the suggested comment a little, FYI.

PR #829, opened 2018-03-09
Simplify stratis_filesystems_unmount()

agrover:
In looking at our mnt-rs usage I happened to look at this, and I don't
believe the closure is needed, although mostly harmless.

### DIFF
--- a/src/engine/strat_engine/tests/util.rs
+++ b/src/engine/strat_engine/tests/util.rs
@@ -70,16 +70,14 @@ fn dm_stratis_devices_remove() -> Result<()> {
 /// Try and un-mount any filesystems that have the name stratis in the mount point, returning
 /// immediately on the first one we are unable to unmount.
 fn stratis_filesystems_unmount() -> Result<()> {
-    || -> Result<()> {
-        let mounts = get_submounts(&PathBuf::from("/"))?;
-        for m in mounts
-                .iter()
-                .filter(|m| m.file.to_str().map_or(false, |s| s.contains("stratis"))) {
-            umount2(&m.file, MntFlags::MNT_DETACH)?;
-        }
-        Ok(())
-    }()
-            .map_err(|e| Error::with_chain(e, "unable to unmount all stratis filesystems"))
+    for m in get_submounts(&PathBuf::from("/"))?
+            .iter()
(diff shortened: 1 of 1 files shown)


src/engine/strat_engine/tests/util.rs
```diff
@@ -70,16 +70,14 @@ fn dm_stratis_devices_remove() -> Result<()> {
...
-        let mounts = get_submounts(&PathBuf::from("/"))?;
-        for m in mounts
-                .iter()
-                .filter(|m| m.file.to_str().map_or(false, |s| s.contains("stratis"))) {
-            umount2(&m.file, MntFlags::MNT_DETACH)?;
-        }
-        Ok(())
-    }()
-            .map_err(|e| Error::with_chain(e, "unable to unmount all stratis filesystems"))
+    for m in get_submounts(&PathBuf::from("/"))?
```
mulkieran: In C code, you will commonly see some construct like this:

```
int method_name(...) {
   ...
   // some code that determines that the time to exist is now
     goto exit;
   exit:
      //stuff that has to be done at the end
}
```

The use of the closure is the only way that I have found to duplicate the same sort of do-some-thing-before-exiting-the-method-behavior.

Your change actually changes the code behavior. In the original, if there is some failure to get the submounts, an error will be returned by the closure. This error will then be chained as the cause of the super-error, with the message "unable to mount all stratis filesystems". In your version, this error will be returned directly, and the final exit action, which was the original intention, will be missed.

I did a good deal of work to develop macros to make this prettier in the upper-layer of devicemapper-rs, but none of that has been committed, so this is one of the very few examples of this structure.

mulkieran: It should be noted that Rust is doing a good thing here. Using the C idiom with the label, it is all to easy to forget to go to exit in some branch in some if in the function and exit directly. Thousands or millions of bugs have no doubt resulted from this error. In this case, it is not up to the programmer to remember, the language does all the work, which is good.

agrover:
Hmm ok. If this isn't trivial then let's close this and come back to it later.

mulkieran:
To be clear: the change certainly introduces a bug; the bug could be considered trivial (by some, but not by me) because it's in test code not the actual stratisd source.

PR #832, opened 2018-03-12
Miscellaneous fixups

mulkieran:
Related: https://github.com/stratis-storage/stratis-docs/pull/81

PR #834, opened 2018-03-13
Get rid of a bunch of check() functions that aren't used anymore

mulkieran:
These function were made dead by PR#703 and previous PRs which introduced
an eventing mechanism supplanting the code that called check() on the
engine every few seconds. They are no longer needed. This change also
eliminates two FIXMEs and the need for some other code improvements having
to do with the instantiation of a devicemapper context.


mulkieran:
Cherry-picked from #832 for a cleaner commit.

mulkieran:
@agrover review, please?

Issue #750, opened 2018-02-09
Don't setup a pool if the metadata tells us that the next step would be to tear it down


mulkieran:
Instead, get the metadata in a prior step. There is a function, ```get_metadata``` which ought to do the trick.

mulkieran:
This ought to be a ten-liner and might as well happen soon.

Issue #661, opened 2017-11-20
Block devices with meta data opened and read too many times


tasleson:
If you do an `strace` of stratisd starting up you will see that for every block device that has stratis meta data on it we appear to open/read/close ~5 times.  It would be beneficial, especially in larger configurations to minimize the number of times we read data off the disk to, ideally reading it up only once.

excerpt for `/dev/loop1`
```
openat(AT_FDCWD, "/dev/loop1", O_RDONLY|O_CLOEXEC) = 5
ioctl(5, FIOCLEX)                       = 0
lseek(5, 0, SEEK_SET)                   = 0
read(5, "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"..., 8192) = 8192
close(5) 
...
openat(AT_FDCWD, "/dev/loop1", O_RDONLY|O_CLOEXEC) = 5
ioctl(5, FIOCLEX)                       = 0
lseek(5, 0, SEEK_SET)                   = 0
read(5, "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"..., 8192) = 8192
close(5)
...
openat(AT_FDCWD, "/dev/loop1", O_RDONLY|O_CLOEXEC) = 4
ioctl(4, FIOCLEX)                       = 0
lseek(4, 0, SEEK_SET)                   = 0
read(4, "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"..., 8192) = 8192
lseek(4, 8192, SEEK_SET)                = 8192
read(4, "s\343C\36^\254\312\16\371\1\0\0\0\0\0\0\305\31\23Z\0\0\0\0\227\3000,\0\0\0\0", 32) = 32
lseek(4, 268288, SEEK_SET)              = 268288
read(4, "i\307\311\367\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0", 32) = 32
close(4) 
...
openat(AT_FDCWD, "/dev/loop1", O_RDONLY|O_CLOEXEC) = 4
ioctl(4, FIOCLEX)                       = 0
lseek(4, 8224, SEEK_SET)                = 8224
[...]

PR #837, opened 2018-03-14
Abstract the action of setting up a pool

mulkieran:
Resolves: #750.
Related: #661.

### DIFF
--- a/src/engine/strat_engine/engine.rs
+++ b/src/engine/strat_engine/engine.rs
@@ -15,7 +15,8 @@ use super::super::errors::{EngineError, EngineResult, ErrorEnum};
 use super::super::structures::Table;
 use super::super::types::{DevUuid, Name, PoolUuid, Redundancy, RenameAction};
 
-use super::backstore::{find_all, is_stratis_device};
+use super::backstore::{find_all, is_stratis_device, setup_pool};
+#[cfg(test)]
 use super::cleanup::teardown_pools;
 use super::devlinks;
 use super::dm::get_dm;
@@ -44,6 +45,7 @@ pub struct StratEngine {
     // we've handled for each
     watched_dev_last_event_nrs: HashMap<DmNameBuf, u32>,
 }
+
 
 
 impl StratEngine {
@@ -80,23 +82,14 @@ impl StratEngine {
         let mut table = Table::default();
         let mut incomplete_pools = HashMap::new();
 
-        for (pool_uuid, devices) in &pools {
-            match StratPool::setup(*pool_uuid, devices) {
+        for (pool_uuid, devices) in pools {
+            match setup_pool(pool_uuid, &devices, &table) {
                 Ok((pool_name, pool)) => {
-                    let evicted = table.insert(pool_name, *pool_uuid, pool);
-                    if evicted.is_some() {
-                        // TODO: update state machine on failure.
-                        let _ = teardown_pools(table);
-                        let err_msg = "found two pools with the same name";
-                        return Err(EngineError::Engine(ErrorEnum::Invalid, err_msg.into()));
-                    }
+                    table.insert(pool_name, pool_uuid, pool);
                 }
-                Err(e) => {
-                    // TODO: Add ability to distinguish expected vs. unexpected errors,
-                    // expected error would be caused by missing block device(s).
-                    warn!("Pool {} failed to setup, reason {:?}.", pool_uuid, e);
-                    // Unable to setup pool, it might become complete later.
-                    incomplete_pools.insert(*pool_uuid, devices.clone());
+                Err(err) => {
+                    warn!("pool failed to setup, reason {:?}", err);
+                    incomplete_pools.insert(pool_uuid, devices);
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


mulkieran:
I plan to merge "Pass metadata to StratPool::seupt() as reference" into "Do not get the metadata" in final version.

mulkieran:
@agrover review, please?

I plan to enhance the error messages returned by ```setup_pool()``` to be more like the log message currently printed out, assuming there are no strong objections to this.

mulkieran:
Some minor clippy errors on Travis...

agrover:
Looks fine to me. @tasleson, do you have time to glance at this too?

src/engine/strat_engine/engine.rs
```diff
@@ -85,23 +87,14 @@ impl StratEngine {
         let mut table = Table::default();
         let mut incomplete_pools = HashMap::new();
 
-        for (pool_uuid, devices) in &pools {
-            match StratPool::setup(&dm, *pool_uuid, devices) {
+        for (pool_uuid, devices) in pools {
+            match setup_pool(&dm, pool_uuid, &devices, &table) {
```
tasleson: `setup_pool` is doing the duplicate name check and returning an error if we already have the name.  With this code change the error will be consumed and a warning message emitted instead of bubbling up the error and causing the daemon to exit.  Is this the behavior we want?  You touch on this in the commit documentation. 

The source code documentation is incorrect for `StratEngine::initialize` with this change of behavior.

src/engine/strat_engine/engine.rs
```diff
@@ -85,23 +87,14 @@ impl StratEngine {
...
                 Ok((pool_name, pool)) => {
-                    let evicted = table.insert(pool_name, *pool_uuid, pool);
-                    if evicted.is_some() {
-                        // TODO: update state machine on failure.
-                        let _ = teardown_pools(&dm, table);
-                        let err_msg = "found two pools with the same name";
-                        return Err(EngineError::Engine(ErrorEnum::Invalid, err_msg.into()));
-                    }
+                    table.insert(pool_name, pool_uuid, pool);
```
tasleson: We are not checking to see if anything is evicted with this change.  I think we can convince ourselves that this isn't possible based on the code flow in getting here (that we don't have a conflicting UUID), but should we assert that?

src/engine/strat_engine/engine.rs
```diff
@@ -85,23 +87,14 @@ impl StratEngine {
...
                 }
-                Err(e) => {
-                    // TODO: Add ability to distinguish expected vs. unexpected errors,
-                    // expected error would be caused by missing block device(s).
-                    warn!("Pool {} failed to setup, reason {:?}.", pool_uuid, e);
-                    // Unable to setup pool, it might become complete later.
-                    incomplete_pools.insert(*pool_uuid, devices.clone());
+                Err(err) => {
+                    warn!("pool failed to setup, reason {:?}", err);
```
tasleson: Is there some reason we need to remove the pool UUID from the message that is created?  Just knowing that some pool failed, isn't as helpful as knowing which pool failed and for what.

src/engine/strat_engine/engine.rs
```diff
@@ -152,73 +145,43 @@ impl Engine for StratEngine {
...
-                        }
+                        self.pools.insert(pool_name, pool_uuid, pool);
+                        Some(pool_uuid)
                     }
-                    Err(e) => {
-                        warn!("udev add: pool {} failed to setup, reason {:?}.",
-                              pool_uuid,
-                              e);
+                    Err(err) => {
+                        warn!("udev add: pool failed to setup, reason {:?}", err);
```
tasleson: Again, the pool UUID is removed, less helpful IMHO.

mulkieran: Pool UUID is in the error returned by ```setup_pool()```.

src/engine/strat_engine/engine.rs
mulkieran: Pool UUID is in the error returned by ```setup_pool()```.

src/engine/strat_engine/engine.rs
mulkieran: Can assert here if desired...should assert in ```block_evaluate()``` as well if that is the plan.

src/engine/strat_engine/engine.rs
mulkieran: One question I'm asking with this PR is "Is there a compelling reason to return an error in ```initialize()``` on this condition but not in ```block_evaluate()```?" I don't see that there is one and to make this distinction makes the code less maintainable.

mulkieran:
No changes address #661 yet. But with this change and added uniformity that it brings that issue should be much easier to address.

src/engine/strat_engine/engine.rs
mulkieran: No, not all, not yet: https://github.com/stratis-storage/stratisd/pull/837#issuecomment-374616448.

src/engine/strat_engine/engine.rs
tasleson: It's probably a good thing to keep the behavior the same and because `stratisd-init` shouldn't exit on errors like this in initramfs.  However, there was a comment with a todo about distinguishing between error types.  There is a difference between being unable to bring up a pool because `thin_check` or `thin_repair` tool is missing vs. one of the devices isn't present.  Do we want the same behavior for those errors too?

mulkieran: Existing code does not distinguish among those causes and PR doesn't change that. I can not imagine a future where the code will not have to distinguish among those different causes and modify its actions accordingly. But, at present, no plan has been developed for that. When such a plan is developed the necessary information will be obtainable by making use of newly invented, non-generic errors returned from ```setup_pool```. This, though, is an extensive undertaking, which will involve modifications to ```ThinPool::Setup``` and other methods, as well as more work on the existing error hierarchy and possibly other internal error heirarchies. It's definitely beyond the scope of this PR.

tasleson: Created https://github.com/stratis-storage/stratisd/issues/854 for tracking this

src/engine/strat_engine/engine.rs
tasleson: > I plan to enhance the error messages returned by setup_pool() to be more like the log message currently printed out, assuming there are no strong objections to this.

I would prefer you defer making changes to the existing messages and include the changes with your future PR to enhance the error messages by `setup_pool`.  IMHO it would be better to have redundant information in the mean time instead of no information until the improvements can be made, thanks.

mulkieran:
rebased.

mulkieran:
Superceded by #864.

PR #841, opened 2018-03-15
Add a method to initialize a devicemapper context just once

mulkieran:
Initialize the context in StratEngine::initialize().
Obtain it everywhere using the method get_dm_context(). Do not pass it as
an argument in stratisd but do instantiate it when passing to methods in
devicemapper.

### DIFF
--- a/src/engine/engine.rs
+++ b/src/engine/engine.rs
@@ -181,7 +181,7 @@ pub trait Engine: Debug {
 
     /// If the engine would like to include an event in the message loop, it
     /// may return an Eventable from this method.
-    fn get_eventable(&self) -> EngineResult<Option<Box<Eventable>>>;
+    fn get_eventable(&self) -> EngineResult<Option<&'static Eventable>>;
 
     /// Notify the engine that an event has occurred on the Eventable.
     fn evented(&mut self) -> EngineResult<()>;
--- a/src/engine/sim_engine/engine.rs
+++ b/src/engine/sim_engine/engine.rs
@@ -121,7 +121,7 @@ impl Engine for SimEngine {
             .collect()
     }
 
-    fn get_eventable(&self) -> EngineResult<Option<Box<Eventable>>> {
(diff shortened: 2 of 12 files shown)


mulkieran:
This is partial, while I get the prelim review. The parts are three:
1. Definition of a function to get a reference to a devicemapper context in new file dm.rs.
2. Calling that function for the first time in StratEngine::initialize().
3. Eliminating passing a reference to the DM context and instead getting that reference exactly where it is needed, i.e., when passing to a method in devicemapper-rs. This is done in a small subset of the methods where it is passed in backstore only. The rest would be exactly the same, but I don't want to do that unless it is decided that this is the way to go.

mulkieran:
clippy is failing, but otherwise Travis checks are fine.

tasleson:
I like this approach, singletons work well for stuff like this.  Could this functionality be pushed into devicemapper-rs?

mulkieran:
(1) could go in devicemapper-rs core, probably dm.rs.

mulkieran:
@agrover review, please?

Some questions:

* Is ```get_dm_context``` the right name? If not, now is the time to fix it.
* Happy with removal of ```mut``` modifiers in Eventable?
* There is no assignment of ```get_dm_context()``` result to any idenfiter in any method except ```StratEngine::initialize()```. Usually, if I want to use the result of a call twice in a single method, I give it a name. The result of ```get_dm_context()``` could be treated differently because of its special nature, as it is here. But the alternative isn't such a problem either. Do you have a preference? If you do prefer the current option, how about inlining the method?

agrover:
How about something even shorter like `get_dm`?

I'd lean against ever binding it to a name. Is there source-level clarity to be gained?? I don't think it matters in what the compiler will output.

I'd lean against notating as inline, I don't see a good reason to second guess whatever the compiler wants to do (probably will inline, I bet).

mulkieran:
rebased.

mulkieran:
Superceded by #858.

PR #843, opened 2018-03-15
Initialize a devicemapper context just once

mulkieran:


mulkieran:
An alternative to #841 where ```get_dm_context``` is now located in devicemapper-rs.

mulkieran:
A further step would involve pushing this change deeper into devicemapper-rs. The idea would be that the client invokes ```get_dm_context()``` once and knows that that guarantees that the methods in the devicemapper-rs outer layer will be able to obtain any DM contexts they might need to invoke methods in the DM implementation. The client, though, has to be the one to manage calling any DM impl methods directly.

mulkieran:
#841 it is.

PR #844, opened 2018-03-15
Get rid of two coverage pragmas

mulkieran:
We do not run coverage tests on this code anymore, because its no longer
part of a library, so the coverage pragmas are no longer useful.


mulkieran:
@agrover review, please?

PR #847, opened 2018-03-16
Ensure only one instance of stratisd runs at a time

### LINKED ISSUE #719: stratisd can have multiple instances running currently
Our assumption is that there can only be one, this is not true!  Badness results when we have existing pools at start up!
```
# ps -ef | grep stratisd
root      8954  1699  0 11:30 pts/0    00:00:00 /home/tasleson/projects/stratis/stratisd/target/debug/stratisd --debug
root      8955  1699  0 11:30 pts/0    00:00:00 /home/tasleson/projects/stratis/stratisd/target/debug/stratisd --debug
root      8958  1699  0 11:30 pts/0    00:00:00 /home/tasleson/projects/stratis/stratisd/target/debug/stratisd --debug
root      8959  1699  0 11:30 pts/0    00:00:00 /home/tasleson/projects/stratis/stratisd/target/debug/stratisd --debug
root      8960  1699  0 11:30 pts/0    00:00:00 /home/tasleson/projects/stratis/stratisd/target/debug/stratisd --debug
```

agrover:
Create, lock, and write our pid to /var/run/stratisd.pid.

Add nix::Error to StratisError.

fixes #719


agrover:
I looked at crates like "daemonize" and "daemon" for this but they also do things like doube-fork and close stdout/err, which we don't currently want.

src/bin/stratisd.rs
```diff
@@ -78,6 +86,33 @@ fn handle_udev_add(event: &libudev::Event) -> Option<(PathBuf, Device)> {
...
+            f.read_to_string(&mut buf)?;
+            // pidfile is supposed to contain pid of holder. But you never
+            // know so be paranoid.
+            let pid_str = buf.split_whitespace()
+                .next()
+                .and_then(|s| s.parse::<pid_t>().ok())
+                .map(|pid| format!("{}", pid))
+                .unwrap_or_else(|| "<unknown>".into());
+            eprintln!("Stratisd is already running with pid {}", pid_str);
+            exit(1);
```
mulkieran: Just return an error with a message. It will be printed out by the ```main()``` method, see below.

src/bin/stratisd.rs
```diff
@@ -78,6 +86,33 @@ fn handle_udev_add(event: &libudev::Event) -> Option<(PathBuf, Device)> {
...
+/// To ensure only one instance of stratisd runs at a time, acquire an
+/// exclusive lock. Exit if lock attempt fails.
+fn create_pid_file() -> StratisResult<File> {
+    let mut f = OpenOptions::new()
+        .read(true)
+        .write(true)
+        .create(true)
+        .open(STRATISD_PID_PATH)?;
+    match flock(f.as_raw_fd(), FlockArg::LockExclusiveNonblock) {
+        Ok(()) => {f.write(format!("{}\n", getpid()).as_bytes())?;
```
mulkieran: ```Ok(())``` -> ```Ok(_)```

mulkieran: Why the "\n"? Is that a rule for these files?

agrover: seems to be what other .pid files do, from a quick survey on my local machine, yeah.

tasleson:
We should update the stratisd.service file to reflect pid file too.

agrover:
> We should update the stratisd.service file to reflect pid file too.

My reading of https://www.freedesktop.org/software/systemd/man/systemd.service.html is that we don't need it because systemd only recommends that if the daemon is of Type = forking.

(Kind of a strange situation.. since stratisd doesn't fork then systemd really doesn't care about the pidfile (it already knows our pid), it's really a stratisd-specific use -- putting the file there so us as developers, or users accidentally somehow running the daemon, don't end up with multiple instances. But the location and contents of the file are a historical convention that implies something that doesn't apply.)

agrover:
fixed a clippy warning.

agrover:
@mulkieran I merged your fixes from #859, much thanks! Due to the scope I decided to merge some commits. I've now included your signed-off-by in the commitlog since it's a shared commit. Please let me know if you'd prefer I not merge so much.

PR #849, opened 2018-03-20
Update travis to build on i686

### LINKED ISSUE #827: Include 32-bit (i686) builds in CI

agrover:
fixes #827

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,24 +7,35 @@ addons:
           - libdbus-1-dev
           - libudev-dev
           - libdbus-glib-1-dev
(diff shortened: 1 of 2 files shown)


agrover:
hah, not that easy. Need to modify makefile to take TARGET. I'll worry about it tomorrow.

agrover:
I can't get this working.

PR #850, opened 2018-03-20
Get rid of a '.'

mulkieran:
In order to avoid trying to format all the Python files in the directory.
We only want to format those that are part of our project.


mulkieran:
We all agreed to this already, so I'll just go ahead and merge it.

PR #851, opened 2018-03-20
Move 1 second sleeps after stratisd setup to setUp function

mulkieran:
There is always a sleep of 1, or in one instance 2 seconds after
every invocation of Service.setUp. These all occur in TestCase.setUp
methods. Just do the sleep in Service.setUp for less verbosity.


mulkieran:
@trgill review, please?

PR #852, opened 2018-03-20
Master unique dm test

mulkieran:


mulkieran:
Just seeing what will transpire...

mulkieran:
Jenkins just had a problem, so this wasn't really helpful.

mulkieran:
The Travis tests are failing sensibly. They don't all fail because the transformation to using ```get_dm_context()``` is partial. But ```get_dm_context()``` fails and presents a fairly reasonable stack trace in the appropriate place.

What the Jenkins tests should demonstrate is some kind of reasonable failure on engine initialization.

mulkieran:
Once getting the DM context has failed once, any further attempts will be "poisoned" and will result in a stack trace that ```catch_unwind()``` can't catch. Testing to see what happens on the first failure.

mulkieran:
Think that correct execution of ```catch_unwrap``` has been confirmed, so this can go.

PR #857, opened 2018-03-21
Make self parameters to Eventable methods non-mut

mulkieran:
They do not need to take a mutable argument, because the DM context
is never mutated and a DM context is the one thing that implements
Eventable in the stratisd code base.

As a consequence it is possible to make Engine::eventable()
return an immutable object, so do that.

Remove some mut modifiers where the devicemapper context is used as
an Eventable.

### DIFF
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -135,12 +135,12 @@ fn run() -> StratisResult<()> {
                  events: libc::POLLIN,
              });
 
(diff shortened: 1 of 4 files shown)


mulkieran:
@agrover review, please?

Cherry-picked from #841 .

agrover:
cool. If we ever need this (doubtful) we can add it back. yagni.

PR #864, opened 2018-03-27
Abstract the action of setting up a pool

mulkieran:
Resolves: #750.


mulkieran:
I think that ```check_metadata``` is a way of specifying an invariant on the pool itself, rather than simply looking for a problem w/ the metadata. Given that fact, it looks like it makes sense to incorporate ```check_metadata```  into an ```invariant``` method that can be used in the pool tests. So I've done gone ahead and done that.

agrover:
thanks!!

PR #866, opened 2018-03-29
Generate UML-diagram like graphs for particular modules in stratis

mulkieran:
These are somewhat useful for code visualization and this approach
seems to be the best available so far.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,9 @@ ${HOME}/.cargo/bin/cargo-tree:
 ${HOME}/.cargo/bin/cargo-outdated:
 	cargo install cargo-outdated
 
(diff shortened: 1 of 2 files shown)


mulkieran:
This doesn't get run automatically. The major question is whether all are comfortable with the output location, which is easily changed.

mulkieran:
I'ld like to use the walkdir crate, to see if I can just generate graphs for every module rather than have an ad-hoc list. This shouldn't be too hard.

agrover:
:+1: if new file goes in scripts/, as we just did over in devicemapper.

mulkieran:
I think this is identical to devicemapper version now.

PR #867, opened 2018-03-29
Use a typical cache block size

mulkieran:
Devicemapper seems uncomfortable with such small blocks on such a large
cache. This should have the effect of postponing the point at which the
cache metadata sub-device runs out of room, for a long time but not
indefinitely. Once a cache device is setup, it is necessary to stick with
the cache block size. Stopping the cache and making a whole new one with
a different cache block size is something we may find that it is advisable
to do, but is a future project.

Excerpt from devicemapper logs:
[Mar29 10:06] device-mapper: cache: You have created a cache device with a lot of individual cache blocks (128183735)
All these mappings can consume a lot of kernel memory, and take some time to read/write.
Please consider increasing the cache block size to reduce the overall cache block count.


mulkieran:
backstore tests failing on Travis and Jenkins w/ an invalid argument on creating the cache device. Devicemapper tells us no more, but it is clear that the problem is that 2056 != 2058. Whoops!

agrover:
:+1:

PR #871, opened 2018-03-30
Reduce reads of Stratis metadata during pool setup

mulkieran:
Resolves: #486

### DIFF
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -68,7 +68,7 @@ fn initialize_log(debug: bool) -> Result<(), SetLoggerError> {
 
 /// Given a udev event check to see if it's an add and if it is return the device node and
 /// devicemapper::Device.
-fn handle_udev_add(event: &libudev::Event) -> Option<(PathBuf, Device)> {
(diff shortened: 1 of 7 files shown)


mulkieran:
So, what really happens?

*  ```find_all()``` pokes twice at each device. The first time, to get its device number and the second when calling

* ```is_stratis_device()```. Currently, ```is_stratis_device()``` reads the whole StaticHeader in order to determine the pool UUID, but only returns the UUID. It is used in ```find_all``` and ```block_evaluate```.

* A pair of pool UUID and a list of component devices are then passed to ```setup_pool````.  ```setup_pool``` in turn invokes ```get_metadata``` on the devices.

* ```get_metadata``` loads the BDA. If the device is not a Stratis device it ignores the device. In theory, all devices it encounters should be known Stratis devices. It needs the BDA in order to get information about the variable length metadata. It investigates the BDAs of all the devices in order to find those w/ the most recently written metadata, and then it reads those. ```get_metadata``` is called exclusively by ```pool_setup```.

mulkieran:
Closing in favor of #872.

PR #872, opened 2018-03-30
Reduce number of metadata reads

mulkieran:
Alternative to #871.

The basic idea is that ```find_all``` or ```block_evaluate``` as the case may be, attempts to load a BDA if it has reason to believe that the device it's working with is a Stratis device. It then stores the result of that attempt as an ```StratisResult<BDA>``` along with the device. Other methods that previously loaded the BDA now get the BDA in the data structure which they take as an argument. These methods themselves return an error if the BDA is an Err(), so their semantics should be more or less the same. There are some TODOs which call out some possible courses of action.

### DIFF
--- a/src/engine/strat_engine/backstore/backstore.rs
+++ b/src/engine/strat_engine/backstore/backstore.rs
@@ -435,7 +435,7 @@ impl Backstore {
                  devnodes: &HashMap<Device, (PathBuf, StratisResult<BDA>)>,
                  last_update_time: Option<DateTime<Utc>>)
                  -> StratisResult<Backstore> {
-        let (datadevs, cachedevs) = get_blockdevs(pool_uuid, backstore_save, devnodes)?;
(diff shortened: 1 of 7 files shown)


mulkieran:
rebased.

mulkieran:
Still awaiting prelim feedback.

agrover:
update commitlog since EngineResult is no more

Is it possible that `HashMap<Device, (PathBuf, StratisResult<BDA>)>` can be `HashMap<Device, (PathBuf, BDA)>`?

mulkieran:
Moving back to "in progress", in order to experiment w/ some deeper algorithm changes.

mulkieran:
rebased.

mulkieran:
This is blocked by libblkid work.

mulkieran:
Closing. Excessive conflicts make it not worth the trouble to try to revive.

PR #877, opened 2018-04-04
Mark rustfmt as an allowed failure

agrover:
Failing just because a line is 101 characters is not cool.

This may be fixed by adopting a newer, better rustfmt but for the moment
it doesn't seem right to modify our code because rustfmt 0.8.3 has this
issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,9 @@ addons:
 language: rust
 
 matrix:
+    allow_failures:
+        - rust: stable
+          env: TASK=fmt-travis
     include:
         - rust: stable
           env: TASK=fmt-travis


mulkieran:
I prefer #878.

agrover:
It just strikes me as very backward and frustrating that the formatter, which is supposed to relieve us from worrying about non-significant whitespace, demands we change the actual structure of our code to pass.  One may find #878 more readable but that doesn't change that the formatter is failing on valid code.

This seems like a good use case for allowed failure, until we can move to the newer better version of rustfmt. I don't mind if we merge #878 too but defanging rustfmt a little in the meantime seems correct, either through travis or if there's some rustfmt config option to make it allow long lines(?) to avoid the issue.

mulkieran:
I don't think it's a good use case. When I'm coding busily, I end up running rustfmt 50 times in one day to bring my code back into line. One fmt failure will break my workflow, just like an errant compiler warning would. Now, that's frustrating, much more frustrating than occasionally coping with a rustfmt inadequacy and incidentally making the code more readable.

It doesn't matter that rustfmt is far from perfect, because it is so much better than nothing and we are not going to fix it. So we might as well live with its little imperfections and get the benefit. If we manage to move up to the new version we'll get new imperfections that we also have to deal with.

agrover:
ok I'm putting up with the status quo a little longer in the hopes that we'll be using a new rustfmt soon.

PR #878, opened 2018-04-04
Fix formattting

mulkieran:
So that we can continue to use our current formatter. As a nice extra,
the code is a bit more readable.


agrover:
ok, thanks for fixing this for me.

PR #881, opened 2018-04-05
Tidy error handling in device cleanup as much as possible

mulkieran:


mulkieran:
I'm putting this up for final review, because it puts it in, I think, the correct state for what we want to express. It will make an easier starting point for failure experiments.

agrover:
lgtm :+1:

PR #884, opened 2018-04-06
Change StratisError::Error to StratisError::Startup

mulkieran:


agrover:
Yes the intention was that that Error would be a generic error. Making this Startup just happens to give a little more information about when the generic error is happening, but how does that help, and why is mapping to dbus INTERNAL ERROR nonsensical? Seems to me if stratisd happens to have another error condition in the future that doesn't occur during startup we'd just be re-renaming this error type again. Or make another error type?

mulkieran:
rebased.

mulkieran:
INTERNAL_ERROR is used to mean that there is an error in our source code. Failing to start up because there is another instance of stratisd already running does not count as an error in our source code.

agrover:
how about adding Startup instead of replacing Error?

PR #888, opened 2018-04-23
Extend a test to verify behavior on non-existent object path

mulkieran:
For greater precision.


mulkieran:
I guess I might as well extend the test a bit.

PR #889, opened 2018-04-24
Build with dbus by default

### LINKED ISSUE #739: Enable building with dbus support by default
Since compiling with dbus is the normal case for stratisd, add dbus_enabled to the `default` feature. See https://github.com/rust-lang/cargo/issues/938#issuecomment-63926940. Building without dbus support should still be possible with `--no-default-features`.

mati865:
Hello,
This is super trivial change.

Verified it works as expected by comparing size of the output binary.
`--no-default-features` mentioned in issue was also tested.

Before this change `clippy` was run without `dbus` enabled so it revealed needless pass by value.
It required changing public API so you might want to hold back merging this PR for some time.

While at it noticed 5 warnings and  1 error from `clippy` in the `src/bin/stratisd.rs` (it doesn't have `clippy` enabled).
Looks like oversight to me.

Fixes #739

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,4 +34,5 @@ quickcheck = "0.6"
 loopdev = "0.2"
 
 [features]
+default = ["dbus_enabled"]
 dbus_enabled = ["dbus"]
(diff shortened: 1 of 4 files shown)


mulkieran:
@agrover review, please? The API change is in a sub-module, so I would not have a problem w/ that going in any time.

PR #891, opened 2018-04-26
Enable i686 Travis build

### LINKED ISSUE #827: Include 32-bit (i686) builds in CI

mati865:
cc @agrover 
Tested it in Docker but let's put it through Travis to confirm.

Fixes #827

Signed-off-by: Mateusz Mikuła <mati865@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,14 +19,12 @@ matrix:
         - rust: stable
           env: TASK=build TARGET=x86_64-unknown-linux-gnu
         - rust: stable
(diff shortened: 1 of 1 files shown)


agrover:
Kickass!!!!

Thanks so much @mati865!

I'm going to leave the two commits, b/c different authors.

PR #894, opened 2018-04-30
Fix a bug and introduce an improvement around ensuring only single instance of stratisd is running

### LINKED ISSUE #893: stratisd should not require setting up pid file when called with CLI options that exit immediately
See #853 for an example of the problem.

mulkieran:
Resolves #893.
Resolves: #895.

### DIFF
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -91,16 +91,11 @@ fn handle_udev_add(event: &libudev::Event) -> Option<(Device, PathBuf)> {
 
 /// To ensure only one instance of stratisd runs at a time, acquire an
 /// exclusive lock. Return an error if lock attempt fails.
-fn create_pid_file() -> StratisResult<File> {
-    let mut f = OpenOptions::new()
-        .read(true)
-        .write(true)
-        .create(true)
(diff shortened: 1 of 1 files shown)


mulkieran:
Waiting for clippy update...

mulkieran:
Still waiting for clippy update...

agrover:
I don't understand why 6e2ee98 is needed, why doesn't the existing code work as intended?

mulkieran:
I got the idea from the flock man page:

> Locks created by flock() are associated with an open file table entry. This means that duplicate file descriptors (created by, for example, fork(2) or dup(2)) refer to the same lock, and this lock may be modified or released using any of these descriptors. Furthermore, the lock is released either by an explicit LOCK_UN operation on any of these duplicate descriptors, or when all such descriptors have been closed.

agrover:
please see #897, I think only a smaller change is all that's required.

I've tested #897 as working for me (i.e. duplicate daemons do not start) with rustc 1.25.0 on F27.

mulkieran:
After one and a half days or thereabouts, building clippy from master branch for CI seems like not a bad idea.

agrover:
let's work on this over in #897. closing.

Issue #895, opened 2018-04-30
Don't lose lock on pid file before stratisd is done


mulkieran:
If the lock is lost, then multiple instances of stratisd can end up running at the same time.

See #853 for example.

PR #897, opened 2018-04-30
Move general startup tasks into main(), and ensure pidfile is held properly

agrover:
The previous code would cause the pidfile to be closed immediately when
control returned to run(). Giving it a name prevents this.

I swear this used to work. Did a newer compiler version perhaps change
this behavior?

### DIFF
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -122,13 +122,11 @@ fn run(matches: &ArgMatches) -> StratisResult<()> {
     initialize_log(matches.is_present("debug"))
         .expect("This is the first and only invocation of this method; it must succeed.");
 
-    // We must setup a udev listener before we initialize the
-    // engine. It is possible that a device may appear after the engine has read the
-    // /dev directory but before it has completed initialization. Unless the udev
-    // event has been recorded, the engine will be unaware of the existence of the
-    // device.
-    // This is especially important since stratisd is required to be able to run
-    // during early boot.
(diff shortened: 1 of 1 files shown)


agrover:
fixes #895

mulkieran:
I think previous code must have been the victim of an overzealous compiler optimization. But I think this version could easily be the victim of a slightly more overzealous optimization. My fix will probably hold for the foreseeable future, so I think it's probably a better choice.

agrover:
I've posed a question: https://users.rust-lang.org/t/unused-variables-that-need-to-not-be-prematurely-dropped/17192

agrover:
Some additional good feedback on that thread. I'm going to see if I can rework this to look more like a held lock (which it is) based on those suggestions.

src/bin/stratisd.rs
```diff
@@ -279,12 +262,25 @@ fn run() -> StratisResult<()> {
...
+        .about("Stratis storage management")
+        .arg(Arg::with_name("debug")
+                 .long("debug")
+                 .help("Print additional output for debugging"))
+        .arg(Arg::with_name("sim")
+                 .long("sim")
+                 .help("Use simulator engine"))
+        .get_matches();
+
+    initialize_log(matches.is_present("debug"))
```
mulkieran: I think that initialize() of log should occur after lock is gained. Otherwise there are two stratisd's both logging at the same time.

agrover:
fixed issue @mulkieran raised, and then added another commit because I couldn't resist.

PR #919, opened 2018-05-03
Disable clippy test while clippy continues broken

mulkieran:


agrover:
:+1:

trgill:
:+1:

mulkieran:
This failed to properly allow the failures, but is identical to previous that worked. Maybe it has to match the entire environment. I don't think that makes sense, but you never know.

mulkieran:
It does have to match the whole environment. Wonder what the reasoning behind that was.

mati865:
I still opt for building clippy from git.

mulkieran:
I want to wait until the clippy people make up there minds as to what they think we should do. Some issue on their GitHub site mentioned that, as of this morning, it was somehow broken on master, too.

PR #920, opened 2018-05-03
Test pool Full -> Good transition

trgill:
Extend the test framework to allow unit tests to request a size range for the test devices.  If the range is not available - the supplied devices may be sliced up into DM linear devs to satisfy the request.

Add a test case that uses the test framework extension to test that once a pool has gone into the "full" state, it can be transitioned back to good when block devices are added.

### DIFF
--- a/src/engine/strat_engine/backstore/backstore.rs
+++ b/src/engine/strat_engine/backstore/backstore.rs
@@ -801,7 +801,8 @@ mod tests {
 
     #[test]
     pub fn real_test_add_cache_devs() {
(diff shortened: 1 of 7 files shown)


trgill:
failure on Travis due to clippy problem

mati865:
You can retrigger build, clippy is fine now.

agrover:
this looks fine. :+1:

trgill:
@mulkieran review please?

mulkieran:
Will try to get to it first thing tomorrow.

mulkieran:
Taking off project since no longer in use.

PR #921, opened 2018-05-04
Prefix commands with entire path

agrover:
When invoked via dbus activation, we don't have /usr/sbin in our path, so
these commands are failing.

This is the cause of create pool failing with "4: Entity not found", execve() is returning ENOENT.


mulkieran:
Superceded by #922.

PR #922, opened 2018-05-07
Organize wrappers for command-line invocations

mulkieran:


mulkieran:
Get clippy benefits now that it is back.

trgill:
we might want to squash these together

mulkieran:
Travis CI is just a lint failure, so it can be ignored.

PR #927, opened 2018-05-09
metadata: Use sync_all() instead of flush()

### LINKED ISSUE #892: Need to ensure data that we write is actually flushed
We were using ```flush()``` to ensure that, but it turns out that ```flush()``` only flushes in app buffering.

This is going to cause problems in the metdata module, because, in order to make testing easier, metadata functions generally take a generic F which implements ```Seek + Write```.

agrover:
Apparently flush() does nothing, so use sync_all() instead to ensure
data is saved.

Since we also want our functions to be generic so we can mock using Cursor,
Add trait SyncAll, which allows Cursor to still work with the new method
call (sync_all() on a Cursor is no-op.)

There were also a couple spots that were taking a reference that didn't
need to, that needed to be corrected to get this working.

fixes #892

### DIFF
--- a/src/engine/strat_engine/backstore/metadata.rs
+++ b/src/engine/strat_engine/backstore/metadata.rs
@@ -18,6 +18,8 @@ use super::super::super::types::{DevUuid, PoolUuid};
 
 use super::super::engine::DevOwnership;
 
+use super::util::SyncAll;
+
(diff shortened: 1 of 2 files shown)


agrover:
obsoleted

PR #930, opened 2018-05-09
Establish SyncAll trait for metadata testing

mulkieran:


agrover:
Nice!

If you want to squash the original commit into yours, we can just go with this PR.

PR #931, opened 2018-05-09
Try to copy back old data if initialize fails

agrover:
Starting in initialize() and proceeding into callees that write data,
read the existing data first and store it, along with where it goes, into
a RestoreState struct. In the event an error occurs before successful
completion of initialize(), attempt to write back the pre-existing data.

Much of the code changes are in order to share F, which is either a File or a
memory Cursor (used for testing.) Wrap it in a `Rc<RefCell>>`. This is mostly
driven by Cursor, which unlike File does not have a cloneable handle.

### DIFF
--- a/src/engine/strat_engine/backstore/metadata.rs
+++ b/src/engine/strat_engine/backstore/metadata.rs
@@ -2,7 +2,9 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-use std::io::{Read, Seek, SeekFrom};
+use std::cell::RefCell;
+use std::io::{Read, Seek, SeekFrom, Write};
+use std::rc::Rc;
 use std::str::from_utf8;
 
 use byteorder::{ByteOrder, LittleEndian};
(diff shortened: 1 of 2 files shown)


agrover:
Thoughts? No CI tests written yet.

mulkieran:
Waiting until conflicts are resolved...

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -505,7 +561,7 @@ mod mda {
                 )))?;
                 f.write_all(&hdr_buf)?;
                 f.write_all(data)?;
-                f.flush()?;
+                f.sync_all()?;
```
mulkieran: Looks like this change and also change immediately above do not belong in this PR. AFAICT, they should have gone in separately, with introduction of SyncAll trait. Please move these and any other changes in the same situation into a separate PR.

agrover:
@mulkieran I've re-pushed the commit w/o the two sync_all changes. It looks like removing all flush()es is a little more work, so I've opened #953 and will tackle that separately.

mulkieran:
I'm waiting until #954 clears.

agrover:
rebased

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -69,7 +114,10 @@ impl BDA {
         f.write_all(&zeroed[..SECTOR_SIZE * 6])?; // Zero 6 unused sectors
         f.sync_all()?;
 
-        let regions = mda::MDARegions::initialize(BDA_STATIC_HDR_SIZE, header.mda_size, f)?;
+        drop(f);
```
tasleson: @agrover The code between the `borrow_mut` and the associated `drop` has many places where code execution could return an error, thus not call `drop`.  So if we have an error in between here, `f` would still be borrowed mut and then we would end up in the error leg that calls initialize where we then call `r_state.restore();` which would cause a run-time error as it tries to do `rc_f.borrow_mut(); or perhaps I'm missing something here?

tasleson: @agrover If what I am outlining here is indeed possible, then utilizing an anonymous function that @mulkieran suggests frequently would be a viable solution.

agrover:
@mulkieran review?

src/engine/strat_engine/backstore/metadata.rs
agrover: We're manually dropping the ref here b/c MDARegions::initialize will borrow it (and re-borrowing is an error) instead of waiting until the end of scope for it to be dropped automatically, but that doesn't affect the places where f will be automatically dropped, such as the error cases you mention. Alternatively we could add another scope around the whole first part of the function that would accomplish the same thing, but I thought this was preferable.

agrover: I should probably go ahead and add a comment there.

agrover:
@tasleson If we find another spot that could use this, then we can make RestoreState usable in both places at that point in time.

agrover:
rebased to add comment for use of manual drop().

mulkieran:
I don't see why it should need Rc at all. I am going to take a deeper look.

mulkieran:
I would say that this is blocked by #978 as well, since the issue has arisen of how precisely to write (and read) the metadata.

mulkieran:
Also blocked by #981.

agrover:
ok, but you can still look for ways to avoid the Rc, right?

agrover:
Out of date and not going to happen for Stratis 1.0. Maybe we can crib some code from this if we decide to tackle this again in the future. Closing.

Issue #822, opened 2018-03-08
Stratis block device cleared with wipefs is unable to be used as stratis device


tasleson:
With the latest changes to `libblkid` we can identify and seemingly clear a device, eg.

```
# ./blkid
/dev/sda1: UUID="cc97b2a8-1f9e-490b-883b-e53da92116b3" TYPE="ext4" PARTUUID="9c220e1f-01"
/dev/sda2: UUID="70f45ac9-1be5-4a75-ab21-f165cd96f759" TYPE="swap" PARTUUID="9c220e1f-02"
/dev/sda3: UUID="565c8028-1774-42c5-b275-c1673f55b8b4" TYPE="ext4" PARTUUID="9c220e1f-03"
/dev/sdb: UUID="a30ef7337e1945498fc9cdafcd0d4b7e" POOL_UUID="ff2a2962d27c489491aa016db808886d" BLOCKDEV_SECTORS="75497472" BLOCKDEV_INITTIME="1520269046" TYPE="stratis"
```

```
# ./wipefs -a /dev/sdb
/dev/sdb: 16 bytes were erased at offset 0x00000204 (stratis): 21 53 74 72 61 30 74 69 73 86 ff 02 5e 41 72 68
```

```
# ./blkid
/dev/sda1: UUID="cc97b2a8-1f9e-490b-883b-e53da92116b3" TYPE="ext4" PARTUUID="9c220e1f-01"
/dev/sda2: UUID="70f45ac9-1be5-4a75-ab21-f165cd96f759" TYPE="swap" PARTUUID="9c220e1f-02"
/dev/sda3: UUID="565c8028-1774-42c5-b275-c1673f55b8b4" TYPE="ext4" PARTUUID="9c220e1f-03"
```

Yet when I try to create a pool using the cleared device I get

```
# stratis pool create some_pool /dev/sdb
Execution failed: 1: Device /dev/sdb appears to belong to another application
```

The output message from `wipefs` indicates that we cleared one copy, maybe not the second.  However, `blkid` fails to recognize the device as a Stratis device even though `blkid` should check both copies.  This could be an issue with `libblkid` failing to recognize second copy as wipefs failed to clear second copy or it actually did clear both copies and Stratis is being very careful to not step on the device.

tasleson:
This was actually a Stratis bug, see: https://github.com/stratis-storage/stratisd/pull/825

tasleson:
With the layout corrected:
```
[...]

PR #936, opened 2018-05-12
Leverage the udev db

tasleson:
With the addition of Stratis support to libblkid, the udev db has
the basic information needed for stratisd.  This change leverages
the udev db where appropriate, to limit when we go to disk to
read up the meta data.

### DIFF
--- a/src/engine/strat_engine/backstore/device.rs
+++ b/src/engine/strat_engine/backstore/device.rs
@@ -5,13 +5,16 @@
 // Functions for dealing with devices.
 
 use std::collections::HashMap;
-use std::fs::File;
+use std::fs::{File, OpenOptions};
 use std::os::unix::prelude::AsRawFd;
 use std::path::Path;
 
 use devicemapper::{devnode_to_devno, Bytes, Device};
-
 use stratis::{ErrorEnum, StratisError, StratisResult};
+
+use super::super::super::types::{DevUuid, PoolUuid};
+use super::metadata::StaticHeader;
+use super::util::get_udev_block_device;
 
 ioctl!(read blkgetsize64 with 0x12, 114; u64);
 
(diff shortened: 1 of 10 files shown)


agrover:
First I have a very basic question I need help understanding: why don't we ever need to call settle in non-test code?

Beyond that, I'd love to see udev settle not require a cmd, if not now then in the future. 

The other big part of this appears to be a move from the DevOwnership struct to a Usage struct, for which having a little more explanation and rationale for would be helpful too.

tasleson:
@agrover 

>First I have a very basic question I need help understanding: why don't we ever need to call settle in non-test code?

Good question, in general we will need too if we do a change and then need to see it reflected in the udev db.

> Beyond that, I'd love to see udev settle not require a cmd, if not now then in the future.

I'll look into this.

> The other big part of this appears to be a move from the DevOwnership struct to a Usage struct, for which having a little more explanation and rationale for would be helpful too.

No real reason, while I was re-factoring I created a new struct while working out the old usage changes.  If you like I could rename it back to it's original name.  The values have changes slightly as I removed the `pool_uuid` and `device_uuid` from `ours` and added a string to `theirs`.

tasleson:
@agrover 

Snippit of udevadm settle
```c
        deadline = now(CLOCK_MONOTONIC) + timeout * USEC_PER_SEC;

        /* guarantee that the udev daemon isn't pre-processing */
        if (getuid() == 0) {
                struct udev_ctrl *uctrl;

                uctrl = udev_ctrl_new(udev);
                if (uctrl != NULL) {
                        if (udev_ctrl_send_ping(uctrl, MAX(5U, timeout)) < 0) {
                                log_debug("no connection to daemon");
                                udev_ctrl_unref(uctrl);
                                return EXIT_SUCCESS;
                        }
                        udev_ctrl_unref(uctrl);
                }
        }

        queue = udev_queue_new(udev);
        if (!queue) {
                log_error("unable to get udev queue");
                return EXIT_FAILURE;
        }

        pfd[0].events = POLLIN;
        pfd[0].fd = udev_queue_get_fd(queue);
        if (pfd[0].fd < 0) {
                log_debug("queue is empty, nothing to watch");
                rc = EXIT_SUCCESS;
                goto out;
        }

        for (;;) {
                if (exists && access(exists, F_OK) >= 0) {
                        rc = EXIT_SUCCESS;
                        break;
                }

                /* exit if queue is empty */
                if (udev_queue_get_queue_is_empty(queue)) {
                        rc = EXIT_SUCCESS;
                        break;
                }

                if (now(CLOCK_MONOTONIC) >= deadline)
                        break;

                /* wake up when queue is empty */
                if (poll(pfd, 1, MSEC_PER_SEC) > 0 && pfd[0].revents & POLLIN)
                        udev_queue_flush(queue);
        }

out:
        udev_queue_unref(queue);
        return rc;
```

The C library exports the `queue` functions, but not the `cntrl` functions.  Furthermore, as stated here: https://www.freedesktop.org/software/systemd/man/libudev.html 

*"... libudev also exports legacy APIs that should not be used by new software (and as such are not documented as part of this manual). This includes the hardware database known as udev_hwdb (please use the new sd-hwdb(3) API instead) and the **udev_queue** object to query the udev daemon (which should not be used by new software at all).*

So even though we could get close to replicating the `settle` functionality, unless we duplicate `cntrl` interface it wouldn't be the same, but according to the docs, the queue functions should be avoided in new software.

I think ideally a settle function should be added to the C library which could be utilized by the `udevadm` command line tool, but I'm not sure how well received that would be.

tasleson:
Re-based

tasleson:
Addresses https://github.com/stratis-storage/stratisd/issues/822

mulkieran:
I'm changing this to review-prelim, because it will definitely need to be squashed, at least.

src/bin/stratisd.rs
```diff
@@ -7,16 +7,16 @@
...
 #![allow(doc_markdown)]
 
-extern crate devicemapper;
-extern crate libstratis;
-#[macro_use]
-extern crate log;
 extern crate clap;
 #[cfg(feature = "dbus_enabled")]
 extern crate dbus;
+extern crate devicemapper;
```
mulkieran: All this is just a change of ordering....perhaps rustfmt did it. I would think that either it is a mistake or it belongs in a separate PR.

mati865: New rustfmt does sort `extern`, `use` and some other things.

tasleson: Those are `rustfmt` changes

mulkieran: That's what I figured, somehow. But I don't think it belongs in this PR if it is just a re-ordering.

agrover: What happened here? Did our mass rustfmt conversion PR miss this file?

Can we just merge this and get on with it?

tasleson: I think what happened is a difference in rustfmt versions.  This PR has spanned the change in versions and I think an older version did this as I reverted this change and the current rustfmt did not mess with it

tasleson:
@mulkieran In a previous comment you requested I squash, all the changes were individual commits before.

mulkieran:
I'm sorry you misunderstood me. All I really meant is that with those typo fixes and miscellaneous bits at the end, it really shouldn't be designated as review-final.

tasleson:
I've rebased and removed the binary handling stuff I originally had in here.

tasleson:
Re-based

tasleson:
I would really like to know how what's in `master` branch ran CI clean with the same test code that is somehow  failing to compile today.

Working: https://travis-ci.org/stratis-storage/stratisd/jobs/393085873

Failing today:
```
error[E0599]: no method named `write_all` found for type `std::io::Cursor<std::vec::Vec<u8>>` in the current scope
    --> src/engine/strat_engine/backstore/metadata.rs:1237:23
     |
1237 |             buf_newer.write_all(&sector).unwrap();
     |                       ^^^^^^^^^
     |
     = help: items from traits can only be used if the trait is in scope
     = note: the following trait is implemented but not in scope, perhaps add a `use` for it:
             candidate #1: `use std::io::Write;`

```

mati865:
Looks pretty odd but should build after you add `use std::io::Write;` to `src/engine/strat_engine/backstore/metadata.rs`.

tasleson:
@mati865 Thanks!  I'm not sure, but for some reason when I did the re-base, the code went from (`metadata.rs`):

```rust
#[cfg(test)]
mod tests {
    use std::io::{Cursor, Write};
```
to

```rust
#[cfg(test)]
mod tests {
    use std::io::Cursor;
```

src/engine/strat_engine/backstore/device.rs
```diff
@@ -43,3 +46,159 @@ pub fn resolve_devices<'a>(paths: &'a [&Path]) -> StratisResult<HashMap<Device,
...
+/// blivet library.
+fn empty(device: &HashMap<String, String>) -> bool {
+    if (device.contains_key("ID_PART_TABLE_TYPE") && !device.contains_key("ID_PART_ENTRY_DISK"))
+        || device.contains_key("ID_FS_USAGE")
+    {
+        false
+    } else {
+        true
+    }
+}
```
agrover: another option:
```
    !((device.contains_key("ID_PART_TABLE_TYPE") && !device.contains_key("ID_PART_ENTRY_DISK"))
        || device.contains_key("ID_FS_USAGE"))
```
might be better? OR go the other way with:
```
    if device.contains_key("ID_PART_TABLE_TYPE") {
        if device.contains_key("ID_PART_ENTRY_DISK") {
            true
        } else {
            false
        }
    } else if device.contains_key("ID_FS_USAGE") {
        false
    } else {
        true
    }
```
might actually be best. Doing a little of both seems odd, though.

src/engine/strat_engine/backstore/device.rs
```diff
@@ -43,3 +46,159 @@ pub fn resolve_devices<'a>(paths: &'a [&Path]) -> StratisResult<HashMap<Device,
...
+        } else {
+            Ok(DevOwnership::Theirs(signature(&device)))
+        }
+    } else {
+        Err(StratisError::Engine(
+            ErrorEnum::NotFound,
+            String::from(format!(
+                "We expected to find the block device {:?} \
+                 in the udev db",
+                devnode
```
agrover: With our current formatting, I think this string can all fit on one line.

src/engine/strat_engine/backstore/device.rs
mulkieran: I prefer first suggested option.

tasleson: Done

src/engine/strat_engine/backstore/device.rs
tasleson: Done

tasleson:
Superseded by https://github.com/stratis-storage/stratisd/pull/999

PR #935, opened 2018-05-10
Attach setting blockdev id to Pool trait

mulkieran:
Related: #614.

PR #940, opened 2018-05-16
Have method in StratEngine implementation return StratBlockDevs

mulkieran:
Previously, these methods returned BlockDevs. At the internal implementation
level, it makes more sense to return objects of the concrete type, rather
than trait objects.

This means that, instead of being cast to BlockDev at the lowest level, these
references are cast to BlockDev at the highest level, i.e., in the impl
of the Pool trait for StratPool.


mulkieran:
Blocks #935 .

agrover:
Yeah this is fine. I think the code was this way because of the previous move to tuck stuff inside Backstore, so this gets us back to where we want to be.

PR #951, opened 2018-05-21
Add static header size to test buffer used for BDA::initialize()

agrover:
sh.mda_size does not include the static header, BDA::initialize() needs
a buffer that is big enough for the static header as well as the defined
MDA regions.


agrover:
blocks #931

PR #954, opened 2018-05-23
Remove some remaining flush() calls in favor of sync_all()

### LINKED ISSUE #953: Remove all uses of flush() in favor of sync_all()
Flush does nothing, whereas sync_all does.

agrover:
flush() does nothing, so we should sync_all() any time we want to ensure
something is saved to permanent media.

Make SyncAll public from mod metadata.

Implement SyncAll for BufWriter.

Don't need nix call to fsync() in MDV code any more, that's what sync_all
does.

fixes #953


agrover:
cool, yeah I was also thinking that might be good.

tasleson:
LGTM

PR #955, opened 2018-05-23
Dbus alerts proxy

trgill:


src/dbus_api/filesystem.rs
```diff
@@ -98,26 +98,29 @@ fn rename_filesystem(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
...
-        Ok(RenameAction::Identity) => {
-            return_message.append3(default_return, msg_code_ok(), msg_string_ok())
-        }
-        Ok(RenameAction::Renamed) => return_message.append3(true, msg_code_ok(), msg_string_ok()),
+        Ok(RenameAction::Identity) => vec![
+            return_message.append3(default_return, msg_code_ok(), msg_string_ok()),
+        ],
+        Ok(RenameAction::Renamed) => vec![
+            return_message.append3(true, msg_code_ok(), msg_string_ok()),
+            build_propchanged("Name", new_name, object_path),
```
mulkieran: This line is the only real change. The rest is just a refactoring.

src/dbus_api/util.rs
```diff
@@ -99,3 +102,14 @@ pub fn get_parent(i: &mut IterAppend, p: &PropInfo<MTFn<TData>, TData>) -> Resul
     i.append(data.parent.clone());
     Ok(())
 }
+
+/// Construct a signal that a property has changed.
+// Currently only handles returning string properties, since that's all that's
+// needed right now
+pub fn build_propchanged(prop_name: &str, new_value: &str, path: &dbus::Path) -> Message {
```
mulkieran: Might make sense to make ```new_value``` have type ```ToOwned```, but this is a minor detail.

mulkieran:
Ok!

So here are the two most important bits, I think.

1. As this PR implements, the engine _must_ have access to the D-Bus connection. This PR uses a call-back. It _might_ be better to make the D-Bus connection a global, Rust style. There are concerns about borrowing either way, I think, but I haven't tried to look at it.
2. Again, as this PR implements, I think it is unavoidable that the pool must hold its D-Bus object path. The way it is set, after the pool is first created, also seems unavoidable to me. But I think that it will also be necessary to set object paths on BlockDev's and Filesystems, and that for consistency, it makes sense to do it as part of this change, rather than to defer it.

I think the rest is detail.

trgill:
@mulkieran could you please explain your thoughts about how this should be split in two?  Not sure I followed.

trgill:
@mulkieran you suggested we split up the work in this PR into 3:

1. Make the dbus connection global
2. Attach dbus object path to engine object
3. Add mechanisms for making needed call back

Just checking to make sure I understand.

mulkieran:
Not quite:

If I understand (2) correctly, I think what you're saying is that any engine object that has a corresponding D-Bus object must get an object path attribute and a way to set the object path from the D-Bus layer after it, the engine object, is created. That is definitely correct.

For (1), what I said is that you really have two possible approaches, and they should be checked out separately. The first is to make the D-Bus connection global, the second is that you pass around a proxy object which holds the D-Bus connection. These are two alternatives to solve the same problem: allowing the engine to put messages on the D-Bus connection.

Finally, and easiest to get wrong, is the step of actually hooking up the two previous steps to get the right messages on the D-Bus under the right circumstances.

(2) is probably easiest, and it probably makes sense to tackle that first, so long as everybody concedes that it is necessary.

agrover:
One way is to make the dbus connection have shared ownership, via our friends(?) `Rc` and `RefCell`. That loosens the reins a bit to make possible what we need to do, but still keeps ownership in the picture, in a way that doing what we did for the dm context (I assume this is what is meant by making it global) doesn't. I think making it shared was what some early version of this PR that I worked on did a long time ago. What are the tradeoffs of that approach versus what we're calling the "global" approach?

PR #960, opened 2018-05-29
Fix test compile errors

tasleson:
Not sure how this happened, but master testing does not compile.


tasleson:
This is required for rustc 1.25.0

PR #961, opened 2018-05-30
fix travis fmt by pinning to 1.26.0

agrover:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,7 @@ fmt:
 	cargo fmt
 
 fmt-travis:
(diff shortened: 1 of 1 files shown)


agrover:
1.26.2 just came out and bumped rustfmt to 0.4.2 from 0.4.1, which apparently changed something.

I think we either need to re-pin or we need a specified rustfmt config that enforces things across rustfmt updates, since the default seems to change sometimes.

agrover:
since this fixes CI (Travis, at least), I'm going to go ahead and push this without review.

PR #966, opened 2018-05-30
Add a test for thin pool transition back to "Good" when space is added

trgill:
Create a test that:

	Fills thin pool - verify it is handled/reported correctly
 	Verify when space is added the pool state transitions back to Good

### DIFF
--- a/src/engine/strat_engine/thinpool/thinpool.rs
+++ b/src/engine/strat_engine/thinpool/thinpool.rs
@@ -871,7 +871,7 @@ fn attempt_thin_repair(
 
 #[cfg(test)]
 mod tests {
-    use std::fs::OpenOptions;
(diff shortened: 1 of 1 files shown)


mulkieran:
All unit tests are passing...

agrover:
CI failures are not this PR's fault. I think it can be merged.

trgill:
@mulkieran I think when sync_data() is called - all IO queued but not on the device will fail.

We've got a commit (not released) in devicemapper-rs to set the thin pool to error_if_no_space to minimize at risk IO.  But even with that set we can have /sys/block/<stratis_name>/queue/nr_requests queued for write.

The application would need to re-try the IO after the pool was restored to writeable.

trgill:
@mulkieran this ok to merge?

PR #967, opened 2018-05-30
Build against the lowest rustc version we claim to support

agrover:


tasleson:
The only thing this doesn't address is the fact that we cannot run all the unit tests on travis, but it's a step in the right direction.

Issue #853, opened 2018-03-20
Explicitly fail to run if not root


agrover:
As suggested by @tasleson,

Currently, if one runs stratisd without root permissions, something will undoubtedly fail and return an error instead of running successfully. But that error message could be something nonintuitive, like a DBus or DM error message.

Maybe we should explicitly check and fail with an appropriate error message instead.

agrover:
Wow, I tested this manually... Looks like I should've finished the job properly and added an actual test.

This and the --help thing, should they be opened as new bugs?

mulkieran:
Regarding the original issue...perhaps it is part of something that could be solved just as well by generic error-handling.

PR #968, opened 2018-05-30
Ask user to run stratisd as root if it cannot access lock file

mati865:
Signed-off-by: Mateusz Mikuła <mati865@gmail.com>

Fixes https://github.com/stratis-storage/stratisd/issues/853

I think it takes root on every distro to create a file in `/run/`.

### DIFF
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -22,7 +22,7 @@ extern crate nix;
 use std::cell::RefCell;
 use std::env;
 use std::fs::{File, OpenOptions};
-use std::io::{Read, Write};
+use std::io::{ErrorKind, Read, Write};
 use std::os::unix::io::AsRawFd;
 use std::path::PathBuf;
 use std::process::exit;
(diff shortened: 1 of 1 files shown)


src/bin/stratisd.rs
```diff
@@ -83,11 +83,22 @@ fn handle_udev_add(event: &libudev::Event) -> Option<(Device, PathBuf)> {
...
-        .open(STRATISD_PID_PATH)?;
+        .open(STRATISD_PID_PATH)
+    {
+        Ok(f) => f,
+        Err(ref e) if e.kind() == ErrorKind::PermissionDenied => {
+            println!(
+                "Permission denied while accessing lock file.\n\
+                 Please make sure to run this command as root."
+            );
+            exit(1);
```
mulkieran: Instead of exiting please return a ```StratisError``` with the text of the error message. The caller will immediately exit w/ that error, see below.

agrover:
I seem to remember a discussion where somebody mentioned that it would be better to have an explicit early check for uid 0 that would error out with a message like this, instead of using the first "thing that actually needs root privs", because what if the code changes and the first thing is no longer first?

agrover:
ok to test

mati865:
@agrover I initially wanted to do it that way but later on I found https://stackoverflow.com/questions/4159910/check-if-user-is-root-in-c
Since I'm far from being expert in POSIX I just did what was suggested on SO.

agrover:
add to whitelist

agrover:
hah it was you! Ok, I'm :+1:

src/bin/stratisd.rs
```diff
@@ -83,11 +83,20 @@ fn handle_udev_add(event: &libudev::Event) -> Option<(Device, PathBuf)> {
...
         .read(true)
         .write(true)
         .create(true)
-        .open(STRATISD_PID_PATH)?;
+        .open(STRATISD_PID_PATH)
+    {
+        Ok(f) => f,
+        Err(e) => {
+            if e.kind() == ErrorKind::PermissionDenied {
+                eprintln!("Please make sure to run stratisd as root.");
```
mulkieran: Please just return the string in the error, as:

```return Err(StratisError::Error("Must be running as root in order to start daemon."));```

mati865: Sorry, I tried that with `Into` trait before.
Should have better checked `StratisError`.

tasleson:
@mati865 Thanks!

Issue #861, opened 2018-03-26
making dbus calls doesn't appear to require root permissions


agrover:
Invoking the cli as a normal user to e.g. create a pool is working, when it should not work due to insufficient permissions.

tasleson:
We need something like: https://sourceware.org/git/?p=lvm2.git;a=blob_plain;f=scripts/com.redhat.lvmdbus1.conf;hb=refs/heads/master

tasleson:
Correction merged

PR #970, opened 2018-05-31
Re-work dbus permissions

tasleson:
Change the permissions so root can do anything and non-root user
can use the following interfaces:

org.freedesktop.DBus.ObjectManager
org.freedesktop.DBus.Introspectable
org.freedesktop.DBus.Properties

On each of the provided interfaces by stratisd.  This change
allows the stratis CLI or any dbus client to query the system, but
not make changes.

### DIFF
--- a/stratisd.conf
+++ b/stratisd.conf
@@ -4,10 +4,50 @@
 <busconfig>
 <policy user="root">
   <allow own="org.storage.stratis1"/>
+  <allow send_destination="org.storage.stratis1"/>
 </policy>
 <policy context="default">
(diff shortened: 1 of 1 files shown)


tasleson:
This should address: https://github.com/stratis-storage/stratisd/issues/861

agrover:
LGTM!

tasleson:
AFAIK all our Properties only allowed read access, however I've updated the change to only allow `Get` & `GetAll` on the properties interface.

tasleson:
Squished

mulkieran:
AFAIC, this can go in at any time.

Issue #530, opened 2017-08-10
Check CLI versions if there is a known minimum version


trgill:
Although we try to avoid invoking CLI tools from Stratis, there are some cases it is needed (XFS, thin provisioning).

Given we've got a known problem with "thin_check" version 2.8 or lower - should we adopt a strategy for a baseline on CLIs?

agrover:
If we document a minimum required version (which we should, prominently), this is the sort of thing that distro packaging scripts would typically enforce.

But, I still am of the opinion this is still worth doing, since developers and do-it-yourselfers won't be using a distro-packaged Stratis, and the consequences for using a too-old version are that it certainly won't be a tested configuration, and we can't guarantee there won't be data loss.

mulkieran:
Our first step should be to reduce the places where we use Command to a very small set, even if it can't be empty.

mulkieran:
Does libblockdev allow client to specify the version of a command-line tool that libblockdev wraps that it requires? Most likely not, although I haven't checked.

PR #974, opened 2018-06-01
Have strat engine verify existence of required binaries

mulkieran:
Related: #530

This is adapted from work in #936. My principle concern in this is to avoid a panic while the engine is executing if there is a missing binary.


mulkieran:
This is really fraught with defects, which I will try to at least enumerate next week.

mulkieran:
rebased.

mulkieran:
This behaves identically to the ```get_dm()``` approach. Using ```lazy_static``` is a bit less verbose, so it makes sense to continue to use it. Can we use ```lazy_static``` for ```get_dm()``` then? It turns out that we can not, because ```error::Error``` is not send. I think that the failure crate solves this problem, although I am not sure.

tasleson:
existance -> existence

src/engine/strat_engine/cmd.rs
```diff
@@ -0,0 +1,155 @@
...
+            return Err(StratisError::Engine(ErrorEnum::NotFound, name.to_string()));
+        }
+    }
+    Ok(())
+}
+
+/// Common function to call a command line utility, returning an Result with an error message which
+/// also includes stdout & stderr if it fails.
+fn execute_cmd(cmd: &mut Command, error_msg: &str) -> StratisResult<()> {
+    let result = cmd.output()?;
```
tasleson: I think is would be better to remove the `exists` check from `get_executable` and place it in the error path for `cmd.output()` so that we can than either raise the missing file error or return the error returned from trying to execute the command.

src/engine/strat_engine/engine.rs
```diff
@@ -52,8 +53,10 @@ impl StratEngine {
     ///
     /// Returns an error if the kernel doesn't support required DM features.
     /// Returns an error if there was an error reading device nodes.
+    /// Returns an error if the binaries on which it depends can not be found.
     pub fn initialize() -> StratisResult<StratEngine> {
         let dm = get_dm_init()?;
+        verify_binaries()?;
```
tasleson: If for some reason a user has an existing stratis setup, they update their system and during the update they mess up the required binaries the stratisd daemon will exit with error.  If we aren't running we cannot monitor the thin pool(s) for out of space and other errors.  Having systemd unit file instructing us to restart us isn't going to help as we will just exit until systemd gives up.  I'm wondering if not exiting here would be better?  This is much better than my proposed PR where we just panic.

src/engine/strat_engine/cmd.rs
```diff
@@ -0,0 +1,155 @@
...
+// and vice-versa.
+const MKFS_XFS: &str = "mkfs.xfs";
+const THIN_CHECK: &str = "thin_check";
+const THIN_REPAIR: &str = "thin_repair";
+const XFS_ADMIN: &str = "xfs_admin";
+const XFS_GROWFS: &str = "xfs_growfs";
+
+lazy_static! {
+    static ref BINARIES: HashMap<String, Option<PathBuf>> = [
+        (MKFS_XFS.to_string(), find_binary(MKFS_XFS)),
```
tasleson: For required binaries that are only needed during test what are you suggesting we do?

mulkieran:
Note: I looked at a tiny Rust library called quale, which claims it implements the functionality of the "which" command. That's a whole different way of finding an executable. Is it better than our way or hard-coding 4 directories, or wrong?

tasleson:
> Note: I looked at a tiny Rust library called quale, which claims it implements the functionality of the "which" command. That's a whole different way of finding an executable. Is it better than our way or hard-coding 4 directories, or wrong?

The `which` command searches the environment PATH which won't exist in our use case, so I don't think it's better in that it likely won't work, but I've not looked at the source code for quale.

src/engine/strat_engine/cmd.rs
```diff
@@ -0,0 +1,161 @@
...
+fn get_executable(name: &str) -> StratisResult<PathBuf> {
+    let executable = BINARIES
+        .get(name)
+        .expect("name arguments are all constants defined with BINARIES, lookup can not fail")
+        .as_ref()
+        .expect("verify_binaries() was previously called and returned no error");
+    if !executable.exists() {
+        return Err(StratisError::Error(format!(
+            "Executable previously located at \"{}\" seems to have been removed since stratisd was started",
+            executable.to_str().expect("All parts of path are constructed from stricly ASCII components")
```
mulkieran: Whoops, there is a typo.

src/engine/strat_engine/cmd.rs
tasleson: My point was that instead of having to `stat` the executable every time to see if it exists before it's called we could just assume that the executable is still there, which it likely is and only if we encounter an error trying to execute it, see if it had gone missing.  Thus the error message would be more helpful.  The end result should be the same, the difference is when the existence check is done, either every time an executable is called or only when we encounter an error.

mulkieran:
My feeling about this is that we shouldn't have to be doing this in the first place. About 50% of the reason that Project Springfield exists is so that this kind of thing can go away. In the meantime, since we have to do it, we should focus more than anything on the quarantine aspect of software engineering, the part where we just make sure that the bad stuff is organized into one place, and not spread all over our application. I think that we're managing that pretty well. However, we shouldn't go about refining it to the nth degree. We should focus on the stuff that makes Stratis interesting and make that very good. I'm going to answer all questions with that in mind.

There is, however, another consideration. It is in the nature of the cmd module to have relatively few dependencies and a fairly simple structure. It is nicely isolated from the rest of stratisd and should remain so. That means that it is likely to be a good object for testing the failure crate. In theory, it should be possible to use the failure crate in this one module and have no difficulty in interacting with the rest of stratisd.

Assuming I give this a try, I will try to make use of principles I have been trying to come up with when working on error handling in the CLI.

One principle is that methods should only take the arguments they need for execution. That is, they should never require additional arguments that are only used for error reporting. If a method fails, then the context in which it failed should be enough to fill in any necessary contextual information.

Another principle is that string-munging should not be used to add additional information to an error. The only way to get that information back out of the error programmatically is via string parsing, which is grim and chronically broken.

I think that these two are pretty obvious.

There is a third bit, which is more tricky. In a previous life, I used to write code to generate error messages from the attributes of a Python exception. I've come to the conclusion that that just leads to really bad prose. Now, I generate the message at the site where the exception is raised. I may also pass a bunch of arguments to the exception, some of which may also appear in some form in the string exception. This may be duplication, but it is, I believe, worth it. I'm going to try to do that here.

src/engine/strat_engine/cmd.rs
mulkieran: * Make a separate HashMap to store the names and absolute paths of these test-only binaries.
* Don't bother to write a ```verify_binaries``` method. We have a deliberate policy of unwrapping all over in tests, that would be consistent with this policy.
* Add a method ```get_test_executable``` which has fewer ```expects()```s and more ```unwraps()``` than ```get_executable``` consistent w/ the way tests are run. Use ```get_test_executable()``` when getting executables in tests.

src/engine/strat_engine/engine.rs
mulkieran: If they just moved the binaries to a place where we expect, stratisd will find them again and all will be well. I don't see why this should happen, but who knows?

If some of the binaries that we rely on are actually gone, then stratisd can not function properly, anymore than it could function properly if devicemapper context is missing. We might as well exit.

In a future, we could work to modify stratisd so that it runs in some kind of crippled mode. But that would be complex. Maybe the future that comes sooner is that all or most of these become available as libraries! That would be marvelous.

src/engine/strat_engine/cmd.rs
mulkieran: This would require some significant alterations to existing code for very moderate gain. I don't think it's worth doing unless this PR is used as a test case for the failure crate.

mulkieran:
In stand-up, the point was made that only the dm executables are available in initramfs. So, with these changes, stratisd would not run properly when that's all that's available.

How would one possibly manage this? Well, we have to assume that if the dm executables are the only ones available, then those are the only ones that will be used. Can we be certain that that is true? We really ought to find out.

But, in the meantime, it was suggested that it is reasonable to go ahead w/ this patch without addressing that difficult issue, which will probably force some changes in code that is quite unrelated to this code. I believe that we should do this.

src/engine/strat_engine/cmd.rs
mulkieran: The suggested change makes the code behave more like the usual exception handling model, which would be an improvement. However, I would still like to merge this as is, and then experiment with the failure crate which I hope will allow me to implement the suggestion in a better manner than could be managed at this time.

src/engine/strat_engine/cmd.rs
mulkieran: Also, as much of the code that I outlined above should be modified with the ```#cfg[test]``` annotation.

mulkieran:
I've put up the final version, so I'm closing this.

PR #975, opened 2018-06-04
try to add some more cross-platform testing

agrover:


agrover:
ubuntu doesn't build pkgs for more than x86 and x86-64, so it's not possible to easily extrapolate from building on x86 to building on less common archs.

PR #976, opened 2018-06-04
Avoid unwrap() in get_dm() method

mulkieran:
It seemed that it was necessary to hav an unwrap() in get_dm() because it
seemed that it was necessary to have get_dm() return a simple type, i.e.,
one without an Option or Result. The problem with this approach is that
it required the use of catch_unwind() at the point that get_dm() was
first called, to ensure an orderly exit.

This solution is exactly equivalent, but leaves out the unwind() and
hence the requirement to use catch_unwrap().

It changes get_dm() so that it returns a StratisResult, and it uses
the macro get_dm in every case where get_dm() is expected to succeed.

The solution is semantically equivalent to the previous approach; under
all circumstances it should behave in an equivalent manner.

### DIFF
--- a/src/engine/strat_engine/dm.rs
+++ b/src/engine/strat_engine/dm.rs
@@ -7,23 +7,35 @@
 use std::os::unix::io::{AsRawFd, RawFd};
 use std::sync::{Once, ONCE_INIT};
 
-use devicemapper::DM;
+use devicemapper::{DmResult, DM};
 
-use stratis::StratisResult;
+use stratis::{ErrorEnum, StratisError, StratisResult};
 
 use super::super::engine::Eventable;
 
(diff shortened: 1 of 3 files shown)


mulkieran:
This is related to #974. Is it desirable to use a virtually identical method to locate external binaries?

agrover:
Do we need the macro?

So with this PR we'd have `get_dm()`, used during init, and `get_dm!()`, used all the other times. Could this be `get_dm_init()` used during init, and then `get_dm()` used all the other times?

mulkieran:
Macro-less version.

mulkieran:
Now with comment fix!

mulkieran:
I've answered the only request, so I'm calling it review final.

agrover:
👍

Issue #210, opened 2017-01-20
Repair primary sigblock or metadata copies if bad and secondary is ok


agrover:
We have two copies of both the sigblock in the static header, and the metadata also stores two copies in 2 MDA regions. If the first one is bad (CRC fails), we can read the second one. Should we then go back and rewrite the bad one immediately with the secondary copy?

This sounds like a good, "self-healing" kinda thing to do, but 1) it's work and 2) we probably want to think about this, maybe run it by others, to make sure this is really a good thing to do and not too clever by half.

tasleson:
Related to: https://github.com/stratis-storage/stratisd/issues/826

Issue #826, opened 2018-03-08
BDA unit test: Corrupt & zero all combinations of disk layout and copies and test for expected outcome(s)


tasleson:
We discovered fairly late in the development process that the BDA layout was incorrect.  This was found when integrating with the external tool `libblkid` where we were not locating the stratis signature for the second copy.  After code inspection the issue was found and corrected: https://github.com/stratis-storage/stratisd/pull/825

The existing Stratis code is correctly reading the location of the second copy, the issue was that the second copy wasn't in the correct location.  Thus, if the first copy didn't pass all the checks the second copy would have offered no additional value.

To ensure that the on disk layout is correct and working as designed it would be good to test the following scenarios:

1. Both BDA copies good => Result in good BDA retrieval
2. Corrupt only primary copy (incorrect checksum) => Result in good BDA retrieval
3. Zero only primary copy => Result in good BDA retrieval
4. Corrupt only secondary copy  (incorrect checksum) => Result in good BDA retrieval
5. Zero only secondary copy => Result in good BDA retrieval
6. Corrupt both copies => Failure to retrieve BDA
7. Zero both copies => Failure to retrieve BDA
8. Corrupt Primary, Zero secondary => Failure to retrieve BDA (And vice versa)

It would also be good to validate that the reserved areas are zero.

Additionally we should have the [discussion](https://github.com/stratis-storage/stratisd/issues/210) on whether or not we want to correct corrupt or missing BDA when we discover it, to avoid the cases where we only have one copy and no backup.

tasleson:
Related to https://github.com/stratis-storage/stratisd/issues/210

mulkieran:
We agreed a while back that issues should be in "in progress" (unless we're being asked to review the issue, I guess), so I'm moving this to "in progress" column.

PR #860, opened 2018-03-23
Have StaticHeader::setup return the buffer it examined

mulkieran:
This way, it's not necessary to duplicate its body in
determine_ownership().

Signed-off-by: mulhern <amulhern@redhat.com>

PR #978, opened 2018-06-05
Verify both copies of BDA, correct if we can

tasleson:
When we read up BDA from a device we will verify both copies and correct in
those cases where we have one good copy and one invalid copy.  This change also
corrects a bug where we would return no Stratis BDA if the magic on the first
copy contains an error even though we have a complete and valid second copy.

Partially resolves: #210 , this doesn't address the MDA, just the BDA

### DIFF
--- a/src/engine/strat_engine/backstore/metadata.rs
+++ b/src/engine/strat_engine/backstore/metadata.rs
@@ -1160,4 +1160,132 @@ mod tests {
             .tests(30)
             .quickcheck(static_header as fn(u64, u32) -> TestResult);
     }
+
+    #[test]
+    /// Verify that we correctly handle reading up the BDA when we walk a byte of corruption through
+    /// the BDA sector for each copy and correct the corrupt copy and return an error when
+    /// we corrupt both copies.
+    fn bda_test_recovery() {
+        fn corrupt_byte<F>(f: &mut F, position: u64) -> io::Result<()>
+        where
+            F: Read + Seek + SyncAll,
+        {
+            let mut byte_to_corrupt = [0; 1];
+            f.seek(SeekFrom::Start(position))?;
+            f.read(&mut byte_to_corrupt)?;
+            byte_to_corrupt[0] = !byte_to_corrupt[0];
+            f.seek(SeekFrom::Start(position))?;
+            f.write(&byte_to_corrupt)?;
+            f.sync_all()?;
+            Ok(())
+        }
+
+        let sh = random_static_header(10000, 4);
+        let buf_size = *sh.mda_size.bytes() as usize + _BDA_STATIC_HDR_SIZE;
+        let mut buf = Cursor::new(vec![0; buf_size]);
+        BDA::initialize(
+            &mut buf,
+            sh.pool_uuid,
+            sh.dev_uuid,
+            sh.mda_size,
+            sh.blkdev_size,
+            Utc::now().timestamp() as u64,
+        ).unwrap();
+
+        let reference_buf = buf.clone();
+
+        for i in 0..SECTOR_SIZE {
+            for primary in &[true, false] {
+                for secondary in &[true, false] {
+                    if !*primary {
+                        // Corrupt primary copy
+                        corrupt_byte(&mut buf, (1 * SECTOR_SIZE + i) as u64).unwrap();
+                    }
+
+                    if !*secondary {
+                        // Corrupt secondary copy
+                        corrupt_byte(&mut buf, (9 * SECTOR_SIZE + i) as u64).unwrap();
+                    }
+
+                    let setup_result = StaticHeader::setup(&mut buf);
+
+                    if *primary || *secondary {
+                        // Setup should work and buffer should be corrected
+                        assert!(setup_result.is_ok() && setup_result.unwrap().is_some());
+
+                        // Check buffer, should be corrected.
+                        assert_eq!(reference_buf.get_ref(), buf.get_ref());
+                    } else {
+                        // Setup should fail to find a usable Stratis BDA
+                        match i {
+                            4...19 => {
+                                // When we corrupt the magics then we believe that the signature is
+                                // not ours and will return Ok(None)
+                                assert!(setup_result.is_ok() && setup_result.unwrap().is_none());
+                            }
+                            _ => {
+                                assert!(setup_result.is_err());
+                            }
+                        }
+
+                        // Check buffer, should be different
+                        {
+                            assert_ne!(reference_buf.get_ref(), buf.get_ref());
+                        }
+
+                        // Reset the buffer as we didn't correct it
+                        buf = reference_buf.clone();
+                    }
+                }
+            }
+        }
+    }
+
+    #[test]
+    /// Test that we re-write the older of two BDAs if they don't match.
+    fn bda_test_rewrite_older() {
+        let sh = random_static_header(10000, 4);
+        let buf_size = *sh.mda_size.bytes() as usize + _BDA_STATIC_HDR_SIZE;
+        let mut buf = Cursor::new(vec![0; buf_size]);
+        let ts = Utc::now().timestamp() as u64;
+
+        BDA::initialize(
+            &mut buf,
+            sh.pool_uuid,
+            sh.dev_uuid,
+            sh.mda_size,
+            sh.blkdev_size,
+            ts,
+        ).unwrap();
+
+        let mut buf_newer = Cursor::new(vec![0; buf_size]);
+        BDA::initialize(
+            &mut buf_newer,
+            sh.pool_uuid,
+            sh.dev_uuid,
+            sh.mda_size,
+            sh.blkdev_size,
+            ts + 1,
+        ).unwrap();
+
+        // We should always match this reference buffer as it's the newer one.
+        let reference_buf = buf_newer.clone();
+
+        for offset in &[SECTOR_SIZE, 9 * SECTOR_SIZE] {
+            // Copy the older BDA to newer BDA buffer
+            buf.seek(SeekFrom::Start(*offset as u64)).unwrap();
+            buf_newer.seek(SeekFrom::Start(*offset as u64)).unwrap();
+            let mut sector = [0u8; SECTOR_SIZE];
+            buf.read_exact(&mut sector).unwrap();
+            buf_newer.write_all(&sector).unwrap();
+
+            assert_ne!(reference_buf.get_ref(), buf_newer.get_ref());
+
+            let setup_result = StaticHeader::setup(&mut buf_newer);
+            assert!(setup_result.is_ok() && setup_result.unwrap().is_some());
+
+            // We should match the reference buffer
+            assert_eq!(reference_buf.get_ref(), buf_newer.get_ref());
+        }
+    }
 }
--- a/src/engine/strat_engine/backstore/metadata.rs
+++ b/src/engine/strat_engine/backstore/metadata.rs
@@ -2,7 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-use std::io::{Read, Seek, SeekFrom};
+use std::io::{self, Read, Seek, SeekFrom};
 use std::str::from_utf8;
 
 use byteorder::{ByteOrder, LittleEndian};
@@ -36,7 +36,65 @@ pub struct BDA {
     regions: mda::MDARegions,
 }
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+enum MetadataLocation {
+    Both,
+    First,
+    Second,
+}
+
 impl BDA {
+    /// Read the BDA from the device and returning a byte array of the entire area.
+    fn read<F>(f: &mut F) -> io::Result<[u8; _BDA_STATIC_HDR_SIZE]>
+    where
+        F: Read + Seek,
+    {
+        #![allow(unused_io_amount)]
+        f.seek(SeekFrom::Start(0))?;
+        let mut buf = [0u8; _BDA_STATIC_HDR_SIZE];
+
+        // FIXME: See https://github.com/stratis-storage/stratisd/pull/476
+        f.read(&mut buf)?;
+        Ok(buf)
+    }
+
+    // Writes bda_buf according to the value of which.
+    // If first location is specified, write zeroes to empty regions in the
+    // first 8 sectors. If the second location is specified, writes zeroes to empty
+    // regions in the second 8 sectors.
+    fn write<F>(f: &mut F, bda_buf: &[u8], which: MetadataLocation) -> io::Result<()>
+    where
+        F: Seek + SyncAll,
+    {
+        let zeroed = [0u8; _BDA_STATIC_HDR_SIZE];
+        f.seek(SeekFrom::Start(0))?;
+
+        // Write to a single region in the header. Zeroes the first sector,
+        // writes bda_buf to the second sector, and then zeroes the remaining
+        // six sectors.
+        fn write_region<F>(f: &mut F, bda_buf: &[u8], zeroed: &[u8]) -> io::Result<()>
+        where
+            F: Seek + SyncAll,
+        {
+            f.write_all(&zeroed[..SECTOR_SIZE])?; // Zero 1 unused sector
+            f.write_all(bda_buf)?;
+            f.write_all(&zeroed[..SECTOR_SIZE * 6])?; // Zero 6 unused sectors
+            f.sync_all()?;
+            Ok(())
+        };
+
+        if which == MetadataLocation::Both || which == MetadataLocation::First {
+            write_region(f, bda_buf, &zeroed)?;
+        } else {
+            f.seek(SeekFrom::Start(8 * SECTOR_SIZE as u64))?;
+        }
+
+        if which == MetadataLocation::Both || which == MetadataLocation::Second {
+            write_region(f, bda_buf, &zeroed)?;
+        }
+        Ok(())
+    }
+
     /// Initialize a blockdev with a Stratis BDA.
     pub fn initialize<F>(
         f: &mut F,
@@ -49,7 +107,6 @@ impl BDA {
     where
         F: Seek + SyncAll,
     {
-        let zeroed = [0u8; _BDA_STATIC_HDR_SIZE];
         let header = StaticHeader::new(
             pool_uuid,
             dev_uuid,
@@ -57,17 +114,8 @@ impl BDA {
             blkdev_size,
             initialization_time,
         );
-        let hdr_buf = header.sigblock_to_buf();
 
-        // Write 8K header. Static_Header copies go in sectors 1 and 9.
-        f.seek(SeekFrom::Start(0))?;
-        f.write_all(&zeroed[..SECTOR_SIZE])?; // Zero 1 unused sector
-        f.write_all(&hdr_buf)?;
-        f.write_all(&zeroed[..SECTOR_SIZE * 7])?; // Zero 7 unused sectors
-        f.sync_all()?;
-        f.write_all(&hdr_buf)?;
-        f.write_all(&zeroed[..SECTOR_SIZE * 6])?; // Zero 6 unused sectors
-        f.sync_all()?;
+        BDA::write(f, &header.sigblock_to_buf(), MetadataLocation::Both)?;
 
         let regions = mda::MDARegions::initialize(BDA_STATIC_HDR_SIZE, header.mda_size, f)?;
 
@@ -78,7 +126,7 @@ impl BDA {
     /// Returns None if no BDA appears to exist.
     pub fn load<F>(f: &mut F) -> StratisResult<Option<BDA>>
     where
-        F: Read + Seek,
+        F: Read + Seek + SyncAll,
     {
         let header = match StaticHeader::setup(f)? {
             Some(header) => header,
@@ -164,7 +212,7 @@ impl BDA {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Eq, PartialEq)]
 pub struct StaticHeader {
     blkdev_size: Sectors,
     pool_uuid: PoolUuid,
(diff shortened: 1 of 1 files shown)


tasleson:
Addresses https://github.com/stratis-storage/stratisd/issues/826

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -36,7 +36,54 @@ pub struct BDA {
     regions: mda::MDARegions,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq)]
+enum Copy {
```
agrover: This is using "copy" as a noun, but it could also be read as a verb, which would be confusing, because we're using it to say which version is overwritten, which is not a "copy" operation.

Anyway, how about calling this MetadataCopy or MetadataLocation so this is unambigous?

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -36,7 +36,54 @@ pub struct BDA {
...
 impl BDA {
+    fn read<F>(f: &mut F) -> Result<[u8; _BDA_STATIC_HDR_SIZE]>
+    where
+        F: Read + Seek,
+    {
+        #![allow(unused_io_amount)]
+        f.seek(SeekFrom::Start(0))?;
+        let mut buf = [0u8; _BDA_STATIC_HDR_SIZE];
+
+        // FIXME: See https://github.com/stratis-storage/stratisd/pull/476
```
agrover: I know this PR just copies this FIXME from another spot, but I went and looked at that PR and wasn't clear what should be fixed here. Should this just be doing a `read_all` to avoid short reads, or what?

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -196,68 +233,76 @@ impl StaticHeader {
     }
 
     /// Try to find a valid StaticHeader on a device.
-    /// If there is no StaticHeader on the device, return None.
-    /// If there is a problem reading a header, return an error.
+    /// We will return the first copy that validates as a Stratis BDA, however we will verify both
+    /// copies and if one validates but one does not, we will re-write the one that is incorrect.
+    /// Return None if the static header's magic does not match for *both* copies.
```
agrover: Suggest cutting the 3 "we will"s in this comment block, I don't think they're needed.

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -196,68 +233,76 @@ impl StaticHeader {
...
-            &buf[SECTOR_SIZE..2 * SECTOR_SIZE],
-            &buf[9 * SECTOR_SIZE..10 * SECTOR_SIZE],
-        ];
-
-        for buf in &sigblock_spots {
-            match StaticHeader::sigblock_from_buf(buf) {
-                Ok(val) => return Ok(val),
-                _ => continue,
+        let buf = BDA::read(f)?;
+        let buf_copy_1 = &buf[SECTOR_SIZE..2 * SECTOR_SIZE];
```
agrover: Yeah, just to follow up on my previous comment, I think "location" or "loc" is more descriptive than "copy".

src/engine/strat_engine/backstore/metadata.rs
tasleson: I'm not exactly sure.  I do know that we will fail the unit tests if we do a `read_exactly` as I tried that as a potential solution.

src/engine/strat_engine/backstore/metadata.rs
tasleson: Done, thank you.

src/engine/strat_engine/backstore/metadata.rs
tasleson: Done, thank you.

tasleson:
@agrover Travis test failed for new unit test, not sure why.  Using the `assert_eq!` with vector byte arrays seems problematic, when they assert the amount of data is huge, gets truncated by travis and not super useful.  I'm going to re-work to only dump the differences between them.

src/engine/strat_engine/backstore/metadata.rs
mulkieran: The reason: https://github.com/stratis-storage/stratisd/pull/860#issuecomment-376331069.

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -36,7 +36,54 @@ pub struct BDA {
...
+        // Write 8K header. Static_Header copies go in sectors 1 and 9.
+        if which == MetadataLocation::Both || which == MetadataLocation::First {
+            f.seek(SeekFrom::Start(0))?;
+            f.write_all(&zeroed[..SECTOR_SIZE])?; // Zero 1 unused sector
+            f.write_all(bda_buf)?;
+            f.write_all(&zeroed[..SECTOR_SIZE * 7])?; // Zero 7 unused sectors
+            f.sync_all()?;
+        }
+
+        if which == MetadataLocation::Second {
```
mulkieran: Nicer if this were an ```else``` block of the first ```if```, as you are doing one or the other.

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -36,7 +36,54 @@ pub struct BDA {
     regions: mda::MDARegions,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq)]
```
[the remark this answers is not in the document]
tasleson: Done, thanks!

src/engine/strat_engine/backstore/metadata.rs
tasleson: Indeed, simplified.  Thanks!

tasleson:
I will squash once be get consensus that this looks good.

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -196,68 +231,74 @@ impl StaticHeader {
     }
 
     /// Try to find a valid StaticHeader on a device.
-    /// If there is no StaticHeader on the device, return None.
-    /// If there is a problem reading a header, return an error.
+    /// Return None if the static header's magic does not match for *both* copies.
```
mulkieran: In final copy, method header comment should document repair work.

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -36,7 +36,52 @@ pub struct BDA {
...
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+enum MetadataLocation {
+    Both,
+    First,
+    Second,
+}
+
 impl BDA {
+    fn read<F>(f: &mut F) -> Result<[u8; _BDA_STATIC_HDR_SIZE]>
```
mulkieran: In final copy, this and other new methods should receive header doc.

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -196,68 +231,74 @@ impl StaticHeader {
...
+        let buf_loc_1 = &buf[SECTOR_SIZE..2 * SECTOR_SIZE];
+        let buf_loc_2 = &buf[9 * SECTOR_SIZE..10 * SECTOR_SIZE];
+
+        match (
+            StaticHeader::sigblock_from_buf(buf_loc_1),
+            StaticHeader::sigblock_from_buf(buf_loc_2),
+        ) {
+            (Ok(loc_1), Ok(loc_2)) => {
+                match (loc_1, loc_2) {
+                    (Some(loc_1), Some(_)) => {
```
mulkieran: Since both values are read it costs virtually nothing to verify that they are equal. So it makes sense to do that and return an error if they are not (because you do not know which one is correct, and one must be false). You'll have to derive ```Eq```, ```PartialEq``` for ```StaticHeader```, but I don't see anything wrong with that.

src/engine/strat_engine/backstore/metadata.rs
tasleson: @mulkieran I had repair notes in original header, @agrover wanted them removed...

src/engine/strat_engine/backstore/metadata.rs
tasleson: They both have a time stamp, should we return the newer of the two?

src/engine/strat_engine/backstore/metadata.rs
tasleson: I wasn't sure if non-public functions needed documentation, but yes I can certainly add some.

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -196,68 +231,74 @@ impl StaticHeader {
...
-
         // Using setup() as a test of ownership sets a high bar. It is
         // not sufficient to have STRAT_MAGIC to be considered "Ours",
         // it must also have correct CRC, no weird stuff in fields,
         // etc!
-        match StaticHeader::setup_from_buf(&buf) {
+        match StaticHeader::setup(f) {
             Ok(Some(sh)) => Ok(DevOwnership::Ours(sh.pool_uuid, sh.dev_uuid)),
             Ok(None) => {
+                let buf = BDA::read(f)?;
```
mulkieran: This is the interesting bit:

```determine_ownership``` now reads the BDA twice, if no Stratis heade ris found, where before it only read once. If computation is structured in the way it currently is, that is necessary, because ```StaticHeader::setup```, since it does its own repair work, must have access to the actual file.

There are obviously a number of ways to restructure the computation, as:
* ```StaticHeader:;setup``` can return instructions on how to fix stuff up and therefore only need to take a buffer argument.
* ```StaticHeader::setup``` can return the buffer that is currently on the disk as a result of its actions, whatever that is, even if it took no action. This too means it can take the allready read buffer, not the file.
* etc.

Or the choice can be made to accept the double read.

File: src/engine/strat_engine/backstore/metadata.rs (hunk 9 above)
Comment (mulkieran) [reply src/engine/strat_engine/backstore/metadata.rs]: Returning most recent isn't unreasonable.

File: src/engine/strat_engine/backstore/metadata.rs (hunk 10 above)
Comment (tasleson) [reply src/engine/strat_engine/backstore/metadata.rs]: This code path is quite a bit different in https://github.com/stratis-storage/stratisd/pull/936 I suggest we just leave as is.

File: src/engine/strat_engine/backstore/metadata.rs [hunk 11]
```diff
@@ -1085,4 +1126,81 @@ mod tests {
...
+    #[test]
+    fn bda_test_recovery() {
+        fn corrupt_byte<F>(f: &mut F, position: u64, mask: u8) -> Result<()>
+        where
+            F: Read + Seek + SyncAll,
+        {
+            let mut byte_to_corrupt = [0; 1];
+            f.seek(SeekFrom::Start(position))?;
+            f.read(&mut byte_to_corrupt)?;
+            byte_to_corrupt[0] ^= mask;
```
mulkieran: Might as well not take any ```mask``` parameter and instead use bitwise negation, ```!```, I would think.

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -1085,4 +1126,81 @@ mod tests {
             .tests(30)
             .quickcheck(static_header as fn(u64, u32) -> TestResult);
     }
+
+    #[test]
```
mulkieran: In final copy, test should have header comment explaining the property it verifies.

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -36,7 +36,52 @@ pub struct BDA {
...
+        // FIXME: See https://github.com/stratis-storage/stratisd/pull/476
+        f.read(&mut buf)?;
+        Ok(buf)
+    }
+
+    fn write<F>(f: &mut F, bda_buf: &[u8], which: MetadataLocation) -> Result<()>
+    where
+        F: Seek + SyncAll,
+    {
+        let zeroed = [0u8; _BDA_STATIC_HDR_SIZE];
```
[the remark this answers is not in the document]
tasleson: This snippet of code you have placed here doesn't make sense to me.  We must have a `sync_all` between copies and the `f.write_all(if which == Both || which == First) { bda_buf } else { &zeroed...})?;` appears to zero out the metadata block.

I'm not sure I understand the value in just writing the zero parts, I think doing that increases risk of not being able to read up either copy.  If we get in the situation of having one good and one bad, we should not re-write the good one at all, even just the zero parts.

In fact I'm going to update this PR for the `BDA::read` because we are doing it incorrectly.  We should actually be issuing 2 reads of the specific BDA blocks with a seek in between instead of doing a single read.  The reason for this is that if we lose power in the middle of updating one of the copies a disk sector could be left in an error state, likely a data sync error (03/1600 which needs to be re-written to correct).  Thus a single read of the first 8K could fail with an EIO when in reality a single block read of each of the specific BDA sectors may succeed for each.  Every sector in the transfer must be good for the entire transfer to succeed, thus it's in our best interest to only read the **sectors** that really matter and read them individually.

src/engine/strat_engine/backstore/metadata.rs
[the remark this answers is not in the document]
tasleson: Yes, good suggestion.

src/engine/strat_engine/backstore/metadata.rs
agrover: I just wanted the "we must" removed, those two words, in three instances...

tasleson: @agrover Ah, OK will do

src/engine/strat_engine/backstore/metadata.rs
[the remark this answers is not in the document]
tasleson: I've tried to achieve the same behavior as what we had before, in that we write the first copy, sync, then write the second copy with the change allowing the update of each part individually.  The individual writes done before the sync are all stream and will be coalesced into a single underlying device write so ultimately I don't think we need to change anything else here.  @agrover can speak to why we zero'd or should still zero the other blocks in the header, but in general the important part is getting one copy down intact.

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -1085,4 +1126,81 @@ mod tests {
...
+            f.seek(SeekFrom::Start(position))?;
+            f.write(&byte_to_corrupt)?;
+            f.sync_all()?;
+            Ok(())
+        }
+
+        let sh = random_static_header(10000, 4);
+        let buf_size = *sh.mda_size.bytes() as usize + _BDA_STATIC_HDR_SIZE;
+        let mut buf = Cursor::new(vec![0; buf_size]);
+        let mut _bda = BDA::initialize(
```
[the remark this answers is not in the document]
tasleson: Done, thanks!

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -1085,4 +1126,81 @@ mod tests {
...
+            sh.dev_uuid,
+            sh.mda_size,
+            sh.blkdev_size,
+            Utc::now().timestamp() as u64,
+        ).unwrap();
+
+        let reference_buf = buf.clone();
+
+        for i in 0..SECTOR_SIZE {
+            for primary in [true, false].iter() {
```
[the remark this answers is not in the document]
tasleson: Done, thanks!

src/engine/strat_engine/backstore/metadata.rs
tasleson: Done, thanks.  Originally I was planning on walking through different values of mask.

src/engine/strat_engine/backstore/metadata.rs
tasleson: Done, thanks!

tasleson:
I've removed the changes for `BDA::read`.  I will open a new issue to track this as it's going to require a number of test changes to make things work.

agrover:
:+1: to be merged after squashing.

mulkieran:
Adding this to Stratis project and putting in review column, so it's among the list of things to be reviewed.

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -36,7 +36,54 @@ pub struct BDA {
...
+
+        // Write 8K header. Static_Header copies go in sectors 1 and 9.
+        if which == MetadataLocation::Both || which == MetadataLocation::First {
+            f.seek(SeekFrom::Start(0))?;
+            f.write_all(&zeroed[..SECTOR_SIZE])?; // Zero 1 unused sector
+            f.write_all(bda_buf)?;
+            f.write_all(&zeroed[..SECTOR_SIZE * 7])?; // Zero 7 unused sectors
+            f.sync_all()?;
+        } else {
+            f.seek(SeekFrom::Start((9 * SECTOR_SIZE as u64) as u64))?;
```
mulkieran: The second cast here is redundant. Just
```
f.seek(SeekFrom::Start(9 * SECTOR_SIZE as u64))?;
```

will do it.

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -196,68 +233,81 @@ impl StaticHeader {
...
+        let buf_loc_2 = &buf[9 * SECTOR_SIZE..10 * SECTOR_SIZE];
+
+        match (
+            StaticHeader::sigblock_from_buf(buf_loc_1),
+            StaticHeader::sigblock_from_buf(buf_loc_2),
+        ) {
+            (Ok(loc_1), Ok(loc_2)) => {
+                match (loc_1, loc_2) {
+                    (Some(loc_1), Some(loc_2)) => {
+                        let latest = if loc_1.initialization_time >= loc_2.initialization_time {
```
mulkieran: I think that if the decision is that the more recent ```initialization_time``` wins, then it makes sense to do a repair of the earlier static header, just as if it were invalid.

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -36,7 +36,54 @@ pub struct BDA {
...
+
+        // FIXME: See https://github.com/stratis-storage/stratisd/pull/476
+        f.read(&mut buf)?;
+        Ok(buf)
+    }
+
+    /// Write the BDA, allowing the caller to specify which part(s) to update.
+    fn write<F>(f: &mut F, bda_buf: &[u8], which: MetadataLocation) -> io::Result<()>
+    where
+        F: Seek + SyncAll,
```
mulkieran: I understand that fixes to ```BDA::read``` are being deferred until a subsequent PR. But I think what regions are zeroed should not be dependent on what location argument is passed. Here's one approach which consistently zeroes nothing and which is very simple:

```
    fn write<F>(f: &mut F, bda_buf: &[u8], which: MetadataLocation) -> Result<()>
    where
        F: Seek + SyncAll,
    {
        if which == MetadataLocation::Both || which == MetadataLocation::First {
            f.seek(SeekFrom::Start(SECTOR_SIZE as u64))?;
            f.write_all(bda_buf)?;
            f.sync_all()?;
        }

        if which == MetadataLocation::Both || which == MetadataLocation::Second {
            f.seek(SeekFrom::Start(9 * SECTOR_SIZE as u64))?;
            f.write_all(bda_buf)?;
            f.sync_all()?;
        }
        Ok(())
    }
```

and here's another approach which consistently zeroes everything and is not too complex:

```
    fn write<F>(f: &mut F, bda_buf: &[u8], which: MetadataLocation) -> Result<()>
    where
        F: Seek + SyncAll,
    {
        let zeroed = [0u8; _BDA_STATIC_HDR_SIZE];

        f.seek(SeekFrom::Start(0))?;
        f.write_all(&zeroed[..SECTOR_SIZE])?; // Zero 1 unused sector
        if which == MetadataLocation::Both || which == MetadataLocation::First {
            f.write_all(bda_buf)?;
        } else {
            f.seek(SeekFrom::Start(SECTOR_SIZE as u64 * 2))?;
        }
        f.sync_all()?;

        f.write_all(&zeroed[..SECTOR_SIZE * 7])?; // Zero 7 unused sectors
        if which == MetadataLocation::Both || which == MetadataLocation::Second {
            f.write_all(bda_buf)?;
        } else {
            f.seek(SeekFrom::Start(SECTOR_SIZE as u64 * 10))?;
        }
        f.write_all(&zeroed[..SECTOR_SIZE * 6])?; // Zero 6 unused sectors
        f.sync_all()?;
        Ok(())
    }
```

Both approaches manage to pass all tests in the metadata module, FWIW.

src/engine/strat_engine/backstore/metadata.rs
agrover: So you're saying we need to differentiate between when the two structs are identical (expected) or not identical. Hard to see how this could accidentally happen because CRC was checked by `sigblock_from_buf`. I guess there are nightmare combinations of dd-ing and power failures that could expose us here.

Anyway I agree with @mulkieran's recommendation. I can't think of anything better to do. Might as well sync em up?

Suggest deriving `Eq`/`PartialEq` for StaticHeader?

src/engine/strat_engine/backstore/metadata.rs
tasleson: @mulkieran Please elaborate on why you believe we should re-write all the zero parts, what value add does that provide?

mulkieran: I don't take any stand on whether the empty parts of the BDA should be zeroed or not. I only think that what is zeroed should not be dependent on the value of the ```which``` argument. I presented two alternatives, one which zeroes nothing and one which zeroes everything.

tasleson: When we have gotten into this situation we have one copy that is good, one that is bad.

IMHO writing to both regions does not improve the odds of having both copies, but decreases it as you are re-writing something known to be good.  You never **know** if a write is actually good unless you re-read it, so if it's known to be good, don't muck with it!

We have two copies, we want to have the ability to write to both or one or the other.  Part of each copy has reserved sectors and @agrover wants reserved sectors zeroed.  I don't know why the function needs to write all or none zero parts instead of just writing just the parts the caller specified. It's bad when function calls have unintended side effects.

agrover: I think some of the stuff in my head didn't make it into the doc! (sorry.)

So what I was thinking was that the StaticHeader was divided into 2 4K chunks. We're only using one sector from each, to prevent partial writes on 512byte blocksize disks. (And the copies aren't right next to each other, because we want them to be separate writes on 4k disks.)

If someone overwrote the first 4K with garbage (but the second one is ok), I'd like to restore the first signature block, but also re-zero-out the rest of the 4K too. Although this is a little sketchy because otherwise we don't look at the other 7 sectors at all, i.e. verify they're zero, I don't think. So maybe now that I think about it we shouldn't zero. Aw hell. Whatever you wanna do, Tony.

mulkieran: What regions are known to be good when ```write``` is invoked?

mulkieran: I've added a third option, which is also perfectly fine w/ me, as it is consistent w/ itself. It also seems to be what @agrover intended, according to discussion in standup.

tasleson: @mulkieran @agrover The third option looks good to me :+1:   I incorporate it with the other suggested changes.

mulkieran: I've made an edit to my third choice. 8K was floating around in the comments where 8 _sectors_ was really meant.

src/engine/strat_engine/backstore/metadata.rs
tasleson: Done indirectly with utilizing re-write of `write`

agrover:
fixes #826

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -196,68 +244,87 @@ impl StaticHeader {
...
-                _ => continue,
+        let buf = BDA::read(f)?;
+        let buf_loc_1 = &buf[SECTOR_SIZE..2 * SECTOR_SIZE];
+        let buf_loc_2 = &buf[9 * SECTOR_SIZE..10 * SECTOR_SIZE];
+
+        match (
+            StaticHeader::sigblock_from_buf(buf_loc_1),
+            StaticHeader::sigblock_from_buf(buf_loc_2),
+        ) {
+            (Ok(loc_1), Ok(loc_2)) => {
```
mulkieran: clippy has noticed that every one of these "return" keywords can be omitted, so I'll mention it too.

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -1085,4 +1152,132 @@ mod tests {
...
+
+    #[test]
+    /// Test that we re-write the older of two BDAs if they don't match.
+    fn bda_test_rewrite_older() {
+        let sh = random_static_header(10000, 4);
+        let buf_size = *sh.mda_size.bytes() as usize + _BDA_STATIC_HDR_SIZE;
+        let mut buf = Cursor::new(vec![0; buf_size]);
+        let ts = Utc::now().timestamp() as u64;
+
+        let _ = BDA::initialize(
```
mulkieran: You can omit the ```let _```'s here and below.

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -196,68 +244,87 @@ impl StaticHeader {
...
+                        return Ok(Some(loc_1));
+                    }
+                    (None, Some(loc_2)) => {
+                        // Copy 2 has valid Stratis BDA, copy 1 has no magic, re-write copy 1
+                        BDA::write(f, buf_loc_2, MetadataLocation::First)?;
+                        return Ok(Some(loc_2));
+                    }
+                }
+            }
+            (Ok(loc_1), Err(_)) => {
```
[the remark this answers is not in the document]
tasleson: Yes I agree, it's better to return `Err` instead of `Ok(None)`, thanks!  I've incorporated this.

src/engine/strat_engine/backstore/metadata.rs
tasleson: Done, thanks

src/engine/strat_engine/backstore/metadata.rs
tasleson: Done, thanks

agrover:
All right! Let's get this squashed and merged asap please.

PR #979, opened 2018-06-05
Improvement to stratisd run() method

mulkieran:
This moves registering pools out of the dbus layer and into bin/stratisd.rs, completely. Previously, in one case the pools were registered in the run() method in stratisd.rs , and in the other they were registered as a side-effect of calling the dbus-api connect() method. This way, the code is more uniform.


mulkieran:
rebased.

mulkieran:
rebased.

PR #983, opened 2018-06-07
Master compiler crash

mulkieran:
This code caused the compiler to crash, so I'm stowing it in a PR for reporting purposes.


mulkieran:
Closing, no longer need the record sitting around.

PR #984, opened 2018-06-07
Get rid of engine_eventable

mulkieran:
It is equivalent to eventable.is_some(), and more cumbersome.
This way we get rid of an expect().


mulkieran:
Cherry-picked from #979.

mulkieran:
Virtually trivial.

PR #986, opened 2018-06-11
Have strat engine verify existence of required binaries

mulkieran:
Assume that all binaries will be required by stratisd at some time,
and that it shouldn't run if they are not available.

This seems better to me than having a panic if we try to use them and they
are not needed.

Hoist cmd.rs up a level, since it seems like the engine really has to
interact with it.

Postpone  dealing with the fact that not all these binaries will actually
be available in the initramfs.


mulkieran:
Supercedes #974 .

Code is identical, but comment on ```verify_binaries``` has been expanded to explain why an explicit list of directories is searched.

agrover:
:+1:

Issue #776, opened 2018-02-19
Improve error text on errors


tasleson:
The daemon code typically takes whatever low level error we have gotten and uses the `from::` to convert it to another error type.  However, when this is done and bubbled up we lose the specifics.  For example when we bubble up: `Io(Error { repr: Os { code: 2, message: "No such file or directory" } }`, it's not apparent what file or directory operation we were trying when we received this error.

Maybe this could be addressed with something like error chain or it's replacement?

mulkieran:
Tangentially related: https://github.com/stratis-storage/devicemapper-rs/issues/246.

mulkieran:
Removing the "help wanted" flag, since this is a necessary and really difficult thing. Removing enhancement as well. There are a lot of less fundamental things that are not labeled as enhancement, so this shouldn't be.

mulkieran:
```

extern crate failure;

extern crate rust_project;

use std::io::Error;

use rust_project::library::errors;

fn error_return() -> errors::StratisResult<()> {
    Err(Error::last_os_error())?
}

fn main() {
    match error_return() {
        Err(err) => {
            println!("{:?}", err);
            match err.downcast::<Error>() {
                Ok(io_error) => {
                    println!("It's an IO error!");
                    println!("Its kind is: {:?}", io_error.kind());
                }
                Err(_) => println!("All that downcasting for nuthin!"),
            }
        },
        Ok(_) => {},
    }
}
```

and 

```
> RUST_BACKTRACE=1 cargo run
```

yields

```
[mulhern@localhost rust-project-2]$ RUST_BACKTRACE=1 cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s                   
     Running `target/debug/main`
Os { code: 0, kind: Other, message: "Success" }

stack backtrace:
   0:     0x557d6c5f26c5 - backtrace::backtrace::libunwind::trace::he5b3a6445bed61ab
                        at /home/mulhern/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.9/src/backtrace/libunwind.rs:53
                         - backtrace::backtrace::trace::h88c78440ea0408ad
                        at /home/mulhern/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.9/src/backtrace/mod.rs:42
   1:     0x557d6c5ee3be - backtrace::capture::Backtrace::new_unresolved::h36a01d8634ce27a4
[...]

Issue #766, opened 2018-02-16
Unhelpful error message when missing executable


tasleson:
If you're missing the `xfsprogs` package when you try to create a pool you will get the message:

`Execution failed: 4: entity not found`

This could be clearer.

mulkieran:
Related: #530.

tasleson:
Related: https://github.com/stratis-storage/stratisd/issues/771

mulkieran:
I'm going to consider this closed by #986. At this point stratis-cli will never report:

```
Execution failed: 4: entity not found
```

because 

1. the engine will have promptly exited, on discovering that the binaries it relies on does not exist. So, stratis-cli, when invoked to do anything, will report instead a D-Bus exception and will summarize its best guess as that the Stratis daemon is not running OR

2. the engine will have found the necessary executables, which will now have gone missing. In that case, the daemon will put an error on the D-Bus, and the error will be the current string, which is: "Executable previously located at ```<absolute path of executable>``` seems to have been removed since stratisd was started".

PR #988, opened 2018-06-13
Master failure

mulkieran:
Related: #530.
Related: #776.
Related: #766.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -199,6 +199,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "failure"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "backtrace 0.3.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "failure_derive 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "failure_derive"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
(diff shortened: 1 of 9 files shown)


mulkieran:
Very wierd Jenkins failure: https://stratis-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/stratisd/1375/.

Error where actual blockdev size is 1/10 the size of the recorded block dev size. I'm going to restart Jenkins and see what happens.
However, this is a definite place where error reporting could be better. If we got the blockdev size, we had a handle on the blockdev itself, for example, its devnode, and this could have been included.

mulkieran:
Blocked by #1004.

mulkieran:
No longer blocked.

mulkieran:
It's been a month and no feedback, so I'm moving it back to in-progress and seeing what more I can do w/ it.

mulkieran:
rebased.

mulkieran:
I'm putting this into Stratis 1.1 preliminary review again as we may consider the problem of identifying the context of errors more urgent. We might want to consider a restricted goal, which does not attempt chaining, but does use failure crate to capture stack trace at point of error.

If we did that, we could perhaps start by sticking in failure errors only at leaf errors, i.e., locations where the error is picked up from an external crate.

mulkieran:
There is another possibility worth exploring. Perhaps we can simply make our existing error types implement ```Fail``` and then we can get the backtrace sort of for free.

mulkieran:
rebased. Keeping around for reference only.

mulkieran:
Superceded by #1499. We can always take a look at this as a reference if need to, but might as well close at this time.

PR #989, opened 2018-06-13
Use ftruncate instead of fallocate to make loopbacked devices

mulkieran:
Because ftruncate makes a sparse file, this helps to avoid running out of
space on the device which backs the loopbacked devices.

The downside is that some failures may yet occur but be postponed until
the devices are actually being filled, i.e., while the test is actually
being run.

Don't wipe the first 1 MiB, with ftruncate, the device should present as
zeroed.

### DIFF
--- a/src/engine/strat_engine/tests/loopbacked.rs
+++ b/src/engine/strat_engine/tests/loopbacked.rs
@@ -9,7 +9,7 @@ use std::os::unix::io::AsRawFd;
 use std::panic;
 use std::path::{Path, PathBuf};
 
-use nix::{self, fcntl::{fallocate, FallocateFlags}};
+use nix;
 use tempfile;
 
 use devicemapper::{Bytes, Sectors, IEC};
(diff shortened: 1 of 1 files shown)


agrover:
I think we had that 1MiB wipe there because we were seeing some failures otherwise. Maybe we don't need it any more. CI is passing, does that mean we're ok without it @trgill ?

trgill:
@agrover yeah, if we use the device as a meta and there is data in the first block DM issues some confusing errors/tried to readopt the device as an existing meta.  

@mulkieran why are you sure the device will present as zeroed?

mulkieran:
@trgill The comment in the corresponding devicemapper PR says:

```
+/// Make sure each loop device is backed by a sparse 1 GiB file. The entire file will read back
+/// as initialized with zero.
```

I interpreted it to mean that the file that was obtained by this means, although sparse, would behave as if zeroed out everywhere.

trgill:
ok :+1:

PR #990, opened 2018-06-14
Do not set revents to 0 when handling engine events

mulkieran:
Expect the kernel to do the correct thing, as with D-Bus and udev
revents.


mulkieran:
Alternatively, if there is a reason why revents is set to 0 in one place and not in another, it should get a comment.

agrover:
Nice catch. We should not have to reinitialize revents to 0 since the kernel never reads, and will always set this field.

:+1:

Issue #794, opened 2018-02-23
Need to evaluate all asserts in code base


mulkieran:
Need to decide which should be asserts, and which should be debug asserts, which get compiled out.

mulkieran:
Need to get rid of ones that should actually be errors, and not asserts at all.

mulkieran:
It turns out I am working on this, but mostly in devicemapper, see:
https://github.com/stratis-storage/devicemapper-rs/pull/417, merged
https://github.com/stratis-storage/devicemapper-rs/pull/430, in progress.

Issue #904, opened 2018-05-01
Move work off the event loop


agrover:
long-running tasks return a job id. task takes as long as it wants and then sets the job as completed with a status
2nd level problem: task with subprocesses have a timeout on subprocessesm b/c crappy subprocesses might hang?

tasleson:
@trgill Makes a good point that the underlying C library that dbus-rs uses isn't thread safe.  Thus all work pertaining to dbus-rs needs to occur in the same thread of execution.

mulkieran:
Obsoleted by https://github.com/stratis-storage/project/issues/86.

PR #991, opened 2018-06-14
Change assertion to an error return

mulkieran:
Related: #794.


agrover:
I like this, but according to the poll(2) man page the value returned on error will always be -1, and then one has to check errno for the actual reason. We could either get errno (based on nix's example, it involves calling libc::__errno_location() and then dereferencing. Will need `unsafe`.) or just not bother including the value, since we're planning on rewriting this using something better, see #656 or #904.

mulkieran:
Possibilities are:

```
ERRORS
       EFAULT The  array  given  as  argument was not contained in the calling
              program's address space.

       EINTR  A signal occurred before any requested event; see signal(7).

       EINVAL The nfds value exceeds the RLIMIT_NOFILE value.

       ENOMEM There was no space to allocate file descriptor tables.
```

EFAULT is impossible without unlikely compiler errors, but the other three are a bit more likely.

For, EINTR there's a strong case to be made for continuing, and letting the default signal handle do its thing, instead of exiting.

For the others, there is a case for continuing, but not as strong.

Issue #735, opened 2018-02-01
We need to add a test to CI, to make sure that compilation w/out dbus enabled doesn't still pull in libdbus


mulkieran:
Related: #734.

mulkieran:
This should be easy to set up on Travis so long as we can, e.g., use ```ld``` to scan for included libraries in executable.

mulkieran:
#997 was an attempt to resolve this problem. But it failed, because even though compilation is correct, libdbus is still included in the executable. #997 comments have a few further details.

mulkieran:
I'm moving this into "blocked", since the obvious thing, which would have been expected to succeed, failed.

PR #996, opened 2018-06-18
Separate out dbus-related expressions in run() method

mulkieran:
There were a few places where identifiers that were only used in dbus related actions were defined in other actions. This approach necessitated prefixing the names with underscores, so that the compiler could ignore them when dbus was not enabled. This PR completely separates the dbus-related actions.


mulkieran:
Maybe I can actually resolve #735.

mulkieran:
This test succeeds on my machine but fails on Travis. That is, libdbus seems to appear in the executable.

mulkieran:
Blocks #1002 .

agrover:
I don't think it'll fix #735 but it's still nice to have. :+1:

PR #997, opened 2018-06-18
Add a Travis task for compilation w/out defaults enabled

mulkieran:
We should always be certain that stratisd compiles w/out defaults enabled.

Related: #735.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,11 @@ build:
 	RUSTFLAGS='-D warnings' \
 	cargo build --target $(TARGET)
 
+build-no-default:
+	PKG_CONFIG_ALLOW_CROSS=1 \
+	RUSTFLAGS='-D warnings' \
+	cargo build --no-default-features --target $(TARGET)
+
 test-loop:
 	sudo env "PATH=${PATH}" RUSTFLAGS='-D warnings' RUST_BACKTRACE=1 RUST_TEST_THREADS=1 cargo test loop_
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,8 @@ matrix:
         - rust: stable
           env: TASK=build TARGET=x86_64-unknown-linux-gnu
         - rust: stable
+          env: TASK=build-no-default TARGET=x86_64-unknown-linux-gnu
+        - rust: stable
           env: TASK=build TARGET=x86_64-unknown-linux-gnu
           install:
             - rustup default 1.26.0


mulkieran:
It compiles, and it doesn't seem to be compiling libdbus-sys or dbus, but libdbus does seem to be included in the executable. Possibly I'm just looking at the wrong executable.

Note that the Makefile target is observed to work correctly on my machine.

mulkieran:
This doesn't work! Any ideas what's wrong?

Makefile
```diff
@@ -32,6 +32,7 @@ build-no-default:
 	PKG_CONFIG_ALLOW_CROSS=1 \
 	RUSTFLAGS='-D warnings' \
 	cargo build --no-default-features --target $(TARGET)
+	! (ldd ./target/${TARGET}/debug/stratisd | grep libdbus)
```
tasleson: I think you have the path incorrect...

`! (ldd ./target/${TARGET}/stratisd | grep libdbus)`

agrover:
libnih uses libdbus-1, libcgmanager uses libnih. Big Q is why we're linking against libcgmanager???
```
ldd ./target/x86_64-unknown-linux-gnu/debug/stratisd
	linux-vdso.so.1 =>  (0x00007ffc217fc000)
	libudev.so.1 => /lib/x86_64-linux-gnu/libudev.so.1 (0x000014e518c15000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x000014e518a11000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x000014e518809000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x000014e5185eb000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x000014e5183d5000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x000014e51800c000)
	/lib64/ld-linux-x86-64.so.2 (0x000014e519a8b000)
	libcgmanager.so.0 => /lib/x86_64-linux-gnu/libcgmanager.so.0 (0x000014e517df1000)
	libnih.so.1 => /lib/x86_64-linux-gnu/libnih.so.1 (0x000014e517bd9000)
	libnih-dbus.so.1 => /lib/x86_64-linux-gnu/libnih-dbus.so.1 (0x000014e5179cf000)
	libdbus-1.so.3 => /lib/x86_64-linux-gnu/libdbus-1.so.3 (0x000014e51778a000)
! (ldd ./target/x86_64-unknown-linux-gnu/debug/stratisd | grep libdbus)
	libdbus-1.so.3 => /lib/x86_64-linux-gnu/libdbus-1.so.3 (0x0000145b66432000)
```

Makefile
tasleson: Guess this isn't it...

mulkieran:
Ok, so I'm removing the check for libdbus. But the rest is working correctly, and might as well go in.

PR #1000, opened 2018-06-19
BDA Test for Ok(None) && Err()

tasleson:
Ensure we return Err instead of Ok(None) when one copy appears
to have no magic and the other copy does, but fails to validate.


agrover:
:+1:

mulkieran:
I never got a chance to review this code. I had actually added a review comment while it was still in progress, but then I figured that was too early to be giving out comments so I deferred until it actually got moved into review. It is not super important.

Issue #987, opened 2018-06-12
Not all binaries needed in initrd binary


agrover:
With the merge of #986, stratisd will fail to run if all binaries aren't found. However, it only uses a subset of binaries from initrd, so it should only check for those if compiled for that use.

mulkieran:
My concern is that this issue about exiting if some binaries that stratisd usually requires, but which aren't available in the initramfs, is just the tip of the iceberg of the problem.

We have a mechanism, conditional compilation, which is supposed to prevent D-Bus actions from being executed at boot. But how do we know that xfs-dependent actions won't be executed at boot, causing stratisd to fail one way or the other? AFAIK, the answer is that we do not.

If we already had a mechanism to deal with that problem; then handling problems around ```verify_binaries()``` would be easy, as we could make alternate version of ```BINARIES``` and ```verify_binaries``` and switch between them based on whether the desired stratisd was the initramfs or the full one. But we do not have that.

agrover:
Since #986 went in, we just will fail to come up in initrd because xfs binaries are absent, so that's not a good state of affairs.

I was thinking we'd just config out XFS binaries being searched-for in BINARIES if building for initrd. Then calling them would return an error. We could also go further and cfg-out the callsites as well. Thoughts?

We also could punt and require xfs binaries in initrd, even though we never(?) will need them! Not saying we should, but this puts in perspective that this is really an optimization to keep our initrd footprint down.

Another tangential thought: right now we have a default feature of "dbus_enabled". Since that isn't the entirety of the changes we're now looking at for stratis-init, we may need to rename it, or add a second default feature.

tasleson:
Should we just edit the dracut plugin to require all of the binaries?  I'm not sure if this will be OK due to space constraints, but it would make this issue go away.

agrover:
Sure, let's just do that for now, and we can refine later if needed.

tasleson:
PR posted upstream, see: https://github.com/dracutdevs/dracut/pull/452

tasleson:
Upstream merged

PR #1003, opened 2018-06-19
RFC: Leverage conditional compilation for full runtime

tasleson:
I renamed the `dbus_enabled` feature to `full_runtime` and prefixed the external executables and functions that aren't in the initrd.  After doing this I just let the compiler flag everything that wouldn't compile or wasn't needed anymore until things compiled clean.  I'm not saying we want to utilize this approach as it's ugly, but I did find it very informative.  For example we basically need to disable the events coming out of `dm` as those code paths utilize binaries that aren't in the initrd.

I'm not really sure how we could restructure the code at the moment so that things could be placed into different modules to make this all much cleaner. 

This was an attempt at an answer for https://github.com/stratis-storage/stratisd/issues/987

**Note: The release code size difference between with/without `full_runtime` is 662440 bytes.**  (3914600 vs. 3252160)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,5 +35,5 @@ loopdev = "0.2"
 either = "1.1.0"
 
 [features]
-default = ["dbus_enabled"]
-dbus_enabled = ["dbus"]
+default = ["full_runtime"]
+full_runtime = ["dbus"]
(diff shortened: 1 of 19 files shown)


agrover:
When the compiler warns something isn't needed, does it still include it in the binary?

tasleson:
> When the compiler warns something isn't needed, does it still include it in the binary?

I don't think so

PR #1004, opened 2018-06-20
Strip message argument out of execute_cmd

mulkieran:
Cherry-picked from #988, but I think it also works well enough on its own.

### DIFF
--- a/src/engine/strat_engine/cmd.rs
+++ b/src/engine/strat_engine/cmd.rs
@@ -20,7 +20,7 @@ use std::process::Command;
 
 use uuid::Uuid;
 
-use stratis::{ErrorEnum, StratisError, StratisResult};
+use stratis::{StratisError, StratisResult};
 
 /// Find the binary with the given name by looking in likely locations.
 /// Return None if no binary was found.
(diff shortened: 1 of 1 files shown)


src/engine/strat_engine/cmd.rs
```diff
@@ -70,16 +70,16 @@ pub fn verify_binaries() -> StratisResult<()> {
 
 /// Common function to call a command line utility, returning an Result with an error message which
 /// also includes stdout & stderr if it fails.
-fn execute_cmd(cmd: &mut Command, error_msg: &str) -> StratisResult<()> {
+fn execute_cmd(cmd: &mut Command) -> StratisResult<()> {
     let result = cmd.output()?;
```
tasleson: I think it would be better if we tested for error here instead of returning it, so that the same additional command information can be added to the error before returning it.

mulkieran: Something like this:

```
use std::process::Command;

fn execute_cmd(cmd: &mut Command) -> Result<(), String> {
    match cmd.output() {
        Err(err) => {
            Err(format!("Failed to execute command {:?}, err: {:?}", cmd, err))
        }
        Ok(result) => {
            if result.status.success() {
                Ok(())
            } else {
                let std_out_txt = String::from_utf8_lossy(&result.stdout);
                let std_err_txt = String::from_utf8_lossy(&result.stderr);
                let err_msg = format!(
                    "cmd: {:?}, stdout: {} stderr: {}",
                    cmd, std_out_txt, std_err_txt
                );
                Err(err_msg)
            }
        }
    }
}

pub fn main() {
    execute_cmd(
        Command::new("junkfs")
            .arg("-f")
            .arg("-q")
            .arg("/dev/sda")
            .arg("-m"),
    ).unwrap();
}
```

would sort of work.

tasleson: :+1:

mulkieran:
This commit wasn't as independent of the other commits in PR #988 as I thought. This is really two commits from that PR, merged, now.

PR #1006, opened 2018-06-21
Leverage the udev db

agrover:
With the addition of Stratis support to libblkid, the udev db has
the basic information needed for stratisd.  This change leverages
the udev db where appropriate, to limit when we go to disk to
read up the meta data.


agrover:
The team agreed to merge this as a first implementation of using udev, to be followed by refactoring & optimization in a subsequent PR.

Some additional commits were needed to build against the just-released "stable" 1.27.0 compiler.

tasleson:
fixes https://github.com/stratis-storage/stratisd/issues/822

PR #1008, opened 2018-06-22
Fix two things in main method

mulkieran:
Resolves: #1015

### DIFF
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -316,13 +316,21 @@ fn main() {
         )
         .get_matches();
 
-    let result = trylock_pid_file()
-        .and_then(|_pidfile| {
-            initialize_log(matches.is_present("debug")).map_err(|_| {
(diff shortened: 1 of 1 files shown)


agrover:
@mulkieran now that clippy is working again you should be able to see the issue in the travis tests. The warning doesn't seem very apt to me, but if there's a better alternative than silencing it, lemme know.

mulkieran:
```
[root@localhost stratisd]# PYTHONPATH=./src ./target/x86_64-unknown-linux-gnu/debug/stratisd --sim --debug &
[2] 14700
Error: Daemon already running with pid: 14691
[2]+  Exit 1                  PYTHONPATH=./src ./target/x86_64-unknown-linux-gnu/debug/stratisd --sim --debug
```

mulkieran:
This PR was augmented to fix a bug where the fix affects the exact same code as the clippy problem which it was originally designed to fix. I don't think there's much point in splitting them into two separate PRs since they are so close together and since the clippy fix would interfere with the bug fix.

mulkieran:
Whoops. clippy found an extra ```.into()```. But I expect all will be fine now.

agrover:
:+1:

PR #1010, opened 2018-06-22
Remove rustfmt skip

mulkieran:
Better just to accept this wierdness than to prevent rustfmt from formatting
the whole module.


mulkieran:
If #1023 goes in, this is totally obsoleted anyway.

Issue #981, opened 2018-06-06
Change how we read up BDA


tasleson:
Today we simply issue an 8KiB read to fetch both copies of the BDA.   We should actually be issuing 2 reads of the specific BDA blocks with a seek in between instead of doing a single read. The reason for this is that if we lose power in the middle of updating one of the copies a disk sector could be left in an error state, possibly a data sync error (03/1600 which needs to be re-written to correct). Thus a single read of the first 8K could fail with an EIO when in reality a single block read of each of the specific BDA sectors may succeed for each. Every sector in the transfer must be good for the entire transfer to succeed, thus it's in our best interest to only read the sectors that really matter and read them individually.

Something similar to this, but better :-)

```rust
        let mut buf = [0u8; _BDA_STATIC_HDR_SIZE];
        let mut loc_1_read = Err(io::Error::new(io::ErrorKind::Other, "Seek error loc 1"));
        let mut loc_2_read = Err(io::Error::new(io::ErrorKind::Other, "Seek error loc 2"));
        if f.seek(SeekFrom::Start(SECTOR_SIZE as u64)).is_ok() {
            loc_1_read = f.read_exact(&mut buf[SECTOR_SIZE..SECTOR_SIZE * 2]);
        }

        if f.seek(SeekFrom::Start(9 * SECTOR_SIZE as u64)).is_ok() {
            loc_2_read = f.read_exact(&mut buf[9 * SECTOR_SIZE..10 * SECTOR_SIZE]);
        }

        if loc_1_read.is_ok() || loc_2_read.is_ok() {
            Ok(buf)
        } else {
            Err(loc_1_read.err().unwrap())
        }
```

PR #1012, opened 2018-06-22
BDA::read limit exposure to hard errors

tasleson:
Read the specific BDA sectors one at a time to prevent probability of
hitting a read error on a non-essential sector.

### DIFF
--- a/src/engine/strat_engine/backstore/metadata.rs
+++ b/src/engine/strat_engine/backstore/metadata.rs
@@ -43,18 +43,37 @@ enum MetadataLocation {
 }
 
 impl BDA {
-    /// Read the BDA from the device and returning a byte array of the entire area.
+    /// Read the BDA from the device and returning a byte array of the entire area.  Only the BDA
+    /// sectors are read up from disk, zero areas are *not* read up.
     fn read<F>(f: &mut F) -> io::Result<[u8; _BDA_STATIC_HDR_SIZE]>
     where
         F: Read + Seek,
     {
-        #![allow(unused_io_amount)]
-        f.seek(SeekFrom::Start(0))?;
+        // Theory of read procedure
+        // We write the BDA in two operations with a sync in between.  The write operation
+        // could fail (loss of power) for either write leaving sector(s) with potentially hard
+        // read errors. It's best to read each of the specific BDA blocks individually, to limit
+        // the probability of hitting a read error on a non-essential sector.
+
         let mut buf = [0u8; _BDA_STATIC_HDR_SIZE];
 
-        // FIXME: See https://github.com/stratis-storage/stratisd/pull/476
-        f.read(&mut buf)?;
-        Ok(buf)
+        /// Read a bda sector worth of data at the specified offset.
+        fn read_sector_at_offset<F>(f: &mut F, offset: usize, buf: &mut [u8]) -> io::Result<bool>
+        where
+            F: Read + Seek,
+        {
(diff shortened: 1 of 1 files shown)


tasleson:
Addresses https://github.com/stratis-storage/stratisd/issues/981

tasleson:
`BDA::read` updated to only return the BDA sectors.

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -43,18 +43,43 @@ enum MetadataLocation {
...
+        // We write the BDA in two operations with a sync in between.  The write operation
+        // could fail (loss of power) for either write leaving sector(s) with potentially hard
+        // read errors. It's best to read each of the specific BDA blocks individually, to limit
+        // the probability of hitting a read error on a non-essential sector.
+
+        let mut buf_loc_1 = [0u8; SECTOR_SIZE];
+        let mut buf_loc_2 = [0u8; SECTOR_SIZE];
+
+        /// Read a bda sector worth of data at the specified offset into buffer.
+        fn read_sector_at_offset<F>(
```
mulkieran: This function doesn't need to return a ```bool```, the value is completely unused. Might as well just return ```()```.

tasleson:
@mulkieran @agrover @trgill  I would really appreciate when a review is done that all review comments are completed at one time.  I think github review facilitates this well if utilized that way.  In this case @mulkieran please let me know with a separate comment when you're completely done reviewing this PR.  I will then make the appropriate updates & corrections at that time.  This will reduce PR churn, thank you!

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -43,18 +43,37 @@ enum MetadataLocation {
...
+        where
+            F: Read + Seek,
+        {
+            f.seek(SeekFrom::Start(offset as u64))?;
+            f.read_exact(&mut buf[offset..offset + SECTOR_SIZE])?;
+            Ok(true)
+        }
+
+        let loc_1_read_result = read_sector_at_offset(f, SECTOR_SIZE, &mut buf);
+        let loc_2_read_result = read_sector_at_offset(f, 9 * SECTOR_SIZE, &mut buf);
```
trgill: @tasleson I ran a test on this code using a flakey target to make the following sector 10 fail.  Read of sector 9 works fine. :+1:

mulkieran:
```BDA::read()``` returns ```Result<(chunk1, chunk2)>```. It returns an error if both fail. (That it more or less arbitrarily picks the first error is odd but not a fundamental problem).

If one succeeds but the other doesn't, one of the chunks it returns will be a buffer with data read from the device, the other will be an all-zero buffer. I believe. Remember that that all-zero buffer represents the fact that there was an error getting the raw bytes off the device.

Now, when ```StaticHeader::setup()``` gets an all zero buffer the result is ```Ok(None)```, meaning that the data was read, but no static header was on the device. So, if there was an error reading exactly one of the sections, one header will be anything, and the other will be ```Ok(None)```. This is the same value as if the region was read succesfully and was found not to have a Stratis magic number.

I think that there are six cases in the ```match``` expression in ```StaticHeader::setup()``` where ```Ok(None)``` is potentially meaning these two very different things: (1), we have no clue, because we couldn't even read the data (2), we're pretty sure this isn't a Stratis header, because there was no magic number.

Is the code simultaneously doing the right thing for both these meanings in all these six cases?

agrover:
oops, didn't see that last comment before merging. Opened #1016.

PR #1011, opened 2018-06-22
Master udev

mulkieran:
Supercedes #1005.

PR #1019, opened 2018-06-28
Master udev intermediate

mulkieran:
Supercedes #1011.

Adapted from #1006. The general approach is maintained. The design is significantly different.

Comments:
"Run Travis...' obviously will not exist in final copy.
"Allow renamed and removed lints..." is necessary for compilation but went into master after this branch. It should be removed in final copy.

The part of the code that uses udev properties to identify other devices for ```DevOwnership::Theirs``` constructor is deliberately incomplete. I think it is important to be able to inform users why we don't want to touch a device, but I don't undertand the approach that I'm trying to adapt, so I thought it best not to finish it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,8 @@ matrix:
           env: TASK=fmt-travis
 
 branches:
-    only: master
+    only:
+        - master
+        - fixup-branch
 
 script: make -f Makefile $TASK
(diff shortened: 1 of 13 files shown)


mulkieran:
Might as well ignore Travis errors...

mulkieran:
Superceded by #1023. Reverting seems to have made the whole thing fairly easy.

PR #1021, opened 2018-06-28
Upgrade to latest version of log and env_logger crates

agrover:
They clean up their APIs a bit. builder.init() can no longer fail, which
is nice.

Note: env_logger 0.5 pulls in ~4 new crates. We're going to have to deal with that in Fedora packaging, but that's just the way it goes.


agrover:
review @trgill @mulkieran ?

PR #1023, opened 2018-06-28
Master udev intermediate 2

mulkieran:
Supercedes #1019.

The idea is to try to work against master, by reverting previous commits.

### DIFF
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -38,7 +38,7 @@ use dbus::WatchEvent;
 
 use devicemapper::Device;
 
-use libstratis::engine::{get_udev_init, Engine, SimEngine, StratEngine};
+use libstratis::engine::{get_device_devnode, get_udev_init, Engine, SimEngine, StratEngine};
 use libstratis::stratis::{StratisError, StratisResult, VERSION};
 
 const STRATISD_PID_PATH: &str = "/var/run/stratisd.pid";
@@ -68,14 +68,10 @@ fn initialize_log(debug: bool) -> () {
 /// devicemapper::Device.
 fn handle_udev_add(event: &libudev::Event) -> Option<(Device, PathBuf)> {
     if event.event_type() == libudev::EventType::Add {
-        let device = event.device();
-        return device.devnode().and_then(|devnode| {
-            device
-                .devnum()
-                .and_then(|devnum| Some((Device::from(devnum), PathBuf::from(devnode))))
-        });
+        get_device_devnode(event.device())
+    } else {
+        None
     }
-    None
 }
 
 /// To ensure only one instance of stratisd runs at a time, acquire an
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -38,7 +38,7 @@ use dbus::WatchEvent;
 
 use devicemapper::Device;
 
-use libstratis::engine::{Engine, SimEngine, StratEngine};
+use libstratis::engine::{get_udev_init, Engine, SimEngine, StratEngine};
 use libstratis::stratis::{StratisError, StratisResult, VERSION};
 
 const STRATISD_PID_PATH: &str = "/var/run/stratisd.pid";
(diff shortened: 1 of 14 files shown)


mulkieran:
rebased.

mulkieran:
So, while we rely somewhat on udev to find Stratis devices, we rely exclusively on udev (at this time) to identify Stratis devices as belonging to some other system. If we have identified them as Stratis device, we try to store the values in the udev database that convinced us. What values are we actually looking up, and why?

mulkieran:
Travis tests are failing on two tests at the first invocation of ```find_all```. My hypothesis is that the failure is on the invocation of ```device_identifiers``` and that, even though udev has obtained a device number and devnode for an unclaimed block device, that devnode is somehow not openable. It seems unlikely that anything else is going wrong, based on the tests of identify that are clearly passing.

mulkieran:
bingo!

mulkieran:
Now that this problem has been fixed in ```find_all```, it remains to be fixed in ```identify()```. My guess is that it didn't show up in ```identify()``` before because ```identify()``` is just harder to test on lots of stuff than ```find_all()```.

mulkieran:
I won't be here tomorrow. I'm hoping for some preliminary feedback by Monday, when I get in.

agrover:
Carry over question from #1011, what's the benefit of having the global udev context? And what are the overall pros/cons of this PR vs what we have now? Thanks.

mulkieran:
Here's a list of ways in the which the current version is better than the old.

SIGNIFICANT:
1. The inconsistencies between ```find_all``` and ```identify``` were eliminated while at the same time ```find_all``` is now linear, rather than quadratic in the number of devices. (Allthough it should be obvious that my primary concern is correctness, I think that this sort of change indicates something significant about the algorithm being used as well.) Nonetheless,  if stratisd is supposed to be efficient, then this change is probably worth making for the sake of efficiency. The key problems in the previous implementation of ```find_all``` were:
  * It invoked the ```identify()``` method, via. ```is_stratis_device()``` to identify the device. This method itself uses libudev to find every block devices so it can locate the block device with a specified device node. This is what makes ```find_all``` quadratic. This method then returns a HashMap which maps all property names to property values. This is unnecessary and must be quite inefficient, also. The current version only invokes ```device_identifiers```, since all the other work done by ```identify()```, which was written to be invoked on a udev add event, has already been done by udev's enumerate mechanism.
* It made use of ``devnode_to_devno``` to get the device number. Since the method is using libudev anyway, getting the device number from udev, in exactly the same way as is done on a udev add event, makes much more sense. Also, using ```devnode_to_devno``` instead of doing the same thing as in ```identify()``` effectively masked a bug ```identify()```. By bringing ```find_all``` into line with ```identify()```, the tests that run on ```find_all```, many and easy to write, can locate the bug in ```identify()```, which is harder to test.

MORE IMPORTANT:
1. It doesn't have an unnecessarily large ```hw_lookup``` function which is called in exactly one place. Instead, ```hw_lookup``` is decomposed into two distinct and generally useful functions, ```get_udev_property``` and ```udev_block_device_apply```. These methods are reused in ```identify()``` and ```unclaimed()``` and can be used elsewhere.
2. It abstract ```get_device_devnode``` out of the ```handle_udev_add``` method and eliminates it. This becomes another generally useful function. It is now used both in the ```run()``` method and the ```find_all()``` method. In the previous PR, there was a lack of consistency between the two methods, ```find_all``` and ```identify``` and udev was not used to get the device number in ```find_all```,
so the reuse opportunity did not come up. That has been fixed. 

MINOR:
1. It doesn't require a ```[cfg_attr(rustfmt, rustfmt_skip)]``` anywhere or a confusing comment about a rustfmt bug. The trick here is just to import cmd and invoke the various executables as ```cmd::<name of function>``` here. This was how it was done in every other place in PR #1006. In my PR, I'm just consistent and always use ```cmd::<name of function>``` and the whole transient rustfmt bug problem (it's already fixed in master), does not have to be recorded for an indefinite time in our code base.

mulkieran:
rebased.

mulkieran:
Spurious test failures. Jenkins was running forever trying to clean up DM devices, so I killed it.

mulkieran:
Superceded by #1043.

PR #1027, opened 2018-07-02
BDA::read use O_DIRECT

tasleson:
It appears that when the read goes through the cache(s) in the
kernel that a bad sector on adjacent reads will cause a read of
a specific good sector to return EIO.  Use O_DIRECT to ensure
we read the specific sector.  This is complicated by the fact
that O_DIRECT read requires reads to be aligned to physical
sector geometry with a buffer that is aligned too, for DMA
transfers.

Note: This change is incomplete in that we haven't handled the
write path too.

### DIFF
--- a/src/engine/strat_engine/backstore/device.rs
+++ b/src/engine/strat_engine/backstore/device.rs
@@ -5,11 +5,12 @@
 // Functions for dealing with devices.
 
 use std::collections::HashMap;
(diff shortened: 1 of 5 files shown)


tasleson:
The write path can be problematic as we need to write in complete units of sector(s) which would necessitate a read, update, write cycle which complicates things quite greatly.

Issue #389, opened 2017-06-05
We should consider using lazy_static crate, judiciously


mulkieran:
We have had a bunch of constants that we would have defined differently if it weren't for the restrictions on expressions allowed in constants. Of course, these restrictions kind of serve a purpose, they prevent you from defiing your constants secretly at runtime w/ code that might fail.

mulkieran:
We have used ```lazy_static``` in one place in our source: for storing a hash of executable names to absolute paths. Now we know a little bit more about it.

The advantage of using ```lazy_static``` for other constants that we currently define in our code is:
If we defined a ```new()``` function for Sectors, Bytes, etc., we could construct a constant, actually static, value for them using ```new```. This might in turn allow us to make the one field of these types private which would eliminate the temptation of extracting that field and using it w/out the benefit of types, and probably getting more errors that way.

The drawback to using ```lazy_static```, though, is that the values have to be dynamically allocated, and must therefore be references. If they are references, then they have to be dereferenced when used, which is annoying since Sectors, Bytes, etc. are copy.

There seems to be no very great benefit to using ```lazy_static``` for these constants, and we've worked out the pitfalls with using ```lazy_static```, so we might as well close this.

PR #1028, opened 2018-07-03
Get rid of an unused public constant

mulkieran:
Calculate the Sectors value of BDA_STATIC_HDR_SIZE from the bytes value
in the one place it is needed.

Related: #389 but only in the distant sense that I was trying to see whether ```lazy_static``` crate could be used to clean up the constants defined in metadata.rs when I realized they could be cleaned up more directly.

### DIFF
--- a/src/engine/strat_engine/backstore/metadata.rs
+++ b/src/engine/strat_engine/backstore/metadata.rs
@@ -20,9 +20,7 @@ use super::super::device::SyncAll;
 
 pub use self::mda::{validate_mda_size, MIN_MDA_SECTORS};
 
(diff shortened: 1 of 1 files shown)


mulkieran:
It turns out that this ```BDA_STATIC_HDR_SECTORS``` was only exported in order to facilitate a test, so it should have been ```cfg(test)``` anyway. And it is no longer even exported for use in a test since commit d6948e16527f0a30b84ac2624653833e7de4a2.

mulkieran:
Rerunning the fmt task, it had an error.

PR #1030, opened 2018-07-05
Use proptest instead of quickcheck for specification based tests

mulkieran:
Switch from using quickcheck to using proptest. The basic motivation is that proptest seems a good deal easier to use to me; no doubt this is partially because I am a veteran Hypothesis user. The other motivation is that it seems to automatically provide a wider range of values in the correct type, and makes it easy to construct ones own strategies, by using strategy composition. Also, I had a nice experience with counter-example reduction.

This also changes two not explicitly specification based tests to one proptest test.

This uses not the most recent minor version of proptest, but the next most recent version. The most recent version can't compile w/ rustc 1.26.0, because it makes use of some new features in fairly recent stable version. I was able to do what I expected w/out the new version, so I don't think that sticking one version behind is a big  deal, at least for a bit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ tests/test_config.json
 
 # coverage output
 htmlcov
+
+# proptest
+proptest-regression
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,125 +1,134 @@
 [[package]]
 name = "aho-corasick"
-version = "0.6.3"
+version = "0.6.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "memchr 1.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr 2.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
 name = "ansi_term"
-version = "0.10.2"
+version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "winapi 0.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
+]
 
 [[package]]
 name = "atty"
-version = "0.2.6"
+version = "0.2.11"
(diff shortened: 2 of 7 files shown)


mulkieran:
Currently, this adds more lines than it subtracts because of Cargo.lock changes. But it subtracts more code lines than it adds because proptest is more succinct than QuickCheck, at least the version of QuickCheck that we were using. That's already one good thing about proptest. The second is that proptest is _much_ better at producing a proper range of values w/out extra work on the client's part. This is why the new version of ```mda_header``` test requires bounds on ```sec``` and ```nsec```. QuickCheck wasn't producing values in the full range of the specified type, but ```proptest``` was. Third, proptest really is Hypothesis-alike in terms of its minimization of failing examples. This has already been a big win, because it helped to quickly localize why tests were failing before assertions were reached when constructing the time stamp in ```mda_header```.

mulkieran:
Now it subtracts way more lines because it is way more succinct.

mulkieran:
This change means that the tests run in about 0.55 sec instead of 0.68 sec. This is a bit surprising actually, I had expected them to take longer.

agrover:
I like that this gets us off of using an old version of env_logger, but I'm wary of all the new build deps. This might need to wait until after 1.0 so we can make this switch and have the time to do all the needed packaging for Fedora.

mulkieran:
There are no new build deps, per se.

New dev deps are, according to cargo tree, proptest, bit-set, rusty-fork, fnv, wait-timeout, i.e., these are the added dev dependencies that aren't normal dependencies already.

mulkieran:
I guess this is blocked by the dependency question.

mulkieran:
rebased.

mulkieran:
Jenkins failure fixed by #1238.

mulkieran:
Travis is actually passed, although it looks like it is still pending.

mulkieran:
Cargo.lock is badly out of date, so I'm going to make up a new version with it redone.

mulkieran:
Closed for #1314.

PR #1039, opened 2018-07-11
Agrover buff log

mulkieran:


mulkieran:
Jenkins failure almost certainly spurious.

mulkieran:
Jenkins failures appear spurious.

mulkieran:
Trying a Jenkins rebuild.

agrover:
Lot of great stuff here. Only quibble is 2b42c28, I prefer configuration of both env_logger and buff_log staying in stratisd.rs. Can you yank that one please?

mulkieran:
I will explain more fully why 2b42c28, which seems like a retrograde step, is actually most probably the best choice.

It may look like a retrograde step for a few reasons. 1. The configuration is split into two parts, which seems less tidy. 2. A dependency on ```env_logger``` is brought into the ```buff_log``` module, which makes it seem like poor modularity. But...

1. These two parts are naturally split. The part that was moved into the ```buff_log``` implementation was introduced solely to support the mechanism of buffered logging. It may be further from the other configuration stuff, but it is closer to the code that requires it in order to work as the writer intended. That is why I think it belongs in the module where I put it.
2.  The dependency on ```env_logger``` in the ```buff_log``` module already existed before the import, it was just hidden. It would be nice to believe that because ```env_logger```'s Logger implements the Log abstraction, that the implementation of ```buff_log``` can ignore anything that is specific to ```env_logger``` and does not belong to Logger. But it can't, because it needs to call a log Builder method in order to configure the logger properly for its own operation. The dependency is sort of hidden if you bury it in the configuration method in stratisd.rs, my change just de-obfuscates what the real dependencies actually are.

mulkieran:
I have a new PR #1048 which modifies 2b42c28 so that it does not place the method in the ```buff_log``` module but does explicitly distinguish the different kinds of configuration. It is otherwise no different. I'm closing this one.

PR #1041, opened 2018-07-11
Format uuids as strings in Debug

agrover:
These commits implement Debug for various structs that include uuids, in order to have them be printed as strings instead of an array of numbers. This benefits #1026 but does not depend on it.

I originally looked at a solution more connected to Uuid itself, but since `uuid::Uuid` is not defined by us, we cannot reimplement `Debug` for it. We could wrap it in a new type (the "newtype" pattern) but that is a much larger change that we may want to pursue in the future, but not right now just to customize `Debug`.


src/engine/structures.rs
```diff
@@ -3,19 +3,31 @@
...
     name_to_uuid: HashMap<Name, Uuid>,
     items: HashMap<Uuid, (Name, T)>,
 }
 
+impl<T: fmt::Debug> fmt::Debug for Table<T> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.debug_map()
+            .entries(
+                self.iter().map(|(name, uuid, item)| {
+                    ((format!("{}", name), uuid.simple().to_string()), item)
```
mulkieran: ```name.to_string()``` is more Rustic.

mulkieran:
Redoing pylint task on Travis. I believe that with just committed pylint fixes it should now pass.

PR #1042, opened 2018-07-12
Rewrite get_blockdevs using the precondition that all devices in devnodes have been determined to belong to the designated pool

mulkieran:
This PR determines and then clearly states that, since ```get_blockdevs``` can only be invoked in a context where all the devices passed as arguments have already been determined to belong to Stratis and to a specific specified pool that it is reasonable:
1. To return an error if no BDA is found on any device.
2. To proceed as if it is certain that the block device has a BDA that indicates that it belongs to the designated pool.

This a preliminary step in a larger project to ensure that blockdevs are ordered at setup in the same order in which they were saved in the metadata. I would like to get this out of the way first.


mulkieran:
Jenkins acting up.

mulkieran:
Trying a Jenkins rebuild.

mulkieran:
rebased...which breaks the prior approval. Whoops!

mulkieran:
Looks like one of those transient failures. Redoing, hopefully it will pass this time.

PR #1046, opened 2018-07-12
Some preliminary backstore refactorings

mulkieran:
This PR is not nearly as big as it looks. The first PR is the largest share of line changes, but it just moves DataTier and CacheTier into separate files and makes their fields public, so it is really not that interesting. The rest are fairly straightforward changes. The biggest decision is that of changing the return type of ```alloc_space``` to conform to the expectation that all allocations will be made from a single linear device, and that therefore an allocation to satisfy a single request will necessarily consist of a single segment.


mulkieran:
Jenkins cleanup failure.

PR #1047, opened 2018-07-13
Avoid turning off rustfmt in code base

mulkieran:


mulkieran:
This revives #1010 slightly more cleanly. I had closed #1010 because I hoped that #1023, which avoids the problem entirely, might be merged soon. Since it won't be, I'm putting #1010 up in a somewhat revised form.

src/engine/strat_engine/backstore/device.rs
```diff
@@ -127,21 +127,19 @@ pub fn is_stratis_device(devnode: &Path) -> StratisResult<Option<PoolUuid>> {
 }
 
 #[cfg(test)]
-// rustfmt bug requires this?
-#[cfg_attr(rustfmt, rustfmt_skip)]
 mod test {
     use std::path::Path;
 
-    use super::super::super::cmd::{create_ext3_fs, udev_settle};
+    use super::super::super::cmd;
```
trgill: These changes related to adding cmd::xxx are intentional?  They caused problems with rustfmt?

mulkieran: rustfmt, for some reason, wasn't alphabetizing ```create_ext3_fs, udev_settle``` in the proper order. It was getting them backwards. This avoids having the imports that puzzle it so, rather than turning off rustfmt for the entire module.

trgill: Ok.  Thanks. :+1:

PR #1051, opened 2018-07-17
Omit explicit "object" in bases lists

mulkieran:
In Python 3, every class implicitly inherits from object, directly
or indirectly if some other class is specified.

Pylint 2.0, which is what is now being run by Travis CI via tox is Python 3
only and has introduced this new useless-object-inheritance check because
being Python 3 only, it can.


mulkieran:
So long as this passes, I'm going to just go ahead and merge it immediately, in order to unstick CI.

PR #1052, opened 2018-07-17
Move up to newer version of rustfmt

mulkieran:


mulkieran:
This, and the corresponding devicemapper PR are somewhat important to my workflow, as I will have to make some adjustments on my newly upgraded machine in order to use the older version of rustfmt, whereas if this is accepted, I won't have to spend time making those adjustments.

agrover:
I'm also using the newer rustfmt and so this would make my life easier. I've opened #1053 because that seems longer-term like a better solution.

mulkieran:
rebased one last time, very few changes.

mulkieran:
A Jenkins test fails because a thin device is busy when tearing down a pool. This is interesting, but obviously irrelevant to these changes. So, I'm going to merge this, and investigate the error somewhat more afterwards.

PR #1061, opened 2018-07-18
Rewrite get_blockdevs method

mulkieran:
This includes a bunch of tidies and clarifications and changes the specification so that the devices are sorted in the order in which they were previously recorded in the Stratis metadata.


mulkieran:
Whoops. Looks like there's a real bug. Cancelling Jenkins cause that would be a waste of time.

mulkieran:
Killing Jenkins test to add a couple of tidies.

PR #1064, opened 2018-07-20
Remove hard coded indices in event loop

tasleson:
Instead of utilizing fixed positions for specific file descriptor
types, lets utilize another vector which contains the type of fd
at that index.  This way we can place the fd in any index and still
process them.  This will allow us to add new fd types easier and
reduce the burden of keeping track of what goes where.

Additionally this change removes some fds churn because everytime we
went through the loop for any reason we would remove and add the dbus
pollfd structures.  This is only needed when we actually process
dbus file descriptors.

### DIFF
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -157,6 +157,39 @@ fn trylock_pid_file() -> StratisResult<File> {
     }
 }
 
+/// What type of file descriptor we have.
+#[derive(PartialEq)]
+enum FdType {
+    Udev,
+    Signal,
+    Engine,
+    #[cfg(feature = "dbus_enabled")]
+    Dbus,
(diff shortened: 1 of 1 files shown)


agrover:
I agree with the badness of hardcoded indices but this PR can be classified as a cleanup/refactoring PR unless we actually plan on adding many more fds. Maybe this should wait. Thoughts?

tasleson:
@agrover I was thinking this would be a nice improvement over having to deal with the hard coded indices when I do the PR for bringing the dbus interface up after dbus is available,  but we can certainly leave as is and deal with the indices values changing at run-time depending on if dbus is available or not.

tasleson:
Closing until we know this change is really needed.

PR #1065, opened 2018-07-23
Remove an unnecessary clippy allow

mulkieran:
This was added because previous version of error-chain had some newly
introduced clippy errors that it didn't either allow or avoid.
Since the upgrade to the new version error-chain has fixed this problem,
and so stratis can skip this allow.


mulkieran:
Trivial if it passes.

PR #1066, opened 2018-07-23
Master backstore

mulkieran:


mulkieran:
Hoping for some interesting failure modes here. Backstore is now totally empty, always, so that needs to be fixed.

mulkieran:
Alas, failure modes are uninteresting. It is illegal to create an empty linear device.

mulkieran:
We hope that failure modes are more interesting now that creating the device is deferred until an allocation actually occurs.

mulkieran:
Broke some obsoleted invariants.

mulkieran:
Excellent! These are problems that can't be solved until ```alloc_space``` is rewritten.

mulkieran:
The next field in DataTier must divide its meaning in two:

* next, which exists in DataTier must move back to Backstore and handle allocations from the linear dev
* another sort of structure must end up in DataTier and represent a starting point for allocations from the blockdevs

Backstore must request chunks from DataTier's blockdevs, add them to the cap device, and then return its own chunks from the cap device to the thin pool.

The hard part here is handling the allocations in the data tier, for they are more complex.

mulkieran:
Superceded by various subsequent work.

Issue #846, opened 2018-03-16
Stratisd: Sequencing issue @ boot


tasleson:
Stratisd when started will assemble block devices and provide a dbus API for user space.  We want to assemble the block devices before the system starts mounting file systems, so that Stratis block devices in the `fstab` are mounted.  The existing service [file](https://github.com/stratis-storage/stratisd/blob/master/stratisd.service) is constructed with both requirements in place, but they are incompatible.  In it's current form when we boot we get 1 or more errors in the journal that look like:

```
Mar 15 16:45:35 strat systemd[1]: ldconfig.service: Found ordering cycle on local-fs.target/start
Mar 15 16:45:35 strat systemd[1]: ldconfig.service: Found dependency on local-fs-pre.target/start
Mar 15 16:45:35 strat systemd[1]: ldconfig.service: Found dependency on stratisd.service/start
Mar 15 16:45:35 strat systemd[1]: ldconfig.service: Found dependency on dbus.socket/start
Mar 15 16:45:35 strat systemd[1]: ldconfig.service: Found dependency on sysinit.target/start
Mar 15 16:45:35 strat systemd[1]: ldconfig.service: Found dependency on systemd-update-done.service/start
Mar 15 16:45:35 strat systemd[1]: ldconfig.service: Found dependency on ldconfig.service/start
Mar 15 16:45:35 strat systemd[1]: ldconfig.service: Job local-fs.target/start deleted to break ordering cycle starting with ldconfig.ser
vice/start
```

If I edit the service file and remove the dbus requirement, then the daemon will start up before the file systems get mounted and exit with:

```
Mar 15 16:56:25 strat stratisd[341]: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
Mar 15 16:56:27 strat systemd[1]: stratisd.service: Main process exited, code=exited, status=1/FAILURE
Mar 15 16:56:27 strat systemd[1]: stratisd.service: Failed with result 'exit-code'.
```
Apparently because dbus services are not yet started.

Possible solutions:
1. We add an option to stratisd to assemble and exit after setting up the dm devices.  We create a `oneshot` service file to run it before file systems get mounted.  We also have the service file that will start the service as a dbus service daemon at the appropriate time.

2. We modify the daemon to periodically check if it's able to bring up the dbus interface and do so when it is available.  This would require us to slow poll in the event loop until it was available.

Other solutions...?  Maybe there is a different target that will get us where we need to be...

[...]

PR #1067, opened 2018-07-24
Bring dbus up when we can

tasleson:
We want stratisd to come up before filesystems get mounted, so
that a user can use Stratis backed filesystems in fstab.  However,
dbus is not available at this time and we need to be able to start
and run without dbus.  When this happens we will periodically
check to see if we can bring up the dbus API.

Fixes: https://github.com/stratis-storage/stratisd/issues/846

### DIFF
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -165,6 +165,11 @@ fn run(matches: &ArgMatches, buff_log: &buff_log::Handle<env_logger::Logger>) ->
     // Ensure that the debug log is output when we leave this function.
     let _guard = buff_log.to_guard();
 
+    // Even if dbus is enabled at compile time, it may not be available at all times depending
+    // on the environment we are running in.
+    #[cfg(feature = "dbus_enabled")]
+    let mut dbus_handle: Option<libstratis::dbus_api::DbusConnectionData> = None;
+
     // Setup a udev listener before initializing the engine. A device may
     // appear after the engine has read the /dev directory but before it has
     // completed initialization. Unless the udev event has been recorded, the
@@ -236,16 +241,12 @@ fn run(matches: &ArgMatches, buff_log: &buff_log::Handle<env_logger::Logger>) ->
 
     let eventable = engine.borrow().get_eventable();
 
-    let poll_timeout = match eventable {
-        Some(ref evt) => {
-            fds.push(libc::pollfd {
-                fd: evt.get_pollable_fd(),
-                revents: 0,
-                events: libc::POLLIN,
-            });
-            -1
-        }
-        None => 10000,
+    if let Some(ref evt) = eventable {
+        fds.push(libc::pollfd {
+            fd: evt.get_pollable_fd(),
+            revents: 0,
+            events: libc::POLLIN,
+        });
     };
 
     #[cfg(feature = "dbus_enabled")]
@@ -254,22 +255,6 @@ fn run(matches: &ArgMatches, buff_log: &buff_log::Handle<env_logger::Logger>) ->
     } else {
         FD_INDEX_ENGINE
     };
-
-    #[cfg(feature = "dbus_enabled")]
-    let (dbus_conn, mut tree, base_object_path, dbus_context) =
-        libstratis::dbus_api::connect(Rc::clone(&engine))?;
-
-    #[cfg(feature = "dbus_enabled")]
-    for (_, pool_uuid, mut pool) in engine.borrow_mut().pools_mut() {
-        libstratis::dbus_api::register_pool(
-            &dbus_conn,
-            &dbus_context,
-            &mut tree,
-            pool_uuid,
-            pool,
-            &base_object_path,
-        )?;
-    }
 
     log_engine_state(&*engine.borrow());
     alarm(STRATISD_ALARM_SECONDS);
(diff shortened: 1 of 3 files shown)


src/bin/stratisd.rs
```diff
@@ -365,30 +364,64 @@ fn run(matches: &ArgMatches, buff_log: &buff_log::Handle<env_logger::Logger>) ->
...
-            fds.extend(dbus_conn.watch_fds().iter().map(|w| w.to_pollfd()));
+                    // Add dbus FD to fds as dbus is now available.
+                    fds.extend(handle.connection.watch_fds().iter().map(|w| w.to_pollfd()));
+                    dbus_handle = Some(handle);
+                }
+            }
         }
 
+        #[cfg(feature = "dbus_enabled")]
+        let poll_timeout = timeout(dbus_handle.is_none(), eventable.is_none());
```
mulkieran: I'm guessing that all the arguments to timeout above and below have the wrong values, except for the literal ```false```. Assuming ```false``` is a correct argument, then it makes sense for the first argument to evaluate to ```false``` if D-Bus is _not_ enabled. So, in the call to timeout above, it makes more sense for the first argument to be ```dbus_handle.is_some()```, which will evaluate to true if stratisd has a D-Bus connection, but false otherwise. I think that a similar argument holds for the second argument.

src/bin/stratisd.rs
```diff
@@ -157,6 +157,18 @@ fn trylock_pid_file() -> StratisResult<File> {
     }
 }
 
+/// Return desired timeout based on if dbus timeout should be selected, non-eventable timeout
+/// selected or default selected
```
mulkieran: I believe that this function should be better documented, to explain the meaning of the arguments and also the reason behind the choices of the return values. This will help understand if it is called correctly and also inform any subequent revisions of the values that may become necessary.

src/bin/stratisd.rs
tasleson: The arguments are correct and the function is doing the correct thing based on them.  However, I can see the confusion.  The `timeout` function simply selects the returned value in a prioritized order of shorter poll when dbus is compiled in but not available vs. an engine that isn't eventable.  When `dbus_handle.is_none()` is `true` dbus is not available, thus we want to the time out to be shorter, 1 second.  When `dbus_handle.is_none()` is true, then we evaluate if engine is eventable.  If `eventable.is_none()` is true then the engine isn't eventable and we want the timeout to be 10 seconds.  However, I need to ask @agrover why this was selected as the comments indicate that  when we aren't eventable that the call to the engine is a no-op.  Lastly when both dbus and eventable are both `false` we return -1 which is to wait forever on socket activity.

Obviously I need to re-work this to make it less confusing, so I'll see what I can do.  Passing in the reverse and negating the comparisons in `timeout` may be better.

src/dbus_api/api.rs
```diff
@@ -205,25 +205,28 @@ fn register_pool_dbus(
...
-) -> Result<
-    (
-        Connection,
-        Tree<MTFn<TData>, TData>,
-        dbus::Path<'a>,
-        DbusContext,
-    ),
-    dbus::Error,
-> {
+pub fn connect<'a>(engine: Rc<RefCell<Engine>>) -> Result<DbusConnectionData<'a>, dbus::Error> {
```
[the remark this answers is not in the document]
tasleson: Agreed, will update

src/bin/stratisd.rs
tasleson: With Andy's comment at the meeting, I think we can make this much simpler.

tasleson:
Reduced poll timeout calculation complexity and removed type complexity allow.

tasleson:
> However, it would be tidier if you started a new PR with all commits squashed into one, or with two commits, one for the refactoring change which changed the return type of connect, and one for all the rest of the changes.
Merge state

Agreed, I can certainly squash.  I've been leaving things as separate commits to allow easier review of subsequent changes.  However, with the auto merge thing, I guess that's not really appropriate.

tasleson:
Squashed

tasleson:
> But to be clear, I had requested that this PR be closed, and a new PR, with changes squashed, be put up for review, not that the commits in this PR just be squashed.
Merge state

I posed this question on IRC and @agrover really said there isn't a defacto process to follow.  What in your opinion is the advantage of creating a new PR which is squashed vs. squashing the existing PR?

mulkieran:
For PRs that are being changed due to review, and are fairly complex, it makes sense just to add change commits to the end of the PR. I can explain the reasoning behind this more fully if necessary.

For PRs that are actually committed to the repo, it makes sense to separate the PR into distinct and coherent individual commits, as far as is possible and within limits of difficulty of the task. I can expand on this more fully if necessary.

It makes sense to make these two PRs separate because:
1. The first type of PR provides a handy record of the reasoning behind decisions made which can be lost or confused by a force push.
2. It provides the person who submitted the PR with an opportunity to reassess their PR in light of the changes that they have been asked to make before making the final copy.
3. It is a courtesy to a hypothetical second reviewer who may still have had their reservations about the PR, even if the first reviewer said that in their opinion it was good to go.

tasleson:
Re-based to correct merge conflict induced by merge of 9b720f99dd651574098af37146ef9472a9a481c9

mulkieran:
It is not clear to me why stratis-cli tests managed to succeed with this change, as they failed on my machine with this change.

PR #1071, opened 2018-07-26
Make allocation from block devices lazy

mulkieran:
Previously, all available segments in data block devices were greedily added to the cap device. Now, they are allocated lazily, on request by the thin pool.

This change required a relaxation of two previous invariants:
1. There was always either a linear DM device or a cache DM device belonging to the backstore. If segments are allocated only on demand, and demand has not occurred, then there are zero segments allocated. It is impossible to make a completely empty linear device, so after the backstore is created and before the thinpool requests space, there is neither a linear nor a cache DM device in the backstore.
2. All segments in the data blockdevs were always fully allocated to the backstore. This is not true any more.

The metadata checks were changed to handle the relaxing of these invariants.

Relaxing the two invariants increased the complexity of handling of the two alternate DM devices. Therefore, all such handling was moved into the backstore, making the *Tier structs just manage block devices and segment allocations. This changed the signatures of some methods in the *Tier implementations from ```Result<T>``` to ```T```.

The last part, making allocation lazy, required that backstore handle making or updating DM devices in its ```alloc_space``` method. This meant that this method could return an error, so its signature was changed accordingly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ test-loop:
 	sudo env "PATH=${PATH}" RUSTFLAGS='-D warnings' RUST_BACKTRACE=1 RUST_TEST_THREADS=1 cargo test loop_
 
 test-real:
-	sudo env "PATH=${PATH}" RUSTFLAGS='-D warnings' RUST_BACKTRACE=1 RUST_TEST_THREADS=1 cargo test real_
+	sudo env "PATH=${PATH}" RUSTFLAGS='-D warnings' RUST_BACKTRACE=1 RUST_TEST_THREADS=1 cargo test real_test_full_pool
 
 test-travis:
 	sudo env "PATH=${PATH}" RUSTFLAGS='-D warnings' RUST_BACKTRACE=1 RUST_TEST_THREADS=1 cargo test travis_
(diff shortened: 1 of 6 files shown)


mulkieran:
Expecting a few changes, so not running stratisd tests yet.

mulkieran:
minor fix to an invariant.

mulkieran:
It _might_ pass all tests.

mulkieran:
Forgot to adjust the backstore invariant, that might have been a rebase mistake.

mulkieran:
Whoops! Got some other invariants to fix as well.

mulkieran:
Looks like fedora-server-2 isn't cleaning up properly...so all stratisd tests are failing.

mulkieran:
This will probably fail somewhere because ```Backstore::alloc_space may anytime require updating the cap device, which means suspend the thinpool.

mulkieran:
Now being more selective about when pool is suspended and resumed. When simply adding datadevs to the pool, it is not necessary to suspend and resume, since this action just tags the devs with a Stratis static header and does some bookkeeping in the data tier. For cachedevs, it is still necessary to suspend the thinpool, because these are immediately added to the cache.

mulkieran:
```real_test_full_pool``` is running for a very long time. Need a hypothesis as to why that should be.

mulkieran:
Superceded by #1079 and #1094 .

PR #1080, opened 2018-08-02
Fallback to smaller request when extending the thin pool

mulkieran:
Instead of having a hard fail if the requested amount is not available, fall back on a smaller request.

Choosing to try for a smaller amount is the only reasonable course. However, this PR introduces the possibility of a situation where an amount was successfully allocated to a thin pool device but that amount was smaller than the calculated amount requested. This PR calls out that issue in a TODO, but does not address it. It is expected that it will be addressed in some form by subsequent PRs, including #1083.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,6 @@ libmount = "0.1.10"
 error-chain = "0.12"
 libudev = "0.2.0"
 lazy_static = "1.0.0"
-either = "1.1.0"
 
 [dependencies.uuid]
 version = "0.6"
@@ -33,6 +32,7 @@ features = ["serde", "v4"]
 [dev-dependencies]
 quickcheck = "0.6"
 loopdev = "0.2"
+either = "1.1.0"
 
 [features]
 default = ["dbus_enabled"]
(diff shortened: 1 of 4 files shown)


mulkieran:
Travis CI failure was spurious. stratisd failures were real, but due to a simple bug, now fixed, most likely.

mulkieran:
Fixed another simple bug...

mulkieran:
There was an error in ```test_meta_expand```, the subsequent errors seems spurious.

mulkieran:
@agrover I've updated to get rid of ```fn try_to_allocate```.

mulkieran:
@agrover 

It might be better to flatten out ```alloc_space``` into two separate methods:

1. alloc_one_segment_max -- which allocates just one, but falls back to smaller value if necessary
2. alloc_multiple_segments_exactly -- which allocates multiple segments, but fails if it can not allocate them all exactly

This would just require lifting the locally defined methods in ```alloc_space``` to the top level, getting rid of ```alloc_space``` entirely, and not using ```either```.

That seems less tortured. I believe you suggested it in our meeting anyway.

agrover:
worth a shot!

mulkieran:
Closing in favor of #1089.

PR #1088, opened 2018-08-07
Add/change properties to D-Bus filesystem interface

agrover:
Does 3 things:
1. Moves some filesystem initialization/setup inside filesystem (from ThinPool)
1. Adds support for Created, Used, and MountPoints filesystem properties in D-Bus
1. Changes filesystem Devnode property to return `/dev/stratis/<poolname>/<fsname>` instead of `/dev/dm-<number>`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,16 +14,21 @@ language: rust
 
 matrix:
     include:
+        # Verify formatting of Rust code
         - rust: stable
           env: TASK=fmt-travis TARGET=x86_64-unknown-linux-gnu
+        # Verify that source builds with standard defaults
         - rust: stable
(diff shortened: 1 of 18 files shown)


agrover:
reference: https://github.com/stratis-storage/stratis-cli/issues/101

agrover:
merged #1100 and added a docstring for devpath_from_names.

agrover:
Pulled in your additional edits, looks good!

Added tests. Didn't actually parse the result of Created since that seems to need the "dateutil" library, and wasn't sure if we want tests to have external dependencies (but if we did, it would be easy to do).

I hear what you're saying about efficiency but yeah, now is not the time to worry about that.

PR #1089, opened 2018-08-07
Rewrite extension of devices belonging to thinpool

mulkieran:
Simultaneously split Backstore::alloc_space into two separate methods
with distinct behaviors. In one case, where a thinpool is being created,
it makes sense to request various different lengths and to fail if the
total request is not satisfied. In the other, when extending a thinpool
device, it makes more sense just to request a maximum value and then
accept the best that the backstore can provide up to that maximum value.

The basic goal is to avoid a hard fail if the initial request can not be
satisifed. The reason for that is that, while it is possible to have a
precise idea of the maximum available in the current setup, as additional
layers are added, that value will come to be less and less reliable. It
is better to request what is wanted, and then to receive as much as can
be made available, modulo other significant constraints. There is nothing
to say that this approach can not be modified somewhat so that the request
itself can be adjusted based on other factors, but avoiding a hard fail
is still a good idea.

This PR also unifies meta- and data- device extending. They do not differ,
except in the segments they update and the devicemapper-rs methods that
they call. It is better to avoid this duplication for better maintainability.
If they start to deviate, then the code can be separated as appropriate.

This PR also calculates the extend size with reference to the space used,
the total space, and the low water mark to ensure that a specific goal for
extension is reached. This is done in calculate_request().

### DIFF
--- a/src/engine/strat_engine/backstore/backstore.rs
+++ b/src/engine/strat_engine/backstore/backstore.rs
@@ -4,6 +4,7 @@
 
 // Code to handle the backing store of a pool.
 
+use std::cmp;
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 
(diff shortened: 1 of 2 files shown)


mulkieran:
Supercedes #1080.

mulkieran:
First commit source is identical to #1080, except for one typo fix in docs: "%" -> "/".

Also, I added the test of newly added blockdev method.

mulkieran:
Blocks #1066.

agrover:
lgtm, but can we consider some shorter names? There's already a convention that "alloc" means all or none and "request" means all, partial, or none in the existing code, so following that can we just call them Backstore::alloc() and ::request()? Single vs multiple segments can be understood from fn signature and/or docstring.

PR #1092, opened 2018-08-07
Use write mode check instead of diff for cargo fmt

mulkieran:
Semantics of diff seems to have changed. It used to return non-zero code
if there was a difference, now it doesn't. Semantics of check seems to
be what we want for a test; it seems to behave like diff used to.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ fmt:
 fmt-travis:
 	rustup default 1.27.0
 	rustup component add rustfmt-preview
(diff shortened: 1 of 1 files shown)


mulkieran:
https://travis-ci.org/stratis-storage/stratisd/builds/413160546?utm_source=github_status&utm_medium=notification is what alerted me to the problem. The "fmt" test is passing, but their is a difference between the code and the expected formatting.

PR #1094, opened 2018-08-08
Rewrite extension of devices belonging to thinpool

mulkieran:
Rewrite these methods to support different ways of allocating from the backstore. In one case, when setting up a new thinpool, allocating exactly makes sense. In the other, when extending a  meta or a data device, requesting a maximum amount makes sense. Unify the code for extending a meta and a data device as much as possible. Come up with a principled way for choosing the amount by which to extend the data or meta data device respectively.


mulkieran:
Supercedes #1089.

PR #1097, opened 2018-08-08
Allocate segments to cap device lazily

mulkieran:
Previously, all available segments in data block devices were greedily added to the cap device. Now, they are allocated lazily, on request by the thin pool.

This change required a relaxation of two previous invariants:
1. There was always either a linear DM device or a cache DM device belonging to the backstore. If segments are allocated only on demand, and demand has not occurred, then there are zero segments allocated. It is impossible to make a completely empty linear device, so after the backstore is created and before the thinpool requests space, there is neither a linear nor a cache DM device in the backstore.
2. All segments in the data blockdevs were always fully allocated to the backstore. This is not true any more.

The metadata checks were changed to handle the relaxing of these invariants.

Relaxing the two invariants increased the complexity of handling of the two alternate DM devices. Therefore, all such handling was moved into the backstore, making the *Tier structs just manage block devices and segment allocations. This changed the signatures of some methods in the *Tier implementations from ```Result<T>``` to ```T```.

The last part, making allocation lazy, required that backstore handle making or updating DM devices in its ```alloc``` and ```request``` methods. This meant that these methods can return an error, so their signatures are changed accordingly.


mulkieran:
Supercedes #1071.

mulkieran:
I am devoting some thought to testing this better, but I'm putting it into preliminary review as is.

mulkieran:
So far, every test failure has been due to the test being poorly written, sigh!

mulkieran:
Bad test again. I forgot that the cache metadata takes up space.

mulkieran:
First interesting failure at https://stratis-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/stratisd/1754/ is ```real_test_add_cachedevs```. Seems relevant. Note that all Travis tests, including the two newly added ones succeeded.

mulkieran:
Some minor updates. Also, would be interesting to see if stratisd test failures repeat themselves. They look a bit spurious, surprisingly.

mulkieran:
Restarted stratisd test, just a complete Jenkins failure.

PR #1098, opened 2018-08-09
Move all handling of top-level DM device out of the Data and Cache Tiers

mulkieran:
Previously DM devices were constructed or updated in *Tier methods. Now, all updating of DM devices is done in the Backstore, using the data in the various Tier structs.

This does turn out to be more concise, which is good.

The principle motivation is that it is necessary in PR #1097 to relax certain invariants that currently hold for the DM devices. In particular, it was previously the case that the backstore held either a linear or a cache device. Now it may hold neither. It is easier to handle this wider range of possibilities in the backstore than when the code is distributed among three files.

One apparent drawback of this change is that the Data or CacheTier segments values are set before the DM devices are constructed or reconstructed. However, if the DM devices are not constructed properly, then the corresponding backstore method will fail, and the metadata will not be written so the change to the segments value is not visible. So the externally visible behavior of stratisd is the same with this change as w/out it.

### DIFF
--- a/src/engine/strat_engine/backstore/backstore.rs
+++ b/src/engine/strat_engine/backstore/backstore.rs
@@ -32,6 +32,47 @@ use super::setup::get_blockdevs;
 /// typical size.
 const CACHE_BLOCK_SIZE: Sectors = Sectors(2048); // 1024 KiB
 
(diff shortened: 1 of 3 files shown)


mulkieran:
Blocks: #1097

Issue #908, opened 2018-05-01
Make sure Stratis DTRT on multipath devices


agrover:
Verify that Stratis's initial device scan and ongoing usage works for multipath devices, and that it doesn't use one of the mpath subdevices by accident.

agrover:
fixed by #1124

PR #1099, opened 2018-08-09
 dm-multipath: Initial support

tasleson:
Verify that a given device as supplied by user is not one of the individual
paths participating in a multipath configuration.  Additionally when
walking the udev db at start up, ensure we filter out these individual
paths too.

### DIFF
--- a/src/engine/strat_engine/backstore/device.rs
+++ b/src/engine/strat_engine/backstore/device.rs
@@ -54,11 +54,13 @@ pub enum DevOwnership {
     Theirs(String), // String is something useful to give back to end user about what's on device
 }
 
-/// Returns true if a device has no signature, yes this is a bit convoluted.  Logic gleaned from
-/// blivet library.
+/// Returns true if a device has no signature and is not one of the paths of a multipath device,
+/// yes this is a bit convoluted.  Logic gleaned from blivet library.
 fn empty(device: &HashMap<String, String>) -> bool {
-    !((device.contains_key("ID_PART_TABLE_TYPE") && !device.contains_key("ID_PART_ENTRY_DISK"))
-        || device.contains_key("ID_FS_USAGE"))
+    !device.contains_key("DM_MULTIPATH_DEVICE_PATH")
+        && !((device.contains_key("ID_PART_TABLE_TYPE")
+            && !device.contains_key("ID_PART_ENTRY_DISK"))
+            || device.contains_key("ID_FS_USAGE"))
 }
 
 /// Generate some kind of human readable text about what's on a device.
@@ -78,9 +80,7 @@ fn signature(device: &HashMap<String, String>) -> String {
 /// Determine what a block device is used for.
 pub fn identify(devnode: &Path) -> StratisResult<DevOwnership> {
     if let Some(device) = get_udev_block_device(devnode)? {
-        if device.contains_key("DM_MULTIPATH_DEVICE_PATH") {
-            Ok(DevOwnership::Theirs(String::from("multipath path")))
-        } else if empty(&device) {
+        if empty(&device) {
             // The device is either really empty or we are running on a distribution that hasn't
             // picked up the latest libblkid, lets read down to the device and find out for sure.
             // TODO: At some point in the future we can remove this and just return Unowned.
@@ -91,6 +91,8 @@ pub fn identify(devnode: &Path) -> StratisResult<DevOwnership> {
             } else {
                 Ok(DevOwnership::Unowned)
             }
+        } else if device.contains_key("DM_MULTIPATH_DEVICE_PATH") {
+            Ok(DevOwnership::Theirs(String::from("multipath path")))
         } else if device.contains_key("ID_FS_TYPE") && device["ID_FS_TYPE"] == "stratis" {
             // Device is ours, but we don't get everything we need from udev db, lets go to disk.
             if let Some((pool_uuid, device_uuid)) = StaticHeader::device_identifiers(
--- a/src/engine/strat_engine/backstore/device.rs
+++ b/src/engine/strat_engine/backstore/device.rs
@@ -78,7 +78,9 @@ fn signature(device: &HashMap<String, String>) -> String {
 /// Determine what a block device is used for.
 pub fn identify(devnode: &Path) -> StratisResult<DevOwnership> {
     if let Some(device) = get_udev_block_device(devnode)? {
-        if empty(&device) {
+        if device.contains_key("DM_MULTIPATH_DEVICE_PATH") {
+            Ok(DevOwnership::Theirs(String::from("multipath path")))
+        } else if empty(&device) {
             // The device is either really empty or we are running on a distribution that hasn't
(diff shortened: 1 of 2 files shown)


tasleson:
Fixes: https://github.com/stratis-storage/stratisd/issues/908

tasleson:
@agrover I made the code for determining the multipaths path quite resilient to errors and if it fails, the user will still get a message to investigate multipath.  I'm not sure I fully understand your concerns with this change.

tasleson:
@agrover As requested I've removed the functionality of returning what the multipath should be and opened an issue to incorporate the functionality at some later date:

https://github.com/stratis-storage/stratisd/issues/1113

src/engine/strat_engine/backstore/setup.rs
```diff
@@ -34,12 +34,14 @@ pub fn find_all() -> StratisResult<HashMap<PoolUuid, HashMap<Device, PathBuf>>>
         match devnode_to_devno(&devnode)? {
             None => continue,
             Some(devno) => {
-                is_stratis_device(&devnode)?.and_then(|pool_uuid| {
+                if let Some((pool_uuid, _)) = StaticHeader::device_identifiers(
+                    &mut OpenOptions::new().read(true).open(&devnode)?,
+                )? {
```
agrover: It looks like there's a bug in `get_stratis_block_devices` so that this is not safe to do.

`.filter(|x| is_stratis_device(&x).ok().is_some())` This converts the Result to an Option (resulting in an `Option<Option<uuid>>` and then checks if the outermost option is some, but it's only a Stratis device if the interior Option is Some. Try `.filter(|x| is_stratis_device(&x).unwrap_or(None).is_some())`.

src/engine/strat_engine/backstore/blockdevmgr.rs
```diff
@@ -394,6 +394,14 @@ fn initialize(
             };
             match ownership {
                 DevOwnership::Unowned => add_devs.push((dev, (devnode, dev_size, f))),
+                DevOwnership::Multipath(_) => {
```
agrover: Why `DevOwnership::Multipath` instead of `DevOwnership::Theirs("multipath")`? In both cases the course of action for users of the enum will be the same -- don't allow its use by Stratis.

src/engine/strat_engine/backstore/util.rs
```diff
@@ -72,7 +82,14 @@ pub fn get_stratis_block_devices() -> StratisResult<Vec<PathBuf>> {
...
         .scan_devices()?
-        .map(|x| x.devnode().expect("block devices have devnode").into())
+        .filter(|dev| {
+            dev.property_value("DM_MULTIPATH_DEVICE_PATH").is_none() && dev.devnode().is_some()
+        })
+        .map(|x| {
+            x.devnode()
+                .expect("we filtered out missing devnode entries")
+                .into()
+        })
```
agrover: for both `get_stratis_block_devices` and `get_all_empty_devices` instead of filtering on `dev.devnode.is_some()` and then doing a `map()` with `expect()`, you can instead do a filter_map:
```
     .filter_map(|i| i.devnode().map(|d| d.into()))
```
This prevents the unfortunate situation where I didn't immediately see the filter was filtering on that (it's all the way over on the right of the filter line (c'mon rustfmt, step it up!)) , and also puts all the devnode stuff, both the filtering and the mapping, together.

src/engine/strat_engine/backstore/blockdevmgr.rs
tasleson: Well now that I've taken out the retrieval of what the actual multipath path actually is, having this as a separate enum value to distinguish that with the corresponding text is not longer useful.  I'll rip this out too.

src/engine/strat_engine/backstore/util.rs
```diff
@@ -82,7 +89,7 @@ pub fn get_stratis_block_devices() -> StratisResult<Vec<PathBuf>> {
 
         Ok(get_all_empty_devices()?
             .into_iter()
-            .filter(|x| is_stratis_device(&x).ok().is_some())
+            .filter(|x| is_stratis_device(&x).unwrap_or(None).is_some())
```
mulkieran: AFAICT, this is just a patch to make ```get_stratis_block_devices()``` match its specification a bit better and doesn't really have anything to do with multipath. Please remove it from this commit.

src/engine/strat_engine/backstore/util.rs
```diff
@@ -55,11 +60,12 @@ fn get_all_empty_devices() -> StratisResult<Vec<PathBuf>> {
...
-            !((dev.property_value("ID_PART_TABLE_TYPE").is_some()
-                && dev.property_value("ID_PART_ENTRY_DISK").is_none())
-                || dev.property_value("ID_FS_USAGE").is_some())
+            dev.property_value("DM_MULTIPATH_DEVICE_PATH").is_none()
+                && !((dev.property_value("ID_PART_TABLE_TYPE").is_some()
+                    && dev.property_value("ID_PART_ENTRY_DISK").is_none())
+                    || dev.property_value("ID_FS_USAGE").is_some())
         })
-        .map(|i| i.devnode().expect("block devices have devnode").into())
+        .filter_map(|i| i.devnode().map(|d| d.into()))
```
mulkieran: AFAICT, filtering on the devnode has nothing to do w/ multipath. Please remove it from this commit. It occurs at this line and also line 82.

src/engine/strat_engine/backstore/device.rs
```diff
@@ -78,7 +78,9 @@ fn signature(device: &HashMap<String, String>) -> String {
 /// Determine what a block device is used for.
 pub fn identify(devnode: &Path) -> StratisResult<DevOwnership> {
     if let Some(device) = get_udev_block_device(devnode)? {
-        if empty(&device) {
+        if device.contains_key("DM_MULTIPATH_DEVICE_PATH") {
```
mulkieran: For consistency it would make more sense to extend function ```empty``` to include a match on ```DM_MULTIPATH_DEVICE_PATH``` and to move this if block below the ```if empty``` block. This would make the code better consistent with the the code in ```get_all_empty_devices``` which is changed at line 63 to filter out multipath devices using ```DM_MULTIPATH_DEVICE_PATH```.

src/engine/strat_engine/backstore/util.rs
```diff
@@ -33,9 +34,12 @@ pub fn get_udev_block_device(
     let mut enumerator = libudev::Enumerator::new(&context)?;
     enumerator.match_subsystem("block")?;
 
+    // Get canonical form to ensure we do correct lookup in udev db
+    let canonical = fs::canonicalize(dev_node_search)?;
```
mulkieran: Why is it necessary to canonicalize the path? Does this have something to do with multipath? If not, it should be taken out of this commit.

tasleson:
Rebased, conflicts corrected

tasleson:
replaced by https://github.com/stratis-storage/stratisd/pull/1124

PR #1101, opened 2018-08-09
Correct poll_timeout, remove dbus type from service file

tasleson:
While verifying a service file correction, I discovered that the `poll_timeout`
was incorrect and we were using `-1` when we had no dbus services.  
Thus the dbus interface was not being brought up until we got a signal 
or some other event made us take a pass through the event loop.

We have conflicting requirements in our service file.  We cannot
be both a dbus service and want to run before file systems get
mounted as dbus is not available until after file systems have been
mounted.  Thus we cannot advertise as providing a dbus service in
the service file and still start up before file systems  get
mounted in fstab.

Notes:
* Tested in VM, stratis backed FS came up with FS UUID in fstab
* For my VM it takes about 3 seconds from the time we start stratisd until we can bring up the dbus interface


mulkieran:
Let's fast-track at least the code part of the fix, since it is preventing testing the stratis-cli. I do not know what made the PR that introduced the bug pass the cli tests.

mulkieran:
restarted a failed job.

PR #1110, opened 2018-08-17
Tgill enginelistenerlist

agrover:


mulkieran:
If rejecting #1114's global changes then it would be merely due diligence to demonstrate that the non-global framework would work for ```Filesystem::Used``` and ```Blockdev::UserInfo```. These are both more challenging than the examples that have been demonstrated so far in this PR.

agrover:
I don't necessarily agree that this is required, but sure, I can take a stab at it and see how it goes.

mulkieran:
Ok. You're adding 329 lines and substracting 33 at this time. I'm interested in seeing the additional lines of code required for actually handling blockdevs and filesystems.

agrover:
@mulkieran I don't think Blockdev UserInfo is really worthy of supporting with a signal. I think State is, so I'm going to try with that.

mulkieran:
@agrover Unfortunately, state isn't a good example, because it doesn't change.

agrover:
it does now :smiley:

mulkieran:
I think it's very important to think carefully about what a global is and to decide _for this particular case_ whether it is a reasonable choice to use a global.

Obviously, a global is globally accessible. The convenience of this is hard to miss. From now on the discussion is restricted to non-constant globals, that is, ones that can be modified. The use of modifiable globals can result in various errors. The simplest example are logical errors. That is one function can assume preconditions about the global that another function somewhere far away in the program invalidates. This generally manifests in a bug. Looking at the event listener list, though, it is clear that currently it can be modified in only way, by adding a listener. Moreover, it is only modified in one place, in the stratisd main method immediately when the dbus channel is set up. It should be clear that logical errors are going to be hard to achieve with exactly one mutation. Now, the idea is that there should be other event listeners that might be added to the list. The question is, how likely is adding another listener at some during execution likely to cause a logical bug. The answer should be, very unlikely. The listener will be added and from then on it will be notified of events.

The other accessing of this listener that is done is non-mutating notification of events. This is only non-mutating because the EngineListener trait's notify() method is non-mutating. That seems pretty reasonable, but might change. Even if notify() becomes a mutating method, however, it ought to difficult to cause a bug from the sharing of this structure. notify() methods will typically match on the event received and do something which may effect some other structure but ought not to effect the listener list.

So, the likelihood of bugs arising from using a global listener is really low while the simplification is really considerable.

What advantage does the non-global approach have in making errors less likely? It is certainly more complex, and that is not a good thing without some corresponding gain.

agrover:
Sometimes I just need to try things and see for myself.

I'm coming around to your perspective.

agrover:
@mulkieran take a look, plz. Merged your branch.

Added blockdev state change and fs used changed signals, and then the last commit I'm hopeful will let us more cleanly handle dbus path within the Engine code.

Hopefully we're nearing a stopping point. Over to you.

mulkieran:
@agrover There are some Rust 2.5 errors on Travis. Also, given that there are actual conflicts I think a rebase is in order.

agrover:
Merging from master to resolve conflicts. I think that's better for work PRs because rebasing would break downstream branches, such as your #1114.

FWIW from `git diff master|diffstat` I'm currently seeing 385 insertions(+), 135 deletions(-). Where is the 804 number you're seeing coming from?

Fixed syntax for 1.25. Had a quibble about a panic message so attempted to remedy.

mulkieran:
> FWIW from git diff master|diffstat I'm currently seeing 385 insertions(+), 135 deletions(-). Where is the 804 number you're seeing coming from?

From the PR page, top right corner. It's gone now, though.

mulkieran:
Please see changes in #1114. With those changes, I would be happy for this PR to be closed and a new, final PR open.

However, this PR has opened a new issue, about updating the BlockDevState. The correctness of the conditional for sending out a signal on allocation of sectors from the blockdev depends on there only being two possible ```BlockDevStates```: ```InUse``` and ```NotInUse```. That's not currently true. We should either make it true, by removing the other three states, as they are not yet in use, or tighten up that code a bit in a subsequent PR.

trgill:
@agrover - ok - will squash for final PR.

PR #1114, opened 2018-08-21
Agrover tgill enginelistenerlist

mulkieran:


agrover:
:+1:  69e2267

agrover:
I really am not a fan of globals. Just because there's only one is not enough to justify.

mulkieran:
Merged into #1125.

PR #1116, opened 2018-08-23
Agrover throttle ng

mulkieran:


mulkieran:
Just a Travis formatting failure.

mulkieran:
Failure of ```real_test_write_throttling``` at https://stratis-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/stratisd/1808/ is worth checking out. Almost certainly not due to my changes.

agrover:
Instead of nested functions, when something is a calculation that is only used in one spot, what do you think about assignment from a scope (not sure if this is what it's really called.) e.g:
```
let blah_plus_foo = {
    let blah = 4;
    let foo = 5;
    blah + foo
};
```
This seems a little lighter syntactically, still encapsulates the statements as being for the purpose of calculating the final value, and keeps it inline with the normal flow of the function.

Also, wrt Jenkins failures: rebooting the server (server-3) seems to help. I'm not sure if we should be blaming the throttle code or if something's wrong with the servers.

mulkieran:
Substituting body of function is fine when the body isn't too interesting, like w/ ```select_state``` method, so I've done it there.  ```used_pct``` though, is a bit more funky of a method and I would like to keep it a thing to itself.

mulkieran:
Spurious failure on Travis, I think.

agrover:
I also don't understand why `remaining  <= low_water` should be the test for if we ask for any additional space at all. How does low_water factor into how much space we ask for? Shouldn't the calc just be the greater of request and remaining?

mulkieran:
@agrover I think that there hasn't been any change in semantics from what you originally wrote wrt. the condition for whether to extend or not. Just some algebra.

mulkieran:
I'm going to back out the ThinPool::check() call in new() and setup() and do it right after those calls instead. I think it'll make testing easier and the code no less safe.

mulkieran:
Merged into #1133.

Issue #1130, opened 2018-08-29
Stratis stacked on lvm results in startup hang (intermittent)


tasleson:
I created 3 LVs to provide the storage for a Stratis pool.  I then created a single Stratis backed FS, mounted it and added the UUID to fstab.  When I boot the system it sometimes hangs and we get dumped to the emergency shell.  This happens as systemd cannot find the UUID of the Stratis back XFS in `/etc/fstab`

lvm configuration:

```
# pvs
  PV           VG       Fmt  Attr PSize     PFree    
  /dev/nvme0n1 stack_it lvm2 a--  <1024.00g <1021.00g
  /dev/nvme0n2 stack_it lvm2 a--  <1024.00g <1024.00g
  /dev/nvme0n3 stack_it lvm2 a--  <1024.00g <1024.00g

# lvs
  LV   VG       Attr       LSize Pool Origin Data%  Meta%  Move Log Cpy%Sync Convert
  lv0  stack_it -wi-ao---- 1.00g                                                    
  lv1  stack_it -wi-ao---- 1.00g                                                    
  lv2  stack_it -wi-a----- 1.00g

```
The journal shows:

```
Aug 29 10:11:17 localhost.localdomain stratisd[451]:  WARN libstratis::engine::strat_engine::engine: udev add: no pool set up, reason: Engine(Error, "failed to set up pool for (pool UUID: ab10b307-edf4-48e1-bf58-6e68d5e21f45, devnodes: /dev/dm-2): reason: Engine(Invalid, \"UUIDs of
 devices found ({Uuid { bytes: [28, 197, 232, 92, 0, 88, 74, 54, 146, 209, 103, 45, 117, 61, 136, 180] }}) did not correspond with UUIDs specified in the metadata for this group of devices ({Uuid { bytes: [28, 197, 232, 92, 0, 88, 74, 54, 146, 209, 103, 45, 117, 61, 136, 180] }, Uu
id { bytes: [101, 188, 85, 160, 82, 134, 65, 53, 165, 149, 220, 24, 195, 15, 158, 206] }, Uuid { bytes: [69, 28, 82, 240, 192, 63, 73, 184, 129, 216, 107, 120, 210, 119, 138, 140] }})\")")
[...]

PR #1131, opened 2018-08-29
Boot hang: Handle missing udev 'add' events

tasleson:
The block devices get 1 or more 'change' events too.  Lets give
ourselves the opportunity to interrogate the devices for these so
that we don't miss bringing up a pool if the 'add' event is never
encountered.

Fixes: https://github.com/stratis-storage/stratisd/issues/1130  ... well I've booted a few times and we have come up cleanly every time :smiley: 

**Another option is to walk the udev db when we fail to bring up a pool to see if we can find the missing device(s)**

### DIFF
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -103,10 +103,12 @@ fn initialize_log(debug: bool) -> buff_log::Handle<env_logger::Logger> {
     }
 }
 
-/// Given a udev event check to see if it's an add and if it is return the device node and
-/// devicemapper::Device.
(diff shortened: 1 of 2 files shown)


agrover:
tbh this approach feels less "hacky" to me than rewalking the udev db.

tasleson:
Rebased

mulkieran:
Added "do not merge" label because it is still in "In Progress" slot. Probably it's supposed to be in "Review" slot, but moving it would be a bigger step, so I'm taking the smaller one because it is not my PR.

tasleson:
Rebased and `warn!` statements modified to remove `udev`

PR #1059, opened 2018-07-18
Use udev to identify devices

mulkieran:
Supercedes #1043 .

PR #1132, opened 2018-08-29
Use udev to identify devices

mulkieran:
Supercedes #1059.


mulkieran:
rebased + a simplification

mulkieran:
Function ```unclaimed``` is kind of screwed up now that multipath member identification has been folded in. It should be split in two, something ```may_ignore``` and ```must_ignore```. Very different, really.

mulkieran:
Closing because this will have to be done from scratch to get it right, and because the code has moved on too far for this to be used for anything but as a reference.

PR #1134, opened 2018-08-29
Massage variable length metadata into a potentially more stable form

mulkieran:


mulkieran:
@agrover Does this look like what you expected to see wrt. to the DataTier only? Notice that cap is empty right now, but I think it would be generated from the ```next``` field of Backstore.

agrover:
Yep, going in the right direction. What's next?

I deleted a bunch of text here but let me just say I'm thinking hard about the layer-to-layer interface. Having just a `next` scalar in cap says something about that interface -- whose job is it to handle mapping grown blockdevs? I don't know if we're going to get to it for 1.0 but we might future proof a bit if next is represented in metadata as a vec of LayeredDev with only 1 entry where the entry's length field is basically "next".

Anyway.. probably a good chat topic for next week sometime, I don't think this blocks immediate next steps for this PR.

mulkieran:
@agrover Most layers in data tier will not have UUID available, just a DM UUID. Thus, need to have two LayeredDev types, one for actual BlockDev which have actual UUID on them, and one called, e.g., DMLayeredDev for all the others. Or something like that.

mulkieran:
rebased, just to be sure. But I think that the correct thing to do is to fall back to before the last five commits.

The new plan is to leave Cap in BackstoreSave, which must continue to exist. Also LayeredDev needs to be changed so that it contains an optional DmUuid or Uuid OR LayeredDev gets a DmUuid and the base layer that gets Stratis UUIDs written to it has a different type.

mulkieran:
rebased, and no longer trying to move next into DataTier. Backstore should continue to manage it, attractive as that is.

mulkieran:
@agrover I'm asking for comments. The ```serde_structs.rs``` file will give you an idea of how the metadata is now.

agrover:
@mulkieran I think serde_structs looks pretty good. How about removing `Blockdev` and inline its two members into CacheTierSave and DataTierSave? Finally, now we have most structs named `*Save` but not all. These could be made uniform, or the rationale for exceptions described in a comment?

mulkieran:
I think it makes sense to leave the ```BlockDev``` bit in. That way it'll be easier to accommodate the other layers, vdo, raid, etc.

Now is probably a good time to change everything to *Save. I'll see about doing that.

mulkieran:
Need ```FlexDevSave``` to use ```LayeredDev```.

mulkieran:
I thought it might be worth a try to see if Cap device could encode ```FlexDevSave``` data. The answer is no. The problem is that there is a meta and a meta spare. If these get swapped, then the Cap device has to be told that their use has been swapped in order to properly track the segments. This seems like the wrong way to go.

mulkieran:
The only thing left is the question of whether FlexDevsSave Sectors pairs should have a type. The answer is probably yes, but it should not be a LayeredDevSave. Call it a FlexDevComponentSave?

agrover:
I'm not sure about the last few commits... it looks to me like the addition of `FlexdevComponentSave` just leads to more work, since we're defining a Save struct that is not actually what we want to serialize/deserialize.

Suggest leaving Save functions as representative of what's given to serde. This doesn't stop using a little struct instead of a tuple in non-Save structs (e.g. struct Thinpool) if desired, with conversions happening in record() (to Save) or setup() (from Save) functions.

mulkieran:
@agrover I think we do want to save a named thing instead of just a pair in the metadata. It's just a bit clearer and consistent w/ our use of LayeredDevSave. We can do whatever we want w/ the internal structures of ThinPool at our convenience. Probably defining a struct for the element type of those segment Vecs does make sense, as well, but it doesn't really need to be in this PR.

agrover:
@mulkieran If I understand you correctly this is going to clutter the json with many instances of objects with "start", "length" fields. I don't see this as adding any value. Please keep json representation of segments as-is.

agrover:
> @agrover So you want me to drop LayeredDevSave and FlexDevComponentSave and use Sectors pairs instead? They are structurally identical.

I suppose so, yes.

> @agrover By "clutter" do you actually mean "bloat"? I mean, are you worried that the names, which will make the JSON easier to read and to process w/out error will also make it unnecessarily large?

Almost the entire contents of the metadata is lists of segments. Anything resembling two numbers is going to be a starting offset and then a length. It would certainly be bloat but it would also be clutter, visually swamping other object property names.

mulkieran:
Done.

mulkieran:
@agrover If you're satisfied w/ this, I will prepare a final PR.

mulkieran:
Superceded.

PR #1145, opened 2018-09-02
Update quickcheck to 0.7

ignatenkobrain:
Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,9 +1,9 @@
 [[package]]
 name = "aho-corasick"
-version = "0.6.3"
+version = "0.6.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "memchr 1.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr 2.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


ignatenkobrain:
Any hint how to fix this issue?

mulkieran:
I'm afraid not.

The only way to go from an unitialized to any other state is to call a ```set_logger``` method. This is only done in one place. The method is very clear and should not fail.

I compiled the code w/ your change, and I can run stratisd and it does not fail. It always initializes the log; if the log were already initialized or there was a bug that said it was initialized when it wasn't, it would fail.

Maybe a side effect of running the tests now is to initialize an ```env_logger```? If that is the case, then the test that we currently have is doomed to fail, always.

mulkieran:
I actually tried running this test in the debugger, and, in fact, it simply looks like a logger has already been initialized, somewhere. That is, the ```STATE``` value when the ```set_boxed_logger``` call is reached is already ```INITIALIZED```. Of course, I may have caused a Heisenbug w/ gdb, but likely something has actually initialized a logger somewhere.

I think we will have to find a different way to test the buffered logger, maybe an integration test.

mati865:
I'm almost sure it's related to `env_logger`, with it's version `0.5` for tests `init()` [should be replaced with `try_init()`](https://github.com/sebasmagri/env_logger/commit/c085ae8e1e72daad5dd12073011844a9ea5d4957#diff-04c6e90faac2675aa89e2176d2eec7d8L106).

It will be a shoot in the dark but try doing it for: https://github.com/stratis-storage/stratisd/blob/5a234865d5fe91e6a2c597e996e880fd59490349/src/engine/strat_engine/tests/logger.rs#L14

mati865:
@agrover I think it's a good alternative to using `RUST_TEST_THREADS=1` which would slow down testing 👍
What is wrong with Jenkins? The url doesn't work.

mulkieran:
This is blocked by #1030. If we move to proptest we won't need to update quickcheck.

mulkieran:
Superceded by #1315 as project has now moved to proptest.

Issue #1137, opened 2018-08-30
Trying to destroy a mounted file system causes daemon to become very confused


tasleson:
If we create a Stratis backed FS and mount it and then try to destroy it while it's mounted madness ensues...

```bash
# df -h
...
/dev/mapper/stratis-1-43957d9811384168ad682e2b0bf6cf0d-thin-fs-774d6b04f8834662848fa9291bb3143d  1.0T  1.1G 1023G   1% /mnt/destroy_while_mounted

# bin/stratis filesystem list
Pool Name  Name  Used      Created            Device                      
teststack  fs1   1.05 GiB  Aug 30 2018 10:54  /dev/stratis/teststack/fs1

# bin/stratis filesystem destroy teststack fs1
Execution failure caused by:
INTERNAL ERROR: low-level ioctl error

# bin/stratis filesystem list
Execution failure caused by:
no name for filesystem with uuid 774d6b04-f883-4662-848f-a9291bb3143d
    which in turn caused:
Error while invoking method "GetManagedObjects" belonging to interface "org.freedesktop.DBus.ObjectManager" with arguments ()

# bin/stratis pool list
Execution failure caused by:
no name for filesystem with uuid 774d6b04-f883-4662-848f-a9291bb3143d
    which in turn caused:
Error while invoking method "GetManagedObjects" belonging to interface "org.freedesktop.DBus.ObjectManager" with arguments ()

# bin/stratis blockdev list
Execution failure caused by:
no name for filesystem with uuid 774d6b04-f883-4662-848f-a9291bb3143d
    which in turn caused:
Error while invoking method "GetManagedObjects" belonging to interface "org.freedesktop.DBus.ObjectManager" with arguments ()

```

If we restart the daemon then all is well and if we un-mount the FS first we can destroy the FS.

```bash
# systemctl restart stratisd
[root@localhost stratis-cli]# bin/stratis filesystem list
Pool Name  Name  Used      Created            Device                      
teststack  fs1   1.05 GiB  Aug 30 2018 10:54  /dev/stratis/teststack/fs1
```

There is absolutely nothing to see in the journal when all this occurs :smiley:

tasleson:
**Note:  The stratis pool is sitting on top of 3 encrypted NVME devices, but I don't think that matters**

agrover:
oops yes, we should not allow mounted filesystems to be destroyed.

agrover:
[...]

PR #1151, opened 2018-09-05
Fix filesystem destroy and setup

agrover:
Make use of the new  `&mut self` argument for destroy() in devicemapper 0.25.0 to recover from `fs.destroy()` failing for a mounted filesystem. There are also some other problems around filesystem setup/destroy that this PR tries to improve.


agrover:
fixes #1137

Issue #1142, opened 2018-08-31
Stratis stacked on multipath results in startup boot hang


tasleson:
We are getting the following message when we try to boot:

```
WARN libstratis::engine::strat_engine::engine: no pool set up, reason: 
Engine(Error, "failed to set up pool for (pool UUID: 85c985d1-4e5b-4b02-8c1c-aeb5ca470fe9, devnodes: /dev/sdj ,/dev/sdd ,/dev/sdc ,/dev/sdk): 
reason: Engine(Invalid, \"The following list of Stratis UUIDs were each claimed by more than one Stratis device: [Uuid { bytes: [189, 15, 99, 222, 235, 237, 74, 210, 137, 134, 76, 198, 180, 69, 199, 239] }, Uuid { bytes: [214, 61, 167, 84, 184, 71, 70, 43, 146, 84, 81, 102, 253, 73, 49, 235] }]\")")
```
The device paths listed correspond to the individual paths that comprise the multipath devices.  We should be ignoring these paths, assuming the udev db has the correct values at the time we evaluate them.

Restarting the stratis daemon when we are in the emergency shell works and we utilize the correct multipath devices `/dev/mapper/mpatha` and `/dev/mapper/mpathb` and the FS gets mounted.

**Note: Code was build off of local branch that has multipath support and the fix for missing udev 'add' events.**

tasleson:
It appears that at the time we get the udev events and we interrogate the udev db for the devices they don't have the values we are looking for, eg.

udev db entry when we interrogate
```
"DEVNAME": "/dev/sdk"
"DEVTYPE": "disk"
"MINOR": "160"
"SUBSYSTEM": "block"
"DEVPATH": "/devices/pci0000:00/0000:00:03.0/0000:08:00.1/host7/rport-7:0-0/target7:0:0/7:0:0:1/block/sdk", "MAJOR": "8"
```

Udev entry later
```
P: /devices/pci0000:00/0000:00:03.0/0000:08:00.1/host7/rport-7:0-0/target7:0:0/7:0:0:1/block/sdk
N: sdk
S: disk/by-id/scsi-360a98000423664653024426830595064
S: disk/by-id/wwn-0x60a98000423664653024426830595064
S: disk/by-path/pci-0000:08:00.1-fc-0x500a098190ba5a64-lun-1
[...]

PR #1153, opened 2018-09-05
Ignore incomplete udev entries during enumeration

tasleson:
When we start up we walk the udev db to see what is available.  When
we do this at early boot we have udev entries that exist, but are not
complete as the udev processing has not completed.  Thus we need to
ignore these db entries until they become complete.  When they do
become complete we will process it in the event handler for udev
events.

### DIFF
--- a/src/engine/strat_engine/backstore/util.rs
+++ b/src/engine/strat_engine/backstore/util.rs
@@ -39,6 +39,7 @@ pub fn get_udev_block_device(
 
     let result = enumerator
         .scan_devices()?
+        .filter(|dev| dev.is_initialized())
         .find(|x| x.devnode().map_or(false, |d| canonical == d))
         .and_then(|dev| Some(device_as_map(&dev)));
     Ok(result)
@@ -59,6 +60,7 @@ fn get_all_empty_devices() -> StratisResult<Vec<PathBuf>> {
 
     Ok(enumerator
         .scan_devices()?
+        .filter(|dev| dev.is_initialized())
         .filter(|dev| {
             dev.property_value("DM_MULTIPATH_DEVICE_PATH").is_none()
                 && !((dev.property_value("ID_PART_TABLE_TYPE").is_some()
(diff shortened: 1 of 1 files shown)


tasleson:
Fixes https://github.com/stratis-storage/stratisd/issues/1142

src/engine/strat_engine/backstore/util.rs
```diff
@@ -78,6 +80,7 @@ pub fn get_stratis_block_devices() -> StratisResult<Vec<PathBuf>> {
 
     let devices: Vec<PathBuf> = enumerator
         .scan_devices()?
+        .filter(|dev| dev.is_initialized())
```
mulkieran: The comment and TODO below are invalidated by this line, because now a third explanation is that stratisd is starting up so early that the devices are not yet initialized. So, it should be fixed up, and I think that the TODO might as well be removed, since under these circumstances it is probably the case that the block of code that it annotates can never be removed.

tasleson:
@mulkieran From my understanding we should have no gaps.  When the kernel emits a udev event the most basic information is added to the udev db.  If a client is listening for "kernel" uevents they would get this limited amount of information then.  If a client is listening for "udev" uevents, then they are notified after all udev processing has occurred.  Thus when we walk the udev db before things have been initialized we will get the uevent sometime later.  We shouldn't have a race condition as we listen for events before we enumerate the db.  Thus if one db entry becomes complete after we interrogated it during enumeration, but before we have completed the enumeration we will have an event waiting for us when we get to the event loop.

src/engine/strat_engine/backstore/util.rs
tasleson: I agree the comment could use an update, but I don't think we want to keep the code it annotates.  For example it's entirely possible that a big configuration at boot has **many** block devices.  If we walk the udev db before any are initialized and thus find none, we then would fall into this annotated code which would falsely think that everything is empty, thus causing block reads for every device looking for Stratis signatures.  IMHO it's OK that at the point in time that we scan things are empty as we will get them when events are generated.  I'm leaning towards removing this code and removing whatever tests that fail on older distribution that run on Travis CI.

tasleson:
Replaced by https://github.com/stratis-storage/stratisd/pull/1156

PR #1156, opened 2018-09-06
Ignore incomplete udev entries during enumeration (final)

tasleson:
When we start up we walk the udev db to see what is available.  When
we do this at early boot we have udev entries that exist, but are not
complete as the udev processing has not completed.  Thus we need to
ignore these db entries until they become complete.  When they do
become complete we will process it in the event handler for udev
events.


tasleson:
Fixes https://github.com/stratis-storage/stratisd/issues/1142

Issue #748, opened 2018-02-09
Handle error case when device(s) show up that belong to an existing activated pool


tasleson:
It is theoretically possible that the following scenario could occur:

1. Pool is modified, a subset of the disks meta data are updated
2. `stratisd` is started and a subset of the disks are available which only have the the **outdated** metadata
3. At some point later the device(s) become available which have the newer metadata

Potential outcome(s):
* If `stratisd` doesn't have udev add support it is blissfully unaware of the issue and continues to operate under the false assumption that all is good.
* If `stratisd` does have udev add support it gets the udev add, evaluates that the device pool id is already active and ignores it, yielding the same outcome as if `stratisd` didn't have udev add support.

Ultimately we need to be able to identify this error case and correctly figure out what actions to take to correct it.  At the moment we could identify it with udev add support, but we don't know what action(s) we should take.  As one previous colleague of mine stated, "Don't check for error conditions you don't know how to handle" :-)

mulkieran:
Note that stratisd can currently distinguish between whether the device is already in the pool, but was found again, and the device is not yet in the pool, and has been found for the first time.

1. The device is already in the pool.
  a. All our data about the device matches the data we already have. Maybe this means everything is fine. Or maybe not.
  b. Something doesn't match between the blockdev we have now and the one in the complete pool. This is a definite panic, I would think.
2. The device is not already in the pool.
  a. It has metadata newer than the pools. In that case the pool must be wrong. We should remove the pool from pools, reinsert it into incomplete pools, and rerun the whole setup thing to see if it is now fine.
  b. Its metadata is older than the pool's metadata. So it belonged to the pool and was removed. Currently, that is impossible, so a panic is the right choice. Later, it may be more possible, as we will be able to remove cache devs from the pool.

tasleson:
> The device is not already in the pool.
a. It has metadata newer than the pools. In that case the pool must be wrong. We should remove the pool from pools, reinsert it into incomplete pools, and rerun the whole setup thing to see if it is now fine.

Once a pool is up and active and IO has been done on it, there is a good case that data loss/corruption has already occurred.  Once cannot simply go from one state to the another and back without ramifications.  IMHO we need to prevent this case from happening, not try to deal with it when it does.  We should know without any ambiguity that a pool is complete or not.

If I understand this correctly, this all stems from an optimization where we choose to not write the latest metadata to all disks, to speed up metadata updates.  IMHO We need to revisit that discussion and determine how we can close this error case.

[...]

PR #1161, opened 2018-09-07
Remove warn! for pool already known

tasleson:
This message gets printed often and is expected with us monitoring
both 'add' and 'change' udev events.  Add additional check for the
more hideous issue of having a device which thinks it belongs in a
pool, but the existing active pool has no knowledge of it.

Removes the `warn` but does not fully address what we should do for https://github.com/stratis-storage/stratisd/issues/748

### DIFF
--- a/src/engine/strat_engine/backstore/device.rs
+++ b/src/engine/strat_engine/backstore/device.rs
@@ -123,9 +123,9 @@ pub fn identify(devnode: &Path) -> StratisResult<DevOwnership> {
 
 /// Determine if devnode is a Stratis device. Return the device's Stratis
 /// pool UUID if it belongs to Stratis.
-pub fn is_stratis_device(devnode: &Path) -> StratisResult<Option<PoolUuid>> {
+pub fn is_stratis_device(devnode: &Path) -> StratisResult<Option<(PoolUuid, DevUuid)>> {
     match identify(devnode)? {
-        DevOwnership::Ours(pool_uuid, _) => Ok(Some(pool_uuid)),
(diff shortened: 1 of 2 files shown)


mulkieran:
Please see https://github.com/stratis-storage/stratisd/pull/1163. There is an alternative approach, which involves hoisting the ```get_by_uuid``` method up and making it the primary if condition, instead of ```contains_uuid```. But because of borrowing restrictions, it needs to change other things more aggressively, so I figured this was better.

tasleson:
@mulkieran AFAIK, `pool.get_blockdev` returns a `BlockDev`, not a `StratBlockDev`.

mulkieran:
@tasleson Ah, I see. You're probably better off doing something like this: https://github.com/stratis-storage/stratisd/pull/1163.

tasleson:
@mulkieran Would it be cleaner to add a method `dev` to `BlockDev` instead?

tasleson:
Replace by: https://github.com/stratis-storage/stratisd/pull/1171

@mulkieran I picked up your changes and squashed.  I did find an issue with the error message:

Changed

```rust
                        if device != *block_dev.device() {
                            error!(
                                "we have a block device with the same uuid as one already in the pool, but the one in the pool has device number {:}, while the one just found has device number {:}",
                                device,
                                block_dev.device()
                            );
```
To
```rust
                       if device != *block_dev.device() {
                            error!(
                                "we have a block device with the same uuid as one already in the \
                                 pool, but the one in the pool has device number {:}, \
                                 while the one just found has device number {:}",
                                block_dev.device(),
                                device,
                            );
                        }
```
The ordering of the variables was incorrect,  `device` contains the just found device and `block_dev.device()` is the device that is in the pool.

PR #1163, opened 2018-09-07
Tasleson remove spurious warn

mulkieran:


tasleson:
@mulkieran @agrover I was thinking about this some more and I think we also need to add a check to ensure that the device we believe is already in the pool because of device UUID, matches the major minor number or some other piece of information that is not written to the disk itself.  Thus we would be able to distinguish an existing disk in the pool and a disk that just showed up that is bitwise identical to a device in the currently active pool, but isn't in the pool at this moment in time.  This could happen if someone replicated one of the disks in the pool and masked & mapped it back to the same host.  The disk contents would be identical, but the path and major & minor number would be different.  As we are filtering out multipath devices we would be confident we weren't going down that path.

mulkieran:
> ...I think we also need to add a check to ensure that the device we believe is already in the pool because of device UUID, like match the major minor number or some other piece of information that is not written to the disk itself.

That sounds to me like a good idea.

agrover:
@tasleson sounds good. Good thinking.

mulkieran:
Merged into #1171.

PR #1166, opened 2018-09-10
Change error return if fs.destroy() failed with EBUSY

### LINKED ISSUE #1160: Do not call failure to destroy a mounted FS an INTERNAL_ERROR
What we get when we try to destroy a mounted FS

```bash
# bin/stratis filesystem destroy testing fs1
Execution failure caused by:
INTERNAL ERROR: low-level ioctl error
```
I think we should be a little more helpful to the user.

agrover:
Using newly-included errno value, match on EBUSY and return a StratisError::Error instead, with a relevant message.

fixes #1160

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -126,7 +126,7 @@ dependencies = [
 [[package]]
 name = "devicemapper"
 version = "0.25.0"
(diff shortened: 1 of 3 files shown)


agrover:
supeseded by #1179

Issue #1168, opened 2018-09-10
Review which signals we should include in Stratis 1.0


agrover:
@trgill @mulkieran @tasleson as per our discussion, we should want to minimize signals we support in Stratis 1.0, since for ones that can trigger frequently, we should specify how often they can trigger, and maybe offer some way for clients to set a threshold, perhaps.

Right now we expose 4 signals:
1. BlockdevStateChanged
1. FilesystemRenamed
1. FilesystemUsedChanged
1. PoolRenamed

The only one whose value will change often is FilesystemUsedChanged. Therefore I propose we push out supporting this signal past Stratis 1.0, until we can specify more details about the behavior that clients can expect from it.

trgill:
Opened PR #1170 - don't send the signal - but leave the EventHandler in place.

agrover:
done

PR #1170, opened 2018-09-10
Don't send signal on FilesystemUsedChanged

trgill:
Since the property can change frequently, don't send a signal on change.

Leave the EventHandler code in place in case we re-enable the signal.


agrover:
fixes #1168

PR #1172, opened 2018-09-11
Make pool.total_physical_used never return an error

agrover:
Having a property included in GetManagedObjects fail is very bad because
if it does then the client gets no info.

Since total_physical_used() makes an ioctl call, theoretically it can
fail. Cover this by caching the last value and returning that if the
ioctl fails. Since this requires changing StratPool, some things in
the dbus code need to use mut refs instead of just refs.

### DIFF
--- a/src/dbus_api/pool.rs
+++ b/src/dbus_api/pool.rs
@@ -296,7 +296,7 @@ fn get_pool_property<F, R>(
     getter: F,
 ) -> Result<(), MethodErr>
 where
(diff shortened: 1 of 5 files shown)


agrover:
will undoubtedly need rebasing if #1158 lands first.

mulkieran:
In that case, I might as well wait until #1158 does land.

agrover:
Rebased on top of #1158.

agrover:
dropping this pr.

PR #1173, opened 2018-09-12
Set throttling in all directories under /sys/fs/cgroup/blkio

agrover:
What we were doing previously did not work in all cases. Using this code
I was able to see blkio throttling on a development machine where
previously it did not work.

Use the "walkdir" crate to iterate through the entire directory hierarchy
and set the throttle in all files that match the filename.


mulkieran:
It looks like there are compilation errors in tests.

mulkieran:
```test_udev.py``` failed because it couldn't set up  a pool.

trgill:
@mulkieran do you still see failures?  Looks like the tests are passing now... (?)

agrover:
#1182 merged and squashed. If that's it, can we go ahead and merge this PR rather than making a final PR? Thanks.

PR #1174, opened 2018-09-12
Change backstore.available() to include space not yet allocated to the cap device

agrover:
We are making incorrect calculations around FreeSpaceState because `backstore available()` is returning 0, meaning that no extra space is left in the cap device, but this ignores all the unallocated space available in the data tier.

The first commit renames `available()` to `available_in_cap()`. The second creates a new `available()` function that includes cap space but also other free space in the data tier. Using this in thinpool results in correct calculations for free space percentage.

### DIFF
--- a/src/engine/strat_engine/backstore/backstore.rs
+++ b/src/engine/strat_engine/backstore/backstore.rs
@@ -317,7 +317,7 @@ impl Backstore {
         sizes: &[Sectors],
     ) -> StratisResult<Option<Vec<(Sectors, Sectors)>>> {
         let total_required = sizes.iter().cloned().sum();
-        let available = self.available();
+        let available = self.available_in_cap();
         if available < total_required {
             if self.data_tier.alloc(total_required - available) {
                 self.extend_cap_device(pool_uuid)?;
@@ -373,7 +373,7 @@ impl Backstore {
             return Ok(None);
         }
 
-        let available = self.available();
+        let available = self.available_in_cap();
         if available < internal_request {
             let mut allocated = false;
             while !allocated && internal_request != Sectors(0) {
@@ -384,7 +384,7 @@ impl Backstore {
             if allocated {
                 self.extend_cap_device(pool_uuid)?;
 
-                let return_amt = cmp::min(request, self.available());
+                let return_amt = cmp::min(request, self.available_in_cap());
                 let return_amt = (return_amt / modulus) * modulus;
                 self.next += return_amt;
                 Ok(Some((self.next - return_amt, return_amt)))
@@ -442,8 +442,8 @@ impl Backstore {
             .unwrap_or(Sectors(0))
     }
 
-    /// The available number of Sectors.
-    pub fn available(&self) -> Sectors {
+    /// The available number of Sectors in the cap device.
+    pub fn available_in_cap(&self) -> Sectors {
         let size = self.size();
         // It is absolutely essential for correct operation that the assertion
         // be true. If it is false, the result will be incorrect, and space
--- a/src/engine/strat_engine/backstore/backstore.rs
+++ b/src/engine/strat_engine/backstore/backstore.rs
@@ -458,6 +458,9 @@ impl Backstore {
         size - self.next
     }
 
+    /// The available number of Sectors in the Backstore. (Note: future
+    /// changes may change the semantics of this value, or may make it
+    /// impossible to calculate.)
     pub fn available(&self) -> Sectors {
         self.datatier_current_capacity() - self.next
     }
--- a/src/engine/strat_engine/backstore/backstore.rs
+++ b/src/engine/strat_engine/backstore/backstore.rs
@@ -458,6 +458,10 @@ impl Backstore {
         size - self.next
     }
 
+    pub fn available(&self) -> Sectors {
+        self.datatier_current_capacity() - self.next
+    }
+
     /// Destroy the entire store.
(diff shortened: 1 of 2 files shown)


src/engine/strat_engine/backstore/backstore.rs
```diff
@@ -458,6 +458,10 @@ impl Backstore {
         size - self.next
     }
 
+    pub fn available(&self) -> Sectors {
```
mulkieran: This needs documenting very carefully. I can't tell what it would eventually mean once other layers are interposed.

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -555,7 +555,7 @@ impl ThinPool {
                 // Update pool space state
                 self.free_space_state = self.free_space_check(
                     usage.used_data,
-                    current_total + sectors_to_datablocks(backstore.available()) - usage.used_data,
+                    current_total - usage.used_data + sectors_to_datablocks(backstore.available()),
```
mulkieran: There's no reason to reverse this order...it looks like it just crept in randomly, and it is kind of confusing.

mulkieran:
I did some high school algebra about the calculation of the low water value a while ago. Looking at my notes, it appears that the lowater calculation for the ```WARN``` state is:
```
max(0.1 * DD - 0.9 * A,  LW)
where DD is the size of the data device
           A is the available value
           LW is the default low water value for the warn state
```

This means that in order to have a low water value greater than ```LW```,
it must be the case that ```DD > 10 LW + 9 A```. With this change now,
```A``` will generally be much larger, so much more often, the low water
value will be ```LW```, the lower of the two possibilities.

However, if a VDO layer is later added in, that could flip-flop, so that
the current ```A``` the amount available in the cap is actually larger
than your new ```A```, the difference between the absolute size of
the blockdevs belonging to the blockdevmgr for the data tier in
this pool and the amount already allocated from the cap device.

The first commit, renaming from ```available``` to ```available_in_cap```
seems fine to me, because obviously there was some confusion about
the meaning of the method.

The second change, which makes the calculations dependent
on this new way of calculating available, I'm not sure what it
does.

File: src/engine/strat_engine/backstore/backstore.rs (hunk 1 above)
Comment (agrover) [reply src/engine/strat_engine/backstore/backstore.rs]: see d6cbc12, yes this will not work in 2.0.

File: src/engine/strat_engine/thinpool/thinpool.rs (hunk 2 above)
Comment (agrover) [reply src/engine/strat_engine/thinpool/thinpool.rs]: see 24f7a05, sorry about that, added some parens. It seemed easier to understand to group this way.

**Comment by agrover** [comment]
@mulkieran yes, it should be returning DATA_LOWATER the great majority of the time. I extended the comments in 83ba317 , is that a better explanation?

70477d4 removes the tighter lowater value for non-Good FreeSpaceStates, I think it's probably not worthwhile.

File: src/engine/strat_engine/thinpool/thinpool.rs [hunk 3]
```diff
@@ -555,7 +560,8 @@ impl ThinPool {
                 // Update pool space state
                 self.free_space_state = self.free_space_check(
                     usage.used_data,
-                    current_total + sectors_to_datablocks(backstore.available()) - usage.used_data,
+                    (current_total - usage.used_data)
```
mulkieran: I think this is the same as ```self.data_tier.current_capacity() - usage.used_data```. I don't know what that means, though.

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -125,8 +125,14 @@ fn coalesce_segs(
...
 
+/// Use the lowater event for two purposes: extending the data dev and also
+/// getting an event when a FreeSpaceState threshold is crossed.
+/// Most of the time, this results in just DATA_LOWATER. The two times it
+/// doesn't is 1) right before a FreeSpaceState threshold (but the threshold
+/// is still above DATA_LOWATER) and 2) If the data dev is already extended
+/// far beyond its current usage -- past the next freespace threshold.
 /// Calculate new low water based on the current thinpool data device size
-/// and the amount of unused sectors available in the cap device.
+/// and the amount of unused sectors available for further extension.
```
mulkieran: The ```L``` in the formula below can be omitted; it's not a variable it's ```DATA_LOWATER```.

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -125,8 +125,14 @@ fn coalesce_segs(
     segments
 }
 
+/// Use the lowater event for two purposes: extending the data dev and also
+/// getting an event when a FreeSpaceState threshold is crossed.
+/// Most of the time, this results in just DATA_LOWATER. The two times it
```
mulkieran: I'm having difficulty making sense of this explanation. But I can see that, when the available amount decreases to none, the lowater will be 0.1 * data dev size, if still in Good state.

mulkieran:
I drew a graph. red axis is size of device, green axis is available, blue plane is DATA_LOWATER constant, orange plane is formula for non-constant low water, i.e., 0.1*data - 0.9 available.

screenshot from 2018-09-13 09-26-47

src/engine/strat_engine/thinpool/thinpool.rs
agrover: fixed in e1abaae7

agrover:
I've decided to break this into two PRs and resubmit.

PR #1179, opened 2018-09-13
Use richer Display for devicemapper Core errors

mulkieran:
Addresses: #1160.


agrover:
#1166 also made BUSY a DbusErrorEnum::ERROR instead of INTERNAL_ERROR. This PR doesn't. I have opened #1181 and suggest it be resolved as part of this PR or a subsequent one.

agrover:
@mulkieran hi, please review and consider #1194?

mulkieran:
@agrover I'ld rather deal with the Enum stuff separately. We once had a clear meaning for ```INTERNAL_ERROR``` which was "internal error". I think this inevitably rotted because of our type-based approach to generating D-Bus error codes. I don't mind throwing errors into other categories, but it is a separate issue.

agrover:
Ok sounds good.

Issue #1175, opened 2018-09-12
Missing /dev/stratis/<pool name>/<fs name> after reboot


tasleson:
When we reboot a system with an existing pool, when we come back up we assemble the pool and the thin devices, thus we are able to mount the Stratis backed FS in the fstab.  However, when we do this the `/dev/stratis/*` entries are missing.

Note: This was found when using a pool backed by an iSCSI logical unit.  However, I don't think that matters.

(Tested with 696b0c07e6e5e6377a009c973bbf7d57e28a07b5)

tasleson:
Confirmed, this has nothing to do with iSCSI.  Using a pool backed by a local disk results in same behavior of missing entries in `/dev/stratis` after a reboot.

tasleson:
**Note: restarting the service results in symlink(s) being created***

Issue #1176, opened 2018-09-12
Destroying a FS and pool with a missing symlink in /dev/stratis/* results in badness


tasleson:
When you encounter issue https://github.com/stratis-storage/stratisd/issues/1175 and then try to destroy the FS and pool you get:

```bash
# ls -l /dev/stratis/
total 0
[root@localhost stratis-cli]# bin/stratis filesystem list
Pool Name   Name       Used     Created            Device                             
iscsi_pool  iscsi_lun  546 MiB  Sep 12 2018 09:37  /dev/stratis/iscsi_pool/iscsi_lun  
[root@localhost stratis-cli]# bin/stratis filesystem destroy iscsi_pool iscsi_lun
[root@localhost stratis-cli]# echo $?
0
[root@localhost stratis-cli]# bin/stratis filesystem list
Pool Name  Name  Used  Created  Device  
[root@localhost stratis-cli]# bin/stratis pool list
Name          Total Physical Size  Total Physical Used
iscsi_pool                  8 GiB               52 MiB
[root@localhost stratis-cli]# bin/stratis pool destroy iscsi_pool
Execution failure caused by:
IO ERROR: entity not found
[root@localhost stratis-cli]# lsblk
NAME   MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
sda      8:0    0   32G  0 disk 
├─sda1   8:1    0    1G  0 part /boot
├─sda2   8:2    0  3.2G  0 part [SWAP]
└─sda3   8:3    0   15G  0 part /
sdb      8:16   0  128G  0 disk 
sdc      8:32   0  128G  0 disk 
sdd      8:48   0  128G  0 disk 
sde      8:64   0  128G  0 disk 
sdf      8:80   0    8G  0 disk 
sr0     11:0    1 1024M  0 rom  
[root@localhost stratis-cli]# bin/stratis pool list
Execution failure caused by:
no pool corresponding to uuid 9304c2fc-0ed2-4651-a7b1-b38095dcec2f
    which in turn caused:
Error while invoking method "GetManagedObjects" belonging to interface "org.freedesktop.DBus.ObjectManager" with arguments ()

```
The journal shows:

```
[...]

PR #1180, opened 2018-09-13
Setup symbolic links on event driven pool bringup

tasleson:
Previously the code was only building up symbolic links for pools
that exist when the daemon starts.  At boot when the device(s)
haven't been discovered or are not ready in udev, the pool(s)
were not present and thus we failed to create the symbolic links
in /dev/stratis for them.  With this change we will setup the
pool directory and filesystem symbolic links as needed when the
pool is assembled when all the devices become available.

Fixes: https://github.com/stratis-storage/stratisd/issues/1175 https://github.com/stratis-storage/stratisd/issues/1176

### DIFF
--- a/src/engine/devlinks.rs
+++ b/src/engine/devlinks.rs
@@ -32,89 +32,115 @@ pub fn setup_dev_path() -> StratisResult<()> {
 /// or filesystem.
 // Don't just remove and recreate everything in case there are processes
 // (e.g. user shells) with the current working directory within the tree.
-pub fn setup_devlinks<'a, I: Iterator<Item = &'a (Name, PoolUuid, &'a Pool)>>(
-    pools: I,
-) -> StratisResult<()> {
-    let mut existing_dirs = fs::read_dir(DEV_PATH)?
-        .map(|dir_e| dir_e.and_then(|d| Ok(d.file_name().into_string().expect("Unix is utf-8"))))
-        .collect::<Result<HashSet<_>, _>>()?;
-
-    for &(ref pool_name, _, pool) in pools {
-        if !existing_dirs.remove(&pool_name.to_owned()) {
-            pool_added(pool_name)?;
-        }
-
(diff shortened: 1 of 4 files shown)


tasleson:
Rebased

src/engine/devlinks.rs
```diff
@@ -17,104 +17,153 @@ use super::engine::DEV_PATH;
 
 /// Set up the root Stratis directory, where dev links as well as temporary
 /// MDV mounts will be created. This must occur before any pools are setup.
-pub fn setup_dev_path() -> StratisResult<()> {
+pub fn setup_dev_path() -> () {
```
agrover: Maybe we leave this returning a result? If this fails, this isn't just the user futzing around with symlinks, this means stratisd has some sort of permission problem, and it might be better to bail.

src/engine/devlinks.rs
```diff
@@ -17,104 +17,153 @@ use super::engine::DEV_PATH;
...
+}
 
-    Ok(())
+/// Set up directories and symlinks under /dev/stratis based on current
+/// config. Clear out any directory or file that doesn't correspond to a pool
+/// or filesystem.
+// Don't just remove and recreate everything in case there are processes
+// (e.g. user shells) with the current working directory within the tree.
+pub fn setup_devlinks<'a, I: Iterator<Item = &'a (Name, PoolUuid, &'a Pool)>>(pools: I) -> () {
+    match || -> StratisResult<()> {
```
agrover: This is a good way to "catch" errors from multiple sources and handle them in one spot, but I now only see one thing ("?") in the closure that can fail. If we just warn on that directly, we can do away with the closure.

tasleson: Actually there is a couple of `?'s` in the function `setup_devlinks` which makes it just like `setup_pool_devlinks`.

tasleson:
Rebased

PR #1183, opened 2018-09-13
Clearer names for size-related functions in backstore

agrover:
`capacity` and `current_capacity` both existing is confusing. Instead, use `size` to mean the total quantity of the resources being managed. Replace `available()` with `available_in_cap()` and `available_in_backstore()`, since various code needs both of these.

This is split-out and extended from #1174.

### DIFF
--- a/src/engine/strat_engine/backstore/backstore.rs
+++ b/src/engine/strat_engine/backstore/backstore.rs
@@ -424,8 +424,8 @@ impl Backstore {
     }
 
     /// The current capacity of all the blockdevs in the data tier.
-    pub fn datatier_current_capacity(&self) -> Sectors {
-        self.data_tier.current_capacity()
+    pub fn datatier_size(&self) -> Sectors {
+        self.data_tier.size()
     }
 
     /// The size of the cap device.
--- a/src/engine/strat_engine/backstore/backstore.rs
+++ b/src/engine/strat_engine/backstore/backstore.rs
@@ -428,12 +428,6 @@ impl Backstore {
         self.data_tier.size()
     }
 
-    /// The current size of all the usable space in blockdevs in the data
-    /// tier.
-    fn datatier_data_size(&self) -> Sectors {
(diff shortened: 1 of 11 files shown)


src/engine/strat_engine/backstore/blockdev.rs
```diff
@@ -173,7 +173,7 @@ impl BlockDev for StratBlockDev {
 
     fn total_size(&self) -> Sectors {
         let start = self.metadata_size();
-        let size = self.current_capacity();
+        let size = self.size();
```
mulkieran: Maybe you should change ```total_size``` to ```net_size``` while you're at it.

src/engine/strat_engine/backstore/backstore.rs
```diff
@@ -423,9 +423,15 @@ impl Backstore {
...
-    pub fn datatier_current_capacity(&self) -> Sectors {
-        self.data_tier.current_capacity()
+    /// The current size of all the blockdevs in the data tier.
+    pub fn datatier_size(&self) -> Sectors {
+        self.data_tier.size()
+    }
+
+    /// The current size of all the usable space in blockdevs in the data
+    /// tier.
+    fn datatier_data_size(&self) -> Sectors {
```
mulkieran: Can you move this to ```DataTier```, i.e., make a method in DataTier called ```data_size```?

src/engine/strat_engine/backstore/blockdev.rs
agrover: See 959dea86; based on the usage of this method for TotalPhysicalSize, I think returning the raw (i.e. including-overhead) value is correct here.

agrover:
@mulkieran nice catch! See a245250.

agrover:
I've opened #1190 for documenting project-specific method naming conventions in Rust code.

agrover:
Update `calc_lowater` comment in 681b7ed1, better?

Issue #1152, opened 2018-09-05
allocating filesystems quickly causes out of space error 


trgill:
Update: Moved to stratisd from docs.

Updated Note: problem is easy to reproduce on dell-r730-011.dsal.lab.eng.rdu2.redhat.com - need to create at least 25 filesystems.

The attached script causes the following on the 3rd filesystem:

    stratis filesystem create pool1 24
    Execution failed: 5: Command failed: cmd: "/usr/sbin/mkfs.xfs" "-f" "-q" "/dev/dm-16" "-m" "uuid=1cf50752-f0e3-4648-b107-31a3c0761ccb", stdout: stderr: mkfs.xfs: libxfs_device_zero write failed: No space left on device

#!/bin/bash
set -x
set -e
DEVS1="<list of block devices here>"

FS_COUNT=25
POOL_NAME=pool1

stratis pool create $POOL_NAME $DEVS1

for((fs=1;fs<=$FS_COUNT;fs++))
do
   echo "create $POOL_NAME-fs-$fs"
   stratis filesystem create $POOL_NAME "$fs"
done

agrover:
One issue is that backstore.available() is returning the remaining space in the backstore cap device, which as the used space increases in the thinpool, the percentage  that is available shrinks until free_space_check goes into `Warn`.

Issue 2 is that afaict throttling is not working -- filesystem creates should be taking longer and that's not what I'm seeing.

Issue 3 is that when creating fs 24 the thinpool status says OutOfSpace, and the xfs mkfs fails. What happened to our lowater event?? It should've triggered and we should've extended. We have different lowater value/behavior when in Warn, so this may be a bad value (too close, esp with non-working throttling) or some algorithm bug.

Anyway, that's what I have so far.

agrover:
@mulkieran since "available" will become increasingly hazy in stratis 2, we may want to think about coming up with an alternative method for determining `FreeSpaceState`s, and possibly move its calculation into `Backstore`? Well, for 1.0, Backstore could just look at blockdev space not yet allocated to the cap device and calc against used. 2.0 is when that stops working and we may need  a heuristic to set PoolSpaceState...

agrover:
`backstore.available()` currently always returns 0, interesting.

[...]

PR #1184, opened 2018-09-13
Increase DATA_LOWATER to 1024

agrover:
I haven't been able to root-cause "25 filesystems" issue (#1152) but
it does appear that raising DATA_LOWATER to 1024 makes the issue no
longer reproduce.

Since performance testing is blocked, let's just do this for now so that
work can proceed.

### DIFF
--- a/src/engine/strat_engine/thinpool/thinpool.rs
+++ b/src/engine/strat_engine/thinpool/thinpool.rs
@@ -41,7 +41,7 @@ use super::mdv::MetadataVol;
 use super::thinids::ThinDevIdPool;
 
 pub const DATA_BLOCK_SIZE: Sectors = Sectors(2 * IEC::Ki);
-const DATA_LOWATER: DataBlocks = DataBlocks(512);
+const DATA_LOWATER: DataBlocks = DataBlocks(1024);
 const META_LOWATER_FALLBACK: MetaBlocks = MetaBlocks(512);
 
 const INITIAL_META_SIZE: MetaBlocks = MetaBlocks(4 * IEC::Ki);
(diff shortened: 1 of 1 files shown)


mulkieran:
The error is real, ```INITIAL_DATA_SIZE``` needs to be adjusted up along with DATA_LOW_WATER. But please see #1185, first. It should be an easy ack.

Given that ```INITIAL_DATA_SIZE``` is local to thinpool after #1185, it might be sensible to make both ```INITIAL_DATA_SIZE``` and ```INITIAL_META_SIZE``` be functions of their ```DATA_LOWATER``` and ```META_LOWATER``` values, rather than the constants they are now.

agrover:
Updated commit. I don't think we need to update INITIAL_DATA_SIZE, we can just modify the data extend test to just extend immediately if DATA_LOWATER is greater than INITIAL_DATA_SIZE.

agrover:
Why should INITIAL_DATA_SIZE be tied to DATA_LOWATER?

I'm also concerned about setting INITIAL_DATA_SIZE to a value in the vicinity of MIN_DEV_SIZE.

mulkieran:
Because ```INITIAL_DATA_SIZE``` is the initial size of the data device when a new pool is made. If ```INITIAL_DATA_SIZE``` is less than ```DATA_LOWATER``` then extension of data device will have to occur immediately after the thin pool is instantiated.

I'm not sure I see how ```MIN_DEV_SIZE``` is related.

agrover:
Not fixing (relnoting) for 1.0, issue #1152 pushed to 1.1. Let's close this and come up with a better patch for that dev cycle.

PR #1185, opened 2018-09-14
Unpublicize a few things that shouldn't be publicized

mulkieran:
These are probably artifacts from some previous testing or from
when thinpool had not yet been separated from Pool.


mulkieran:
Whoops! I forgot to reformat.

PR #1200, opened 2018-09-18
Change specification of variable length metadata

mulkieran:
Since DataTierSave and CacheTierSave exist, pass them directly to
CacheTier and DataTier setup methods.

Make check_metadata check properties of the CapSave part of BackstoreSave.

Do not pass next to Backstore::setup since with this change it can now
be extracted from BackstoreSave.

Adjust tests appropriately.

Reorder type definitions in serde_structs and document them and their
ordering more.


mulkieran:
Supercedes #1134. Code is identical, except that name change of ```DataTier::alloc``` is omitted from this PR since it is not really related, doesn't accomplish much, and is basically noise.

mulkieran:
I don't think Jenkins failures are related, so I'm redoing the build.

mulkieran:
Alas, I'm rebooting the machine.

mulkieran:
Worse, looks like it needs investigation.

PR #1202, opened 2018-09-18
Add versions to Sigblock, metadata header, and metadata

agrover:
Use previously unused bytes as version fields. All are currently version 1. Check when reading that value is as expected, and write proper value.

Corresponding doc update is: https://github.com/stratis-storage/stratis-docs/pull/98

### DIFF
--- a/src/engine/strat_engine/backstore/metadata.rs
+++ b/src/engine/strat_engine/backstore/metadata.rs
@@ -28,6 +28,8 @@ const MDA_RESERVED_SECTORS: Sectors = Sectors(3 * IEC::Mi / (SECTOR_SIZE as u64)
 
 const STRAT_MAGIC: &[u8] = b"!Stra0tis\x86\xff\x02^\x41rh";
 
+const STRAT_SIGBLOCK_VERSION: u8 = 1;
+
 #[derive(Debug)]
 pub struct BDA {
     header: StaticHeader,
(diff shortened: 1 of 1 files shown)


agrover:
Instead of the second commit, we also could just remove that test -- it may not have much value any more.

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -754,6 +785,13 @@ mod mda {
                     "MDA region data CRC".into(),
                 ));
             }
+
```
mulkieran: I'm not fully understanding the entire intent of this PR, but I'm still pretty sure that this chunk of code is in the wrong place. It doesn't make sense to read the buffer that contains the variable length metadata, check its crc, and then after that decide that there was no point in reading it in the first place. I think it makes sense to move this check elsewhere.

However, I also think that this is the wrong kind of check. This is a private method, buried in a sub-module of metadata.rs. This means that it can not be invoked by any method outside the module. It is only invoked in one place in the outer module ```MDAREgions::load_state```. That method first locates the appropriate mda, of two, and only then invokes this method,```MDAHeader::load_region```. It's at the point where the newer mda is located where it makes most sense to verify if the ```STRAT_METADATA_VERSION``` is the correct one. Probably the error check and return should be moved up to ```load_state``` and ```load_region``` should contain only an assertion.

agrover:
I don't think we have to worry in this PR about old metadata, because there is none yet. What we need to worry about is how this code will handle blockdevs with newer metadata. If `find_all` fails entirely, that's not correct -- a system should be able to have a pool with v1 metadata that is handled, and also a pool with v2 metadata, and v2 metadata blockdevs should be ignored, not cause an error.

src/engine/strat_engine/backstore/metadata.rs
agrover: Yep, you're right. fixed in 0ddf9edf. Do you think the check should return Ok(None) or Err()?

agrover:
@mulkieran We are currently defining only one metadata format -- pre-1.0 metadata is not supported and should fail. Can you please tell me more about remaining concerns you have?

agrover:
sounds good, I've added your commit from #1220. Are we good to go for final here?

PR #1204, opened 2018-09-19
Remove dbus enums

### LINKED ISSUE #1167: Remove bad DbusErrorEnum enums such as NIX_ERROR
We haven't looked at DbusErrorEnum and I think it is showing its age. For example:

* NIX_ERROR: we probably shouldn't expose the crate source of an error. Why should API users depend on this?
* INTERNAL_ERROR: How is this different from ERROR? It just sounds scarier (to me). See #1160. System-breaking errors should also consider putting the pool into an error state (#1121)
* IO_ERROR: Why would API users differentiate between this and ERROR?

I like the other ones -- API users may programmatically want to check if an API call failed for reasons like BUSY, ALREADY_EXISTS, or NOTFOUND, and then do a fallback action for example, but the ones above we should consider removing before API freeze for Stratis 1.0.

agrover:
fixes #1167

### DIFF
--- a/src/dbus_api/types.rs
+++ b/src/dbus_api/types.rs
@@ -23,7 +23,6 @@ macro_attr! {
 
         ALREADY_EXISTS,
         BUSY,
-        IO_ERROR,
         INTERNAL_ERROR,
         NOTFOUND,
     }
@@ -44,7 +43,6 @@ impl DbusErrorEnum {
             DbusErrorEnum::ALREADY_EXISTS => "Already exists",
             DbusErrorEnum::BUSY => "Operation can not be performed at this time",
             DbusErrorEnum::INTERNAL_ERROR => "Internal error",
-            DbusErrorEnum::IO_ERROR => "IO error during operation",
             DbusErrorEnum::NOTFOUND => "Not found",
         }
     }
--- a/src/dbus_api/types.rs
+++ b/src/dbus_api/types.rs
@@ -25,7 +25,6 @@ macro_attr! {
         BUSY,
         IO_ERROR,
         INTERNAL_ERROR,
-        NIX_ERROR,
         NOTFOUND,
     }
 }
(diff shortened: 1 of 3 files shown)


agrover:
The requirement I've been considering for these enums is: are different values something that a client could use to do something useful? e.g. a client may need a given filesystem to exist. The race-free way to ensure this is to try and create it and then continue if it gets ALREADY_EXISTS. (i.e. it's a Stratis error, but it's not really an error to the client) Any other error would cause the client to stop what it was doing.

Can you think of any case where a client would treat an INTERNAL_ERROR differently then an ERROR?

mulkieran:
If _I_ say an internal error I would probably consider trying to bring stratisd down and then start it up again. If I saw a not internal error I would try to poke around in the environment, see what caused the problem. An automatic monitoring or reporting tool might do something similar.

agrover:
So you are ok with never currently returning INTERNAL_ERROR, but just keep it around as documented valid return value for possible future use?

agrover:
Ok, I misread what you said above. You said:

> Getting rid of D-Bus IO and NIX_ERRORs and throwing most of the engine errors into D-Bus ERROR and none of them into D-Bus INTERNAL_ERROR is the part that seems right to me.

So what would be left to return INTERNAL_ERROR? If we tried to lookup a device that we were exporting a dbus object for and it wasn't there, i.e. things that should never happen?

mulkieran:
Wherever ```INTERNAL_ERROR``` is explicitly decided on in the D-Bus layer, i.e., not decided on because mapped to from an engine error it should remain, as far as I can tell.   ```dbus_api/filesystem.rs```, ```dbus_api/macros.rs```, ```dbus_api/pool.rs``` shouldn't be changed.

agrover:
OK, kept INTERNAL_ERROR and just mapped engine errors to ERROR.

Issue #1197, opened 2018-09-18
Handling unexpected blockdev removal


agrover:
As mentioned in #1195, we should probably change the state of the pool and possibly perform some cleanup actions when this happens.

Things that might affect what Stratis does are: 1) Were any of the pool's filesystems mounted at the time? and 2) Did the removal result in a partial pool (i.e. remaining members of the now-incomplete pool are still present on the system) or complete removal, where no part of the pool remains?

mulkieran:
From #1195 it seems clear that calling GetManagedObjects brought stratisd down. It would be nice to know why that is.

tasleson:
I just did a test with master (47234e249c5e743339c7ddfed3d112bfa9deac7b).  I created a pool on a usb drive, created a fs and mounted it.  When I yank the jump drive out nothing apparent happens until I run the stratis cli.  When I do that stratisd panics:

```
thread 'main' panicked at 'Kernel must return at least 8 values from thin pool status', /home/tasleson/.cargo/registry/src/github.com-1ecc6299db9ec823/devicemapper-0.26.1/src/thinpooldev.rs:488:9
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
   1: std::sys_common::backtrace::print
   2: std::panicking::default_hook::{{closure}}
   3: std::panicking::default_hook
   4: std::panicking::rust_panic_with_hook
   5: std::panicking::begin_panic
             at /builddir/build/BUILD/rustc-1.32.0-src/src/libstd/panicking.rs:425
   6: devicemapper::thinpooldev::ThinPoolDev::status
             at /home/tasleson/.cargo/registry/src/github.com-1ecc6299db9ec823/devicemapper-0.26.1/src/thinpooldev.rs:488
   7: libstratis::engine::strat_engine::thinpool::thinpool::ThinPool::total_physical_used
             at src/engine/strat_engine/thinpool/thinpool.rs:795
   8: <libstratis::engine::strat_engine::pool::StratPool as libstratis::engine::engine::Pool>::total_physical_used
             at src/engine/strat_engine/pool.rs:391
   9: libstratis::dbus_api::pool::get_pool_total_physical_used::get_used
             at src/dbus_api/pool.rs:344
  10: core::ops::function::Fn::call
             at /builddir/build/BUILD/rustc-1.32.0-src/src/libcore/ops/function.rs:78
  11: libstratis::dbus_api::pool::get_pool_property
[...]

PR #1211, opened 2018-09-21
No longer supply properties on the D-Bus by means of external calls

mulkieran:
This PR's goal is to eliminate all possibility of failure in methods that are used to supply property values on the D-Bus. Currently, the consequence of a failure in obtaining one of these properties is a message with the form:

```no total physical size computed for pool with uuid <some UUID>```

when most or all of the stratis CLI commands that require interaction with stratisd are run. One of these methods supplies its value via a devicemapper ioctl, another by means of a stat call.

The approach taken here is to cache the values used at the time when these methods are necessarily called, which occurs in the corresponding ```check``` methods for each module. The reason this seems like a good idea is that the view that the user gets of the stratisd system is consistent with the view that stratisd has of the system.

Related: #1197.

### DIFF
--- a/src/dbus_api/filesystem.rs
+++ b/src/dbus_api/filesystem.rs
@@ -232,9 +232,5 @@ fn get_filesystem_used(
     i: &mut IterAppend,
     p: &PropInfo<MTFn<TData>, TData>,
 ) -> Result<(), MethodErr> {
-    get_filesystem_property(i, p, |(_, _, fs)| {
-        fs.used()
-            .map(|v| (*v).to_string())
-            .map_err(|_| MethodErr::failed(&"fs used() engine call failed".to_owned()))
-    })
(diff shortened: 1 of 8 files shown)


mulkieran:
Alternative to #1172.

mulkieran:
Retrying Jenkins...

mulkieran:
restarted.

agrover:
Please consider https://github.com/stratis-storage/stratisd/pull/1226, this addresses my concern that value returned by `total_physical_used` is current.

mulkieran:
rebased, did same for thindev status as for thinpool dev status.

mulkieran:
Fixed a bug...was pritning out "bytes".

mulkieran:
rebased.

mulkieran:
rebased.

mulkieran:
rebased.

mulkieran:
rebased.

mulkieran:
Superceded by #1539.

PR #1216, opened 2018-09-24
Build release binaries

tasleson:
Include symbol information to make profiling etc. easier.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,25 +25,25 @@ fmt-travis:
 
 build:
 	PKG_CONFIG_ALLOW_CROSS=1 \
-	RUSTFLAGS='-D warnings' \
-	cargo build --target $(TARGET)
+	RUSTFLAGS='-g -D warnings' \
+	cargo build --target $(TARGET) --release
 
 build-no-default:
 	PKG_CONFIG_ALLOW_CROSS=1 \
-	RUSTFLAGS='-D warnings' \
-	cargo build --no-default-features --target $(TARGET)
+	RUSTFLAGS='-g -D warnings' \
+	cargo build --no-default-features --target $(TARGET) --release
 
 test-loop:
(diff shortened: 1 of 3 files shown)


tasleson:
IMHO It would be good to test what end users will be using, especially as we ready for release 1.  Release binary changes timings considerably.  Now if for some reason we aren't shipping a release binary, then we should disregard this.

tasleson:
From a cursory look, the total time to build and run a release build of stratisd may be a bit faster than running a debug build version, but that's comparing across nodes.  Don't have data for same node at the moment.

tasleson:
@trgill Have you seen a case where the travis is showing `in progress`, but if you click on details it's completed successfully?

tasleson:
@mulkieran Perhaps we should leverage https://doc.rust-lang.org/cargo/reference/manifest.html#the-profile-sections and configure our tests to run with a better optimization level and over flow checks enabled instead?

mulkieran:
@tasleson Yes, we could use a custom profile, that would just mean that we would have to figure out how to set it, etc. 

Random thoughts:
=====================================================================
There's a strong tradition of ambitious optimizations resulting in bugs, so, setting opt-level to three might flush out more bugs. OTOH, it might mask bugs by optimizing away the checks that we're interested in getting, like overflow-checks. OTOH, if we had a single profile, switching between ```make build``` and ```make test``` would not require all dependencies to be recompiled, I think, which would be nice for my personal work flow. So it seems like it might be worth a try. 

Then again, it might turn out that ```overflow_checks=true```, and ```opt-level=3``` are just not allowed in the same profile.

But someday, we will probably have to use a custom profile anyway, for some as yet unthought of reason.
=======================================================================

agrover:
Removing from 1.0 project, this is not a release blocker.

tasleson:
https://github.com/stratis-storage/stratisd/issues/1223

PR #1222, opened 2018-09-25
Remove dbus enums final

### LINKED ISSUE #1167: Remove bad DbusErrorEnum enums such as NIX_ERROR

agrover:
final PR of #1204

fixes #1167

### DIFF
--- a/src/dbus_api/types.rs
+++ b/src/dbus_api/types.rs
@@ -23,7 +23,6 @@ macro_attr! {
 
         ALREADY_EXISTS,
         BUSY,
-        IO_ERROR,
         INTERNAL_ERROR,
         NOTFOUND,
     }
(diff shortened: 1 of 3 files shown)


mulkieran:
Whoops! You should fix up test infrastructure, too: ```tests/client-dbus/src/stratisd_client_dbus/_stratisd_constants.py```. It looks like it was just the random luck of alphabetization that tests didn't fail. That is tests check for ```ALREADY_FOUND```, ```BUSY```, but they keep their numeric values after the change.

agrover:
fixes #1160

agrover:
I don't understand, this PR updates _stratisd_constants.py. What further needs to be changed?

PR #1226, opened 2018-09-25
Mulhern master cache thinpool status

agrover:


mulkieran:
Could you explain why you dismiss consistency within the data that is presented to the user as a worthwhile objective and make greater up-to-dateness (nothing can be truly up-to-date) the dominant objective at the cost of repeated dm calls while the GetManagedObjects method is running?

The notion presented in meeting of calling ```check()``` from the D-Bus API layer seems more sensible. That way you get consistency, and only the same number of invocations. But it would be a good idea to make sure ```check()``` can't fail.

Note that ThinPool::check also invokes status for all the filesystems belonging to the pool as well. Stash that as well, and ```Filesystem::Used``` can't fail.

agrover:
Consistency: Where is consistency hurt? The only place ThinPool::thin_pool_status is used is in total_physical_used(), afaics.

One dm call per pool has insignificant cost -- no I/O is being performed, it's just basically reading data out of kernel memory.

"Nothing can be truly up to date": A used value that was just read from the kernel is way better than the last time check() was called, which could have been a week ago.

> The notion presented in meeting of calling check() from the D-Bus API layer seems more sensible. That way you get consistency, and only the same number of invocations. But it would be a good idea to make sure check() can't fail.

I looked at this but it would have required changes to `Engine`, and the approach this PR takes seemed better.

agrover:
Since we haven't reached consensus on this and time is super short for 1.0, I'm guessing this will need to move to 1.1. Let's plan on that if not resolved by EOD today (Wed).

Issue #1188, opened 2018-09-14
Fix/Document stratis pool add-cache causes IO to stop for the duration of the add


tasleson:
Using `fio` test like:

```ini
[write-and-verify]
rw=randwrite
bs=4k
buffered=0
ioengine=posixaio
fallocate=none
iodepth=16
verify=crc32c
size=6G
max_latency=15s
```
`fio` will fail because the latency gets exceeded if we are using a debug build of `stratisd`.  I need to set the `max_latency` to 2 seconds to not fail the `fio` test with a release build when adding a cache device.

I'm not sure what improvements we can do to minimize the amount of time IO gets suspended, but we should likely instruct users to not add a cache device when the work load is sensitive to IO latency.

tasleson:
@agrover I'll try to get some profiling data for this to see what's going on with debug version.

tasleson:
For a release binary:
```
- 69.27% libstratis::engine::strat_engine::device::wipe_sectors                                                                ▒
               - libstratis::engine::strat_engine::device::write_sectors (inlined)                                                         ▒
                  + 46.79% std::io::Write::write_all                                                                                       ▒
                  + 15.60% <std::io::buffered::BufWriter<T> as libstratis::engine::strat_engine::device::SyncAll>::sync_all (inlined)      ▒
                  + 6.88% core::ptr::drop_in_place (inlined)
```

For debug binary

```
-   97.39%     0.00%  stratisd  stratisd                [.] libstratis::engine::strat_engine::device::wipe_sectors                         
     libstratis::engine::strat_engine::device::wipe_sectors                                                                                
   - libstratis::engine::strat_engine::device::write_sectors                                                                               
      - 95.65% std::io::Write::write_all                                                                                                   
         - 95.23% <std::io::buffered::BufWriter<W> as std::io::Write>::write                                                               
            - 94.85% <std::io::buffered::BufWriter<W>>::flush_buf                                                                          
               - 91.91% core::ptr::drop_in_place                                                                                           
                    <alloc::vec::Drain<'a, T> as core::ops::drop::Drop>::drop                                                              
                    core::iter::iterator::Iterator::for_each                                                                               
                    core::iter::iterator::Iterator::fold                                                                                   
                  + core::iter::iterator::Iterator::try_fold                                                                               
               + 2.94% <std::fs::File as std::io::Write>::write                                                                            
[...]

PR #1244, opened 2018-10-02
Wipe only the first 4K of each metadata device

mulkieran:
Related: #1188 .

If we only need to zero out the first 4K of the metadata device, we might as well do that instead of zeroing out the whole thing. That's not a complete fix to the problem of #1188, but it will go a long way to making it less obvious and unpleasant. Note that the current metadata device size in the backstore is 1 MiS where S stands for Sectors, i.e., 512 MiB.


agrover:
I think that should be #1188

Issue #1251, opened 2018-10-03
Bug 1635424 - stratis - creating a snapshot with "stratis filesystem snapshot" fails with "xfs_db: command not found" due to PATH issue


DaveWysochanskiRH:
#!/bin/sh
rm -f /tmp/file[123456].bin
for i in $(seq 1 3); do truncate /tmp/file$i.bin --size=1G; losetup -f /tmp/file$i.bin; done
for i in $(seq 4 6); do truncate /tmp/file$i.bin --size=1G; losetup -f /tmp/file$i.bin; done
stratis pool create pool1 /dev/loop[012]
stratis pool create pool2 /dev/loop[345]
stratis pool list
stratis blockdev list pool1
stratis blockdev list pool2
stratis filesystem list pool1
stratis filesystem list pool2
stratis filesystem create pool1 fs1
stratis filesystem create pool2 fs2
stratis filesystem list pool1
stratis filesystem list pool2
stratis filesystem snapshot pool1 fs1 fs1-snap1

Actual results:
INTERNAL ERROR: Command failed: cmd: "/usr/sbin/xfs_admin" "-U" "2877e0f7-9916-4eff-83cf-ee5c6439cabc" "/dev/dm-14", stdout:  stderr: /usr/sbin/xfs_admin: line 38: xfs_db: command not found

mulkieran:
I don't see how the version of stratisd being run can be the just released one, 1.0. If it were, the error would not say ```INTERNAL_ERROR```.

This looks like two bugs to me:

* There is an an xfs problem. It looks like ```xfs_admin``` has developed a dependency on ```xfs_db``` in a recent version. Or it maybe lost that dependency in all the versions we tested on, or something.

* stratis-cli should check the version of stratisd before it runs any commands to verify that it is one that stratis-cli is compatible with, but as far as I know, it doesn't.

DaveWysochanskiRH:
> I don't see how the version of stratisd being run can be the just released one, 1.0. If it were, the error would not say `INTERNAL_ERROR`.
> 

It was stratisd-0.5.5-2 but Andy indicated he thought this problem existed upstream as well so I filed the bug.  Sorry I should have tried an upstream version but I didn't have one handy.


> This looks like two bugs to me:
> 
>     * There is an an xfs problem. It looks like `xfs_admin` has developed a dependency on `xfs_db` in a recent version. Or it maybe lost that dependency in all the versions we tested on, or something.
> 
[...]

PR #1252, opened 2018-10-03
Invoke xfs_db directly

mulkieran:
Resolves: #1251 .


mulkieran:
We might have fewer constraints on how we run testing w/ this, as well. One of the reasons we can't run many tests on Travis is that Travis has an old version of ```xfs_admin``` that does not allow setting the UUID. If it has a version of ```xfs_db``` that does allow setting the UUID, than we can revive those tests on Travis, which will probably be a good thing. This might be worth checking out.

mulkieran:
test failed, I expect -x flag is actually needed.

agrover:
`man xfs_db` says:
```
xfs_db is used to examine an XFS filesystem. Under  rare  circumstances
it  can also be used to modify an XFS filesystem, but that task is nor‐
mally left to xfs_repair(8) or to scripts such as xfs_admin(8) that run
xfs_db.
```
I think we should keep using xfs_admin for now.

mulkieran:
The "-x" flag fixed it. It's funny that when we call ```xfs_admin``` we immediately are promoted to "expert" ```xfs_db``` users in almost all cases.

Pros and cons of switching to invoking ```xfs_db``` directly that I am aware of.

Pros:
* #1251 would have been impossible, so it would never have been filed. stratisd will succeed in making snapshots, even in a situation where the environment is kind of messed up.
* We expect to have stratisd running in situations where PATH variable is not even set. ```xfs_admin``` is a shell script which redirects to executables in all or almost all cases. It relies implicitly on the PATH variable being set. Calling ```xfs_admin``` in these circumstances will always fail to work, given the code as it is now. Note: This point is a bit more complicated than it looks, as one can make arguments about the probability that stratisd will be making snapshots in these early environments, which is kind of low, AFAIK. But that discussion could get really complicated, so I won't follow it up here.
* Knowledge of what we're actually doing, i.e., invoking ```xfs_db``` with "-x" flag set.

Cons:
None

Note: Being able to run more tests on Travis than formerly was a red herring. Switching to invoking ```xfs_db``` directly will not change anything there, AFAICT.

mulkieran:
We agreed that this was actually 1.0.1 in standup, so moving it there.

PR #1253, opened 2018-10-04
Run a test that can't be run on Travis on Travis

mulkieran:
I have this feeling that our testing may have been overconservative.


mulkieran:
Whoops, I misremembered. It's setting the UUID in mkfs.xfs which Travis didn't, and still doesn't support. So this is no help.

PR #1266, opened 2018-10-09
Call thin pool's check method after adding data devs

mulkieran:
This may allow a space starved thin pool to be expanded.

### DIFF
--- a/src/engine/strat_engine/pool.rs
+++ b/src/engine/strat_engine/pool.rs
@@ -310,8 +310,17 @@ impl Pool for StratPool {
         } else {
             // If just adding data devices, no need to suspend the pool.
             // No action will be taken on the DM devices.
-            self.backstore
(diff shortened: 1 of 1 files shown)


mulkieran:
This needs a test...

mulkieran:
Added a test. fmt target will fail, because I have not got the correct version of rustfmt yet.

mulkieran:
New test is running forever! That can happen sometimes, w/ loops.

mulkieran:
Jenkins failure. However reached point where extending the thin device's data device fails, so trying again. Maybe this time the test itself will proceed a bit further.

mulkieran:
The test passed. The warning that the extend failed was printed out twice, but that is only because there are two tests running, with different numbers of devices.

mulkieran:
Ok. Guess this is ready for a final PR.

PR #1224, opened 2018-09-25
Master cache split

mulkieran:
Related: #1188.

Issue #1014, opened 2018-06-26
Design, document, implement and on-disk metadata format for Stratis 1.0


agrover:
This is important because we will need to support this for a long time.

kocoten1992:
As a user, as long as I could upgrade stratisd without problem, metadata could change however needed I think

agrover:
Yeah. But from a coding perspective, 1.0 is when compatibility is something we need to worry about. Sometimes during development, we have made changes that cause the new version to fail to work against metadata created by the version from the previous day. That doesn't fly after 1.0. If we did make such a change, we would need to be able to read the new AND the old format. And maybe we'd write the metadata only ever using the new format, or maybe not, but it becomes something we need to think about a lot more than currently.

Therefore, we want to nail down the metadata format for 1.0 as much as possible. If we design our 1.0 metadata so that it more easily incorporates additions then it will make future development easier.

kocoten1992:
Hi, there is a trick I've learn when install gitlab, like when user are at 9.x version and want to upgrade to 11.x version (using ```apt-get upgrade gitlab```), they force me to go through 10.x before go to 11.x versions.

We could have apply that technique here, so user stratisd at 1.0 and want to go strange to 3.0, we said no, let them upgrade to 2.0 first, then we migrate the metadata, and then let them upgrade to 3.0

Hope this help :+1:

PR #1269, opened 2018-10-10
Eliminate Backstore::add_blockdevs

mulkieran:
This function made sense when add_datadevs and add_cachedevs did the same
thing. But now that they are so radically different, the shared function
is just no help at all.


mulkieran:
This should have gone in w/ other PRs associated with #1014, probably w/ #1071, but I missed it at the time. It might as well go in now. Cherry-picked from #1224.

mulkieran:
rebased.

PR #1270, opened 2018-10-10
Correct release version and version document link

tasleson:


mulkieran:
@tasleson Any objection to putting this in 1.0.1 project and just clearing it? It seems like it's just a correction to an oversight, so might as well go in now.

Issue #1268, opened 2018-10-10
test_thinpool_expand needs to be rewritten


mulkieran:
With changes in source from 1.0 release as well as changes in the test itself, it only accidentally tests anything.

Changes in the underlying source that effect this test that I know of:

* the data lowater value is no longer a constant since it can be, and is, set dynamically in ```ThinPool::setup``` and ```ThinPool::new```. So, using ```DATA_LOWATER``` in determining whether it is time to extend is now incorrect.

Changes in the test code that effect the test that I know of:

* The addition of BufWriter means that the data being written is buffered and that the pool has not necessarily reached the low water mark when the amount of data corresponding to the difference between the original size of the data device and the low water mark has been written.

The way I think that pool extension test should be written instead:
* Setup as before.
* Loop while writing data.
       * Use thinpool status result to check data total, data usage.
        * Use thinpool dev info to check data lowater
        * Stop writing when free space in thinpool dev falls below its data lowater. This simulates catching the event.
* Call ```ThinPool::extend_thin_data_device``` to extend the pool.
* Use thinpool status to verify change in total data.

mulkieran:
There are more problems in this test than I realized. One of the problems is that assumption of exactitude. As we expand the file, we can actually guarantee that the filesystem won't be expanded more,, so that the amount of data written to the file is actually less than the expansion of the filesystem and thindev.

PR #1260, opened 2018-10-04
Increase DATA_LOWATER to 1024

agrover:
I haven't been able to root-cause "25 filesystems" issue (#1152) but
it does appear that raising DATA_LOWATER to 1024 makes the issue no
longer reproduce.

Rename INITIAL_DATA_SIZE to DEFAULT_INITIAL_DATA_SIZE, and bump it up to
2 GiB. Let's not be stingy, and this will work better with the higher
DATA_LOWATER. But, still allow pools consisting of one blockdev of
MIN_DEV_SIZE by allocating via a request instead of alloc.

This may fix #1242.

Signed-off-by: Andy Grover <agrover@redhat.com>

PR #1284, opened 2018-10-18
Rewrite test_thinpool_expand

mulkieran:
Resolves: #1268.

### DIFF
--- a/src/engine/strat_engine/cmd.rs
+++ b/src/engine/strat_engine/cmd.rs
@@ -163,3 +163,8 @@ pub fn udev_settle() -> StratisResult<()> {
 pub fn create_ext3_fs(devnode: &Path) -> StratisResult<()> {
     execute_cmd(Command::new("mkfs.ext3").arg(&devnode))
 }
+
+#[cfg(test)]
+pub fn xfs_repair(devnode: &Path) -> StratisResult<()> {
+    execute_cmd(Command::new("xfs_repair").arg("-n").arg(&devnode))
+}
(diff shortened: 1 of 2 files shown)


mulkieran:
Note that this will resolve some of the problems w/ the tests in #1260. It is reasonable to fix the test first, so that it is not necessary to alter #1260 to fix up a test that is not actually working. It is more efficient and makes the commit log more aesthetic to fix the test first.

mulkieran:
However, there is one big flaw in the test...the file system is not being expanded.

mulkieran:
```
---- engine::strat_engine::thinpool::thinpool::tests::real_test_thinpool_expand stdout ----
	thread 'engine::strat_engine::thinpool::thinpool::tests::real_test_thinpool_expand' panicked at 'assertion failed: `(left == right)`
  left: `DataBlocks(766)`,
 right: `DataBlocks(768)`', src/engine/strat_engine/thinpool/thinpool.rs:1771:21
```

the amount written is 2 data blocks less than the original total.

mulkieran:
rebased.

mulkieran:
rebased again, to get line numbers right.

mulkieran:
> Maybe I'm not understanding the unit test fully, but it appears that it creates a file system, but then proceeds to write to the underlying block device directly. If this is indeed correct, that will corrupt the FS. I believe we should mount the FS and write to a file on it instead. Then we could umount FS when done and do a fsck to ensure that it's still intact while doing the other checks too.

I think that is a good idea, because the operation of stratis does not expect that users will be writing directly to the device node. I'll make that change as well while I'm at it.

mulkieran:
Whoops, remember to create the file.

mulkieran:
Note that fsck.xfs does nothing, successfully according to its man page. Again according to the man page, to check consistency of the filesystem in situ must use ```xfs_repair```.

Issue #1289, opened 2018-10-23
creating snapshot does not check for name conflict


agrover:
A tester reports:

When creating snapshot with the same name as previously created snapshot, the
snapshot is created. This produces name conflict in metadata and prevents any
further listing on any stratis part (blockdev, pool, fs). Restarting stratisd
does not load the whole pool because of metadata name conflict. Manual cleanup
is needed to fix this.

PR #1294, opened 2018-10-23
Snapshot dupe name check

tasleson:
Ensure that the file system snap shot name doesn't already exist in the existing file system names.

Fixes: https://github.com/stratis-storage/stratisd/issues/1289


tasleson:
Rebased

PR #1297, opened 2018-10-25
Ensure total cache size is less than 32 TiB

### LINKED ISSUE #1292: stratisd fails when cache grows to ~40TiB
A tester reports:

Adding lots of cache to pool and listing pools results in stratisd process stuck and unkillable. The addition fails with ENOSPC somewhere around 40T, it seems the pool runs out of metadata, device mapper dm_cache_resize fails with -28 error. At this point it is possible to restart stratisd service, but the pool will not come online. But if user tries to list pools with 'stratis pool', the command itself fails, because it gets no response. At this point the stratisd process is stuck and unkillable even by SIGKILL. 'systemctl stop stratisd' leads to hung task timeout calltraces, stratisd process persists.

agrover:
This is a short-term fix until we can support growing the cache metadata
device, or rebuilding the cache from scratch.

Does not limit the number or size of devices that can be added to the
cache tier, but constrains their use to a max of 32 TiB.

fixes #1292

### DIFF
--- a/src/engine/strat_engine/backstore/backstore.rs
+++ b/src/engine/strat_engine/backstore/backstore.rs
@@ -10,7 +10,7 @@ use std::path::{Path, PathBuf};
 
 use chrono::{DateTime, Utc};
 
-use devicemapper::{CacheDev, Device, DmDevice, LinearDev, Sectors};
+use devicemapper::{CacheDev, Device, DmDevice, LinearDev, Sectors, IEC, SECTOR_SIZE};
 
 use stratis::{ErrorEnum, StratisError, StratisResult};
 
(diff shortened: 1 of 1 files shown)


tasleson:
@agrover I think we should re-base this

agrover:
> EDIT: Yes, on further reflection I think it makes more sense to return an error if an effort is made to add more cache device than the metadata can handle, and that should be done in the cache tier module. Fewer surprises all around.

Ok, if you think so. Can you put together a PR that does this, asap?

agrover:
Based on the above, I'm going to close this PR and #1300.

PR #1299, opened 2018-10-29
Some tidies

mulkieran:
Two minor fixes to #1294. I'm putting it at 1.0.1 since #1294 was 1.0.1.


mulkieran:
If it passes tests, it qualifies as trivial.

agrover:
LGTM, merging as trivial.

PR #1300, opened 2018-10-29
Agrover issue 1292

mulkieran:


agrover:
against a now-moot PR, taking a different approach, closing.

Issue #1293, opened 2018-10-23
Handle cache tiers beyond ~40TiB


agrover:
see #1292, very large cache tiers run into trouble. stratisd should do what needs to be done (maybe blow away existing cache, increase block size, and recreate cache?) if given more than 40 TiB.

PR #1302, opened 2018-10-30
Limit the size of the cache sub-device

mulkieran:
Resolves: #1292.

The better solution is to learn to dynamically expand the cache sub-device, see #1293. This is just a way to place a restriction on the use of stratisd, for safety.


mulkieran:
This is not quite ready yet.

mulkieran:
It is very hard for me to come up with a Jenkins test that would test this functionality specifically, due to the large sizes required, so I'm putting it up for review as is.

Issue #1309, opened 2018-11-07
udevadm is not one of our required binaries


mulkieran:
We use it in source, so you might think it ought to be made a required binary. But we only use it in cleanup, so maybe we don't really need it to be a required binary. But we must be explicit about this in our comments somewhere, possible on ```udev_settle``` method and ```fs_settle``` method as well.

agrover:
I don't understand why it being used in cleanup makes a difference. The thing I think may make a difference is that we have a fallback if it's not present -- just wait 5 seconds. Cheezy, but effective.

FWIW on Fedora, udevadm is provided by systemd-udev, which is a protected package. and I'm guessing other things will break if it's not present. So it's kinda moot whether we officially list it as a required binary or not.

Ok, how about we NOT list it as a required binary (b/c it's not required, since we have a fallback).

mulkieran:
> I don't understand why it being used in cleanup makes a difference. The thing I think may make a difference is that we have a fallback if it's not present -- just wait 5 seconds. Cheezy, but effective.

_Since_ we only use it in cleanup, falling back to just waiting, as we do, is a reasonable thing to do. If it played an essential function, then we couldn't fall back to just waiting. My meaning is the same as yours, I just don't spell out the inference.

For consistency, let us list it, then, as it will do no harm and make the code/comments less complicated.

I'll do that, since I already have a commit for it.

agrover:
OK just to be clear to future theoretical readers, you're saying we don't think udevadm technically needs to be listed as a required binary, based on the above discussion, but you want to list it anyway for consistency's sake. If so, I am in agreement.

PR #1312, opened 2018-11-08
Make udevadm a required binary

mulkieran:
Resolves: #1309.


mulkieran:
Couldn't cd to ```/root/workspace/stratis-cli``` in ci-with-stratisd. Rerunning Jenkins just in case that fixes it.

mulkieran:
This is trivial if we agree that listing it as a binary is a good idea.

PR #1314, opened 2018-11-09
Use proptest instead of quickcheck for specification based tests

mulkieran:
Switch from using quickcheck to using proptest. The basic motivation is that proptest seems a good deal easier to use to me; no doubt this is partially because I am a veteran Hypothesis user. The other motivation is that it seems to automatically provide a wider range of values in the correct type, and makes it easy to construct ones own strategies, by using strategy composition. Also, I had a nice experience with counter-example reduction.

This also changes two not explicitly specification based tests to one proptest test.

This uses not the most recent minor version of proptest, but the next most recent version. The most recent version can't compile w/ rustc 1.26.0, because it makes use of some new features in fairly recent stable version. I was able to do what I expected w/out the new version, so I don't think that sticking one version behind is a big  deal, at least for a bit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ tests/test_config.json
 
 # coverage output
 htmlcov
+
+# proptest
+proptest-regression
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -45,6 +45,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "bit-set"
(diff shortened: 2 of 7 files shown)


mulkieran:
Identical to #1030 except that Cargo.lock changes are much less. Assuming this passes tests, I'm going to go ahead and merge it.

PR #1315, opened 2018-11-12
Run test_buffering by itself

mulkieran:
Previous tests must be setting the logger, so that when test_buffering
is run, it cannot set the buffer like it needs to.


mulkieran:
This doesn't seem to be necessary yet, i.e., test_buffering is passing regardless, so I"m letting this sit.

mulkieran:
We haven't proved that we need this, so I'm going to close it. We'll always know how to do it if we do need it.

PR #1206, opened 2018-09-20
Some ThinPool organization

mulkieran:
Cherry-picked from #1208 

Related: #1135.

PR #1316, opened 2018-11-12
Group segments in ThinPool struct

mulkieran:
It's a bit more organized that way.


mulkieran:
Cherry-picked from #1206 .

mulkieran:
Just another funky ```test_udev.py``` failure: https://stratis-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/stratisd/2252/console.

mulkieran:
I only group things together that belong together. However, I feel that grouping things is more urgently needed when there are many things than when there are few in a struct, class, etc.

Issue #1109, opened 2018-08-15
dbus-client tests should not use example() function to generate lists of devices


mulkieran:
```example()``` is not truly random, it is more "interesting", according to Hypothesis developers.

**Comment by mulkieran** [comment]
This is a really benign misuse, though. It isn't something that Hypothesis will deprecate, since it's not being used in any Hypothesis tests. The quality of the items drawn w/ example() doesn't really matter to the tests. We could improve the examples by using the external Hypothesis strategy fspaths. But that would add an extra dependency which isn't worth the trouble, because we are not testing filesystem paths.

We could eliminate the use of Hypothesis entirely, by writing a function that directly generates the random values wanted.

```
def random_legal_path_names(min, max):
  # choose a random number between min and max, n
      v = []
      for 0..n:
          # generate p strings
           # mush them together with a "/"
           # prepend a  "/"
          # add to list, v
  return v
```

That's avoiding using heavyweight machinery for a lightweight problem, but is it worth the trouble?

PR #1320, opened 2018-11-12
Remove Hypothesis dependency in tests

mulkieran:
Resolves: #1109.

### DIFF
--- a/tests/client-dbus/tests/dbus/_misc.py
+++ b/tests/client-dbus/tests/dbus/_misc.py
@@ -16,25 +16,30 @@ Miscellaneous methods to support testing.
 """
 
 import os
+import random
 import string
 import subprocess
 import time
 import unittest
 
-from hypothesis import strategies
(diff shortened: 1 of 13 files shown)


mulkieran:
Blocked by #1318, since the code is in such close proximity.

mulkieran:
rebased.

mulkieran:
This is a pretty trivial test infrastructure change. It's probably worth doing even though the abuse of Hypothesis wasn't causing an deprecation warnings. The reason for that is that Hypothesis was being abused so much, i.e., not even used in the context of a Hypothesis test, that Hypothesis couldn't notice it, and therefore didn't deprecate it.

mulkieran:
If this is acceptable, should do the same for stratis-cli tests.

mulkieran:
Whoops. I missed a pair of parens somehow. Wierd that linter didn't catch that.

PR #1321, opened 2018-11-13
Added help text to error message when a device is already in use

carzacc:
Addresses https://github.com/stratis-storage/stratis-cli/issues/208 along with https://github.com/stratis-storage/stratis-cli/pull/211

### DIFF
--- a/src/engine/strat_engine/backstore/blockdevmgr.rs
+++ b/src/engine/strat_engine/backstore/blockdevmgr.rs
@@ -432,10 +432,12 @@ fn initialize(
             match ownership {
                 DevOwnership::Unowned => add_devs.push((dev, (devnode, dev_size, f))),
                 DevOwnership::Theirs(signature) => {
+                    let dev = devnode.display();
                     let err_str = format!(
-                        "Device {} has an existing signature {}",
-                        devnode.display(),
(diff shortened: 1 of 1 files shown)


agrover:
Hmm. I think we've had stratisd report errors, but stratis-cli catch them and make helpful suggestions like this one. Should (and could) something like this go in stratis-cli, @mulkieran ?

mulkieran:
Related: https://github.com/stratis-storage/stratis-cli/pull/223.

mulkieran:
It has just occurred to me that maybe it wouldn't be wrong to enrich the message so that it returns more stuff. We could use a D-Bus dictionary to stash various components, i.e., the "message", the backtrace, or whatever is desired. This way, we do offer more information to the CLI. I don't know if the CLI would necessarily be able to make much use of the backtrace, though, if it was formatted as text. Anyway, I think any further discussion of this should probably be in a separate issue.

mulkieran:
I opened an issue about stratis-cli: https://github.com/stratis-storage/stratis-cli/issues/224.

mulkieran:
We shouldn't really do this, but I've added another issue: #1421 about better error reporting from the engine.

Issue #1218, opened 2018-09-25
Stabilize D-Bus API


agrover:
Based on discussion in #1191, let's not declare API as stable in 1.0, so we have some more time to make sure we're happy with it.

mulkieran:
I am still in favor of making the D-Bus properties be as inexpensive and safe to calculate as possible.

mulkieran:
Closed in favor of: https://github.com/stratis-storage/project/issues/52.

PR #1326, opened 2018-11-15
Get rid of some unnecessary boiler plate about enum constructors and numeric values

mulkieran:
This gets rid of a bunch of boiler plate, but preserves behaviour.

~~1. Since the value of an enum constructor as u16 depends upon the order in which it appears in the enum definition, it might be safer to number all the enum constructors, even the ones that use the default numbering: 0, 1, 2, etc. This would make it robust to changes in ordering.~~ It seemed better just to do this, so I did.
2. Some of the numbering for the constructors that don't use the default is a little odd and this should be reviewed in case the oddness is unintentional. But, might as well defer this question to #1218, since that's what it is about.


mulkieran:
If this passes it is a very easy ack.

mulkieran:
Whoops, there was some extra code left in.

agrover:
see [here](https://doc.rust-lang.org/1.29.1/reference/items/enumerations.html#custom-discriminant-values-for-field-less-enumerations). This makes it hard to add fields to enum values. I can imagine we'd never need to do this for e.g. `DbusErrorEnum` but for all the *State enums in engine/types.rs, these happen to need a value to be exposed in D-Bus but I imagine they're also going to be used heavily in program logic, when fields may be very useful.

Upon reflection, I think this is fine for now. These are in engine/types.rs, they're part of the Engine API basically. If we ever want to add fields to an enum inside StratEngine, we can define an engine-private enum, and then pay the expense of mapping the internal one to the API enum only when returning values via the generic trait methods, e.g. `Pool::free_space_state()`.

mulkieran:
> see [here](https://doc.rust-lang.org/1.29.1/reference/items/enumerations.html#custom-discriminant-values-for-field-less-enumerations). This makes it hard to add fields to enum values. I can imagine we'd never need to do this for e.g. `DbusErrorEnum` but for all the *State enums in engine/types.rs, these happen to need a value to be exposed in D-Bus but I imagine they're also going to be used heavily in program logic, when fields may be very useful.
> 
> Upon reflection, I think this is fine for now. These are in engine/types.rs, they're part of the Engine API basically. If we ever want to add fields to an enum inside StratEngine, we can define an engine-private enum, and then pay the expense of mapping the internal one to the API enum only when returning values via the generic trait methods, e.g. `Pool::free_space_state()`.

Yes. You're quite correct about that one drawback. We have some options if we find we need to make the constructors non-nullary.

1. Do as you suggest.
2. Implement ```From<the particular enum> for u16```.
3. Make a special trait ```ExportsDbusValue``` and put a special method on it.

But for now, this is all we need and the variegated approaches that we were using previously were just creating muddle.

Issue #1189, opened 2018-09-15
test_udev.py has begun failing intermittently


mulkieran:
This is the second time this has occurred, but here is one example: https://stratis-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/stratisd/1953/console.

mulkieran:
Recent: https://stratis-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/stratisd/2275/console

tasleson:
We got another failure with latest fixes, ref. https://stratis-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/stratisd/2328/console

mulkieran:
https://stratis-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/stratisd/2575/console

mulkieran:
This might be relevant: https://github.com/stratis-storage/stratisd/pull/1477#issuecomment-478593419. This is not about ```test_udev.py``` per se, but about device discovery failing in a regular test.

tasleson:
Addressed with https://github.com/stratis-storage/stratisd/pull/1510

mulkieran:
Assuming this was all caused by #1446, and fixed by #1510, this can be closed.

PR #1328, opened 2018-11-17
Fix test_udev.py intermittent errors

tasleson:
See if we can make this unit test run more reliably.

Fixes: https://github.com/stratis-storage/stratisd/issues/1189


tests/client-dbus/tests/dbus/pool/test_udev.py
```diff
@@ -121,6 +121,9 @@ def _start_service(self):
         """
 
         if self._service is None:
+            self.assertEqual(UdevAdd._process_exists("stratisd"), None)
```
mulkieran: AFAICT: This isn't part of any test. What you're doing here is verifying that the internal data of the test infrastructure matches reality as well as possible. Instead of using the test assertions you should use regular assertions or possibly return an exception. Same is true for ```_stop_service_remove_dm_tables```.

PR #1331, opened 2018-11-20
Greedily alloc all space to meta/data devices

agrover:
While we work on perfecting the thinpool extension code, one sure way to
avoid issues is to just fully extend from the start. That's what this
does: it bypasses much of the use of ThinPoolSizeParams, asks how much
space is available in the backstore, and allocates it to the meta and
data devices with a ~1:1000 ratio.

### DIFF
--- a/src/engine/strat_engine/pool.rs
+++ b/src/engine/strat_engine/pool.rs
@@ -314,6 +314,16 @@ impl Pool for StratPool {
             // No action will be taken on the DM devices.
             let bdev_info = self.backstore.add_datadevs(pool_uuid, paths)?;
 
+            // Temporary: Immediately fully extend thin meta/data onto new
+            // blockdevs
+            let available = self.backstore.available_in_backstore();
+            let meta_size: Sectors = available / 1000u16;
+            let data_size = available - (meta_size * 2u16);
+            self.thin_pool
+                .extend_thin_meta_device(pool_uuid, &mut self.backstore, meta_size)?;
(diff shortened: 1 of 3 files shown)


src/engine/strat_engine/pool.rs
```diff
@@ -314,6 +314,16 @@ impl Pool for StratPool {
             // No action will be taken on the DM devices.
             let bdev_info = self.backstore.add_datadevs(pool_uuid, paths)?;
 
+            // Temporary: Immediately fully extend thin meta/data onto new
+            // blockdevs
+            let available = self.backstore.available_in_backstore();
```
mulkieran: Note that ```available_in_backstore``` means all space on all blockdevs in the data tier that is not allocated to Stratis metadata. So, if the pool is big, and the new blockdevs added are relatively small, it is possible to vastly overestimate the amount of additional allocation that can be done.

The consequence of this would be that the thin meta device is extended proportionately much more than is really appropriate given the amount that can be allocated, since it gets the first request. This could result in an error return from the call to extend the thin data device.

I have just realized, looking at uses of ```extend_thin_data_device``` and ```extend_thin_meta_device``` that there is probably a bug, in that the calls expect different behavior from the callee than actually happens. I think I will look into that first, and then resume this review.

agrover:
Please look at #1336 for an alternative implementation.

agrover:
Closing in favor of #1336

Issue #1257, opened 2018-10-04
Extending meta and data devices in ThinPool::check code has diverged unnecessarily


mulkieran:
Also, there is a bug. That is, ```should_save``` is set to ```true``` when the meta device is extended by 0 sectors. The code that does the same task for the data device gets this right.

This whole action should be unified and abstracted to eliminate bugs and unnecessary code duplication.

mulkieran:
I guess I should label this as a bug because of the redundant and unnecessary metadata save.

agrover:
A separate issue for either the bug or the refactoring would be nice, since they will likely be fixed in separate PRs.

mulkieran:
I don't think you can bring meta- into synchrony w/ data- w/out fixing the bug (or introducing a new bug in data- to match the one in meta-).

agrover:
but maybe we want to fix the bug sooner than refactoring, yes?

mulkieran:
Well, I have this problem sometimes, where these sores show up on my left hip. When this happens I could go out and buy some fancy ointment or something to put on the sores. But I happen to know that those sores are actually a symptom of infectious bursitis. Bursa are these little fluid filled sacks that sit around your bones and muscles and allow the muscles to slide over the bones. But, things can go badly, and the bursa can become inflamed and then, if things go really badly, infected and eventually the infection will manifest as these sores on the skin. There are a lot of reasons for that problem, but in my case it is caused by a misalignment of the many joints between my sacrum and my femur which causes the muscles to be dragged against the joints and the bursa and result in this inflammation. So, when this happens, I treat it as a wake up call, a warning that my musculo-skeletal system is way out of whack. I pay no attention to the sores, unless they actually start to bleed, and focus with a fanatical intensity on getting my bones and joints realigned because I know that the consequences of my joints being poorly aligned can be things that are far more painful than mere infectious bursitis and that if I get the joints realigned the bursitis will almost certainly go away.

My analogy here is:
1. The bug in meta-expansion handling <=> the sores which are the visible symptom of the infectious bursitis
2. The divergence in code that should be identical, modulo meta- vs. data- distinctions <=> the misalignment of my various bones and joints

agrover:
but maybe we want to fix the bug sooner than refactoring, yes?

agrover:
For the  1.0.1 release we just want to include bug fixes, not refactoring. That's why I want the bugfix you identified to be tracked separately from the refactoring.

mulkieran:
[...]

PR #1334, opened 2018-11-21
Tidy up meta- and data- thinpool sub-device extension

mulkieran:
Resolves: #1257.


mulkieran:
There is still a pending question, however. Should meta/data extend be considered to have failed if the extension was by 0 sectors? Previously, meta said yes, and data said no, or the other way around, I can't remember which. Now they both say "No", i.e., if extended by 0 then data extend succeeded. Should they both say yes, i.e., that extend failed?

mulkieran:
Gadzooks! This is all wrong! Needs a redo!

mulkieran:
This looks mostly correct to me.

Issue #1329, opened 2018-11-19
Daemon failed to start (Invalid argument)


tasleson:
During testing of `udev_test.py` I ran into a situation where the daemon started, and then exited with an exit value of (1).  The only information I have is:

```
DEBUG libstratis::stratis::buff_log: BuffLogger: pass_through: true hold time: none
INFO stratisd: Using StratEngine
IO error: Invalid argument (os error 22)
```

The use of `?` and propagating errors isn't very helpful when we don't have any context.  Being able to debug field failures is important.

I tried reproducing this issue for hours by looping the test, with and without `strace` and was unable to reproduce.

related: https://github.com/stratis-storage/stratisd/issues/776

tasleson:
I added a bunch of code to check for errors and report them when encountered instead of just bubbling them up.

```
DEBUG libstratis::stratis::buff_log: BuffLogger: pass_through: true hold time: none
 INFO stratisd: Using StratEngine
ERROR libstratis::engine::strat_engine::backstore::metadata: f.seek(SeekFrom::Start(512) failed: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }
ERROR libstratis::engine::strat_engine::backstore::metadata: f.seek(SeekFrom::Start(4608) failed: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }
ERROR libstratis::engine::strat_engine::backstore::metadata: BDA::read failed Os { code: 22, kind: InvalidInput, message: "Invalid argument" }
ERROR libstratis::engine::strat_engine::backstore::setup: StaticHeader::device_identifiers failed: Io(Os { code: 22, kind: InvalidInput, message: "Invalid argument" })
ERROR libstratis::engine::strat_engine::engine: find_all failed Io(Os { code: 22, kind: InvalidInput, message: "Invalid argument" })
ERROR stratisd: StratEngine::initialize failed: Io(Os { code: 22, kind: InvalidInput, message: "Invalid argument" })
 INFO libstratis::stratis::buff_log: HandleGuard.drop()
IO error: Invalid argument (os error 22)
```

At the moment I have no idea why we cannot seek into the block device.

mulkieran:
We do not expect to be able to reproduce this.

PR #1335, opened 2018-11-26
RFC: Possible way to capture error context

tasleson:
We have a lot of code that uses the `?` operator.  When an issue occurs deep in the call stack and bubbles up we can be left with a simple error message like the one seen in https://github.com/stratis-storage/stratisd/issues/1329 
`IO error: Invalid argument (os error 22)` . 

The call stack looked something like:
```
main 
  -> run 
    -> initialize 
      -> find_all 
        -> device_identifiers 
          -> setup 
            -> BDA::read 
              -> read_sector_at_offset 
                -> seek
``` 

But we had no idea that this is where the error occurred as there are a number of different paths that could of lead to this error of invalid argument.  Having the information about where an error occurs would be helpful and thus the motivation for this PR.  This would add some useful information, ideally it would be better to have which function and it's arguments when the failure occurs.

### DIFF
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -6,6 +6,7 @@
 #![allow(doc_markdown)]
 
 extern crate devicemapper;
+#[macro_use(tll)]
 extern crate libstratis;
 #[macro_use]
 extern crate log;
@@ -145,12 +146,12 @@ fn trylock_pid_file() -> StratisResult<File> {
     };
     match flock(f.as_raw_fd(), FlockArg::LockExclusiveNonblock) {
         Ok(_) => {
-            f.write_all(format!("{}\n", getpid()).as_bytes())?;
+            tll!(f.write_all(format!("{}\n", getpid()).as_bytes()));
             Ok(f)
         }
         Err(_) => {
             let mut buf = String::new();
-            f.read_to_string(&mut buf)?;
+            tll!(f.read_to_string(&mut buf));
             // pidfile is supposed to contain pid of holder. But you never
             // know so be paranoid.
             let pid_str = buf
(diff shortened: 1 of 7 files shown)


tasleson:
Actually the `error!` macro gives us the location, so adding `file!` is redundant.

mulkieran:
I think this raises some issues:

* When to log.
* What to capture.
* What to log, i.e, error, warn, etc?

The failure crate (and error-chain as well) yield an error that has all the facilities to hold the entire stack trace. This gives you much more information than just the file/line and can be vastly more useful. This part can be programmatically accessed. Also, it is a Rust expert's problem, then, and not ours, which is somewhat good. So, when considering what to capture, failure crate is still sort of promising.

There is a problem with chaining, however, which apparently needs to be fixed by a fix to standard lib's Error. This fix has gone in, but it hasn't reached stable yet. chains can be printable, but they can not be programmatically accessed. This makes chaining sort of self-defeating for us at this time, as we found out when we wanted to programmatically access a an item in the chain that was not the head, and couldn't: the commit is 4fe59dd099b2beb14f56b67154a4a29864e2dd72.

If we capture the stack trace, then it is not necessary to log at the point where the function is called, because the information that the log infrastructure would capture is already captured by the stack trace. It is possible to log the propagated error at the top level and get the same information.

What to log is also a question. If we always log something as an error at every return site of a function we maybe over stating the case. The motivating example sure looks like an error, because it stopped the program. But if it is possible to recover from an error somewhere lower in the stack than it is generated, then it is less obvious that it should be logged as an error.

I've revived #988 which is a first attempt at seeing how it might be possible to use the failure crate in stratisd. It's exactly as it was before I found out the bad news about chaining, so it might be worth revising in terms of that new information.

I realize now that the idea of this logging activity is that bits of a stack trace would end up embedded in the log as individual log entries as the stack frames were dropped. But I'm pretty sure that they wouldn't necessarily be contiguous.

tasleson:
Rebased

PR #1336, opened 2018-11-27
Greedy allocate the thinpool (take 2)

agrover:
Instead of allocating space for the thinpool meta & data devices as they grow, allocate them to take all available space. This will avoid timing issues around data extensions and out-of-space under heavy write loads, but being "greedy" does mean we are committed to those enlarged allocations permanently, which may become limiting in the future.

* Change check() to allocate available space to meta/data device with a 1:1000 ratio
* Change lowater event calculation to also be based on events we need to check if filesystems need to grow
* Ensure metadata device segments are of a certain minimum size
* ~~Fix a bug in `check()` that did not update pool status for extensions made during `check()`~~. Merged in #1361.


agrover:
Probably also needs rebasing once #1334 goes in.

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -555,19 +493,11 @@ impl ThinPool {
...
-                    usage.total_meta.sectors(),
-                    usage.used_meta.sectors(),
-                    meta_lowater.sectors(),
-                    false,
-                ) {
-                    match self.extend_thin_meta_device(pool_uuid, backstore, request) {
+                // Ensure meta blocks is 1/1000th of total usable size
+                let target_meta_size = backstore.datatier_usable_size() / 1000u16;
+                if usage.total_meta.sectors() < target_meta_size {
+                    let meta_request = target_meta_size - usage.total_meta.sectors();
```
trgill: should we bump the meta size by some minimum?  maybe a MB?  otherwise we could trigger small meta resize.

trgill:
FYI - the Jenkins CI had trouble with this PR - log indicates repeated:

ERROR 2018-11-27T12:33:00Z: libstratis::engine::strat_engine::thinpool::thinpool: Thinpool out of space! -> OutOfSpace

https://stratis-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/stratisd/2306/console

src/engine/strat_engine/thinpool/thinpool.rs
agrover: should be resolved by e5a8698

agrover:
Seeing intermittent Jenkins failures that I cannot reproduce locally.

mulkieran:
Proof that "Address issue with real_test_fail_single_signature" commit should be in a separate PR.

1. Assume that the commit is trivial. Since it is a fix to a test (that apparently is failing intermittently, since it is clearly not failing all the time) it would be kindest to merge it as rapidly as possible into the master branch so that other developers get the benefit of the fix for their PRs regardless of the fate of the rest of the PR. Therefore, if the commit is trivial, it should be in a separate PR.
2. Assume that the commit is not trivial. If it is not trivial, that means it requires more work in order to make it acceptable. In that case it should be in a separate PR, so as not to tangle up this PR with difficult considerations about an orthogonal matter. Therefore, if the commit is not trivial it should be in a separate PR.

Therefore, "Address issue with real_test_fail_single_signature" commit should be in a separate PR.

agrover:
Opened #1347 for single_signature CI failure fix.

mulkieran:
AFAICT, "Fix test_add_datadevs() to also check PoolState" is orthogonal to this change. That is, checking the pool state in the loop in ```test_add_datadevs``` is the right thing to do, regardless of whether allocation is greedy or not. I've moved that commit into a separate PR: #1350, and added a few changes. If you can ack PR #1350, then the "Fix test_add_datadevs() to also check PoolState" commit can be dropped from this PR.

agrover:
Ok that's fine, but it's not exactly orthogonal (an independent, unrelated change) if this PR, absent that change, can't be merged (doesn't pass CI) without the fix, because this PR's changes caused the bug to manifest.

I've acked #1350 and will drop the commit from this PR as soon as that PR is merged.

mulkieran:
Please add an overall explanation of what this PR does. It is possible that an overall description will assist with understanding the purpose of the individual commits.

agrover:
I've updated the explanation.

agrover:
rebased on top of #1350 and other stuff.

mulkieran:
I believe that "Re-fetch thinpool status if there were thinpool extensions in check()" is intended to be permanent, i.e., it will not be removed when stratisd resumes lazy allocation.

It changes the paradigm a bit. Previously, stratisd checked the status whenever the ```check()``` method was invoked (generally becuase of an event), or most times a user used the CLI. If the ```check()``` method was invoked, the pool state, which is mostly determined by the thin pool status, was set, but in the CLI, not. (As I have said before, that the thin pool status is checked in an accessor method seems very inappropriate to me, so I'm not saying this is correct, I'm just saying it is what happens.) In the check method, some error messages are logged if the thin pool device state is bad and I believe in the accessor method invoked by the CLI as well.

This commit adds a conditional invocation of status and a yet more conditional state setting at the end of the ```check``` method. It omits all logging of bad state if one exists. So, it adds a third behavior, different from what happens at the top of the ```check()``` method and also different from the accessor method.

I think the behavior should be modified a bit so that a few things happen:
* It should check the status unconditionally. This is a reasonable thing to do, because it is less complex, and it detects changes that may have happened regardless of any actions taken or whether the pool was Ok before. We can imagine that the pool might, for example, have transitioned from running to out of data while our method was running. Why use extra logic to avoid catching that occurrence?
* It should set the state to the one determined by the status, regardless of what the status is. So, for example, if the thin pool had transitioned from out of data space to fail, the code in this commit would not set the new state, even though it had obtained the status and could set the state. I think it should set the state.
* Our log message should help the log reader, so it makes sense to decouple the information about the pool state from alarming messages about the thin pool device state. If that is accomplished then a log reader will have a correct understanding of the state of the pool. If the state of the pool is logged every time it is changed in the ```set_state()``` method, then the reader of the log will be informed if the action taken by the ```check()``` method resulted in a change of status or not.

I have made an attempt to add some modifications to achieve these in #1355. Please take a look.

agrover:
No, we don't need to handle arbitrary state changes while check() is running. If the thinpool transitions to a bad state while check() is running, we will get another event. The only thing we need to handle is changes check() is directly responsible for, which is why I believe only doing anything if `should_save` is appropriate -- it means we did something.

A better approach that addresses the concerns raised and also updates status after an extension may be to re-run `check()` if it returns true for `should_save`. Let's see how that looks...

mulkieran:
> No, we don't need to handle arbitrary state changes while check() is running. If the thinpool transitions to a bad state while check() is running, we will get another event. The only thing we need to handle is changes check() is directly responsible for, which is why I believe only doing anything if `should_save` is appropriate -- it means we did something.
> 
> A better approach that addresses the concerns raised and also updates status after an extension may be to re-run `check()` if it returns true for `should_save`. Let's see how that looks...

Ok! Please try it out in a separate PR.

mulkieran:
I think that redoing check if a change was made is reasonable, whatever ```check()``` itself does. I split the redoing stuff out into a separate PR: #1360. I believe that refetching the status within ```check()``` is no longer necessary, since check is being redone anyway, so I removed that commit. I tightened up the loop a bit. In case these changes are acceptable, and in the interests of moving forward quickly, I also put up a final version, #1361, which squashes these changes into one. If that PR can be merged, then it will be easier to move forward w/ the remainder of this PR.

agrover:
rebased on top of #1361.

agrover:
@mulkieran instead of removing the check() within thinpool setup and initialize, I believe it it better to remove the check() outside thinpool setup and initialize.  i.e. if these methods succeed, you get a ThinPool that is working without further actions needed. I pushed a commit to do this.

The other thing I'm not sure about is removing the extension tests. I know, I authored a commit that removes unused extension-related code, but I feel unsure if removing tests instead of just turning them off is a good idea, even if the same rationale somewhat applies. We should come up with a clear policy for situations like this.

Finally, we're really getting down to the wire on this one. If there are other review comments or additions please raise them immediately, and please indicate if there are remaining issues that should block merging, or ones that may continue to be worked separately. (e.g. neither of the two above mentioned issues I care about enough to hold up merging if we can't bottom out on them.)

mulkieran:
> This will avoid timing issues around data extensions and out-of-space under heavy write loads, but being "greedy" does mean we are committed to those enlarged allocations permanently, which may become limiting in the future.

I'm puzzled by all of the sentence that occurs after the comma. My understanding is that greedy allocation is a holding measure and not permanent, but the sentence makes it seem more permanent.

mulkieran:
Regarding removing the tests...the problem with keeping them in is that they may stay in their dead state long enough that they will require correction while they are dead, due to API  changes or other things that might cause them to fail to compile. They will be "fixed" w/out any attention to what they do, and when they come back on line they are likely to be wrong. Although I am generally a big fan of rebasing early and often, this one instance where I think it's best to wait until the code is actually being run.

mulkieran:
I have peeled off moving the check() method calls into a separate commit. As far as I can tell, this change isn't tied to greedy allocation at all.

mulkieran:
Superceded by #1369.

PR #1337, opened 2018-11-27
Set interface name in propchanged signal

agrover:
Apparently everyone does this, but we currently do not. This is useful for
filtering dbus signals.

### DIFF
--- a/src/dbus_api/util.rs
+++ b/src/dbus_api/util.rs
@@ -17,9 +17,6 @@ use super::super::stratis::{ErrorEnum, StratisError};
 
 use super::types::{DbusErrorEnum, TData};
 
-pub const STRATIS_BASE_PATH: &str = "/org/storage/stratis1";
-pub const STRATIS_BASE_SERVICE: &str = "org.storage.stratis1";
-
 /// Convert a tuple as option to an Option type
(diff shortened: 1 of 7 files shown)


src/dbus_api/util.rs
```diff
@@ -123,6 +123,7 @@ where
     prop_changed
         .changed_properties
         .insert(prop_name.into(), Variant(Box::new(new_value)));
+    prop_changed.interface_name = STRATIS_BASE_SERVICE.into();
```
mulkieran: Maybe this is supposed to be the interface specific to the property instead? e.g., ```org.storage.stratis1.filesystem```, ```org.storage.stratis1.pool```. I didn't see much information in the source code about this, but that seems more plausible to me, as we know that objects can support multiple interfaces and it is possible that an identical property name might occur in each interface, so the particular interface would be the best way to disambiguate among the two properties.

mulkieran:
If I'm right, then it's probably not trivial.

src/dbus_api/util.rs
agrover: fixed

trgill:
I'll restart the Jenkins tests now.

PR #1339, opened 2018-11-28
Set RUST_BACKTRACE=1 if debug

agrover:
We can programatically set this environment variable, which may be
helpful in ensuring full stack traces when debugging.

### DIFF
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -604,7 +604,11 @@ fn main() {
         match lock_file {
             Err(err) => Err(err),
             Ok(_) => {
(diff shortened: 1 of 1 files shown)


agrover:
@mulkieran @tasleson if we want to set this, it looks like it's easy. This is an example PR so please take and use this in an actual PR if you deem it worthwhile.

tasleson:
> @mulkieran @tasleson if we want to set this, it looks like it's easy. This is an example PR so please take and use this in an actual PR if you deem it worthwhile.

I believe we could export the environmental variable in the service file too, eg. `Environment=RUST_BACKTRACE=1`

PR #1343, opened 2018-11-29
RFC: Simplify main loop, remove some exiting daemon paths, fix clippy error

tasleson:
This turned out to be more involved than I had planned, especially with the need to build without dbus support.  This will need to be rebased and squashed, but I wanted to preserve the history of how I got to where it is.

This also fixes a clippy error for the complexity of `run`.

### DIFF
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -394,6 +394,88 @@ fn process_signal(
     }
 }
 
+/// Handle any client dbus requests.
+fn process_dbus(
+    handle: &mut libstratis::dbus_api::DbusConnectionData,
+    engine: &Rc<RefCell<Engine>>,
+    fds: &mut Vec<libc::pollfd>,
+    dbus_client_index_start: usize,
+) -> StratisResult<()> {
+    for pfd in fds[dbus_client_index_start..]
+        .iter()
+        .filter(|pfd| pfd.revents != 0)
+    {
+        for item in handle
+            .connection
+            .borrow()
+            .watch_handle(pfd.fd, WatchEvent::from_revents(pfd.revents))
+        {
+            if let Err(r) = libstratis::dbus_api::handle(
+                &handle.connection.borrow(),
+                &item,
+                &mut handle.tree,
+                &handle.context,
+            ) {
+                log_engine_state(&*engine.borrow());
+                print_err(&From::from(r));
+            }
+        }
+    }
+
+    // Refresh list of dbus fds to poll for. This can change as
+    // D-Bus clients come and go.
+    fds.truncate(dbus_client_index_start);
+    fds.extend(
+        handle
+            .connection
+            .borrow()
+            .watch_fds()
+            .iter()
+            .map(|w| w.to_pollfd()),
+    );
+
+    Ok(())
+}
+
+/// Iterate through all the pools and register them with the dbus interface and update the vector
+/// of file poll file descriptors.
+fn initialize_dbus(
+    handle: &mut libstratis::dbus_api::DbusConnectionData,
+    engine: &Rc<RefCell<Engine>>,
+    fds: &mut Vec<libc::pollfd>,
+) -> StratisResult<()> {
+    info!("DBUS API is now available");
+    let event_handler = Box::new(EventHandler::new(Rc::clone(&handle.connection)));
+    get_engine_listener_list_mut().register_listener(event_handler);
+    // Register all the pools with dbus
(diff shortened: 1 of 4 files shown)


agrover:
Definitely seems to be a big step towards taming `run()`! I hacked a little on it this afternoon in #1349, if either of those two commits look appealing.

mulkieran:
Putting in Review slot, so I remember to take a look.

tasleson:
@agrover I cherry picked your 2 commits and rebased, thanks!

mulkieran:
"Update comment on udev monitorng" is obviously correct.
"Don't exit on dbus pool registration errors" seems to me like it is could be a bad idea. Now stratisd engine has pools, apparently, but can't show any evidence of them. So, what does the user do? If they had an idea that the pools should be there, they panic. If they don't, they could start doing things that put them in a serious state of confusion. Until our code is prepared to deal with that sort of situation I don't think it's reasonable to cause it to happen.

I haven't had a chance to look at the other commits, yet.

agrover:
@tasleson I added one last commit in #1358 to put more stuff in UdevMonitor, for consideration.

Not much more fat to be trimmed in the main loop, that I can see.

agrover:
Looking at the code, the only thing that should cause a dbus registration error is duplicate object path, which should never happen because of our id-based object path generation scheme. If we could convince ourselves of this, then register_pool() could stop returning a Result, eliminating the issue.

tasleson:
Rebased, merge conflict corrected, and I picked up the latest commit from @agrover branch, Thanks!

tasleson:
Re-based and picked up latest commit `Do not return an error if register_object_path fails ` from @agrover

agrover:
@tasleson see #1384, which does as we discussed in today's standup.

tasleson:
@agrover Re-based and picked up your `expect` add.

mulkieran:
I'm still working on it. I think that the ```register_object_path``` commits should be in a separate PR.

agrover:
> My preference for this PR would be leave as is as it spans more than one file and adds new data structures and changes signatures, etc. However, if you like I can try to consolidate it a little bit, but I'm reluctant to squash it into one.

Yeah, no I didn't mean to squash it into one, just needs a rebase, and then whatever squashing makes sense, if any, based on your judgment.

I'm fine with including the changes in #1386 (but for one quibble) and would be fine pulling that out into a separate PR but I'm afraid dropping 4cf1b90 from this PR will prove to be a rebase conflict nightmare. (Just tried it. Yep.) Since the end result will be the same, let's give in to practicality for this one and keep the changes together.

mulkieran:
> "Update comment on udev monitorng" is obviously correct.
> "Don't exit on dbus pool registration errors" seems to me like it is could be a bad idea. Now stratisd engine has pools, apparently, but can't show any evidence of them. So, what does the user do? If they had an idea that the pools should be there, they panic. If they don't, they could start doing things that put them in a serious state of confusion. Until our code is prepared to deal with that sort of situation I don't think it's reasonable to cause it to happen.
> 
> I haven't had a chance to look at the other commits, yet.

At this point it seems that:

* "Update comment on udev monitoring" is still correct. Why not just pull it out and merge it now, since it is actually a bug fix, however minor? I'ld ack it immediately.
* "Don't exit on dbus pool registration errors." is fixed by subsequent commits in this PR. However, commit https://github.com/stratis-storage/stratisd/pull/1386/commits/91a76110149582f9855893dda63bc5fc57eb36a9 from #1386 should also be added to this PR, to complete it.
* "Try to encapsulate udev events" commit doesn't seem to do much good, and should probably be dropped. It seems to add greater complexity, but with what looks like no benefit at this time.

I'll continue looking at the other commits tomorrow.

mulkieran:
It looks like "Avoid callee..." and "Make handle..." are follow ons to the "Try to encapsulate udev events" commit. I'm going to try to look at them as if they were a coherent whole.

mulkieran:
AFAICT, separating out separate parts of this PR would be a massive quantify of work. I tried with the UdevMonitor work and also with the ```register_object_path``` commits. Neither one was less than grim.

mulkieran:
See #1389 for suggested changes.

mulkieran:
I've realized that there might be a bug in ```UdevMonitor``` stuff. Taking a look.

tasleson:
Replaced with https://github.com/stratis-storage/stratisd/pull/1391 which contains all the changes from @agrover  and @mulkieran and sqashed as suggested via https://github.com/stratis-storage/stratisd/pull/1343#issuecomment-451595737

PR #1344, opened 2018-11-30
Use matches crate for tests where possible

mulkieran:
The error reporting should the test fail is much better that way.


mulkieran:
matches crate is already packaged for Fedora 28.

PR #1345, opened 2018-11-30
Ensure udev db is as expected

tasleson:
If the udev db matches our expectations, then we are left with the conclusion that the issue resides with stratisd itself.

Hopefully helps with: https://github.com/stratis-storage/stratisd/issues/1189


tasleson:
Modified to handle the case where we don't have `libblkid` Stratis support.

tasleson:
Rebased

mulkieran:
This udev testing ended up in the wrong subdirectory to start with. But at that time it seemed like it was only in one file, so probably not worth making such a big deal about. Now that it is growing

tests/client-dbus/tests/udev/test_udev.py
```diff
@@ -31,8 +31,11 @@
 
 from stratisd_client_dbus._constants import TOP_OBJECT
 
-from .._loopback import LoopBackDevices
-from .._dm import remove_stratis_setup, _get_stratis_devices
+import pyudev
```
mulkieran: pyudev import should go above ```stratisd_client_dbus``` imports.

mulkieran:
It is necessary still to coordinate w/ @trgill changing the Jenkins tests, I think.

Issue #1255, opened 2018-10-04
Information about the pool state should be logged in the ```set_state``` method only


mulkieran:
Except in cases where the ThinPool struct does not exist and therefore the ```set_state``` method can not be called.

If this is done then the information that appears in the log will contain no duplicates and no redundancies.
Moreover, it will be possible for it to be reported in some kind of a canonical form that a human being can locate using mechanical means, e.g., grep, rather than just by visual scanning.

mulkieran:
This really applies to the other setting methods as well.

PR #1355, opened 2018-12-04
Fetch thinpool status and set pool state at end of check() method

mulkieran:
Cherry picked from #1336.

Related: #1255.


mulkieran:
It would be better to log an error only the first time. An endless repetition of the same error is kind of boring.

mulkieran:
Superceded by #1361. Keeping the branch around for a bit, since it makes a lot of changes that might be useful in other contexts.

PR #1361, opened 2018-12-07
Re-run thinpool::check() if it changed thinpool configuration (final)

mulkieran:
Make check() a private method and rename to do_check(). Make check()
a wrapper that re-runs do_check() until it returns false, to ensure
status values are up to date given thinpool changes.


agrover:
ok let's do this.

PR #1365, opened 2018-12-09
add basic debian packaging

gertux:
debian packaging support.
---------------------------------

The debian packaing files are in their standard location, the debian folder.

There's also a wrapper script provided that builds the debian package for stratisd.

This script can be used to build the package, but I provided it more as a convenient way to show how to use the setup but as it uses docker images it has the advantage of not polluting any systems and should be easy usable in a CI system.

The wrapper script uses the docker images provided by the debian project. Default it builds on/for buster, which is the next version. If you want to build for stretch just set DISTRO_VERSION=stretch before running the build script.

The setup produces a debian package with about the same contents as the rpm package (systemd and dbus config) the pre- and postinstall scripts to enable the systemd script are automatically generated using the systemd plugin for the debian packaging system.

Debian normally uses the debian/changelog file to determine the version of the package, which can be different from the software version. In the build script I generate this changelog file with one entry for the current version as known by cargo. I also used my own name and email address so this should probably be changed to a team member reponsible for packaging. Depending on the project's workflow the changelog file should be checked in.

The debian packaging cargo plugin is not used because that plugin doesn't use crates.io as repository, forces you to use debian provided deb packages for every dependency, so you have to duplicate dependency management for every rust dependency and quite a few dependencies were not available.


agrover:
Much thanks!!!

FYI before merging this, I'm going to contact the Debian pkg-rust-maintainers and feel out how hard it would be to get stratisd's rust dependencies packaged as well, as would be required for it to be included as part of a Debian release someday.

Thanks again! I now have hope :smiley:

gertux:
If all dependencies are upstream provided you can use dh-cargo and everything should be honky dory.
The problem with the 'every depency should be a debian rust library package' is that it's IMHO not really sustainable. This means that every version of every crate should be packaged which is a big task with +20K crates and will only get worse as rust gains more traction.

BTW, during my little adventure this weekend with dh-cargo I tried to circumvent the crates.io remapping. This worked using a rather ugly hack but then I got stuck on the fact that dh-cargo expects that for every included static library the license is known. This poses a problem with all crates that wrap a C library.

mulkieran:
@drckeefe I'm assigning this to you, solely on the basis that it is about packaging.

mulkieran:
I feel that likely this is stale, and that Debian Rust packaging processes will have changed in all this time, so I'm going to close this.

PR #1366, opened 2018-12-10
Make logs about extension less alarmist

mulkieran:
The amount by which the device is extended will always be less than
the amount requested unless the requested amount is a multiple of the
modulus. Don't warn in this case, only warn if it is less than the
maximum possible amount to satisfy the request.

### DIFF
--- a/src/engine/strat_engine/thinpool/thinpool.rs
+++ b/src/engine/strat_engine/thinpool/thinpool.rs
@@ -834,6 +834,7 @@ impl ThinPool {
         existing_segs: &mut Vec<(Sectors, Sectors)>,
         data: bool,
     ) -> StratisResult<Sectors> {
+        assert!(modulus != Sectors(0));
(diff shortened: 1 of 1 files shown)


mulkieran:
I think unnecessarily alarmist warnings qualify as a bug.

Related: #1336 .

mulkieran:
cli-with-stratisd failure was due to a missing directory.

mulkieran:
A clearly unrelated error in stratisd tests. It might be caused by the test getting a stale usage value, so that assertion in line 1709 in thinpool.rs fails. rerunning.

mulkieran:
The error is clearly unrelated, but it recurred. Kind of worrisome.

mulkieran:
The recurring error is in a test, ```test_meta_expand```, that will be removed by PR#1336 (or one of its descendants). So, we really ought to ignore the error at this time.

PR #1367, opened 2018-12-10
Lower ThinPool::check() methods into ThinPool::setup()/new()

mulkieran:
Change Thinpool::setup() to return if metadata needs saving. (Metadata
*always* needs saving after ::initialize().)


mulkieran:
I'm going to let this sit for a bit, it never seems compelling.

mulkieran:
obsolete

PR #1368, opened 2018-12-10
Agrover greedy allocate bis 2

mulkieran:


mulkieran:
stratisd tests failed due to that intermittent udevadm failure.

mulkieran:
Now, they're failing due to test_udev.py failure. But the new, simple test for greedy allocation is succeeding, at least.

PR #1372, opened 2018-12-12
Update to newer versions of proptest and uuid crates

### LINKED ISSUE #1371: RFE: update proptest to 0.8.x and uuid to 0.7.x
Would appreciate for Fedora packaging ;)

agrover:
fixes #1371

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -64,7 +64,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "byteorder"
-version = "1.2.1"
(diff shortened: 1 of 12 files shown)


ignatenkobrain:
Thanks!

mulkieran:
Not having success compiling new version of rand on new stable version of compiler.

agrover:
OK, we're going to advance the minimum rustc version to 1.31. For our convenience, this will happen as soon as 1.31 is available in Fedora (should be just a few days) since Jenkins CI uses rustc from repos, not rustup, and will need to be updated.

PR #1373, opened 2018-12-13
Tasleson more test udev

mulkieran:


mulkieran:
Tiny fixes.

mulkieran:
Tests are running now.

mulkieran:
Ok. Now tests are separated into two separate directories and have two separate targets in the Makefile. I think that the next step would be to:

1. Remove the use of ```STRATIS_DESTRUCTIVE_TEST``` from ```test_udev.py```. Since the tests exist in separate directories, distinguishing the destructive one by an environment variable is no longer necessary. All tests in udev subdirectory are destructive, those tests in dbus subdirectory are not.

2. Add some READMEs to help clarify the structure of these tests.

3. Add a Makefile target for tests to be run on Jenkins. Call it ```tests``` and have it do both ```dbus-tests``` and ```udev-tests``` regardless of whether either of the tests fails. Change Jenkins tests so that they just call ```make tests``` if possible. Otherwise, call the target for both tests ```dbus-tests``` and call the ```dbus-tests``` target ```dbus-layer-tests```, or something like that.

mulkieran:
Merged into #1345.

Issue #993, opened 2018-06-15
We are no longer running clippy on our test code but would like to


mulkieran:
This is hard to do, though, see: https://github.com/rust-lang-nursery/rust-clippy/issues/1436.

mulkieran:
We would like to run clippy on our test code, so that we don't develop bad habits that we are surprised to be corrected for in our source.

PR #1375, opened 2018-12-18
Make clippy a mandatory test once more

mulkieran:
Resolves: #993.

All are allows or do what clippy recommends except for https://github.com/stratis-storage/stratisd/pull/1375/commits/eaace9b567cc4042189e1b3418207e382205f3e8, to which you might want to pay special attention.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,6 @@ addons:
 language: rust
 
 matrix:
-    allow_failures:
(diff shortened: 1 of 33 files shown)


mulkieran:
This was failing on fedora-server-2 with udevadm issues, so I rebooted fedora-server-2 and re-ran.

PR #1379, opened 2018-12-19
Tasleson pr 1343

agrover:
This PR adds a commit to #1343, for your consideration. Commit message:

Do not return an error if register_object_path fails

It should never fail -- the only time it should fail is if we attempt to
register a duplicate object path, which should never happen because
object paths include a serial id.

Catching that possibility with a log message (hey, you never know)
results in a series of functions then not needing to return Result,
all the way up to the main event loop, where we can see that handling
dbus events now cannot fail. Cool.


agrover:
Tony pulled this commit in, closing.

PR #1386, opened 2019-01-03
Expect dbus method register_object_path to always succeed

mulkieran:
It must succeed because each new object path generated is guaranteed unique. Abstract object path generation into a separate method, in order to make it more likely that the uniqueness property will continue to hold as the code is maintained.


agrover:
Only suggestion I have is naming the helper function `make_object_path` (or maybe "create_". "build_"?) "get" doesn't quite capture it, IMO.

mulkieran:
Merged into #1391.

Issue #1136, opened 2018-08-30
Duplicate code in DataTier/CacheTier::setup can not yet be eliminated


mulkieran:
This code maps from the UUID of a blockdev and its segment information to a BlkDevSegment, which really means just picking up the device number. It is realized as a closure in both methods. The closures are identical. It is possible to write a BlockDevMgr method with the type:

```
BlockDevMgr -> (Box<Fn(DevUuid) -> StratisResult<BlkDevSegment>>)
```
or to write a function with the type

```
(Box<Fn(DevUuid) -> Option<Device>) -> (Box<Fn(&LayeredDev) -> StratisResult<BlkDevSegments>>)
```

These compile, but can not be used in the context they are wanted because the compiler insists that they are FnOnce, although they shouldn't be since the compiler also agrees that they are Fn. I think this may be a compiler bug, but it is also the reality, because I bet that returning a StratisResult makes the function FnOnce somehow. Maybe returning an Option would get around the problem, come to think of it. In any case, here's the issue, I hope to address it sometime.

mulkieran:
This would be a fun thing to check into, but I don't really think it's all that important. Perhaps we should postpone it.

mulkieran:
It's probably worth checking into again, now that our lowest supported compiler is much less low.

mulkieran:
This method:

```
pub fn junk(
    uuid_to_devno: Box<Fn(DevUuid) -> Option<Device>>,
) -> (Box<Fn(&BaseDevSave) -> Option<BlkDevSegment>>) {
    Box::new(move |base_dev_save: &BaseDevSave| {
        let parent = base_dev_save.parent;
        uuid_to_devno(parent).map(|device| {
            BlkDevSegment::new(
                parent,
                Segment::new(device, base_dev_save.start, base_dev_save.length),
            )
        })
    })
}
```

compiles w/ no problem and ought to do the right thing if used. But, it cannot be used, because in order to downcast it from the Box, it's type must be specified, as:

```
    pub fn setup(block_mgr: BlockDevMgr, data_tier_save: &DataTierSave) -> StratisResult<DataTier> {
        let mapper = junk(block_mgr.uuid_to_devno());
        let segments = data_tier_save.blockdev.allocs[0]
            .iter()
            .map(mapper.downcast::<Fn(&BaseDevSave) -> Option<BlkDevSegment>>>().unwrap())
            .collect::<StratisResult<Vec<_>>>()?;

        Ok(DataTier {
[...]

PR #1387, opened 2019-01-03
Abstract function for looking up segment

mulkieran:
Resolves: #1136.


mulkieran:
It seems that the only way to define a closure that can be passed to ```map``` is to define it in the body of the same function that invokes ```map```. This means it is impossible to write a function that returns a closure that the ```map``` function will accept.

The approach here is to define a simple function, not a closure, and iterate explicitly in the loop rather than using ```map```. This compiles and should work.

This is in preliminary review, because I'ld like to find out, before I proceed further, whether this would be considered to be worth doing.

mulkieran:
Superceded by #1399.

PR #1391, opened 2019-01-07
Simplify main loop (final)

tasleson:
Squashed version of https://github.com/stratis-storage/stratisd/pull/1343 which should be ready for merging.


tasleson:
@mulkieran Picked up your latest commits.  I tried squashing, but ran into conflicts so leaving them as they are.

Issue #1394, opened 2019-01-08
Bug: function set_write_throttling does not work for partitions


tasleson:
If you run `make test-real` and specify block devices that are partitions, the unit test `real_test_write_throttling` will fail:

```
---- engine::strat_engine::throttle::tests::real_test_write_throttling stdout ----
thread 'engine::strat_engine::throttle::tests::real_test_write_throttling' panicked at 'called `Result::unwrap()` on an `Err` value: Io(Os { code: 19, kind: Other, message: "No such device" })', libcore/result.rs:1009:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
   1: std::sys_common::backtrace::print
   2: std::panicking::default_hook::{{closure}}
   3: std::panicking::default_hook
   4: std::panicking::rust_panic_with_hook
   5: std::panicking::continue_panic_fmt
   6: rust_begin_unwind
   7: core::panicking::panic_fmt
   8: core::result::unwrap_failed
             at /builddir/build/BUILD/rustc-1.31.0-src/src/libcore/macros.rs:26
   9: <core::result::Result<T, E>>::unwrap
             at /builddir/build/BUILD/rustc-1.31.0-src/src/libcore/result.rs:808
  10: libstratis::engine::strat_engine::throttle::tests::test_write_throttling
             at src/engine/strat_engine/throttle.rs:103
  11: core::ops::function::Fn::call
             at /builddir/build/BUILD/rustc-1.31.0-src/src/libcore/ops/function.rs:78
  12: libstratis::engine::strat_engine::tests::real::test_with_spec::{{closure}}
             at src/engine/strat_engine/tests/real.rs:243
  13: std::panicking::try::do_call
             at /builddir/build/BUILD/rustc-1.31.0-src/src/libstd/panicking.rs:310
  14: __rust_maybe_catch_panic
  15: std::panicking::try
             at /builddir/build/BUILD/rustc-1.31.0-src/src/libstd/panicking.rs:289
  16: std::panic::catch_unwind
             at /builddir/build/BUILD/rustc-1.31.0-src/src/libstd/panic.rs:392
  17: libstratis::engine::strat_engine::tests::real::test_with_spec
             at src/engine/strat_engine/tests/real.rs:243
  18: libstratis::engine::strat_engine::throttle::tests::real_test_write_throttling
[...]

Issue #1393, opened 2019-01-08
Bug: Stratis on partitions end up being missing at boot


tasleson:
When you create a pool using partitions they don't get brought up at boot.  This is because the udev listener is configured to skip block devices that are partitions.  Tested correction below.

``` diff
diff --git a/src/bin/stratisd.rs b/src/bin/stratisd.rs
index 01592fe..2df6a0b 100644
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -317,7 +317,7 @@ fn run(matches: &ArgMatches, buff_log: &buff_log::Handle<env_logger::Logger>) ->
     // This is especially important since stratisd must run during early boot.
     let context = libudev::Context::new()?;
     let mut monitor = libudev::Monitor::new(&context)?;
-    monitor.match_subsystem_devtype("block", "disk")?;
+    monitor.match_subsystem("block")?;

```

PR #1396, opened 2019-01-09
Integration test fixes

tasleson:
This PR addresses the following issues:
* https://github.com/stratis-storage/stratisd/issues/1393
* https://github.com/stratis-storage/stratisd/issues/1394
* Incorrect State value in stratis cli due to endian issue in dbus library

PR tested on x86_86 & Z using partitions instead of entire disks.


agrover:
Looks great. Only change: we no longer need the walkdir crate, so drop it?

PR #1398, opened 2019-01-10
Tighten up computations connected w/ FreeSpaceState

mulkieran:
Tidies up ```free_space_check``` and eliminates a now redundant state from ```FreeSpaceState``` enum.

### DIFF
--- a/src/engine/strat_engine/thinpool/thinpool.rs
+++ b/src/engine/strat_engine/thinpool/thinpool.rs
@@ -639,30 +639,21 @@ impl ThinPool {
         self.set_free_space_state(new_state);
 
         match (self.free_space_state, new_state) {
-            (FreeSpaceState::Good, FreeSpaceState::Warn) => {
-                // TODO: other steps to regain space: schedule fstrims?
-            }
-            (FreeSpaceState::Good, FreeSpaceState::Crit) => {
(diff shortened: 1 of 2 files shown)


mulkieran:
This is a follow-up to #1396, which removed all the throttling action, since it didn't work on partitions, and thereby all distinctions between the Warn state and the Good state. For the sake of simplicity, use of the Warn state should be dropped. Unfortunately, free space state is exported on the D-Bus, so, without changing the stratis-cli, it must stay as it is. This is a good argument for keeping the external enums that are exported on the D-Bus and those that are used internally separate.

mulkieran:
```test_udev,py``` failure, all else succeeding.

agrover:
We want to expose (and signal) a different value (Warn) for the FreeSpaceState property to the user, even though it (currently) no longer changes Stratis's internal behavior. We want the user to know Stratis is getting full, even though things aren't critical yet.

mulkieran:
> We want to expose (and signal) a different value (Warn) for the FreeSpaceState property to the user, even though it (currently) no longer changes Stratis's internal behavior. We want the user to know Stratis is getting full, even though things aren't critical yet.

Ok. I have a new PR which just includes the first, tidy commit.

PR #1400, opened 2019-01-11
Deny all future incompatible lints

mulkieran:
This sets up the Makefile so that all future incompatible constructs are denied for all build and test targets. The only future incompatible construct stratisd has is its floating imports. The result of this PR is to cause all imports to have the smallest path possible whether relative or absolute. Where there is a tie in length, the relative path is preferred. This PR also remove a re-export which does not appear to be useful. It also denies one additional lint category "unused", which only flags two lints.

It doesn't deny rust-2018-compatibility or non-standard-style categories, because these are a subset of future-incompatible and warnings respectively.

It doesn't deny rust-2018-idioms, because that brings in a whole lot of new and possibly interesting lint messages, and it seems better to stop at this point.

### DIFF
--- a/src/dbus_api/api.rs
+++ b/src/dbus_api/api.rs
@@ -15,15 +15,15 @@ use dbus::tree::{
 use dbus::{BusType, Connection, ConnectionItem, Message, NameFlag};
 use uuid::Uuid;
 
+use crate::dbus_api::consts;
 use crate::engine::{Engine, Pool, PoolUuid};
 use crate::stratis::VERSION;
 
-use super::blockdev::create_dbus_blockdev;
-use super::consts;
-use super::filesystem::create_dbus_filesystem;
-use super::pool::create_dbus_pool;
-use super::types::{ActionQueue, DbusContext, DbusErrorEnum, DeferredAction, TData};
-use super::util::{
+use crate::dbus_api::blockdev::create_dbus_blockdev;
+use crate::dbus_api::filesystem::create_dbus_filesystem;
+use crate::dbus_api::pool::create_dbus_pool;
+use crate::dbus_api::types::{ActionQueue, DbusContext, DbusErrorEnum, DeferredAction, TData};
+use crate::dbus_api::util::{
     engine_to_dbus_err_tuple, get_next_arg, msg_code_ok, msg_string_ok, tuple_to_option,
 };
 
--- a/src/dbus_api/blockdev.rs
+++ b/src/dbus_api/blockdev.rs
@@ -11,12 +11,11 @@ use dbus::Message;
 
 use uuid::Uuid;
 
+use crate::dbus_api::consts;
 use crate::engine::{BlockDev, BlockDevTier, MaybeDbusPath};
 
-use super::consts;
-use super::types::{DbusContext, DbusErrorEnum, OPContext, TData};
-
-use super::util::{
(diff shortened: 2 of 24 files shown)


trgill:
looks like the stratisd tests failed due to the test node being in a bad state....

agrover:
I like three of these but have two quibbles with a8cb070. First, this doesn't give developers a clear way to do things -- "prefer shortest" requires everyone to try both and see which is shorter, and there's no way by looking at an import to know if the import should be in the other form or not.

I favor always using absolute imports, because the main advantage of one over the other is how they behave when code is moved or reorganized. Relative imports require imports in moved code to change (i.e. add or remove "super"s). Absolute imports require imports to possibly change if the *imported* code changes location. Since code is called more places than it's written (we hope) this should result in less total lines changed as code moves around. And, I personally find top-down imports easier to get right in new code, with less guessing about how many supers are needed etc.

mulkieran:
> I like three of these but have two quibbles with [a8cb070](https://github.com/stratis-storage/stratisd/commit/a8cb0701264722fd83428b680a2a3bf8613a6f62). First, this doesn't give developers a clear way to do things -- "prefer shortest" requires everyone to try both and see which is shorter, and there's no way by looking at an import to know if the import should be in the other form or not.
> 
> I favor always using absolute imports, because the main advantage of one over the other is how they behave when code is moved or reorganized. Relative imports require imports in moved code to change (i.e. add or remove "super"s). Absolute imports require imports to possibly change if the _imported_ code changes location. Since code is called more places than it's written (we hope) this should result in less total lines changed as code moves around. And, I personally find top-down imports easier to get right in new code, with less guessing about how many supers are needed etc.

I think relative imports have the same advantage over absolute imports that you say absolute imports have over relative imports, _when the relative import path is short_. There are many changes that can be made in the directory structure. For example, the name of a directory can be changed. In that case, the relative paths will be untouched, but the absolute paths that include the changed directory name will all have to be changed. Or an entire directory may be lifted up or moved down. In that case, all the relative imports that were within the directory will be unchanged, but if they were absolute imports they would all have to be mass-changed. So I think having distant imports be absolute and near imports be relative is the best way to minimize import edits during code refactoring.

The rule of minimum length has the good property of ensuring that short relative imports are used, but that long relative imports are transformed to absolute imports. But it's not the only rule that will do this.  A simpler rule, like "use a relative import if the path length is <= (n), use an absolute import otherwise" 
would work as well, supposing n were a nice small number like 1 (or 2). I prepared a patch that makes this number 1 generally (but 2 in test modules). I tried it out in the ```thinpool``` module only.

mulkieran:
I must as well see how nested imports look, while I'm here.

mulkieran:
rustfmt has a nested imports configuration parameter, but it is not yet stable. It can already format nested imports reasonably well, in its default, regular setting, but it can not convert except on nightly.

mulkieran:
We could consider converting to nested imports using nightly and then formatting using our current rustfmt. The risk, of course, is that, rustfmt will change its mind about what constitutes the correct way to format nested imports in between now and when the configuration parameter becomes stable. That would cause spontaneous reformatting of our nested imports to occur --- which maybe wouldn't be such a big deal. However, we have been really enjoying rustfmt stability in recent months, so it feels like a bad idea to jeopardize it. Votes are on the side of waiting.

PR #1406, opened 2019-01-16
Deny all future-incompatible and unused lints

mulkieran:
The basic motivation is to ensure future compatibility.
The only interesting lint is the one that forbids floating imports.
This PR changes the imports so that absolute imports are used everywhere, except when importing the names in a module being tested from a test module.

Nested imports were considered but rejected at least for now. In themselves, they look like a very desirable change; they seem a good deal clearer than line-by-line imports. However, there needs to be considered the cost of making the change and also rustfmt's handling of nested imports. rustfmt can convert to nested imports but at this time only on nightly. Currently, the conversion mechanism is under discussion and may change. There is no external auto-conversion mechanism that I can find. Doing the conversion by hand would be so time-consuming as to be practically morally wrong. Writing one's own converter would take time, and rustfmt's would eventually supercede it anyway. The option of using nightly version of rustfmt to do the conversion and then using stable version of rustfmt to format the already converted code seems a bit risky, due to the instability of the conversion in rustfmt. It may look better as the conversion stabilizes.

Supercedes #1400.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+DENY = "-D warnings -D future-incompatible"
+
 TARGET ?= "x86_64-unknown-linux-gnu"
 
 ${HOME}/.cargo/bin/cargo-tree:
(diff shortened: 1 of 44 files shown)


mulkieran:
udevadm failure on fedora-server-3 - reboot.

mulkieran:
rebased.

mulkieran:
The only change since prior approvals is a rebase, so this might as well go in now.

PR #1411, opened 2019-01-18
Fix some clippy lints

mulkieran:


mulkieran:
> If lints are going to be based on a specific version, then that version needs to be documented and pinned too.

In order to pin clippy it is necessary to pin the version of the compiler that clippy is running as the recommended way to run clippy is as a rustc component. It turns out that Travis makes this easy to do.
I have added a commit that should accomplish this.

PR #1414, opened 2019-01-18
Set fast-finish to true

mulkieran:
Build is marked as finished as soon as all required tasks complete, i.e.,
the allowed failures are not waited for.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ addons:
 language: rust
 
 matrix:
+    fast-finish: true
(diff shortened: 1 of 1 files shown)


mulkieran:
rebased.

mulkieran:
I'm merging this now as I don't think anybody cares. I expect it would have been the default if Travis had thought of it when they thought of allowed failures. Why wait for allowed failures to complete when they don't affect the summary value anyway?

PR #1416, opened 2019-01-23
Get rid of a redundant assertion in a test

mulkieran:


mulkieran:
rebased.

mulkieran:
mergify is not working at all, already approved by one person, I don't think anybody would argue that this change isn't trivially correct, so merging.

PR #1418, opened 2019-01-23
Use v2 of mergify configuration syntax

mulkieran:
V1 will become unsupported at the end of January.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,12 +1,13 @@
-rules:
-  default:
-    disabling_label: "do not merge"
-    protection:
-      required_status_checks:
-        strict: true
-        contexts:
-          - continuous-integration/travis-ci
-          - stratisd
-          - cli-with-stratisd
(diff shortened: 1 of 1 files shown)


mulkieran:
It's trivial assuming that we trust the mergify developers to convert properly.

mulkieran:
> Did we want to try out using `rebase` for `method`?

I think it makes most sense just to merge this PR as is before Jan 31 arrives (assuming that there is interest in continuing to use mergify), so that our CI doesn't break on that day or the subsequent.

Then we can approach changing the policy in a separate PR, which would have a less strict deadline.

It also might not be so simple; according to my initial reading, the important key is ```strict_method```, not ```method```.

mulkieran:
I would have expected it to auto-merge. I'm taking a look to see why not.

mulkieran:
It doesn't auto merge via mergify because it is a change to the mergify configuration. merging by hand.

PR #1420, opened 2019-01-28
Add pr to end of Travis build

mulkieran:
So that it will match.


mulkieran:
This is already tested in devicemapper, so might as well merge right now. It's funny, the name of the Travis tasks changes when it starts actually running, and that's when it gets the "pr" suffix. But that's what has to be matched on. Perhaps mergify started matching more precisely w/ new version.

Issue #1291, opened 2018-10-23
Unclear error message when trying to add an existing member as cache dev


agrover:
It looks like this has the proper behavior -- attempting to add a data dev already in a pool as a cache dev fails, whereas trying to (re)add it as a data dev idempotently succeeds. But the error message is not great. It would be better if it said something about the blockdev already being a member of the pool, but in a different tier.

```
# stratis pool create test_pool /dev/sdj
# stratis pool add-cache test_pool /dev/sdj
Execution failure caused by:
ERROR: Device /dev/sdj with pool UUID is unknown to pool
# stratis blockdev
Pool Name  Device Node    Physical Size   State  Tier
test_pool  /dev/sdj            9.10 TiB  In-use  Data
# stratis pool add-data test_pool /dev/sdj
# stratis blockdev
Pool Name  Device Node    Physical Size   State  Tier
test_pool  /dev/sdj            9.10 TiB  In-use  Data
```

agrover:
Not adding to 1.0.1 because code has correct behavior other than the error message returned.

jkrysl:
Doing the vice versa operation shows the same result:
# stratis pool create test_pool /dev/sda
# stratis pool add-cache test_pool /dev/sdb
# stratis pool add-data test_pool /dev/sdb
Execution failure caused by:
ERROR: Device /dev/sdb with pool UUID is unknown to pool

mulkieran:
When we initialize a new set of devices we pass to the initialize method the uuids of the devices that the currently being addressed tier owns. In order to correctly identify it as a member in good standing of the opposite tier, a fair amount of restructuring would have to be done. However, the error message itself can be improved to indicate that there is a possibility that the device belongs to the opposite tier.

mulkieran:
Blocked by https://github.com/stratis-storage/project/issues/51, at least partially.

PR #1429, opened 2019-02-06
Precondition check for block device usage

tasleson:
Check that a block device isn't already in use for the engine.

This works, but isn't ideal as the engine caller needs to validate
block device preconditions instead of the calls themselves. This occurs
as we don't have the entire engine context known when we are executing
code for a specific pool for example.  The following additional approaches
were attempted.

* We cannot pass the engine to the calls that need it as it is already
  borrowed mutably.
* Refactor the interface, so that we call methods on the engine instead of
  dispatching to the individual items contained in the engine, this results
  in the engine knowing too much of the internals of the pool.

Addresses: https://github.com/stratis-storage/stratisd/issues/1291

### DIFF
--- a/src/dbus_api/pool.rs
+++ b/src/dbus_api/pool.rs
@@ -209,11 +209,19 @@ fn add_blockdevs(m: &MethodInfo<MTFn<TData>, TData>, tier: BlockDevTier) -> Meth
         .get(object_path)
         .expect("implicit argument must be in tree");
     let pool_uuid = get_data!(pool_path; default_return; return_message).uuid;
+    let blockdevs = devs.map(|x| Path::new(x)).collect::<Vec<&Path>>();
+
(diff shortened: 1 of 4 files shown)


mulkieran:
Note: Trying to characterize the problem. It first showed up because someone tried to add a device to the cache tier of one pool, but that was already being used by the data tier of the same pool. The behavior, not adding, was correct, but the error message, ```"Device {} with pool UUID is unknown to pool"``` was unhelpful. What would happen now if the device belonged to a wholly different pool? The behavior would be the same, but the error message would be a good deal more helpful: ```"Device {} already belongs to Stratis pool {}"```.

mulkieran:
I don't think we should add a parallel mechanism for checking ownership of devices. We already have one built on a mix of udev and other stuff. We should try to work within that, if possible.

PR #1430, opened 2019-02-07
Log start and end of event handling activities in the main loop

mulkieran:
This ought to allow us to compare timing of, for example, D-Bus handling in
stratisd w/ D-Bus invocations in stratis CLI.


mulkieran:
Getting the time stamps fixed is so much more important...I'm taking this off review.

mulkieran:
Getting the time stamps fixed is so much more important than this, which was introduced to help diagnose a bug which is long since past. I'm closing this.

PR #1434, opened 2019-02-13
Update libc to the lowest version that stratisd will compile with

mulkieran:
It will require this higher version once it depends on the newer version
of devicemapper which will require nix 0.13.


mulkieran:
rebased.

PR #1435, opened 2019-02-13
Update allowed failures

mulkieran:
Remove cross-compilation task as Travis bug is not manifesting.

Add all Python-related tasks, since they use 3.4 and 3.4 no longer seems to
be available on Travis. They use 3.4 in order to install dbus-python easily.
Some other solution may be necessary.


mulkieran:
Merging, in order to get CI happy again.

PR #1438, opened 2019-02-14
Udev remove unit test

tasleson:
Automated unit test which reproduces
https://github.com/stratis-storage/stratisd/issues/1197

This will require the following changes on the CI test nodes:

* nbd kernel module is loaded
* `/usr/lib/udev/rules.d/60-block.rules` needs `|nbd*`  see: https://github.com/systemd/systemd/pull/11676
* nbd and nbdkit packages installed `dnf install nbd nbdkit nbdkit-basic-plugins`


tasleson:
Rebase

mulkieran:
Note: Behavior is what is to be expected if running on most recent devicemapper version: https://stratis-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/stratisd/2696/console, since stratisd will fail on an assertion. If running off master devicemapper, though, stratisd would not hit the assertion, since now the function returns an error in that case.

tests/client-dbus/tests/udev/remove/test_remove.py
```diff
@@ -0,0 +1,66 @@
...
+        """
+        super(UdevRemove, self).setUp()
+
+        result = flight_check()
+        if result:
+            raise unittest.SkipTest(result)
+
+        self._blk_mgr = NbdDevices()
+
+    def test_single_device_in_pool(self):
```
mulkieran: Please decorate test with ```@unittest.expectedFailure``` and run it again.

tasleson:
@mulkieran 

> The test itself also has an error, because stratisd is no longer running. Let's assume that stratisd should always be running at the end of a successful test, for now.

The test is ensuring that the daemon is running when we are told to stop, so I'm confused about this statement?

tasleson:
> Is there any way to get rid of newstyle/oldstyle nbd warnings?

Done

mulkieran:
Seems like Travis failure is just due to a formatting error.

tests/client-dbus/tests/udev/remove/test_remove.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def setUp(self):
+        """
+        Common needed things
+        """
+        super(UdevRemove, self).setUp()
+
+        result = flight_check()
+        if result:
+            raise unittest.SkipTest(result)
```
mulkieran: Try just calling ```self.skip()``` here.

tasleson: Done with https://github.com/stratis-storage/stratisd/pull/1438/commits/b9b3a738fa9b89bab27f1766e4d7f5741df35e24

tasleson:
@mulkieran @trgill Rebased, re-worked a couple things and made suggested edits.

tasleson:
@trgill @mulkieran Updated and squashed a bit.  Local testing is showing this to be very stable.

trgill:
@bgurney-rh @drckeefe @mulkieran please discuss in standup?

mulkieran:
We should consider this blocked by #1502, since none of the tests being changed are being run at all.

mulkieran:
No longer blocked by #1502.

mulkieran:
@tasleson Please rebase.

tasleson:
> @tasleson Please rebase.
@mulkieran Done

tasleson:
> Please remove requirement that STRATIS_DESTRUCTIVE_TEST be set on new remove test.

Yes, forgot I had that in there, thanks.  Added removal.

tasleson:
Re-based, conflicts resolved, commits squashed some more.

tasleson:
I remembered that some testing was done with Stratis which yielded similar test results as this one, but it requires less functionality for the unit test.  Basically you can do a less invasive remove by doing

```
echo offline > /sys/block/<devname>/device/state
```

You can restore operation by doing:
```
echo running > /sys/block/<devname>/device/state
```
Note: No udev events are created by this 

Thus it would be easier to add an existing test to the existing udev tests which creates a pool with one or more loop back devices and then does the `offline` operation on one or more of them.  Our expectation is that the stratisd process continues to run and you can query the pool(s) and get the expected number returned.  Today I believe we don't crash the service, but when we retrieve all the managed objects we get an error as dm cannot talk to the device to retrieve information needed to fill out the properties.

tasleson:
Wrote up issue to add this unit test using easier method than this: https://github.com/stratis-storage/stratisd/issues/1512

Issue #1016, opened 2018-06-27
Check that behavior is correct in all cases in BDA::read()


agrover:
from @mulkieran in #1012:
BDA::read() returns Result<(chunk1, chunk2)>. It returns an error if both fail. (That it more or less arbitrarily picks the first error is odd but not a fundamental problem).

If one succeeds but the other doesn't, one of the chunks it returns will be a buffer with data read from the device, the other will be an all-zero buffer. I believe. Remember that that all-zero buffer represents the fact that there was an error getting the raw bytes off the device.

Now, when StaticHeader::setup() gets an all zero buffer the result is Ok(None), meaning that the data was read, but no static header was on the device. So, if there was an error reading exactly one of the sections, one header will be anything, and the other will be Ok(None). This is the same value as if the region was read succesfully and was found not to have a Stratis magic number.

I think that there are six cases in the match expression in StaticHeader::setup() where Ok(None) is potentially meaning these two very different things: (1), we have no clue, because we couldn't even read the data (2), we're pretty sure this isn't a Stratis header, because there was no magic number.

Is the code simultaneously doing the right thing for both these meanings in all these six cases?

mulkieran:
I tend to think honesty is the best policy on this one. Let ```BDA::read()``` return ```(Result<chunk1>, Result<chunk2>)``` rather than merging two results into one and let ```StaticHeader::setup()``` sort that one out however it needs to. But perhaps the point will become moot.

PR #1439, opened 2019-02-21
BDA::read change function return values

tasleson:
Change signature to return
(io::Result<[u8; SECTOR_SIZE]>, io::Result<[u8; SECTOR_SIZE]>) so
that we can distinguish the difference from an IO error and a
device that simply has nothing but zeros.

Addresses: https://github.com/stratis-storage/stratisd/issues/1016

### DIFF
--- a/src/engine/strat_engine/backstore/metadata.rs
+++ b/src/engine/strat_engine/backstore/metadata.rs
@@ -46,7 +46,7 @@ impl BDA {
     /// Read the BDA from the device and return 2 SECTORS worth of data, one for each BDA returned
     /// in the order of layout on disk (location 1, location 2).
     /// Only the BDA sectors are read up from disk, zero areas are *not* read.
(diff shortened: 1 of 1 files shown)


tasleson:
@mulkieran  FYI: The stray changes occurred by cargo fmt.  I'm guessing there is differences between 1.31.0 and 1.32.0, however I find it interesting that the PR passed all checks.

mulkieran:
> @mulkieran FYI: The stray changes occurred by cargo fmt. I'm guessing there is differences between 1.31.0 and 1.32.0, however I find it interesting that the PR passed all checks.

We should discuss at some time whether we should move the recommended development compiler up to 1.32. If we do, we should prepare the CI (and the README), first.

tasleson:
@mulkieran Picked up your changes and rebased

mulkieran:
I think this should go back in review slot.

tasleson:
@mulkieran Picked up your latest changes & rebased.  Good catch on double IOError case.

PR #1440, opened 2019-02-25
Tasleson read bda signature

mulkieran:
The first three commits just get rid of stray formatting changes.

The next three are straightforward tidies.

The last just adds more precision to ```setup()``` function specification.

There are also two review comments.

### DIFF
--- a/src/engine/strat_engine/backstore/metadata.rs
+++ b/src/engine/strat_engine/backstore/metadata.rs
@@ -46,7 +46,7 @@ impl BDA {
     /// Read the BDA from the device and return 2 SECTORS worth of data, one for each BDA returned
     /// in the order of layout on disk (location 1, location 2).
     /// Only the BDA sectors are read up from disk, zero areas are *not* read.
-    fn read<F>(f: &mut F) -> io::Result<([u8; SECTOR_SIZE], [u8; SECTOR_SIZE])>
+    fn read<F>(f: &mut F) -> (io::Result<[u8; SECTOR_SIZE]>, io::Result<[u8; SECTOR_SIZE]>)
     where
         F: Read + Seek,
     {
@@ -69,13 +69,10 @@ impl BDA {
             Ok(())
         }
 
-        let loc_1_read_result = read_sector_at_offset(f, SECTOR_SIZE, &mut buf_loc_1);
-        let loc_2_read_result = read_sector_at_offset(f, 9 * SECTOR_SIZE, &mut buf_loc_2);
-
-        match (loc_1_read_result, loc_2_read_result) {
-            (Err(loc_1_err), Err(_)) => Err(loc_1_err),
-            _ => Ok((buf_loc_1, buf_loc_2)),
-        }
+        (
+            read_sector_at_offset(f, SECTOR_SIZE, &mut buf_loc_1).map(|_| buf_loc_1),
+            read_sector_at_offset(f, 9 * SECTOR_SIZE, &mut buf_loc_2).map(|_| buf_loc_2),
+        )
     }
 
     // Writes bda_buf according to the value of which.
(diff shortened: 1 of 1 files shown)


src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -267,70 +264,114 @@ impl StaticHeader {
     /// Return the latest copy that validates as a Stratis BDA, however verify both
     /// copies and if one validates but one does not, re-write the one that is incorrect.  If both
     /// copies are valid, but one is newer than the other, rewrite the older one to match.
-    /// Return None if the static header's magic does not match for *both* copies.
+    /// Return None if it's not a Stratis device.
+    /// Return an Err if the metadata seems to indicate that the device is
+    /// a Stratis device, but no well-formed signature block could be read.
```
mulkieran: This returns an error in two situations:
* If the metadata seems to indicate that the device is a Stratis deivce, but no well-formed signature could be found.
* If a well-formed signature was found but an error occurred during the repair of a second, ill-formed, old, or unreadable signature.

I think it is right to return an error in the first case, but I think the second should possibly be handled differently. I think a good option is to enrich the return type to
```StratisResult<Option<StaticHeader>, Option<some error thing if repair fails>>```.
Alternatively, it is fine to open an issue to address this later.

File: src/engine/strat_engine/backstore/metadata.rs [hunk 2]
```diff
@@ -267,70 +264,114 @@ impl StaticHeader {
...
+                            Err(loc_1)
+                        }
                     }
-                    (None, Some(loc_2)) => {
-                        // Copy 2 has valid Stratis BDA, copy 1 has no magic, re-write copy 1
-                        BDA::write(f, &buf_loc_2, MetadataLocation::First)?;
-                        Ok(Some(loc_2))
+                    (Err(_), Err(_)) => {
+                        let err_str =
+                            "Appeared to be a Stratis device, but no valid sigblock found";
```
mulkieran: Might as well enhance this error message. The contents of both buffers are available, so they might as well be included.

src/engine/strat_engine/backstore/metadata.rs
tasleson: New issue: https://github.com/stratis-storage/stratisd/issues/1443

src/engine/strat_engine/backstore/metadata.rs
tasleson: New issue: https://github.com/stratis-storage/stratisd/issues/1444

mulkieran:
Jenkins error is just the udevadm error.

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -285,10 +287,18 @@ impl StaticHeader {
                             (Some(loc_1), Some(loc_2)) => {
                                 if loc_1 == loc_2 {
                                     Ok(Some(loc_1))
+                                } else if loc_1.initialization_time == loc_2.initialization_time {
+                                    // Inexplicable disagreement among static headers
+                                    let err_str = "Appeared to be a Stratis device, but signature blocks disagree.";
+                                    Err(StratisError::Engine(ErrorEnum::Invalid, err_str.into()))
```
tasleson: IMHO this would be extremely unlikely, but if it did happen I would like to have the contents of each super blocks dumped to the logs.

src/engine/strat_engine/backstore/metadata.rs
```diff
@@ -373,7 +383,8 @@ impl StaticHeader {
             }
             (Err(_), Err(_)) => {
                 // Unable to read the device at all.
-                Ok(None)
+                let err_str = "Unable to read data at sigblock locations.";
+                Err(StratisError::Engine(ErrorEnum::Invalid, err_str.into()))
```
tasleson: Returning None was the previous behavior.

mulkieran:
Merged into #1439 .

Issue #1147, opened 2018-09-04
udevadm settle failing in tests


mulkieran:
See Jenkins record https://stratis-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/stratisd/1861/.

stdout is empty, stderr is miscellaneous punctuation.

mulkieran:
The machine wasn't loginable, so it got rebooted. We may never know what the problem was, and if it doesn't resurface again, I think we can close it.

mulkieran:
Whoops it did resurface.

mulkieran:
Some excerpts from the journal, sorry it's a graphic:

![screenshot from 2018-09-12 09-27-10](https://user-images.githubusercontent.com/1218366/45427921-159bf400-b66e-11e8-9adf-0c6caa6a4e06.png)

mulkieran:
The code fails on udevadm settle before the tests have even started to run. I think that means that stratisd hasn't been able to do anything. ```test_add_cache_devs``` which is the first test requires 4 devices, which is unusually high. ```real_test_add_cachedevs``` places no limits on the size of the devices, though, so only normal, i.e., no dm devices are given to the test. Moreover, no set up of the devices is done until after the ```udevadm settle``` call is made. In fact, nothing is done, but reading from a file of designated test devices and calculaton.

mulkieran:
Looked at first interesting previously run sequence: https://stratis-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/stratisd/1883/console.

It turns out that ```test_empty_pool``` failed, but a few subsequent succeeded. This is wierd. According to my code inspection an error should be returned by ```ThinPool::new``` because it failed to allocate the necessary amount from the backstore, which has zero devices. Holding this though, in case it shows up again, and rebooting this machine.

mulkieran:
It hasn't done this in a very long time, best to move on.

PR #1441, opened 2019-02-25
cmd: Include exit code/reason on failure

tasleson:
When looking at some logs of udevadm settle failing, there is no
stdout or stderr.  It would be good to know what the exit code is
to see if we can understand this failure better.

### DIFF
--- a/src/engine/strat_engine/cmd.rs
+++ b/src/engine/strat_engine/cmd.rs
@@ -83,11 +83,17 @@ fn execute_cmd(cmd: &mut Command) -> StratisResult<()> {
             if result.status.success() {
                 Ok(())
             } else {
+                let exit_reason = result
+                    .status
+                    .code()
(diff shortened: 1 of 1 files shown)


src/engine/strat_engine/cmd.rs
```diff
@@ -83,11 +83,17 @@ fn execute_cmd(cmd: &mut Command) -> StratisResult<()> {
             if result.status.success() {
                 Ok(())
             } else {
+                let exit_reason = result
+                    .status
+                    .code()
+                    .map_or(String::from("process terminated by signal"), |ec| {
+                        format!("{}", ec)
```
mulkieran: ```format!("{}", ec)``` -> ```ec.to_string()```

tasleson: Done, thanks for the suggestion.

mulkieran:
This is related to https://github.com/stratis-storage/stratisd/issues/1147.

PR #1451, opened 2019-02-28
Ignore everything starting with proptest-regression

mulkieran:
The name of the file or directory to ignore used to be proptest-regression,
now the directory we want to ignore is certainly called proptest-regressions.
Ignore it.


mulkieran:
rebased for tidyness.

Issue #1356, opened 2018-12-06
Let's split Pool state into two states (thinpool state and overall)


mulkieran:
overall state would just be ```Initializing```, ```Stopping```, and ```Running```.
thin pool state would just mirror the thinpool summary more or less, ```Fail```, ```ro```, ```rw``` (that's the good state), ```out of data space```.

I feel that this would make things easier for the user to understand and easier for the developer to implement.

Related: #1351.

agrover:
For this one and for #1351, this is making me think more and more we should have the overall api-exposed state that is pretty high-level, and then we should feel free to use multiple state enums internally, but only make them visible to the user in how they affect the api-exposed state. The existence of a thinpool within the pool is an implementation detail. WDYT?

mulkieran:
Not having a direct mapping between the internally recorded state and the state exposed on the D-Bus API is fine w/ me.

But we should be very careful what we expose on the D-Bus API. Right now, we do expose the "pool state" which is _very_ closely tied to the thin pool status. So, that we use a thin pool is currently exposed on the D-Bus API even if it shouldn't be.

mulkieran:
A summary of states in Stratis and statuses available from devicemapper.

[states.md.txt](https://github.com/stratis-storage/stratisd/files/2691317/states.md.txt)

mulkieran:
I think that the suggestion of making the states that we expose on the D-Bus API and the states that we use internally distinct naturally leads to the second suggestion which is to make the internal states as orthogonal as possible. That is what I am proposing in the specific case of the pool state in this issue.

I have drawn up a document showing the available devicemapper states and the existing stratisd pool states, some of which are also API states. I think it would be valuable, in the course of this effort, to draw up a document that also maps internal states or combinations of states to D-Bus exposed states.
This would probably require a table for every mapping.

mulkieran:
I'm closing this as obsolete. We no longer export these on the D-Bus and #2106 and #2103 are addressing the internal implementation.

Issue #1267, opened 2018-10-09
Pool::total_physical_used is on its way to becoming unhelpful


mulkieran:
When we add another layer, say VDO, we may find ourselves in the disconcerting situation where total_physical_used is greater than total physical size the way we calculate it.

I think we would be better off going for a stabler and more accurate if not as precise value and simply returning the result of ```DataTier::allocated()``` for this value. This can overestimate, because some of the segments allocated may not actually be being used. But it can't actually exceed the total physical size, by definition.

Before there was lazy allocation from the data tier this suggestion wouldn't have made any sense. But now that there is lazy allocation, I think it is really the best suggestion.

I also think that these two values should be supplemented with another pair of values, which would be specific to the thin pool and not to the backstore. Again, with the VDO example, we can imagine a total thin pool size which is very large, and a value for used which is a lot less than the thin pool size, but still a lot larger than the total physical size.

mulkieran:
I started to think about this problem while trying to figure out how to test #1266 .

mulkieran:
Removing from project, in favor of parent issue.

jbaublitz:
This will be resolved by #1662

mulkieran:
We can close this. We decided that we were in a position to maintain this value accurately, so we will do that. The alternative suggestions are ones we will support naturally as the design and implementation proceeds.

Issue #1243, opened 2018-10-02
Last chance to change units for size properties


mulkieran:
Right now they are all in Sectors or Bytes. Should they all be Bytes...or are we certain that the ones that are Sectors will always require or have only that amount of precision?

Recall that the CLI can _display_ the values however it wants.

Ones of interest currently exposed:
* BlockDev
  - physical size (Sectors)
* Filesystem
   - used (Bytes)
* Pool
   - total physical used (Sectors)
   - total physical size (Sectors)

agrover:
I'm fine with this. If anything, I'd change filesystem.used to sectors, for consistency.

jbaublitz:
@mulkieran Is your concern with `Sectors` that we could theoretically have different sector sizes at some point? If that's the concern, I'm on board with keeping that open as a possibility and changing to `Bytes`.

If not, what is the concern with precision? I don't think it's possible to run into an issue with anything more granular than sectors if we only allow sectors of the same size simply because you can't really have part of a sector to my knowledge.

mulkieran:
Ok, we decided in standup that all sizes will be in bytes now (and that work can be added onto existing PRs for https://github.com/stratis-storage/projec.t/issues/52.

jbaublitz:
@mulkieran I've played around with this a bit, and my recommendation is that we use sectors internally and convert to bytes right before sending these properties over the DBus. I think this makes the most sense as number of sectors is an important construct in stratisd and this would potentially break some things if we change all of the `.size()` methods to move away from returning `Sectors`. If we only change it at the top level of the `Engine` trait, some of the internal parts of stratisd use some of the methods that calculate size in hard to predict ways. I think leaving `Filesystem::used()` as `Bytes` makes sense from a contextual perspective, and we'll simply convert all of the sector return values from `.size()` calls to bytes right before sending them to the user. How does that sound?

mulkieran:
I think that is the most reasonable thing to do at this time. We can adjust the internal representation if and when we need to. For right now, just making sure that all sizes sent across the D-Bus by the daemon represent bytes is what we're after.

jbaublitz:
Hm, we seem to have a problem. When we convert to bytes using the `.bytes()` method in devicemapper for default pool size, multiplication by `SECTOR_SIZE` overflows the `u64` and stratisd panics. How do we want to address this?

jbaublitz:
[...]

Issue #1215, opened 2018-09-24
Right now the main method can crash if check() methods fail


mulkieran:
The check() methods need a good review, to decide how they should actually be used and how errors should be handled at the high level.

mulkieran:
Not changed w/ multi-threading.

mulkieran:
I believe this has been addressed by more recent changes to our multi-threading implementation.

PR #1461, opened 2019-03-07
Add changes necessary for addition of Error constructor to status types

mulkieran:


mulkieran:
This is labeled do not merge, because it will not compile against most recent release of devicemapper.

mulkieran:
Note that this PR in a sense compounds the following existing problems:
* https://github.com/stratis-storage/stratisd/issues/1356. I do not choose to add an additional state corresponding with the Error return, because the Error return doesn't mean any state for the pool. But our choice of pool states is already very dubious.
* https://github.com/stratis-storage/stratisd/issues/1267 I added more lines to ```total_physical_used```. It continues to produce a potentially bogus value, regardless of these new lines, obviously.
* https://github.com/stratis-storage/stratisd/issues/1255 It doesn't improve the logging to give the log reader an accurate idea of the state of the pool/thinpool.
* https://github.com/stratis-storage/stratisd/issues/1243 It does not change the value of the units returned from any size methods (and then exported on the D-Bus). See also #1218.
* https://github.com/stratis-storage/stratisd/issues/1215 It adds another of many allready existing possible error returns to the check method.
* https://github.com/stratis-storage/stratisd/issues/1148 It does not make D-Bus properties that could fail safe. It could even be said to make them slightly more likely to fail, by adding another possible source of an error return.
* https://github.com/stratis-storage/stratisd/issues/1148 It doesn't fix log messages that currently aren't very useful, although it tries to make newly introduced log messages somewhat helpful.
* https://github.com/stratis-storage/stratisd/issues/512 I didn't try to improve the FilesystemStatus enum in any way. Luckily, it is not yet exported on the D-Bus.

mulkieran:
Whoops! Trivial clippy error.

mulkieran:
Jenkins failure was udevadm problem.

mulkieran:
udevadm again. Reboot fedora-server-1!

mulkieran:
Trying again.

mulkieran:
Now trying fedora-server-2.

mulkieran:
fedora-server-3 is also failing on something...but not everything, so maybe it's not udevadm problem.

mulkieran:
rebased.

mulkieran:
Also had to update Cargo.lock.

Cargo.toml
```diff
@@ -7,7 +7,7 @@ authors = ["Stratis Developers <stratis-devel@lists.fedorahosted.com>"]
 dbus = {version = "0.6.4", optional = true}
 clap = "2"
 nix = "0.13"
-devicemapper = "0.26.0"
+devicemapper = {git="https://github.com/mulkieran/devicemapper-rs", branch="master-error"}
```
trgill: don't forget to remove this

mulkieran:
rebased.

Cargo.toml
tasleson: New version of devicemapper-rs is available, so you can bump the version for this PR.

mulkieran:
Now pulling in devicemapper version 0.27.0, since release is made.

mulkieran:
I've removed the "do not merge" flag, since there will be no conflict if this is merged.

src/engine/strat_engine/thinpool/filesystem.rs
```diff
@@ -220,6 +220,13 @@ impl StratFilesystem {
                 // TODO: do anything when filesystem is not mounted?
                 // TODO: periodically kick off fstrim?
             }
+            ThinStatus::Error => {
+                let error_msg = format!(
+                    "Unable to get status for filesystem thin device {}",
+                    self.thin_dev.device()
+                );
```
tasleson: Would it be useful to include the fs name in this error message?  Also asking for all the other error messages that are similar in PR.

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -545,6 +545,14 @@ impl ThinPool {
                 self.resume()?;
                 self.set_extend_state(data_extend_failed, meta_extend_failed);
             }
+            ThinPoolStatus::Error => {
+                // Do not set the state, because it is unknown.
+                warn!(
+                    "Devicemapper could not obtain the status for devicemapper thinpool device {} belonging to pool with UUID {}",
+                    self.thin_pool.device(),
+                    pool_uuid
+                );
```
tasleson: I believe adding the pool name would be useful for this message too.

src/engine/strat_engine/thinpool/filesystem.rs
mulkieran: Filesystem name is not available here.

src/engine/strat_engine/thinpool/thinpool.rs
mulkieran: Pool name is not available here.

PR #1467, opened 2019-03-18
Try out 1.32

mulkieran:
Just seeing.


mulkieran:
This _seems_ to go just fine. Need to do a test with 1.32.

Issue #1475, opened 2019-03-29
Make sure that RUST_BACKTRACE is set to 1 in systemd configuration file


mulkieran:
This will ensure that if stratisd panics, as on a panic, or a failed expect or assert, that the backtrace is displayed. Otherwise, the display will be terse and very uninformative.

mulkieran:
I'll give this a go.

mulkieran:
Closed by #1479.

PR #1479, opened 2019-04-09
Set RUST_BACKTRACE to 1 in systemd service file

mulkieran:
closes: #1475.


mulkieran:
This is basically impossible to test directly, but stratisd ought _never_ to panic. But all the docs I've found seem to agree that this is the correct change.

Issue #1474, opened 2019-03-27
MDV not properly updated when filesystem is auto-extended


ingothierack:
Strange behavior now here.

First Problem with autoextending the FS from 1TB to 2 TB see #1456.
Restart of stratisd daemon after fs was full extend it to 2 TB(no autoextending).
Do fill up the fs to 1.3 TB, works till now.

Reboot the server, it fails now on mount this filesystem with invalid superblock. 

If it check the fs, it is set to only 1 TB size

from lsblk
stratis-1-210eb8dac6cb440993454dac6a885dae-thin-fs-f4a9283f89c3457bb5980d94eb82cc34 253:37   0     1T  0 stratis 

output stratis fs list

stratis_pool01  fs_share01       1.22 TiB     Mar 15 2019 14:41  /stratis/stratis_pool01/fs_share01       f4a9283f89c3457bb5980d94eb82cc34  

from logfile stratisd start output
```
Mar 27 21:42:03 nas001 stratisd[11431]:                        "fs_share01",
Mar 27 21:42:03 nas001 stratisd[11431]:                        SimpleRef(
Mar 27 21:42:03 nas001 stratisd[11431]:                            f4a9283f-89c3-457b-b598-0d94eb82cc34
Mar 27 21:42:03 nas001 stratisd[11431]:                        )
Mar 27 21:42:03 nas001 stratisd[11431]:                    ): StratFilesystem {
Mar 27 21:42:03 nas001 stratisd[11431]:                        thin_dev: ThinDev {
Mar 27 21:42:03 nas001 stratisd[11431]:                            dev_info: DeviceInfo {
Mar 27 21:42:03 nas001 stratisd[11431]:                                hdr: Struct_dm_ioctl {
Mar 27 21:42:03 nas001 stratisd[11431]:                                    version: [
Mar 27 21:42:03 nas001 stratisd[11431]:                                        4,
Mar 27 21:42:03 nas001 stratisd[11431]:                                        39,
Mar 27 21:42:03 nas001 stratisd[11431]:                                        0
Mar 27 21:42:03 nas001 stratisd[11431]:                                    ],
Mar 27 21:42:03 nas001 stratisd[11431]:                                    data_size: 305,
Mar 27 21:42:03 nas001 stratisd[11431]:                                    data_start: 312,
Mar 27 21:42:03 nas001 stratisd[11431]:                                    target_count: 1,
[...]

PR #1480, opened 2019-04-15
If we extend the DM table return an indicator to save the fs metadata (final)

trgill:
If we extend the DM table return a bool to indicate the MDV needs update
    
When the thin device is extended and XFS is grown to fill additional
storage, the MDV must be saved so when the device is reconstructed it is
the expected size.

Update test to make sure rebuild of the filesystem has extended size.

closes: #1474.


mulkieran:
Again experiencing an error in Travis cross-compilation target, should be fixed soon: https://github.com/stratis-storage/stratisd/pull/1484.

mulkieran:
@trgill Please rebase once more (to pull in update to Travis CI).

PR #1483, opened 2019-04-15
Use O_SYNC for BDA open

tasleson:
We really want the BDA writes to make it all the way to the
hardware.  The O_SYNC flag should ensure this.


tasleson:
May address: https://github.com/stratis-storage/stratisd/issues/1446

tasleson:
I was able to re-create https://github.com/stratis-storage/stratisd/issues/1446 with this change applied.  Thus I can conclude this change is not helping with correction of the missing super block signatures.

PR #1488, opened 2019-04-23
Make error-chain a dev dependency

mulkieran:
It has never been anything else. It was included as a regular dependency
in error.


mulkieran:
Discreases size by 1/1000th %.

PR #1491, opened 2019-04-24
Downgraded 'Data tier percent used' message from info to debug

waltdisgrace:
The message 'Data tier percent used' was being displayed to the user at the
info level, which is unnecessary for users who don't use up a significant
amount of space.

In this PR, I downgraded the original message to the debug level
and added an info level message which displays the data tier percent used
only when the said percent has reached a critical level.

### DIFF
--- a/src/engine/strat_engine/thinpool/thinpool.rs
+++ b/src/engine/strat_engine/thinpool/thinpool.rs
@@ -633,7 +633,13 @@ impl ThinPool {
         }
 
         let overall_used_pct = used_pct(*used, *used + *available);
-        info!("Data tier percent used: {}", overall_used_pct);
+        debug!("Data tier percent used: {}", overall_used_pct);
+        if overall_used_pct >= 90 {
+            info!(
+                "Percent used has reached a critical threshold level; data tier percent used: {}",
+                overall_used_pct
+            );
+        }
 
         let new_state = if overall_used_pct < SPACE_WARN_PCT {
             FreeSpaceState::Good
--- a/src/engine/strat_engine/thinpool/thinpool.rs
+++ b/src/engine/strat_engine/thinpool/thinpool.rs
@@ -633,7 +633,10 @@ impl ThinPool {
         }
 
         let overall_used_pct = used_pct(*used, *used + *available);
-        debug!("Data tier percent used: {}", overall_used_pct);
+        debug!(
+            "Data tier percent used; thinpool device {} percent used: {}",
+            overall_used_pct, self.thin_pool.device()
+        );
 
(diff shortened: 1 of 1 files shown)


mulkieran:
add to whitelist

waltdisgrace:
I will fix the formatting tomorrow morning.

mulkieran:
@waltdisgrace If you look at the Makefile in the project, you'll see a Make target, "fmt".

> make fmt

should run and auto-format everything for you. If it doesn't, you must need to install rustfmt.

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -633,7 +633,13 @@ impl ThinPool {
         }
 
         let overall_used_pct = used_pct(*used, *used + *available);
-        info!("Data tier percent used: {}", overall_used_pct);
+        debug!("Data tier percent used: {}", overall_used_pct);
+        if overall_used_pct >= 90 {
```
tasleson: There is a constant you can use here instead of a magic number `const SPACE_WARN_PCT: u8 = 90;`.  Also according to the constants themselves, critical is defined as >= 95% `SPACE_CRIT_PCT` so I'm wondering if the info statement should be slightly reworded.

tasleson: Also instead of introducing a new conditional check, we could insert the info message into the existing conditional assignment, something to consider.

```rust
        let new_state = if overall_used_pct < SPACE_WARN_PCT {
            FreeSpaceState::Good
        } else if overall_used_pct < SPACE_CRIT_PCT {
            FreeSpaceState::Warn
        } else {
             info!(
                "Percent used has reached a critical threshold level; data tier percent used: {}",
                overall_used_pct
            );
            FreeSpaceState::Crit
        };
```

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -633,7 +633,13 @@ impl ThinPool {
         }
 
         let overall_used_pct = used_pct(*used, *used + *available);
-        info!("Data tier percent used: {}", overall_used_pct);
+        debug!("Data tier percent used: {}", overall_used_pct);
```
mulkieran: @waltdisgrace Please focus on giving this message more useful information. There are some models in the file that should help. I think that the other messages, which are legitimate info messages, contain a lot more information. The issue here is that stratisd can have multiple pools at a given time. This message is too bare, if it appears in the logs the person scrutinizing the logs will have to guess what pool the particular data tier belongs to.

mulkieran:
Travis failure is just a format error.

mulkieran:
Superceded by #1588.

PR #1498, opened 2019-04-29
remove hs-dbus-signature from stratisd CI test

bgurney-rh:
The "hs-dbus-signature" repository is a Hypothesis strategy for
testing other Python libraries, so it doesn't need to be cloned
as part of the stratisd CI test.

Signed-off-by: Bryan Gurney <bgurney@redhat.com>


mulkieran:
add to whitelist

PR #1510, opened 2019-05-06
test_udev.py: Correct pool create

tasleson:
When we dynamically create loop back devices we need to ensure that
the devices exist in the udev db before we issue a dbus call to
create the pool.  Otherwise stratisd will be unable to find the
block device and return an error on the pool create.

This correction adds settles after the loop back devices are
created and also adds retries when creating the stratis pool.


tasleson:
Resolves: https://github.com/stratis-storage/stratisd/issues/1446

PR #1515, opened 2019-05-07
Remove setting default TARGET from Makefile

mulkieran:
Also do not set TARGET in travis config file where it is ignored by the make task.

The basic idea is to let _cargo_ decide what the target architecture should be. If that is set up, then the CI script does not need to accept the _Makefile_'s target architecture, but can instead accept cargo's default. This is simpler, and also more general.

This also means that in stratisd.sh there is only one ```TARGET``` variable, the one that specifies what kind of test to run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,12 +25,12 @@ matrix:
         - rust: 1.33.0
           before_script:
               - rustup component add rustfmt
-          env: TASK=fmt-travis TARGET=x86_64-unknown-linux-gnu
+          env: TASK=fmt-travis
         # clippy
         - rust: 1.33.0
           before_script:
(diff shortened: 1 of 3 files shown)


mulkieran:
@bgurney-rh, @drckeefe If you accept these changes here, I'll put together the equivalent for the CI repo. We'll have to coordinate a bit about these changes, since we'll require changes to the ci repo and here.

bgurney-rh:
The stratisd test failed because there's a change necessary in the ci repo.

mulkieran:
rebased.

bgurney-rh:
The stratisd test succeeds after the ci pull request was merged.

PR #1516, opened 2019-05-09
Remove a useless check when reading metadata

mulkieran:
The check does some harm and no good.

* It diminishes encapsulation. The parameter, ```per_region_size``` is passed to ```MDAHeader::from_buf``` solely to enable the check, even though it is not a necessary value for any action that method has to take. If the check is eliminated, then the parameter can be as well.
* It forces test code to synthesize some acceptable value for ```per_region_size``` for tests when testing other properties of the method.
* It causes an additional test, to check that an error is actually returned in the method if the value for the parameter ```per_region_size``` is too small, to be run.
* It makes it possible for ```from_buf``` to fail when parsing the MDAHeader, even when the header is parsed correctly, the checksums match, and the metadata versions are as expected, which makes no real sense. Generally speaking, methods should only return an error when they can't perform the action requested.
* If it were ever to fail, it would assign the fault to the ```used``` value when the failure would be just as likely to be caused by a logic error somewhere else in the program, and possibly occurring in some other code entirely and at a different point in time. The proper way to make sure that the metadata is consistent is not to insert a check at the point where the data is read, but to define an invariant function on the metadata which can be run when suitable, i.e., once the whole BDA has been loaded. This PR does not take on that task, but that would be the correct approach.
* Since it causes the check to be run in two places, both when writing the variable length metadata and when reading the MDAHeader, the check is abstracted into a method. This makes the code that calls the method less clear. If the very simple code is inlined instead, the check in the method which saves the variable length metadata is much clearer and more readable.
* It should never fail, because the code that writes the variable length metadata verifies that it is not too large _before_ performing the write.

MDAHeader should also be made more consistent, so that either it is responsible for both reading and writing the variable length metadata or it is responsible for neither. Currently it is responsible for reading, but not for writing. This PR does not address that problem, either.

### DIFF
--- a/src/engine/strat_engine/backstore/metadata.rs
+++ b/src/engine/strat_engine/backstore/metadata.rs
@@ -582,7 +582,7 @@ mod mda {
                     per_region_size,
                 )))?;
                 f.read_exact(&mut hdr_buf)?;
-                Ok(MDAHeader::from_buf(&hdr_buf, per_region_size)?)
+                Ok(MDAHeader::from_buf(&hdr_buf)?)
             };
 
             // Get an MDAHeader for the given index.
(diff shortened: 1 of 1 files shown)


mulkieran:
rebased.

PR #1527, opened 2019-05-15
Do not use MIN_MDA_SECTORS in inappropriate contexts

mulkieran:


mulkieran:
Whoops, needs more work! The tests should not be wiping in either place.

mulkieran:
Superceded by #1529.

PR #1529, opened 2019-05-16
Remove some instances of constant misuse

mulkieran:
This PR addresses two places where the constant MIN_MDA_SECTORS was being used in tests not for testing MDA code or anything related to it, but because it must have seemed to be roughly the correct size. In both cases it was used as an argument to wipe_sectors, and in both cases, the starting offset was 0; the MDA itself does not start from zero, which was a hint that something might be wrong. It turns out that in both cases the call to wipe_sectors is unnecessary and the tests can be improved by removing the call entirely.


mulkieran:
Supercedes #1527.

PR #1526, opened 2019-05-15
Master region size

mulkieran:

PR #1534, opened 2019-05-20
Use types to distinguish between sizes that represent different parts of the MDA

mulkieran:
This PR accomplishes the following:

* It makes the MDA size always valid, eliminating the need for the validate_mda_size method. It does this by changing the parameter to higher level methods to represent the required size for the variable length metadata, which is more relevant to these methods, and then _calculating_ a valid MDA size from that value in some code belonging to the metadata module.
* It makes a class of bugs virtually impossible by using types to distinguish between the size of the MDA, the size of an MDA region, and the size of the region that contains the variable length metadata. Bugs like these:
   * https://github.com/stratis-storage/stratisd/pull/1524
   * https://github.com/stratis-storage/stratisd/pull/1530
* It enforces better encapsulation by not exposing internal details of the metadata externally, thereby avoiding misuse of MDA specific constants in other contexts: https://github.com/stratis-storage/stratisd/pull/1529.

It does not address the possible problem that the variable length metadata could be too large for its device. A device has minimum size of 1 GiB, so the variable length metadata would have to be huge to overrun this limit.

### DIFF
--- a/src/engine/strat_engine/backstore/backstore.rs
+++ b/src/engine/strat_engine/backstore/backstore.rs
@@ -21,8 +21,9 @@ use crate::{
                 blockdevmgr::{map_to_dm, BlockDevMgr},
                 cache_tier::CacheTier,
                 data_tier::DataTier,
+                metadata::MDADataSize,
                 setup::get_blockdevs,
-                StratBlockDev, MIN_MDA_SECTORS,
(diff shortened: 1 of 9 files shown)


mulkieran:
Supercedes #1526, code is identical.

PR #1399, opened 2019-01-10
Abstract function for looking up segment (2nd try)

mulkieran:
Resolves: #1136.

PR #1535, opened 2019-05-21
Get rid of obsolete cleanup module

mulkieran:


mulkieran:
This is cherry picked from #1399, the rest of which will be coming up soon, but it's really orthogonal, so I pulled it out.

PR #1536, opened 2019-05-21
Eliminate duplicate closures

mulkieran:
Resolves: #1136.

This eliminates an instance of code duplication which was caused by an apparently insurmountable restriction on the definition of closures in Rust. It turns out that with enough ingenuity, the restriction on closures can be avoided, but it seemed simpler to drop the function return type from a single method, in order to simplify the use of its result. Since the bodies of the two closures are no longer duplicated, they need a single place to be put, so this adds a new module called shared. Another method, which is shared between the cache_tier and data_tier modules is also placed there.


mulkieran:
Supercedes #1399.

Issue #1148, opened 2018-09-04
D-Bus properties that can fail should be eliminated


mulkieran:
Expensive D-Bus properties are properties that are computationally expensive or, even more important, require interacting with the environment. Cheap properties are ones that are obtained with a limited amount of computation and do not require any interaction with the environment.

Definitely cheap properties:
* Manager.Version

* Blockdev.Devnode
* Blockdev.HardwareInfo
* Blockdev.UserInfo
* Blockdev.InitializationTime
* Blockdev.State
* Blockdev.Pool
* Blockdev.Uuid
* Blockdev.Tier

* Filesystem.Devnode
* Filesystem.Name
* Filesystem.Pool
* Filesystem.Uuid
* Filesystem.Created

* Pool.Name
* Pool.Uuid

In-between Properties:
* BlockDev.TotalPhysicalSize - this may be an artifact of how this data is held now, but currently the computation _is_ cheap

Expensive Properties:
* Filesystem.Used - ~~May require mounting the filesystem, also statting~~. EDIT: Since #1149 requires DM status call.
* Pool.TotalPhysicalSize - Somewhat computationally expensive, as it is neccessary to iterate through all blockdevs belonging to the pool
* Pool.TotalPhysicalUsed - DM call to get status

See also #1267. Pool.TotalPhysicalUsed will be less and less meaningful as more layers are added.

mulkieran:
The above should be an exhaustive list of currently published properties.

I propose that Blockdev.TotalPhysicalSize, Filesystem.Used, Pool.TotalPhysicalSize, Pool.TotalPhysicalUsed all be removed as properties and methods that obtain these values be substituted instead. The principal motivation is to avoid the problem of making invoking ```GetManagedObjects``` prohibitively expensive for large systems. Since almost every method in the CLI relies on a ```GetManagedObjects``` call, this is an important consideration.

Moreover, the four properties that I'm suggesting removing are not, unlike many of the cheap properties, useful for identifying particular objects of interest. It is very likely that a query for identifying Blockdev based on their pool would be constructed. But a situation where a Blockdev is identified by its total physical size is much less likely to arise programmatically. The only use that the CLI currently has for any of these expensive properties is to display them. This could be achieved quite easily with method calls.

[...]

PR #1539, opened 2019-05-22
No longer supply properties on the D-Bus by means of external calls

mulkieran:
This PR's goal is to eliminate all possibility of failure in methods that are used to supply property values on the D-Bus. Currently, the consequence of a failure in obtaining one of these properties is a message with the form:

```no total physical size computed for pool with uuid <some UUID>```

when most or all of the stratis CLI commands that require interaction with stratisd are run. None of these CLI commands can succeed when this happens. One of these methods supplies its value via a devicemapper ioctl, another by means of a stat call.

The approach taken here is to cache the values used at the time when these methods are necessarily called, which occurs in the corresponding ```check``` methods for each module. The reason this seems like a good idea is that the view that the user gets of the stratisd system is consistent with the view that stratisd has of the system.

However, there is currently a defect. The view that stratisd has based on the information it got in the ```check``` methods is the last information that could be obtained. Perhaps subsequently, the ```check``` method was called, but the result returned was ```Fail``` or ```Error```. In that case, the information that is placed on the D-Bus and is cached in the objects, is not just slightly stale as is inevitable w/ a cached value, but potentially totally wrong.

Resolves: #1148.

### DIFF
--- a/src/dbus_api/filesystem.rs
+++ b/src/dbus_api/filesystem.rs
@@ -232,9 +232,5 @@ fn get_filesystem_used(
     i: &mut IterAppend,
     p: &PropInfo<MTFn<TData>, TData>,
 ) -> Result<(), MethodErr> {
-    get_filesystem_property(i, p, |(_, _, fs)| {
-        fs.used()
-            .map(|v| (*v).to_string())
-            .map_err(|_| MethodErr::failed(&"fs used() engine call failed".to_owned()))
-    })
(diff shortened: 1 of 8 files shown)


mulkieran:
Please see #1211 for prior discussion.

mulkieran:
Ok, thanks. I'm going to examine the problem and see if the total staleness problem can be mitigated. Also, verify that signals are being set where appropriate.

mulkieran:
rebased.

mulkieran:
Closing as probably obsolete.

PR #1540, opened 2019-05-23
No longer supply two properties that require real-time environment ac…

mulkieran:
…cess

This solves two problems:
1. The environment access can fail, resulting in a failure of
GetManagedObjects and hence a failure of any client action that requires
GetManagedObjects, which is most of them.
2. Pool.TotalPhysicalUsed will stop being meaningful in the general case
as other layers are added.


mulkieran:
Should fail some D-Bus tests and some tests in cli-with-stratisd.

mulkieran:
The fact that no tests in cli-with-stratisd failed is worth investigating. We stop supplying two properties, and at least some of them should be used by the CLI tests. Maybe we're skipping a good one.

mulkieran:
rebased and fixed up tests.

mulkieran:
rebased.

mulkieran:
Also remove properties from interface files.

mulkieran:
rebased.

mulkieran:
rebased.

mulkieran:
rebased.

mulkieran:
@bgurney-rh, @drckeefe: I believe this is complete in the sense that it does remove all the dubious properties from the D-Bus. But if you can identify more dubious properties, please let me know :)

We should discuss the effect on the CLI, etc.

mulkieran:
The next step would be to harden blockdev's ```TotalPhysicalSize``` with types, similarly to the metadata changes, since at present the plan is to keep it in.

mulkieran:
rebased, no longer exporting blockdev size either.

bgurney-rh:
(I'm not sure why, but the stratisd and cli-with-stratisd checks didn't trigger properly after the last rebase.  I manually re-ran those two tests, and they both pass.)

mulkieran:
rebased. conflicts resolved.

mulkieran:
rebased.

mulkieran:
rebased, conflicts resolved.

mulkieran:
rebased.

mulkieran:
Superceded by #1679.

PR #1545, opened 2019-05-29
Make BDA_STATIC_HDR_SIZE constant private

### LINKED ISSUE #1543: BDA_STATIC_HDR_SIZE should not be used in tests in mda module
The offset of the MDA from the start of the device is always abstracted in the source. The tests should be abstracted, too, and use any valid offset. In this way, BDA_STATIC_HDR_SIZE can be made private to the bda module, created in #1541.

GuillaumeGomez:
Fixes #1543.


mulkieran:
ok to test

PR #1546, opened 2019-05-30
Create specific type for MDAHeader's used field

### LINKED ISSUE #1544: MDAHeader's used field should probably have a special type indicating its semantics
It is the actual size of the variable length metadata written (as opposed to the space allocated for that metadata, which has type MDADataSize). Obviously, it should always be less than that value. A type name like MetaDataSize might be about right.

GuillaumeGomez:
Fixes #1544.

### DIFF
--- a/src/engine/strat_engine/backstore/metadata/mda.rs
+++ b/src/engine/strat_engine/backstore/metadata/mda.rs
@@ -287,11 +287,11 @@ impl MDARegions {
                 used, max_available
             );
             return Err(StratisError::Engine(ErrorEnum::Invalid, err_msg));
-        };
+        }
 
         let header = MDAHeader {
             last_updated: *time,
-            used,
+            used: MetaDataSize::new(used),
             data_crc: crc32::checksum_castagnoli(data),
         };
         let hdr_buf = header.to_buf();
@@ -389,8 +389,24 @@ impl MDARegions {
         for mda in self.mda_headers.iter() {
             assert!(mda
                 .as_ref()
-                .map_or_else(|| true, |mda| mda.used != Bytes(0)));
+                .map_or_else(|| true, |mda| mda.used.bytes() != Bytes(0)));
         }
+    }
+}
+
+/// A type representing the actual size of variable length metadata written within this metadata
+/// region. This amount must never be greater than the size of the region allocated for the variable
+/// length metadata, which has `MDADataSize` type.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct MetaDataSize(Bytes);
+
+impl MetaDataSize {
+    pub fn new(used: Bytes) -> MetaDataSize {
+        MetaDataSize(used)
+    }
+
+    pub fn bytes(self) -> Bytes {
+        self.0
     }
 }
 
@@ -398,8 +414,9 @@ impl MDARegions {
 pub struct MDAHeader {
     last_updated: DateTime<Utc>,
 
-    /// Size of region used for pool metadata.
-    used: Bytes,
+    /// Size of the variable length metadata last written in the corresponding MDA region.
+    /// If no variable length has been written, the size is 0 bytes.
+    used: MetaDataSize,
 
     data_crc: u32,
 }
@@ -412,7 +429,7 @@ impl Default for MDAHeader {
     fn default() -> MDAHeader {
         MDAHeader {
             last_updated: Utc.timestamp(0, 0),
-            used: Bytes(0),
+            used: MetaDataSize::new(Bytes(0)),
             data_crc: 0,
         }
     }
@@ -433,7 +450,7 @@ impl MDAHeader {
                 assert!(secs <= std::i64::MAX as u64);
 
                 Some(MDAHeader {
-                    used: Bytes(used),
+                    used: MetaDataSize::new(Bytes(used)),
                     last_updated: Utc.timestamp(secs as i64, LittleEndian::read_u32(&buf[24..28])),
                     data_crc: LittleEndian::read_u32(&buf[4..8]),
                 })
@@ -481,7 +498,7 @@ impl MDAHeader {
         let mut buf = [0u8; mda_size::_MDA_REGION_HDR_SIZE];
 
         LittleEndian::write_u32(&mut buf[4..8], self.data_crc);
-        LittleEndian::write_u64(&mut buf[8..16], *self.used as u64);
+        LittleEndian::write_u64(&mut buf[8..16], *self.used.bytes() as u64);
         LittleEndian::write_u64(&mut buf[16..24], self.last_updated.timestamp() as u64);
         LittleEndian::write_u32(&mut buf[24..28], self.last_updated.timestamp_subsec_nanos());
         buf[28] = STRAT_REGION_HDR_VERSION;
@@ -509,8 +526,8 @@ impl MDAHeader {
         // where usize is u64, which is usual. It is not absurd when
         // compiled in an environment where usize is u32.
         #![allow(clippy::absurd_extreme_comparisons)]
-        assert!(*self.used <= std::usize::MAX as u64);
-        let mut data_buf = vec![0u8; *self.used as usize];
+        assert!(*self.used.bytes() as u64 <= std::usize::MAX as u64);
+        let mut data_buf = vec![0u8; *self.used.bytes() as usize];
 
         f.read_exact(&mut data_buf)?;
 
@@ -583,7 +600,7 @@ mod tests {
 
             let header = MDAHeader {
                 last_updated: Utc.timestamp(sec, nsec),
-                used: Bytes(data.len() as u64),
+                used: MetaDataSize::new(Bytes(data.len() as u64)),
                 data_crc: crc32::checksum_castagnoli(&data),
             };
             let buf = header.to_buf();
@@ -602,9 +619,10 @@ mod tests {
     #[test]
     fn test_from_buf_crc_error() {
         let data = [0u8; 3];
+
         let header = MDAHeader {
             last_updated: Utc::now(),
-            used: Bytes(data.len() as u64),
+            used: MetaDataSize::new(Bytes(data.len() as u64)),
             data_crc: crc32::checksum_castagnoli(&data),
         };
         let mut buf = header.to_buf();


GuillaumeGomez:
I took a look at #1534 but it didn't seem to match exactly what we wanted to do in here. Also, I wonder if implementing `Deref` over `MetadataSize` wouldn't be better that implementing the functions like I did... Anyway, please tell me if I can improve this code in any way.

mulkieran:
ok to test

bgurney-rh:
Among the stratisd test failures was this failure, which contains an error message that was affected in one of the commits in this PR:
```
thread 'engine::strat_engine::pool::tests::real_test_basic_metadata' panicked at 'called `Result::unwrap()` on an `Err` value: Engine(Invalid, "metadata length 442 bytes exceeds region available 32 bytes")', libcore/result.rs:1009:5
```

GuillaumeGomez:
I'll check what went wrong.

src/engine/strat_engine/backstore/metadata/mda.rs
```diff
@@ -367,12 +362,38 @@ impl MDARegions {
     }
 }
 
+#[derive(Debug, Default, PartialEq, PartialOrd)]
+pub struct MetaDataSize(Bytes);
+
+impl MetaDataSize {
+    pub fn new(used: Bytes, max_available: Bytes) -> StratisResult<MetaDataSize> {
```
mulkieran: I think it's best to not have the ```max_available``` parameter here. It made sense to have a minimum for MDADataSize, but the minimum for that is a constant, so doesn't need to be a parameter. Here, though, the minimum for the MetaDataSize depends on the value for the MDADataSize, which is not a constant. Better to just construct the value as it is, and leave the comparison check in ```save_state```.

src/engine/strat_engine/backstore/metadata/mda.rs
```diff
@@ -367,12 +362,38 @@ impl MDARegions {
...
+            Ok(MetaDataSize(used))
+        }
+    }
+
+    #[allow(dead_code)]
+    pub fn bytes(&self) -> Bytes {
+        self.0
+    }
+
+    pub fn size(&self) -> u64 {
```
mulkieran: Please drop this method. You can use ```*value.bytes()``` instead of ```value.size()``` everywhere you are using ```value.size()```.

GuillaumeGomez:
Updated.

src/engine/strat_engine/backstore/metadata/mda.rs
```diff
@@ -367,12 +367,25 @@ impl MDARegions {
     }
 }
 
+#[derive(Debug, Default, PartialEq, PartialOrd)]
```
mulkieran: Please derive the same set of properties as the other examples.

mulkieran: Please add a doc comment, like:

A type representing the actual size of variable length metadata written within this
metadata region. This amount must never be greater than the size of the region allocated for the variable length metadata, which has type MDADataSize.

src/engine/strat_engine/backstore/metadata/mda.rs
```diff
@@ -367,12 +367,25 @@ impl MDARegions {
...
+    }
+}
+
 #[derive(Debug)]
 pub struct MDAHeader {
     last_updated: DateTime<Utc>,
 
     /// Size of region used for pool metadata.
-    used: Bytes,
+    used: MetaDataSize,
```
mulkieran: Please remove or fix the comment above ```used```. It is too ambiguous to be useful. Maybe,

/// Size of the variable length metadata last written in the corresponding MDA region.
/// If no variable length has been written, the size is 0 bytes.

mulkieran:
@GuillaumeGomez I'm afraid you'll have to deal w/ a conflict, but it should be striaghtforward. I'ld like to ask you to augment this PR w/ one further step, by making the type of ```used``` Option<MetaDataSize>```. Then it should be ```None``` if no variable length metadata has ever been written, and ```Some(...)``` otherwise.

src/engine/strat_engine/backstore/metadata/mda.rs
GuillaumeGomez: That's what I did: I took the derive from `MDADataSize` which is basically the same. Or are you referring to another type?

GuillaumeGomez:
If you prefer using an `Option`, why not. But what should I do for the `bytes` method in case it's `None`? Should I return 0?

src/engine/strat_engine/backstore/metadata/mda.rs
```diff
@@ -367,12 +367,29 @@ impl MDARegions {
...
+
 #[derive(Debug)]
 pub struct MDAHeader {
     last_updated: DateTime<Utc>,
 
-    /// Size of region used for pool metadata.
-    used: Bytes,
+    /// Size of the variable length metadata last written in the corresponding MDA region.
+    /// If no variable length has been written, the size is 0 bytes.
+    used: MetaDataSize,
```
GuillaumeGomez: If this becomes an `Option<MetaDataSize>`, I'll update the doc comment.

GuillaumeGomez:
Updated with usage of `Option` for the `used` field (in the second commit).

src/engine/strat_engine/backstore/metadata/mda.rs
```diff
@@ -574,9 +602,10 @@ mod tests {
     #[test]
     fn test_from_buf_crc_error() {
         let data = [0u8; 3];
+
         let header = MDAHeader {
             last_updated: Utc::now(),
-            used: Bytes(data.len() as u64),
+            used: MetaDataSize::new(Bytes(data.len() as u64)),
```
bgurney-rh: The Jenkins test failed with this error:
```
   --> src/engine/strat_engine/backstore/metadata/mda.rs:608:19
    |
608 |             used: MetaDataSize::new(Bytes(data.len() as u64)),
    |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |                   |
    |                   expected enum `std::option::Option`, found struct `engine::strat_engine::backstore::metadata::mda::MetaDataSize`
    |                   help: try using a variant of the expected type: `engine::strat_engine::serde_structs::_IMPL_SERIALIZE_FOR_PoolSave::_serde::export::Some(<MetaDataSize>::new(Bytes(data.len() as u64)))`
    |
    = note: expected type `std::option::Option<engine::strat_engine::backstore::metadata::mda::MetaDataSize>`
               found type `engine::strat_engine::backstore::metadata::mda::MetaDataSize`

```

Did you mean to say this?
```
            used: Some(MetaDataSize::new(Bytes(data.len() as u64))),
```

GuillaumeGomez: Weird that I didn't get the error when running locally `make build`... I'll update, thanks for the logs!

GuillaumeGomez: Ah it was in a test, that's why!

mulkieran: > Ah it was in a test, that's why!

```
> make test
```

will compile all the tests and run the ones that are purely computational (don't build devices or actually have an effect on your machine). I always make sure not to be root when I do this on my own machine, just in case there's an error in how the tests are set up, but that has never been the case so far.

mulkieran:
@GuillaumeGomez I think I misled you about the desirability of the Option type. Please don't do anymore work on this, until I figure out what's best. This work tipped me off to the bug that #1551 fixes, but I think that also means that the Option type is undesirable (and that we need better docs on MDAHeader).

GuillaumeGomez:
Ok, keep me up to date once you have something definitive in mind!

mulkieran:
Blocked by #1551, but that should clear soon.

mulkieran:
@GuillaumeGomez Now that #1554 has cleared, this is ready to go forward again. Based on what was figured out in #1554, the commit that adds the Option type should be dropped. I think that then, all you need to do is a rebase, and that this should be ready to go.

GuillaumeGomez:
Ok, will update soon!

GuillaumeGomez:
Updated.

src/engine/strat_engine/backstore/metadata/mda.rs
```diff
@@ -394,12 +394,29 @@ impl MDARegions {
     }
 }
 
+/// A type representing the actual size of variable length metadata written within this metadata
```
bgurney-rh: The stratisd test failed on something slightly above this, on line 392:
```
   --> src/engine/strat_engine/backstore/metadata/mda.rs:392:57
    |
392 |                 .map_or_else(|| true, |mda| mda.used != Bytes(0)));
    |                                                         ^^^^^^^^ expected struct `engine::strat_engine::backstore::metadata::mda::MetaDataSize`, found struct `devicemapper::units::Bytes`
    |
    = note: expected type `engine::strat_engine::backstore::metadata::mda::MetaDataSize`
               found type `devicemapper::units::Bytes`
```

bgurney-rh: Strange; this is part of what was added in stratisd PR #1554, but the stratisd test passed for that PR.  I wonder why it's failing now.

GuillaumeGomez: Interesting, it didn't pass `cargo check` but `make build` compiled just fine. I wonder what's going on... I fixed it.

bgurney-rh: Looks good so far; the stratisd test is still running, but it's progressed past this point.

bgurney-rh: The stratisd test passed.

src/engine/strat_engine/backstore/metadata/mda.rs
```diff
@@ -389,17 +389,34 @@ impl MDARegions {
...
+/// length metadata, which has `MDADataSize` type.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct MetaDataSize(Bytes);
+
+impl MetaDataSize {
+    pub fn new(used: Bytes) -> MetaDataSize {
+        MetaDataSize(used)
+    }
+
+    pub fn bytes(&self) -> Bytes {
```
bgurney-rh: There was a travis-ci test failure:
```
error: this argument is passed by reference, but would be more efficient if passed by value
   --> src/engine/strat_engine/backstore/metadata/mda.rs:408:18
    |
408 |     pub fn bytes(&self) -> Bytes {
    |                  ^^^^^ help: consider passing by value instead: `self`
    |
    = note: `-D clippy::trivially-copy-pass-by-ref` implied by `-D warnings`
```

mulkieran: @GuillaumeGomez We have a clippy make target, which you can run on your local.

mulkieran:
I think this is ready to go.

PR #1547, opened 2019-05-31
Make more understandable error message

GuillaumeGomez:
Took me longer than it should have to understand that it was talking about a binary (`thin_repair` in my case).

### DIFF
--- a/src/engine/strat_engine/cmd.rs
+++ b/src/engine/strat_engine/cmd.rs
@@ -24,13 +24,15 @@ use uuid::Uuid;
 
 use crate::stratis::{StratisError, StratisResult};
 
+const BINARIES_PATHS: [&str; 4] = ["/usr/sbin", "/sbin", "/usr/bin", "/bin"];
+
 /// Find the binary with the given name by looking in likely locations.
 /// Return None if no binary was found.
 /// Search an explicit list of directories rather than the user's PATH
 /// environment variable. stratisd may be running when there is no PATH
 /// variable set.
 fn find_binary(name: &str) -> Option<PathBuf> {
-    ["/usr/sbin", "/sbin", "/usr/bin", "/bin"]
+    BINARIES_PATHS
         .iter()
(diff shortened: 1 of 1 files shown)


mulkieran:
ok to test

mulkieran:
Could you do something slightly different?

We search in a few places for the executable, specified at about line 33. Could you make that array a constant, and improve the error message so that it specifies where the code looked for the executable using that array? Something like "Could not find executable <the name> in any of <the paths>."

GuillaumeGomez:
Sure.

GuillaumeGomez:
It now looks like this:

```
Unable to find executable "thin_repair" in any of "/usr/sbin", "/sbin", "/usr/bin", "/bin". Did you install it?
```

Is it good for you or do you prefer something else?

src/engine/strat_engine/cmd.rs
```diff
@@ -67,8 +69,13 @@ pub fn verify_binaries() -> StratisResult<()> {
     match BINARIES.iter().find(|&(_, ref path)| path.is_none()) {
         None => Ok(()),
         Some((ref name, _)) => Err(StratisError::Error(format!(
-            "Unable to find absolute path for \"{}\"",
-            name
+            "Unable to find executable \"{}\" in any of {}. Did you install it?",
```
mulkieran: Please give the bare information here, without the follow-up question or suggestion. In future, this could end up in several possible contexts, like a log message, or some further larger message, where the suggestion would be out of place.

GuillaumeGomez: Ok, I'll remove the second part.

GuillaumeGomez:
Updated!

GuillaumeGomez:
No problem, I'll update shortly.

GuillaumeGomez:
Fixed clippy, squashed and rebased on master.

GuillaumeGomez:
That's what happens when playing around on different computers, sorry about that. Updated!

PR #1499, opened 2019-04-30
[skip ci] Add an error type for the command module

mulkieran:
Signed-off-by: mulhern <amulhern@redhat.com>

PR #1549, opened 2019-06-03
Introduce a simple method to parse an MDAHeader buffer

mulkieran:
This PR simplifies a spot in the code and makes future error management easier by better isolating code that can not fail from code that can.


mulkieran:
This was cherry-picked from #1499. I thought that it had been enabled by the changes in #1499, but actually it has always been possible, so it seems reasonable to put it up now and keep it out of #1499, since it is orthogonal to those changes.

mulkieran:
rebased.

mulkieran:
looks like udev db is too slow for our tests.

PR #1551, opened 2019-06-04
Harden implementation of MDARegions as much as possible (prelim)

mulkieran:
This hardens the implementation of MDARegions to avoid future bugs.

Most of the work consists of fixing incorrect header comments and adding further explanation to
some implementations. Also:

* ```MDAHeader::parse_buf``` now checks the value of the length of the data to determine whether or not the sectors read correspond to an existing header, rather than checking the value of the timestamp, as that seems like a more direct and meaningful value.
* There is an invariant on ```MDARegions``` which is used in the one relevant test.
* Renames a field in the ```MDARegions``` struct to be less confusing.

### DIFF
--- a/src/engine/strat_engine/backstore/metadata/bda.rs
+++ b/src/engine/strat_engine/backstore/metadata/bda.rs
@@ -159,9 +159,9 @@ impl BDA {
         // Assume that, since a valid StaticHeader was found on the device,
         // that this implies that BDA::initialize() was succesfully executed
         // sometime in the past. Since that is the case, valid MDA headers
-        // were written to the device. If there is an error loading the
-        // MDARegions, which can only be caused by an I/O error or invalid
-        // MDA headers, return an error.
+        // were written to the device. Returns an error if there is an error
+        // when loading the MDARegions, which can only be caused by an I/O
+        // error or invalid MDA headers.
         let regions = mda::MDARegions::load(BDA_STATIC_HDR_SIZE, header.mda_size, f)?;
 
         Ok(Some(BDA { header, regions }))
--- a/src/engine/strat_engine/backstore/metadata/bda.rs
+++ b/src/engine/strat_engine/backstore/metadata/bda.rs
@@ -156,6 +156,12 @@ impl BDA {
             None => return Ok(None),
         };
 
+        // Assume that, since a valid StaticHeader was found on the device,
+        // that this implies that BDA::initialize() was succesfully executed
+        // sometime in the past. Since that is the case, valid MDA headers
+        // were written to the device. If there is an error loading the
+        // MDARegions, which can only be caused by an I/O error or invalid
+        // MDA headers, return an error.
         let regions = mda::MDARegions::load(BDA_STATIC_HDR_SIZE, header.mda_size, f)?;
 
         Ok(Some(BDA { header, regions }))
--- a/src/engine/strat_engine/backstore/metadata/bda.rs
+++ b/src/engine/strat_engine/backstore/metadata/bda.rs
@@ -496,12 +496,7 @@ impl fmt::Debug for StaticHeader {
 mod tests {
     use std::io::{Cursor, Write};
 
-    use proptest::{
-        collection::{vec, SizeRange},
-        num, option,
-        prelude::BoxedStrategy,
-        strategy::Strategy,
-    };
+    use proptest::{collection::vec, num, option, prelude::BoxedStrategy, strategy::Strategy};
     use uuid::Uuid;
 
     use devicemapper::{Bytes, Sectors, IEC};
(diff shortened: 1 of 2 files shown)


mulkieran:
Pulling this back...there is more than one misleading header comment.

mulkieran:
rebased. add an additional comment.

mulkieran:
rebased, add an invariant.

mulkieran:
@bgurney-rh , @drckeefe There have been several additions as more things came up. Please take a look again.

src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -156,6 +156,12 @@ impl BDA {
             None => return Ok(None),
         };
 
+        // Assume that, since a valid StaticHeader was found on the device,
+        // that this implies that BDA::initialize() was succesfully executed
+        // sometime in the past. Since that is the case, valid MDA headers
+        // were written to the device. If there is an error loading the
```
GuillaumeGomez: The last sentence is a bit complicated to read. What about `Returns an error if there is an error when loading the MDARegions, which can only be caused by an I/O error or invalid MDA headers.`?

src/engine/strat_engine/backstore/metadata/mda.rs
```diff
@@ -155,7 +155,11 @@ pub struct MDARegions {
     /// The MDA headers which contain information about the variable
     /// length metadata. NUM_PRIMARY_MDA_REGIONS is 2: in the general
     /// case one is more recently written than the other.
-    mdas: [Option<MDAHeader>; mda_size::NUM_PRIMARY_MDA_REGIONS],
+    /// A value of None indicates that no variable length metadata has been
+    /// written to the MDA regions corresponding to a given MDA header.
+    /// If there is Some value, then variable length metadata has been read;
+    /// the MDA header's used attribute therefore can not be 0 bytes.
```
GuillaumeGomez: "\`used\` field" instead maybe?

src/engine/strat_engine/backstore/metadata/bda.rs
mulkieran: Ok, I'll go w/ your version.

src/engine/strat_engine/backstore/metadata/mda.rs
mulkieran: Yes, field is probably better.

src/engine/strat_engine/backstore/metadata/mda.rs
```diff
@@ -197,15 +204,21 @@ impl MDARegions {
...
-    /// well-formed MDAHeaders for this device.
+    /// Construct an MDARegions struct from data on the disk.
+    /// The individual MDAHeaders in the struct may all be None, as it is
+    /// possible that no variable length metadata has been written to the
+    /// device on which the metadata has been written.
+    ///
+    /// Returns an error if there is an I/O error or if the MDA header data
+    /// on the device is invalid.
+    //
+    // TODO: Consider whether the return type of this method should be
```
GuillaumeGomez: Create an issue so this isn't forgotten? (returning a `Result` containing an enum looking like this: `enum Error<E: Error, T> { IOError(E), Error(E), Ok(T) }` should do the trick I guess?)

src/engine/strat_engine/backstore/metadata/mda.rs
```diff
@@ -391,19 +420,20 @@ impl Default for MDAHeader {
...
 impl MDAHeader {
     /// Parse a valid MDAHeader from buf.
-    /// If the timestamp value is 0, then return None. This means that
-    /// No variable length metadata has been written, so there is no
-    /// corresponding header.
+    /// If the amount used by the variable length metadata is 0, return None,
+    /// as this means that no variable length metadata has been written.
     fn parse_buf(buf: &[u8; mda_size::_MDA_REGION_HDR_SIZE]) -> Option<MDAHeader> {
-        match LittleEndian::read_u64(&buf[16..24]) {
+        match LittleEndian::read_u64(&buf[8..16]) {
```
GuillaumeGomez: Is it normal that it changed from `16..24` to `8..16`?

src/engine/strat_engine/backstore/metadata/mda.rs
mulkieran: I thought of creating an issue, but I don't think that the question has quite that status yet. It's more a hint to the reader at this point, since I'm not sure if it would or would not be useful to distinguish in that way.

src/engine/strat_engine/backstore/metadata/mda.rs
mulkieran: Exactly!

src/engine/strat_engine/backstore/metadata/mda.rs
GuillaumeGomez: Fine by me.

src/engine/strat_engine/backstore/metadata/mda.rs
mulkieran: > On line 160 can the word some have a lower case s?

@drckeefe: That's a Rust-specific "Some", it's one of the two constructors of the Option type.
https://doc.rust-lang.org/std/option/enum.Option.html. That's short for saying something like "if the value has been constructed with the Some constructor", or something like that.

GuillaumeGomez: Not a constructor, a variant. (sorry for nitpicking) :)

mulkieran:
Superceded by #1554 .

src/engine/strat_engine/backstore/metadata/mda.rs
mulkieran: It is both! But that would be somewhat easier to say: "If the value is the Some variant of the Option type".

PR #1552, opened 2019-06-04
Update nix to 0.14

cuviper:
This is needed to fix breaking changes in libc on arm and s390x.
See: https://github.com/nix-rust/nix/pull/1055


cuviper:
Note -- we'll also need published versions of these to completely update `nix`:

- https://github.com/tailhook/libmount/pull/12
- https://github.com/stratis-storage/devicemapper-rs/pull/475

mulkieran:
ok to test

mulkieran:
@cuviper Can you rebase on master? We plan to cut new releases of stratisd and devicemapper tomorrow.

cuviper:
Rebased, and I bumped `libmount` for its `nix` update. I think we should also update `devicemapper` here once it is published, so the lock file is consistent with only the newest `nix`.

mulkieran:
> Rebased, and I bumped `libmount` for its `nix` update. I think we should also update `devicemapper` here once it is published, so the lock file is consistent with only the newest `nix`.

Yes, thanks! The complete plan is:

1. Merge your devicemapper PR.
2. Cut a new release of devicemapper.
3. Publish release on crates.io.
4. Merge this PR.
5. Merge a new PR to update stratisd devicemapper requirement to new version.
6. Cut a new release of stratisd.

cuviper:
OK, if you want to do it as two separate PRs here, that's up to you. :)

PR #1554, opened 2019-06-06
Harden implementation of MDARegions as much as possible

mulkieran:
This hardens the implementation of MDARegions to avoid future bugs.

Most of the work consists of fixing incorrect header comments and adding further explanation to
some implementations. Also:

* ```MDAHeader::parse_buf``` now checks the value of the length of the data to determine whether or not the sectors read correspond to an existing header, rather than checking the value of the timestamp, as that seems like a more direct and meaningful value.
* There is an invariant on ```MDARegions``` which is used in the one relevant test.
* Renames a field in the ```MDARegions``` struct to be less confusing.

### DIFF
--- a/src/engine/strat_engine/backstore/metadata/bda.rs
+++ b/src/engine/strat_engine/backstore/metadata/bda.rs
@@ -156,6 +156,12 @@ impl BDA {
             None => return Ok(None),
         };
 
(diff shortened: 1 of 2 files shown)


mulkieran:
Supercedes #1551 .

mulkieran:
rebased.

PR #1562, opened 2019-06-17
[WIP] StaticHeader::setup Improve error reporting

### LINKED ISSUE #1443: StaticHeader::setup  Improve error reporting
Note:  Taken from https://github.com/stratis-storage/stratisd/pull/1440#discussion_r259868293

```rust
fn setup<F>(f: &mut F) -> StratisResult<Option<StaticHeader>>
    where
        F: Read + Seek + SyncAll,
```
This returns an error in two situations:

 * If the metadata seems to indicate that the device is a Stratis deivce, but no well-formed signature could be found.
 * If a well-formed signature was found but an error occurred during the repair of a second, ill-formed, old, or unreadable signature.

I think it is right to return an error in the first case, but I think the second should possibly be handled differently. I think a good option is to enrich the return type to
```rust
StratisResult<Option<StaticHeader>, Option<some error thing if repair fails>>
```

Additionally, taken from: https://github.com/stratis-storage/stratisd/pull/1440#discussion_r259869237
```rust
let err_str =  "Appeared to be a Stratis device, but no valid sigblock found";
Err(StratisError::Engine(ErrorEnum::Invalid, err_str.into()))
```
Actually, better to include both errors, and improve the error messages in the errors returned from `sigblock_from_buf`.

GuillaumeGomez:
Fixes #1443.

Before finalizing this: do we want to provide two different kind of errors in here or consider the second case of error as non-fatal and providing it alongside the returned data (like I did currently)?


mulkieran:
> Before finalizing this: do we want to provide two different kind of errors in here or consider the second case of error as non-fatal and providing it alongside the returned data (like I did currently)?

Yes, this is exactly right. We should consider the error on repair attempt as non-fatal, because if we tried to repair that means we think we have good data on hand.

GuillaumeGomez:
@mulkieran suggested to use an enum instead and also to add `#[must_use]` usage. I'll check with those solutions.

GuillaumeGomez:
Should I do something with the errors or...?

src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -283,7 +300,7 @@ impl StaticHeader {
     /// Return an error if the sigblocks differ in some unaccountable way.
     /// Returns an error if a write intended to repair an ill-formed,
     /// unreadable, or stale signature block failed.
-    fn setup<F>(f: &mut F) -> StratisResult<Option<StaticHeader>>
+    fn setup<F>(f: &mut F) -> StratisResult<Option<SetupResult>>
```
mulkieran: Note that your change also requires a change to the header comment, since the method no longer just returns an error if the repair failed.

src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -298,41 +315,44 @@ impl StaticHeader {
...
+                                    Ok(Some(
+                                        SetupResult::new_with_error(
+                                            loc_1,
+                                            StratisError::Engine(ErrorEnum::Invalid, err_str.into()))))
                                 } else if loc_1.initialization_time > loc_2.initialization_time {
                                     // If the first header block is newer, overwrite second with
                                     // contents of first.
                                     BDA::write(f, &buf_loc_1, MetadataLocation::Second)?;
-                                    Ok(Some(loc_1))
+                                    Ok(Some(SetupResult::new(loc_1)))
```
mulkieran: _Here_ is where you have to split into two possibilities: if the write succeeded, then ```SetupResult::Ok```, else ```SetupResult::OkWithErr```.

A rule to follow is that ```Err```'s in the old version map to ```Err```s in the new version and the same w/ ```Ok```'s. It's that internal ```SetupResult``` that can now vary based on whether the repair succeeds or not. There are a few more places that need similar adjustment.

src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -298,41 +315,44 @@ impl StaticHeader {
                         match (loc_1, loc_2) {
                             (Some(loc_1), Some(loc_2)) => {
                                 if loc_1 == loc_2 {
-                                    Ok(Some(loc_1))
+                                    Ok(Some(SetupResult::new(loc_1)))
                                 } else if loc_1.initialization_time == loc_2.initialization_time {
                                     // Inexplicable disagreement among static headers
                                     let err_str = "Appeared to be a Stratis device, but signature blocks disagree.";
-                                    Err(StratisError::Engine(ErrorEnum::Invalid, err_str.into()))
+                                    Ok(Some(
```
mulkieran: You'll need to stick w/ ```Err``` here, because no acceptable sigblock has been found.

GuillaumeGomez:
Updated. Is there anything else to do in here?

src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -380,10 +413,16 @@ impl StaticHeader {
                 // Copy 2 read OK, 1 resulted in IO Error
                 match StaticHeader::sigblock_from_buf(&buf_loc_2) {
                     Ok(loc_2) => {
-                        if loc_2.is_some() {
-                            BDA::write(f, &buf_loc_2, MetadataLocation::First)?;
+                        if let Some(loc_2) = loc_2 {
```
mulkieran: Note that in this if-expression you always return ```Ok(...)```, so you can move the ```Ok``` out to surround the ```if-expression```, like ```Ok(if let Some(loc_2) = loc_2 {...} else { None })```. It's a bit clearer that way, because it reinforces the fact that if you found one good signature buffer and the other just had an IO error, then you consider the succesfully read one to be correct.

src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -411,7 +450,11 @@ impl StaticHeader {
         // it must also have correct CRC, no weird stuff in fields,
         // etc!
         match StaticHeader::setup(f) {
-            Ok(Some(sh)) => Ok(Some((sh.pool_uuid, sh.dev_uuid))),
+            Ok(Some(SetupResult::Ok(sh))) => Ok(Some((sh.pool_uuid, sh.dev_uuid))),
+            Ok(Some(SetupResult::OkWithError(sh, err))) => {
+                warn!("StaticHeader::setup didn't completely succeed: {:?}", err);
```
mulkieran: The I/O error that you picked up doesn't really explain what happened. So, a clearer explanation would be something like:

```warn!("Experienced an I/O error while attempting to repair an ill-formed, unreadable, or stale signature block: {:?}", err);```

This has the flaw that it does not identify which signature block, on which device, etc., so it is more alarming than helpful in that simple state. 

But you can improve it quite a bit, by including the good StaticHeader, the one that was succesfully read, in the error message (or possibly just selected identifying information from that StaticHeader). This would probably be worth a separate function, since you do it in two places.

Comment (GuillaumeGomez) [reply src/engine/strat_engine/backstore/metadata/bda.rs]: I extended your text a bit. Tell me if it's what you had in mind otherwise I'll just update. :)

File: src/engine/strat_engine/backstore/metadata/bda.rs [hunk 6]
```diff
@@ -376,12 +412,11 @@ impl StaticHeader {
...
                 // Copy 2 read OK, 1 resulted in IO Error
                 match StaticHeader::sigblock_from_buf(&buf_loc_2) {
-                    Ok(loc_2) => {
-                        if loc_2.is_some() {
-                            BDA::write(f, &buf_loc_2, MetadataLocation::First)?;
-                        }
-                        Ok(loc_2)
-                    }
+                    Ok(loc_2) => Ok(if let Some(loc_2) = loc_2 {
+                        bda_write_check(f, &buf_loc_2, MetadataLocation::First, loc_2)
```
bgurney-rh: There was a compilation failure here: 
```
   --> src/engine/strat_engine/backstore/metadata/bda.rs:416:25
    |
416 |                         bda_write_check(f, &buf_loc_2, MetadataLocation::First, loc_2)
    |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected enum `std::option::Option`, found enum `std::result::Result`
    |
    = note: expected type `std::option::Option<_>`
               found type `std::result::Result<std::option::Option<_>, stratis::errors::StratisError>`
```

mulkieran:
@GuillaumeGomez Could you rebase this one? ```device_identifiers``` has moved and changed its structure, so there is a conflict. I believe it shouldn't be hard for you to adapt your changes to the new structure, and location.

GuillaumeGomez:
Sure!

GuillaumeGomez:
Rebased.

mulkieran:
Makes sense to call this blocked b #1575.

mulkieran:
unblocked.

GuillaumeGomez:
I need to rebase once again. :)

mulkieran:
yup. conflicts galore, unfortunately, but ultimately a better situation.

src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -377,28 +423,36 @@ impl StaticHeader {
...
-                (Err(_), Err(_)) => {
-                    let err_str = "Appeared to be a Stratis device, but no valid sigblock found";
-                    Err(StratisError::Engine(ErrorEnum::Invalid, err_str.into()))
+                (None, None) => Ok(None),
+                (Some(loc_1), None) => {
+                    // Copy 1 has valid Stratis BDA, copy 2 has no magic, re-write copy 2
+                    BDA::write(f, &buf_loc_1, MetadataLocation::Second)?;
+                    Ok(Some(loc_1))
+                }
+                (None, Some(loc_2)) => {
```
bgurney-rh: Here's the compilation failure from the stratisd Jenkins test:
```
   --> src/engine/strat_engine/backstore/metadata/bda.rs:436:18
    |
436 |                 (None, None) => Ok(None),
    |                  ^^^^ expected enum `std::result::Result`, found enum `std::option::Option`
    |
    = note: expected type `std::result::Result<std::option::Option<engine::strat_engine::backstore::metadata::bda::StaticHeader>, stratis::errors::StratisError>`
               found type `std::option::Option<_>`

   --> src/engine/strat_engine/backstore/metadata/bda.rs:436:24
    |
436 |                 (None, None) => Ok(None),
    |                        ^^^^ expected enum `std::result::Result`, found enum `std::option::Option`
    |
    = note: expected type `std::result::Result<std::option::Option<engine::strat_engine::backstore::metadata::bda::StaticHeader>, stratis::errors::StratisError>`
               found type `std::option::Option<_>`

   --> src/engine/strat_engine/backstore/metadata/bda.rs:437:18
    |
437 |                 (Some(loc_1), None) => {
    |                  ^^^^^^^^^^^ expected enum `std::result::Result`, found enum `std::option::Option`
    |
    = note: expected type `std::result::Result<std::option::Option<engine::strat_engine::backstore::metadata::bda::StaticHeader>, stratis::errors::StratisError>`
               found type `std::option::Option<_>`

   --> src/engine/strat_engine/backstore/metadata/bda.rs:437:31
    |
437 |                 (Some(loc_1), None) => {
    |                               ^^^^ expected enum `std::result::Result`, found enum `std::option::Option`
    |
    = note: expected type `std::result::Result<std::option::Option<engine::strat_engine::backstore::metadata::bda::StaticHeader>, stratis::errors::StratisError>`
               found type `std::option::Option<_>`

   --> src/engine/strat_engine/backstore/metadata/bda.rs:442:18
    |
442 |                 (None, Some(loc_2)) => {
    |                  ^^^^ expected enum `std::result::Result`, found enum `std::option::Option`
    |
    = note: expected type `std::result::Result<std::option::Option<engine::strat_engine::backstore::metadata::bda::StaticHeader>, stratis::errors::StratisError>`
               found type `std::option::Option<_>`

   --> src/engine/strat_engine/backstore/metadata/bda.rs:442:24
    |
442 |                 (None, Some(loc_2)) => {
    |                        ^^^^^^^^^^^ expected enum `std::result::Result`, found enum `std::option::Option`
    |
    = note: expected type `std::result::Result<std::option::Option<engine::strat_engine::backstore::metadata::bda::StaticHeader>, stratis::errors::StratisError>`
               found type `std::option::Option<_>`

error: aborting due to 6 previous errors
```

GuillaumeGomez: Thanks! I failed the rebase pretty badly apparently. :laughing:

GuillaumeGomez:
cargo fmt once again. T_T

src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -272,10 +276,49 @@ impl BDA {
...
+                setup_warn(&sh, err);
+                Ok(Some((sh.pool_uuid, sh.dev_uuid)))
+            }
+            Ok(None) => Ok(None),
+            Err(err) => Err(err),
+        }
     }
 }
 
+fn setup_warn(header: &StaticHeader, err: StratisError) {
```
mulkieran: Please add some explanatory header comments.

src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -272,10 +276,49 @@ impl BDA {
     where
         F: Read + Seek + SyncAll,
     {
-        StaticHeader::setup(f).map(|sh| sh.map(|sh| (sh.pool_uuid, sh.dev_uuid)))
+        match StaticHeader::setup(f) {
```
mulkieran: Please go with:

```
    where
        F: Read + Seek + SyncAll,
    {
        StaticHeader::setup(f).map(|sh| {
            sh.map(|sh| match sh {
                SetupResult::OkWithError(sh, err) => {
                    setup_warn(&sh, err);
                    (sh.pool_uuid, sh.dev_uuid)
                }
                SetupResult::Ok(sh) => (sh.pool_uuid, sh.dev_uuid),
            })
        })
    }
```

preserving previous structure.

src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -272,10 +276,49 @@ impl BDA {
...
+            Err(err) => Err(err),
+        }
     }
 }
 
+fn setup_warn(header: &StaticHeader, err: StratisError) {
+    warn!(
+        "Experienced an I/O error while attempting to repair an ill-formed, \
+         unreadable, or stale signature block: {:?}.\n\
+         Returning device \"{:?}\" from pool \"{:?}\".",
```
mulkieran: How about ```Read and returned static header {:?}, sh``` for the second sentence. We implemented ```Debug``` for ```StaticHeader``` so that is doable; we had to implement it explicitly since the Debug info for UUIDs is far too long.

src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -272,10 +276,49 @@ impl BDA {
...
+        err, header.dev_uuid, header.pool_uuid
+    );
+}
+
+fn bda_write_check<F>(
```
mulkieran: Please put this method inside ```StaticHeader::setup``` definition.

src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -332,43 +381,40 @@ impl StaticHeader {
...
-                            // If the first header block is newer, overwrite second with
-                            // contents of first.
-                            BDA::write(f, &buf_loc_1, MetadataLocation::Second)?;
-                            Ok(Some(loc_1))
-                        } else {
-                            // The second header block must be newer, so overwrite first
-                            // with contents of second.
-                            BDA::write(f, &buf_loc_2, MetadataLocation::First)?;
-                            Ok(Some(loc_2))
+                (Ok(loc_1), Ok(loc_2)) => {
```
mulkieran: Please drop the extra brace here, you lost that change on the rebase, and I'ld definitely like to keep it.

src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -272,10 +276,49 @@ impl BDA {
...
+        "Experienced an I/O error while attempting to repair an ill-formed, \
+         unreadable, or stale signature block: {:?}.\n\
+         Returning device \"{:?}\" from pool \"{:?}\".",
+        err, header.dev_uuid, header.pool_uuid
+    );
+}
+
+fn bda_write_check<F>(
+    f: &mut F,
+    bda_buf: &[u8],
```
mulkieran: Please change ```bda_buf``` to ```sh_buf```. I know the original calls it ```bda_buf```, but it's wrong, and since this is new code we might as well get it right here.

src/engine/strat_engine/backstore/metadata/bda.rs
GuillaumeGomez: I added a comment. Considering the function was just emitting a warning, I didn't think it was necessary to kinda repeat what it was doing but I guess a comment cannot hurt. Also, don't hesitate if the comment isn't well written.

src/engine/strat_engine/backstore/metadata/bda.rs
GuillaumeGomez: I didn't get one so I guess it was on a previous version?

src/engine/strat_engine/backstore/metadata/bda.rs
GuillaumeGomez: I updated the text but not sure if it was how you had it in mind?

GuillaumeGomez:
Updated.

mulkieran:
Travis failed due to infrastructure errors, not test failiures, I tried restarting it.

src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -344,31 +395,27 @@ impl StaticHeader {
...
-                        Ok(Some(loc_2))
+                        bda_write_check(f, &buf_loc_2, MetadataLocation::First, loc_2)
                     }
                 },
                 (Ok(loc_1), Err(loc_2)) => {
-                    if loc_1.is_some() {
-                        BDA::write(f, &buf_loc_1, MetadataLocation::Second)?;
-                        Ok(loc_1)
+                    // Re-write copy 2
+                    if let Some(loc_1) = loc_1 {
```
mulkieran: Thanks for introducing the ```bda_write_check``` abstraction. It opens up opportunities for some further abstraction as this match arm and the one below are virtually identical. I made an attempt at an abstraction:

```
        // Action to take if there appeared to be one malformed sigblock
        // on the device.
        // If the other sigblock appears not to exist at all, return an error.
        // If the other sigblock exists, attempt a repair of the malformed
        // sigblock and return the other sigblock.
        // sh_buf are the bytes of the other sigblock
        // sh is the optional other sigblock
        // sh_error is the error indicating a malformed sigblock 
        // write_location is where to write the optional repair.
        fn repair_on_sigblock_read_error<F>(
            f: &mut F,
            sh_buf: &[u8],
            sh: Option<StaticHeader>,
            sh_error: StratisError,
            write_location: MetadataLocation,
        ) -> StratisResult<Option<SetupResult>>
        where
            F: Read + Seek + SyncAll,
        {
            match sh {
                Some(sh) => bda_write_check(f, sh_buf, write_location, sh),
                None => Err(sh_error),
            }
        }
```

which seems to work when used here. Please try that out, or something similar.

src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -395,10 +442,11 @@ impl StaticHeader {
             // Copy 1 read OK, 2 resulted in an IO error
             (Ok(buf_loc_1), Err(_)) => match StaticHeader::sigblock_from_buf(&buf_loc_1) {
                 Ok(loc_1) => {
-                    if loc_1.is_some() {
-                        BDA::write(f, &buf_loc_1, MetadataLocation::Second)?;
+                    if let Some(loc_1) = loc_1 {
```
mulkieran: I think that the match arm here, and the one immediately below it can also be abstracted, kind of like suggestion above.

GuillaumeGomez: Not as is. I'll change the function parameters to allow it.

GuillaumeGomez:
Updated.

mulkieran:
@GuillaumeGomez I've realized that there is yet another preliminary step that should happen before you try to move forward w/ this. I'ld already worked it out, but now I realize that it will conflict w/ this very badly. So, please put this on hold, and I'll assign to you this additional pre-project.

mulkieran:
Blocked by #1579.

mulkieran:
Blocked by #1586.

PR #1563, opened 2019-06-17
Eliminate some dead code that has persisted

mulkieran:


mulkieran:
@bgurney-rh, @drckeefe: There are a few other bits of dead code in stratisd, but all of them requiring more justification for removing, for various reasons. These, on the other hand, have definitely passed their sell-by date.

PR #1568, opened 2019-06-19
Move static header functionality into a separate module

mulkieran:
This accomplishes the following:

* Better encapsulation:
  * Since all static header related functionality is now in a separate module, the dependencies of the bda module on static header functionality is made more explicit and obvious.
  * Some constants that were previously shared among all parts of the BDA are now confined to the static header module. The two constants are the stratis magic number and the metadata version.
  * Since it is now obvious that StaticHeader need not be exported, the metadata module has better encapsulation.
  * A number of tests that solely tested static header behavior are now incorporated into the static header tests module and simplified.

* Better documentation: It is now explicitly stated that all methods that read or write the static header seek to the 0 location before beginning. This was not at all obvious before.

* StaticHeader::write is made an instance method, insuring, at the level of the StaticHeader impl, that only a valid signature block can be written to the device by our code. Previously it was necessary to rely on all code that invoked the write method, since the method could take an arbitrary buffer.

### DIFF
--- a/src/engine/strat_engine/backstore/blockdevmgr.rs
+++ b/src/engine/strat_engine/backstore/blockdevmgr.rs
@@ -499,7 +499,7 @@ mod tests {
         tests::{loopbacked, real},
     };
 
-    use crate::engine::strat_engine::backstore::metadata::StaticHeader;
+    use crate::engine::strat_engine::backstore::metadata::BDA;
 
     use super::*;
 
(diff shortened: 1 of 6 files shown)


mulkieran:
The final goal is to use the same tricks used in mda.rs with types to ensure that our sizes are correct.

mulkieran:
Blocked by #1571.

mulkieran:
This will be obsoleted by #1572 and #1574 , perhaps others. It might be useful as a reference for #1573 .

PR #1572, opened 2019-06-20
Use types and defined constant for BDA sizes

mulkieran:
Use types and defined constants for all sizes associated with Stratis metadata.

The metadata stores some sizes which do not describe Stratis metadata regions, they describe Stratis blockdev's, etc. These sizes are still just Sectors or Bytes, they will be distinguished from each other using types in a later PR. All sizes which represent sizes of regions in the metadata or of values stored in those regions are distinguished by types, and related in functions between those types as much as possible.

### DIFF
--- a/src/engine/strat_engine/backstore/blockdev.rs
+++ b/src/engine/strat_engine/backstore/blockdev.rs
@@ -14,7 +14,10 @@ use crate::{
     engine::{
         event::get_engine_listener_list,
         strat_engine::{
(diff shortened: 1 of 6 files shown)


mulkieran:
rebased. Fixed a minor conflict.

mulkieran:
Whoops! Missed a spot.

mulkieran:
Ok, think this is ready to go.

mulkieran:
Just cancelled out a commit and its reversion, code is identical.

Issue #1573, opened 2019-06-24
Place static header specific functionality in a separate module, static_header in metadata module


mulkieran:
Goals:
 * Move all static header specific functions into a static header module. Some of these methods are currently part of the ```BDA``` implementation, these will need to be identified and moved. ```BDA::write``` and ```BDA::save``` are some of these.
 * Move all static header specific tests into a test submodule in the static header module.
 * Currently, all methods that read and write static headers implicitly or explicitly seek to 0 before beginning. Make this explicit, by adding a parameter to these methods for the offset of the static header from the start of the device. All BDA methods will invoke these methods with a value of 0.
  * Probably, as a last step, hoist metadata module up a level so that it is on the same level as backstore module.

Procedure:
  1. ~~Many static header specific tests currently in the bda tests module make use of unnecessarily general BDA methods to set up their test condition. These should be changed so they make use of static header specific methods. An obvious example is that ```BDA::initialize``` which initializes MDA headers is used, when all that is necessary is ```BDA::write``` which only initializes the static headers.
NOTE: ```BDA::write``` is static header specific functionality which should eventually be moved into the proposed ```static_header``` module, but for this initial step, the best course is to leave it where it is.
It should be possible to make this a separate PR.~~ Outdated.
  2. Additional PRs.

You'll know when you're done when: 
 * The ```static_header``` module is the only module which explicitly imports sizes::static_header_sizes.
 * ...

Component issues:
  - [x] #1579
  - [x] #1582
  - [x] #1584
  - [x] #1586
  - [x] #1794

Reference:
  https://stratis-storage.github.io/StratisSoftwareDesign.pdf, p.13 has a useful diagram of the metadata, helping to distinguish between the BDA, the MDA, the static header and the signature blocks.

mulkieran:
@GuillaumeGomez I think that all possible preparations that can be made have been made. I think it should now be possible to split out the relevant parts of the ```bda``` module, which is the StaticHeader definition and implementation and all StaticHeader-specific tests and any StaticHeader specific constants, into a ```static_header``` module. Moving onto the main October board. Happy refactoring!

GuillaumeGomez:
I think it'll have to wait mi-november at best. We're in the middle of the conferences rush! :)

mulkieran:
Ok. Happy conferencing!

mulkieran:
@GuillaumeGomez Would you like to come back to this? If yes, great! If not, please let us know, and thanks for all the work that you have done!

[...]

PR #1574, opened 2019-06-24
Move device_identifiers from StaticHeader to BDA

mulkieran:


mulkieran:
This should also help a bit w #1573, by removing one obvious cross-cutting constraint.

PR #1581, opened 2019-07-14
Move BDA::write to StaticHeader implementation

### LINKED ISSUE #1579: Make BDA::write an instance method in the StaticHeader implementation
Note that BDA::write doesn't write a BDA, but only the signature blocks in the static header which means that it really is in the wrong implementation.

There are two steps to this task:

* Move the method to the StaticHeader implementation and fix up the call sites. Fix up the method header, which uses "BDA" when it should properly use "signature block". There may be mistaken inline comments as well that confuse the BDA with the signature block.
* Make the method an instance method. That is, it should take a reference to self as an argument, it should not take a buffer as an argument, and it should construct the buffer it needs to write within the method body. The advantage of this is that the method ```StaticHeader::write``` created in the previous step can only write well-formed sigblocks to the designated locations, i.e., the program can not write arbitrary bytes to these spots. Fix up all call sites.

Related: #1573.

GuillaumeGomez:
Fixes #1579.

### DIFF
--- a/src/engine/strat_engine/backstore/metadata/bda.rs
+++ b/src/engine/strat_engine/backstore/metadata/bda.rs
@@ -106,43 +106,6 @@ impl BDA {
         )
     }
 
-    // Writes bda_buf according to the value of which.
-    // If first location is specified, write zeroes to empty regions in the
-    // first 8 sectors. If the second location is specified, writes zeroes to empty
-    // regions in the second 8 sectors.
-    fn write<F>(f: &mut F, bda_buf: &[u8], which: MetadataLocation) -> io::Result<()>
-    where
-        F: Seek + SyncAll,
-    {
-        let zeroed = [0u8; bytes!(static_header_size::POST_SIGBLOCK_PADDING_SECTORS)];
-        f.seek(SeekFrom::Start(0))?;
-
-        // Write to a static header region in the static header.
-        fn write_region<F>(f: &mut F, bda_buf: &[u8], zeroed: &[u8]) -> io::Result<()>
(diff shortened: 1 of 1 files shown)


src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -344,31 +307,31 @@ impl StaticHeader {
                         } else if loc_1.initialization_time > loc_2.initialization_time {
                             // If the first header block is newer, overwrite second with
                             // contents of first.
-                            BDA::write(f, &buf_loc_1, MetadataLocation::Second)?;
+                            loc_1.write(f, MetadataLocation::Second)?;
```
GuillaumeGomez: I wondered about one thing: we call `sigblock_from_buf` just above in order to get in here and then I call it again inside `StaticHeader::write`. Should I let it as is or do I rewrite the whole thing differently? (not sure if it's really necessary but asking just in case...)

mulkieran:
ok to test

GuillaumeGomez:
I moved back the `write` method. :)

mulkieran:
> I moved back the `write` method. :

Whoops! Too far! I think you moved it right back into BDA implementation, somehow.

GuillaumeGomez:
What failed in the test btw? `cargo fmt` seems ok so no clue what failed...

GuillaumeGomez:
Ah, that might explains what failed.

GuillaumeGomez:
Ok, sorry about the mess. Should be good again.

PR #1585, opened 2019-07-16
Use map to simplify structure of setup a bit

mulkieran:
Related: #1573.


mulkieran:
@GuillaumeGomez I was finding this change hard to explain...so I thought it would be more efficient to just do it. I think it will facilitate #1586.

PR #1588, opened 2019-07-18
Tidy up imports

mulkieran:
Related: https://github.com/stratis-storage/project/issues/39

The imports weren't that bad so the main accomplishments were to improve encapsulation in the backstore module by no longer defining ```blkdev_size``` within that module and no longer publicly exporting its ```device``` sub-module.


mulkieran:
rebased.

Issue #1509, opened 2019-05-04
Stratis pool becomes broken after being activated on machine with newer stratisd and incorrect date/time.


sergeystepanovx:
Hi, guys.

I have been experimenting with Stratis in new Linux distribs and encountered some problems with pool going into some kind of degraded state.
I've managed to get it 100% reproducible, though it requires rather complex configuration.

Basically, I have 2 machines, one with RHEL 8 Beta (with stratisd 1.0.0) and the other one with Fedora 30 Beta (stratisd 1.0.3).
**Machine with Fedora 30 has time set in the past** (doesn't matter, a few hours or a year).
Steps are the following:
1. Create a simple stratis pool on one disk on RHEL machine
````
[root@localhost ~]# stratis pool list
Name       Total Physical Size  Total Physical Used
rh8pool              16.00 GiB             1.12 GiB
[root@localhost ~]# stratis filesystem list
Pool Name  Name    Used     Created            Device
rh8pool    testfs  546 MiB  May 01 2019 20:28  /dev/stratis/rh8pool/testfs
rh8pool    lalala  546 MiB  May 01 2019 20:28  /dev/stratis/rh8pool/lalala
[root@localhost ~]# stratis blockdev list
Pool Name  Device Node    Physical Size       State  Tier
rh8pool    /dev/sdb1              4 GiB  Not-in-use  Data
rh8pool    /dev/sdb2          12.00 GiB      In-use  Data
````

2. Turn off the machine, detach the disk and attach it to the Fedora machine
3. The pool becomes active and on the first glance everything looks OK, all devices are activated and file systems are mountable.
````
sdb                                                                8:16   0    16G  0 disk
├─sdb1                                                             8:17   0     4G  0 part
│ └─stratis-1-private-bcf7b4d98a284e58a602834045a6d604-physical-originsub
│                                                                253:3    0    16G  0 stratis
│   ├─stratis-1-private-bcf7b4d98a284e58a602834045a6d604-flex-thinmeta
│   │                                                            253:4    0    16M  0 stratis
│   │ └─stratis-1-private-bcf7b4d98a284e58a602834045a6d604-thinpool-pool
│   │                                                            253:6    0    16G  0 stratis
│   │   ├─stratis-1-bcf7b4d98a284e58a602834045a6d604-thin-fs-2310877ed848427bbbcf76d7af3fe80a
│   │   │                                                        253:8    0     1T  0 stratis
[...]

PR #1590, opened 2019-07-22
Fix issue with Stratis metadata handling when the two host machines have different times

jbaublitz:
This PR should fix the problem we were seeing in #1509 by changing `get_metadata` to always return the associated timestamp with the parsed metadata. Only the timestamp is returned from the `MDAHeader` as it is the only part of that struct that is useful outside of the `impl MDAHeader` block. Given that there should always be a timestamp if metadata is present, the timestamp and parsed metadata are wrapped in an `Option` as you should never get one back but not the other. `Backstore::setup` is always provided with `Some(timestamp)` as the precondition for this function is that metadata verification has always been done according to the comments so it should always be present if the metadata is present.

### DIFF
--- a/src/engine/strat_engine/backstore/setup.rs
+++ b/src/engine/strat_engine/backstore/setup.rs
@@ -94,22 +94,25 @@ pub fn get_metadata(
     // Try to read from all available devnodes that could contain most
     // recent metadata. In the event of errors, continue to try until all are
     // exhausted.
-    for &(devnode, ref bda) in bdas
+    let poolsave = bdas
         .iter()
-        .filter(|&&(_, ref bda)| bda.last_update_time() == Some(most_recent_time))
-    {
-        let timestamp = bda.last_update_time();
-        let poolsave = OpenOptions::new()
-            .read(true)
-            .open(devnode)
-            .ok()
-            .and_then(|mut f| bda.load_state(&mut f).ok())
-            .and_then(|opt| opt)
-            .and_then(|data| serde_json::from_slice(&data).ok());
+        .filter_map(|&(devnode, ref bda)| {
+            if bda.last_update_time() == Some(most_recent_time) {
+                OpenOptions::new()
+                    .read(true)
+                    .open(devnode)
+                    .ok()
+                    .and_then(|mut f| bda.load_state(&mut f).ok())
+                    .and_then(|opt| opt)
+                    .and_then(|data| serde_json::from_slice(&data).ok())
+            } else {
+                None
+            }
+        })
+        .next();
 
-        if let (Some(ts), Some(psave)) = (timestamp, poolsave) {
-            return Ok(Some((*ts, psave)));
-        }
+    if let Some(psave) = poolsave {
+        return Ok(Some((*most_recent_time, psave)));
     }
 
     // If no data has yet returned, we have an error. That is, we should have
--- a/src/engine/strat_engine/backstore/setup.rs
+++ b/src/engine/strat_engine/backstore/setup.rs
@@ -94,8 +94,7 @@ pub fn get_metadata(
     // Try to read from all available devnodes that could contain most
     // recent metadata. In the event of errors, continue to try until all are
     // exhausted.
-    let poolsave = bdas
-        .iter()
+    bdas.iter()
         .filter_map(|&(devnode, ref bda)| {
             if bda.last_update_time() == Some(most_recent_time) {
                 OpenOptions::new()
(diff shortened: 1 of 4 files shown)


mulkieran:
add to whitelist

src/engine/strat_engine/backstore/setup.rs
```diff
@@ -97,6 +98,7 @@ pub fn get_metadata(
         .iter()
         .filter(|&&(_, ref bda)| bda.last_update_time() == Some(most_recent_time))
     {
+        let timestamp = bda.last_update_time();
```
mulkieran: ```bda.last_update_time()``` is necessarily equal to ```most_recent_time``` so you should not need to get it here.

mulkieran: Challenge problem: can this entire for-loop be made into a single expression?

src/engine/strat_engine/pool.rs
```diff
@@ -191,9 +192,10 @@ impl StratPool {
     pub fn setup(
         uuid: PoolUuid,
         devnodes: &HashMap<Device, PathBuf>,
+        timestamp: DateTime<Utc>,
         metadata: &PoolSave,
     ) -> StratisResult<(Name, StratPool)> {
-        let mut backstore = Backstore::setup(uuid, &metadata.backstore, devnodes, None)?;
+        let mut backstore = Backstore::setup(uuid, &metadata.backstore, devnodes, Some(timestamp))?;
```
mulkieran: I think that it is possible to make ```Backstore::setup```'s timestamp parameter not an ```Option```. The argument is that ```setup``` takes a ```BackstoreSave``` argument, which is a component of the variable length metadata. Since that is not an ```Option```, then the variable length metadata must have been written, hence have a timestamp, so the timestamp is not an ```Option``` here.

jbaublitz: Let me take a look. I just was hesitant in case it was initialized with `None` elsewhere in the case of initialization of metadata on a device. If that's the only place it's used, I'll remove the `Option` here.

jbaublitz: Interesting. That does seem to be the only place we have to worry about that call. I'll remove the `Option`.

jbaublitz:
I've made a few changes:
* I changed `Option<DateTime<Utc>>` to `DateTime<Utc>` but only in `Backstore::setup()`. This was based on what I saw in comments and code as the timestamp is passed to `BlockDevMgr::new()`. `BlockDevMgr::new()` will still take `Option<DateTime<Utc>>` as a parameter as it appears that there is no restriction on `BlockDevMgr::new()` being called on only blockdevs with existing metadata.
* I eliminated the for loop to retrieve `PoolSave` and it is now all one expression.
* I now return `most_recent_time` from `get_metadata` instead of re-fetching the latest metadata time from `bda`.

Does this look good?

src/engine/strat_engine/backstore/setup.rs
```diff
@@ -93,21 +94,25 @@ pub fn get_metadata(
...
+                    .open(devnode)
+                    .ok()
+                    .and_then(|mut f| bda.load_state(&mut f).ok())
+                    .and_then(|opt| opt)
+                    .and_then(|data| serde_json::from_slice(&data).ok())
+            } else {
+                None
+            }
+        })
+        .next();
```
mulkieran: Could you consolidate the rest of the method body on the end of the expression above? I know it's a bit of mission creep, but it'll be a clearer structure.

jbaublitz: Change pushed. Lemme know if that looks good.

mulkieran:
A clippy error on Travis...

src/engine/strat_engine/backstore/setup.rs
```diff
@@ -109,17 +108,12 @@ pub fn get_metadata(
...
-        return Ok(Some((*most_recent_time, psave)));
-    }
-
-    // If no data has yet returned, we have an error. That is, we should have
-    // some metadata, because we have a most recent time, but we failed to
-    // get any.
-    let err_str = "timestamp indicates data was written, but no data successfully read";
-    Err(StratisError::Engine(ErrorEnum::NotFound, err_str.into()))
+        .next()
+        .ok_or(StratisError::Engine(
```
mulkieran: clippy, quite correctly, would like to see an ```ok_or_else``` here. I agree, otherwise looks fine to me.

src/engine/strat_engine/backstore/setup.rs
```diff
@@ -93,28 +94,26 @@ pub fn get_metadata(
...
-    let err_str = "timestamp indicates data was written, but no data successfully read";
-    Err(StratisError::Engine(ErrorEnum::NotFound, err_str.into()))
+    bdas.iter()
+        .filter_map(|&(devnode, ref bda)| {
+            if bda.last_update_time() == Some(most_recent_time) {
+                OpenOptions::new()
+                    .read(true)
+                    .open(devnode)
+                    .ok()
+                    .and_then(|mut f| bda.load_state(&mut f).ok())
```
mulkieran: Err, could you do one more thing? Apparently this construction has troubled me since two years ago, PR#444. Change these two lines to the equivalent single line:

```
.and_then(|mut f| bda.load_state(&mut f).unwrap_or(None))
```

jbaublitz:
Superseded by #1595

PR #1591, opened 2019-07-23
Define a type for total blockdev size

mulkieran:
This PR accomplishes the following:

* Every value stored in any struct defined in the metadata that represents a size has a type which identifies its function, not just its units.
* The new type, BlockdevSize, is used in several contexts outside the metadata, but within the internal implementation of the backstore. This helps with code comprehension, by identifying those places in the StratBlockDev or RangeAllocator implementation where a size that is used is exactly the size of an entire blockdev.

It does not attempt to use this type yet in a larger context. The intention is to define types to represent other distinct sizes, like the portion of the blockdev which is not set aside for Stratis metadata, but this is left for future work.

Related: https://github.com/stratis-storage/project/issues/50.

### DIFF
--- a/src/engine/strat_engine/backstore/blockdev.rs
+++ b/src/engine/strat_engine/backstore/blockdev.rs
@@ -159,7 +159,7 @@ impl BlockDev for StratBlockDev {
 
     fn size(&self) -> Sectors {
         let size = self.used.size();
-        assert_eq!(self.bda.dev_size(), size);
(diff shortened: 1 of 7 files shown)


mulkieran:
Need to reboot stratis-tester-1.

drckeefe:
stratis-tester-1 rebooted.

mulkieran:
udev too slow test error, rebuilding.

PR #1593, opened 2019-07-25
Remove RangeAllocator::remove_ranges

mulkieran:
This PR removes the method remove_range from the RangeAllocator implementation. The method has never been used, because Stratis does not de-allocate blocks from a blockdev once they are allocated.

It seems like it's time to remove this method and the accompanying tests because it is likely that it will never be used in its current form, and it interferes with ongoing development which uses types to track the meaning of certain sizes throughout the codebase.


mulkieran:
rebased.

PR #1596, opened 2019-07-31
Add a target to build but not run tests

mulkieran:
Related: https://github.com/stratis-storage/project/issues/43.

PR #1598, opened 2019-07-31
Add a target to build but not run tests

mulkieran:
Related: https://github.com/stratis-storage/project/issues/43.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,10 @@ Once toolchain and other dependencies are in place, run `make build` to build, a
 `stratisd` executable in `./target/x86_64-unknown-linux-gnu/debug` as root.
 Pass the `--help` option for more information on additional developer options.
 
+##### Building tests
+The Makefile provides a target, `build-tests` which allows compiling the
+tests without running any of them, as a convenience to developers.
+
 ##### Secondary dependencies
 The rust library dbus-rs has an external dependency on the C dbus library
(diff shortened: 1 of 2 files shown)


mulkieran:
Supercedes #1596 .

README.md
```diff
@@ -62,6 +62,10 @@ Once toolchain and other dependencies are in place, run `make build` to build, a
 `stratisd` executable in `./target/x86_64-unknown-linux-gnu/debug` as root.
 Pass the `--help` option for more information on additional developer options.
 
+##### Building tests
+The Makefile provides a target, `build-tests` which allows compiling the
+tests without running any of them, as a convenience to developers.
```
GuillaumeGomez: Just wondering: shouldn't it be "as a convenience **for** developers"? If there is a rule/exception about this in english, I'd really like to learn it.

mulkieran: Either is correct. There's probably a very subtle difference in meaning. Something like "convenient to" would make more sense when thinking of a tool that could be grabbed quickly (probably my thought at the time). "convenient for" would emphasize the person and their experience more.

mulkieran:
Whoops. Same error of pulling from wrong PR recurring. This definitely requires investigation now.

README.md
GuillaumeGomez: I see, thanks for the explanation!

mulkieran:
rebased.

PR #1599, opened 2019-07-31
Move BDA::read into StaticHeader::read

### LINKED ISSUE #1582: Move BDA::read out of BDA implementation
It only reads the locations of the sigblocks and the only method that depends on it is ```StaticHeader::setup```.

~~We should probably make it a member of the static_header module, rather than a member of the ```StaticHeader``` impl, also.~~ We'll handle this part of the big move later if it seems indicated. Let's just move it into the ```StaticHeader``` impl, so that it can be with its friend, ```StaticHeader::write```.

GuillaumeGomez:
Fixes #1582.

### DIFF
--- a/src/engine/strat_engine/backstore/metadata/bda.rs
+++ b/src/engine/strat_engine/backstore/metadata/bda.rs
@@ -223,9 +223,15 @@ impl StaticHeader {
         }
     }
 
-    /// Read the BDA from the device and return 2 SECTORS worth of data, one for each BDA returned
-    /// in the order of layout on disk (location 1, location 2).
-    /// Only the BDA sectors are read up from disk, zero areas are *not* read.
+    /// Read the data at both signature block locations.
+    ///
+    /// Return the data from each location as an array of bytes
+    /// or an error if the read fails. The values are returned
+    /// in the same order in which they occur on the device.
+    ///
+    /// Read the contents of each signature block separately,
+    /// as this increases the probability that at least one read
+    /// will not fail.
     fn read<F>(
         f: &mut F,
     ) -> (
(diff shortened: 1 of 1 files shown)


mulkieran:
ok to test.

src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -270,6 +223,53 @@ impl StaticHeader {
         }
     }
 
+    /// Read the BDA from the device and return 2 SECTORS worth of data, one for each BDA returned
```
mulkieran: Please add an additional commit to change the header comment to something like this:

/// Read the data at both signature block locations.
/// Return the data from each location as an array of bytes
/// or an error if the read fails. Return the values
/// in the order in which they occur on the device.
/// Read the contents of each signature block separately,
/// as this increases the probability that at least one read
/// will not fail.

src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -270,6 +223,53 @@ impl StaticHeader {
...
+    fn read<F>(
+        f: &mut F,
+    ) -> (
+        io::Result<[u8; bytes!(static_header_size::SIGBLOCK_SECTORS)]>,
+        io::Result<[u8; bytes!(static_header_size::SIGBLOCK_SECTORS)]>,
+    )
+    where
+        F: Read + Seek,
+    {
+        // Theory of read procedure
```
mulkieran: Please remove this comment, too. The topic is covered in the suggested header comment, and a lot of the nomenclature is wrong in this comment.

src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -270,6 +223,53 @@ impl StaticHeader {
...
+        // We write the BDA in two operations with a sync in between.  The write operation
+        // could fail (loss of power) for either write leaving sector(s) with potentially hard
+        // read errors. It's best to read each of the specific BDA blocks individually, to limit
+        // the probability of hitting a read error on a non-essential sector.
+
+        let mut buf_loc_1 = [0u8; bytes!(static_header_size::SIGBLOCK_SECTORS)];
+        let mut buf_loc_2 = [0u8; bytes!(static_header_size::SIGBLOCK_SECTORS)];
+
+        /// Read a bda sector worth of data at the specified offset into buffer.
```
mulkieran: Might as well omit this header comment as well. It is a bit misleading also.

src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -270,6 +223,53 @@ impl StaticHeader {
...
+        fn read_sector_at_offset<F>(f: &mut F, offset: usize, mut buf: &mut [u8]) -> io::Result<()>
+        where
+            F: Read + Seek,
+        {
+            f.seek(SeekFrom::Start(offset as u64))?;
```
mulkieran: In another separate commit, could you tighten up the body to one expression, as:

```f.seek(SeekFrom::Start(offset as u64)).and_then(|_| f.read_exact(&mut buf))```

GuillaumeGomez:
Updated.

GuillaumeGomez:
Updated.

mulkieran:
Drat. I accidentally pressed the merge button.

GuillaumeGomez:
Removed merge commit and updated.

mulkieran:
Thanks!

mulkieran:
Travis is just being a bit flaky.

PR #1605, opened 2019-08-06
Use black instead of yapf for Python auto-formatting

mulkieran:
Related: https://github.com/stratis-storage/project/issues/35.


mulkieran:
rebased.

PR #1607, opened 2019-08-08
Move bda wipe

### LINKED ISSUE #1584: Move the implementation of BDA::wipe into the static header module
~~I think that ```BDA::wipe``` is invoked from outside the metadata module, so the method itself should be left. A new method should be created inside the ```static_header``` module.~~

Here's the new proposal:
1. Hoist the ```device_identifiers``` method right out of the BDA implementation and into the module top-level. This can be done by defining it, as a public method, in metadata/mod.rs. A lot of callers which currently call ```BDA::device_identifiers``` will need to be updated accordingly. Internal tests which call ```device_identifiers``` should be updated to call ```StaticHeader::setup``` instead.

2. Simultaneously:
  * Add another method in metadata/mod.rs called ```disown_device```. Have this method call the ```StaticHeader::wipe``` method directly.

  * Move ```BDA::wipe``` into the ```StaticHeader``` implementation.

  * Update calls to ```BDA::wipe``` so that they instead call ```disown_device``` wherever necessary (or ```StaticHeader::wipe``` if they are metadata internal testing code).

Note that, after this, all StaticHeader-specific methods should have been moved into the ```StaticHeader``` implementation. So in the next step it should be possible to move the entire implementation into a separate module, one of the main parts of the goal of #1573.

GuillaumeGomez:
Fixes #1584.

I didn't write the functions into the `bda/mod.rs` file because it seemed ugly. However, if you prefer to have them there, I'll make the move.

### DIFF
--- a/src/engine/sim_engine/engine.rs
+++ b/src/engine/sim_engine/engine.rs
@@ -314,4 +314,5 @@ mod tests {
             Ok(RenameAction::NoSource)
         );
     }
+
 }
--- a/src/engine/sim_engine/engine.rs
+++ b/src/engine/sim_engine/engine.rs
@@ -314,5 +314,4 @@ mod tests {
             Ok(RenameAction::NoSource)
         );
     }
-
 }
--- a/src/engine/strat_engine/backstore/setup.rs
+++ b/src/engine/strat_engine/backstore/setup.rs
@@ -19,7 +19,10 @@ use devicemapper::{devnode_to_devno, Device, Sectors};
 use crate::{
     engine::{
         strat_engine::{
-            backstore::{blockdev::StratBlockDev, metadata::BDA, util::get_stratis_block_devices},
+            backstore::{
+                blockdev::StratBlockDev, metadata::device_identifiers, metadata::BDA,
+                util::get_stratis_block_devices,
+            },
             device::blkdev_size,
             serde_structs::{BackstoreSave, BaseBlockDevSave, PoolSave},
         },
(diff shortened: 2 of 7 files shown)


mulkieran:
> I didn't write the functions into the `bda/mod.rs` file because it seemed ugly. However, if you prefer to have them there, I'll make the move.

That position is ok, too. My goal is generally to make those two actions more abstract, and not specifically part of the BDA implementation.

mulkieran:
ok to test

mulkieran:
Looks like this change just created some unused imports...

mulkieran:
@GuillaumeGomez Can you rebase and remove the unused imports?

GuillaumeGomez:
Sure!

src/engine/sim_engine/engine.rs
```diff
@@ -314,5 +314,4 @@ mod tests {
             Ok(RenameAction::NoSource)
         );
     }
-
```
mulkieran: Please remove stray line removal.

GuillaumeGomez: A reason behind undoing this change?

mulkieran: It's the only line you touched in ```sim_engine/engine.rs```, because there were no changes needed for the sim engine. It would be better if the file just wasn't part of the PR at all.

GuillaumeGomez: Ah indeed. Might have changed when looking around...

GuillaumeGomez:
Removed line removal.

src/engine/strat_engine/backstore/setup.rs
```diff
@@ -19,7 +19,10 @@ use devicemapper::{devnode_to_devno, Device, Sectors};
 use crate::{
     engine::{
         strat_engine::{
-            backstore::{blockdev::StratBlockDev, metadata::BDA, util::get_stratis_block_devices},
+            backstore::{
+                blockdev::StratBlockDev, metadata::device_identifiers, metadata::BDA,
```
mulkieran: Please make as nested as possible, like ```metadata::{device_identifiers, BDA}```.

GuillaumeGomez:
Updated.

mulkieran:
@GuillaumeGomez I think you must not have pushed your changes.

PR #1609, opened 2019-08-08
Add macro uuid_to_string

jbaublitz:
#1597 is going to use a new macro, `uuid_to_string!()`. Per requested changes in that PR, I've broken out this macro definition and the use of it where `.to_simple_ref().to_string()` was used previously into a separate PR to be merged before #1597.

### DIFF
--- a/src/dbus_api/macros.rs
+++ b/src/dbus_api/macros.rs
@@ -42,3 +42,10 @@ macro_rules! get_mut_pool {
         }
     };
 }
+
+// Macro for formatting a Uuid object for transport on the D-Bus as a string
+macro_rules! uuid_to_string {
+    ($uuid:expr) => {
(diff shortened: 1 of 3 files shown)


src/dbus_api/macros.rs
```diff
@@ -42,3 +42,9 @@ macro_rules! get_mut_pool {
         }
     };
 }
+
```
mulkieran: I think it'll help if there's a comment here, something like:

```// Format a UUID for transport on the D-Bus.```

Just two slashes, we found out eventually that doc comments on macros don't make sense.

jbaublitz: Rebased and resolved.

jbaublitz:
Rebased and ready for merge.

PR #1625, opened 2019-09-03
Add a bloat target to the Makefile

mulkieran:


mulkieran:
For anybody who wants to satisfy their curiosity about what is adding most to the size of stratisd...

PR #1626, opened 2019-09-04
Add a formatting task for the beta build

mulkieran:


mulkieran:
We do not have to address this until next release, so we might as well just wait and see.

mulkieran:
This will only matter when we update our recommended development compiler.

mulkieran:
Very puzzled by the test failures here.

mulkieran:
Old version of stratisd, maybe?

mulkieran:
> Old version of stratisd, maybe?

We have facilities to prevent that in these tests.

PR #1627, opened 2019-09-16
Add a changelog

mulkieran:


mulkieran:
This initiates a changelog for stratisd. Some choices I made are:
* Name: CHANGES.txt
* Some important specs at the top of the file, which I don't associate with any particular PR.
* Essentially no use of text markup. I think plain text is a good idea, unless there's a required format, like for man pages. If we want a really nice document, there is always TeX -> LaTeX -> LyX. But having to know many different markup formats is really a burden. I already have to try to remember sphinx markup, and that's probably enough.
* All PRs are grouped by theme. The idea is to communicate w/ users, and they probably are happy to skip that spot where we were able to delete an obsolete clippy annotation or get rid of some code that had been dead for 15 months.
* No links to bugzillas. We might as well keep the changelog upstream only.
* No credits to individual contributors. We hope this will all be visible in the git log and on GitHub, and it is a burden and not always quite accurate to include contributors in the changelog.

Issue #1482, opened 2019-04-15
Use abstract name space unix domain socket instead of locking pid file


tasleson:
Today stratisd uses a locking pid file to determine if another instance is already running.  This could be handled cleaner with the use of a unix domain socket that is using abstract name space.

ref. https://blog.petrzemek.net/2017/07/24/ensuring-that-a-linux-program-is-running-at-most-once-by-using-abstract-sockets/

bgurney-rh:
I tried a test where I overwrote the /run/stratisd.pid file with another number, not corresponding to the PID of a running stratisd process.  (I also stopped stratisd.)

```
# echo 3141592653589793 | tee /var/run/stratisd.pid 
3141592653589793
```
After starting a new stratisd process, it starts (probably after determining that there is no stratisd PID with the value in the file), but only seems to overwrite enough bytes to contain the new PID:
```
# while true; do date; cat /var/run/stratisd.pid; sleep 1; done
Mon Jun  3 10:32:28 EDT 2019
3141592653589793
Mon Jun  3 10:32:29 EDT 2019
3141592653589793
Mon Jun  3 10:32:30 EDT 2019
3141592653589793
Mon Jun  3 10:32:32 EDT 2019
3141592653589793
Mon Jun  3 10:32:33 EDT 2019
3141592653589793
Mon Jun  3 10:32:34 EDT 2019
12365
2653589793
Mon Jun  3 10:32:35 EDT 2019
12365
2653589793
Mon Jun  3 10:32:36 EDT 2019
12365
2653589793
```

mulkieran:
It looks like there's a latent bug in the PID reading mechanism as well. It reads the whole string available, but parses using ```s.parse::<pid_t>()```, and the ```pid_t``` type is defined as an i32. So, in the presence of larger PIDs than the standard maximum PID, the value will not be parsed properly.

jbaublitz:
I'm going to put up a preliminary PR that we can take a look at and see if it's something we want to consider adding.

mulkieran:
Great!

jbaublitz:
[...]

PR #1628, opened 2019-09-16
Swap out PID file for Unix socket in abstract namespace

jbaublitz:
This is a prototype for how we could use a Unix socket in the abstract namespace instead of a PID lockfile for preventing multiple stratisd instances from running at once.

### DIFF
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -10,8 +10,6 @@ extern crate log;
 use std::{
     cell::RefCell,
     env,
-    fs::{File, OpenOptions},
-    io::{ErrorKind, Read, Write},
     os::unix::io::{AsRawFd, RawFd},
     path::PathBuf,
     process::exit,
@@ -21,15 +19,15 @@ use std::{
 use chrono::Duration;
 use clap::{App, Arg, ArgMatches};
 use env_logger::Builder;
-use libc::pid_t;
 use log::LevelFilter;
 use nix::{
-    fcntl::{flock, FlockArg},
+    errno::Errno,
     sys::{
         signal::{self, SigSet},
         signalfd::{SfdFlags, SignalFd},
+        socket::{bind, socket, AddressFamily, SockAddr, SockFlag, SockType, UnixAddr},
     },
-    unistd::getpid,
+    unistd::{close, Uid},
 };
 use timerfd::{SetTimeFlags, TimerFd, TimerState};
 use uuid::Uuid;
@@ -47,7 +45,7 @@ use libstratis::{
     stratis::{buff_log, StratisError, StratisResult, VERSION},
 };
 
-const STRATISD_PID_PATH: &str = "/var/run/stratisd.pid";
+const STRATISD_SOCKET_ADDR: &[u8] = b"org.storage.stratis1";
 
 /// Interval at which to have stratisd dump its state
 const DEFAULT_STATE_DUMP_MINUTES: i64 = 10;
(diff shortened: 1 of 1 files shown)


src/bin/stratisd.rs
```diff
@@ -511,19 +499,24 @@ fn main() {
...
-    let lock_file = trylock_pid_file();
-
-    let result = {
-        match lock_file {
-            Err(err) => Err(err),
-            Ok(_) => {
-                let log_handle = initialize_log(matches.is_present("debug"));
-                run(&matches, &log_handle)
-            }
+    let socket = match try_abstract_socket() {
```
mulkieran: You'll want to drop the comment above, if it does not apply. It was necessary to extend the scope of the lock file. otherwise it got closed, and it was possible to start a new stratisd process. This was a problem for us, hence the comment.

I think https://docs.rs/nix/0.15.0/nix/unistd/fn.close.html is relevant here, as it is possible to drop the socket twice.

src/bin/stratisd.rs
```diff
@@ -102,47 +100,37 @@ fn initialize_log(debug: bool) -> buff_log::Handle<env_logger::Logger> {
...
+    )
+    .map_err(|e| {
+        StratisError::Error(format!(
+            "Failed to create socket and test if process is already running: {}",
+            e
+        ))
+    })?;
+    bind(sock_fd, &SockAddr::Unix(match UnixAddr::new_abstract(STRATISD_SOCKET_ADDR) {
+        Ok(ua) => ua,
+        Err(_) => unreachable!("The abstract path for the unix PID lock socket is static and should work - please report this as a bug"),
```
mulkieran: I'ld rather not say unreachable here...this being true is up to kernel developers, not to us.

src/bin/stratisd.rs
mulkieran: But you have a raw fd, so I think that doesn't apply, actually, and you're only dropping it once.

src/bin/stratisd.rs
jbaublitz: I'm curious what you mean here. Are you referring to whether or not the kernel chooses to change the interface? This call in itself is preparing data for the kernel for sure, but it does not interact with the kernel at all. If you're making the argument that the kernel interface could change, I agree that it's possible. I'm open to making it an error if we want to fail without a panic if the kernel interface gets changed and nix decides to validate that too before passing it to the kernel, but it appears the only validation that this call does is on whether the path name is too long for the buffer. If you're concerned that nix might add additional validation, I'm happy to change this.

src/bin/stratisd.rs
jbaublitz: Yes, that message is a bit confusing. `RawFd` does not implement `Drop` so this is only applicable when calling `.as_raw_fd()` in the case of an object such as `File` which does implement `Drop`. I will remove the comment. That was an oversight on my part.

jbaublitz:
> I think that this change is an improvement, in itself. However, I'ld like to find an alternative to commit (2), (which I haven't reviewed at all deeply).

What is your initial reaction for what you'd like to do instead of the second commit? Are you looking to use a library instead? I would also prefer to use a library if I'm able to find one that can help here.

mulkieran:
> > I think that this change is an improvement, in itself. However, I'ld like to find an alternative to commit (2), (which I haven't reviewed at all deeply).
> 
> What is your initial reaction for what you'd like to do instead of the second commit? Are you looking to use a library instead? I would also prefer to use a library if I'm able to find one that can help here.

Yes. We should try to find a library or possibly consider if we are willing to do w/out the pid in the error message.

src/bin/stratisd.rs
mulkieran: I misread the message and thought you meant it was a property of the unix socket path, not of the string that _represents_ the unix socket path.

Using ```unreachable``` makes sense here. But since it really should be unreachable, please make the message more terse, something like: ```STRATISD_SOCKET_ADDR is a valid socket address name".

jbaublitz: Resolved.

jbaublitz:
I've removed the commit that you mentioned and it is now on the branch [display-pid-no-deps](https://github.com/jbaublitz/stratisd/commits/display-pid-no-deps). We can pull it in at a later time if desired.

mulkieran:
Closing, see discussion in related issue: #1482 .

PR #1631, opened 2019-09-19
Change pid file path and improve failure error message

### LINKED ISSUE #1630: Use /run instead of /var/run for PID file
If the /var/run symlink gets removed or /var is in a separate partition that's not loaded prior to stratisd starting, stratisd will fail to start with with:

```
IO error: No such file or directory (os error 2)
```

We should change the pid file to use the path `/run/stratis.pid` instead of `/var/run/stratisd.pid` as that should always be there.

jbaublitz:
Closes #1630.

This commit resolves two problems:
1. stratisd uses the legacy symlink `/var/run` for the location of its pid file. This PR moves the pid file path to `/run` which should always be there. This should resolve issues where `/var` is not yet mounted when stratisd starts and similar problems.
2. The error message on failure to create the pid file is more informative and makes it clear why stratisd is failing and where PID file creation was attempted.

### DIFF
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -11,7 +11,7 @@ use std::{
     cell::RefCell,
     env,
     fs::{File, OpenOptions},
-    io::{ErrorKind, Read, Write},
+    io::{Read, Write},
     os::unix::io::{AsRawFd, RawFd},
     path::PathBuf,
     process::exit,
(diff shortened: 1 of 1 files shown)


mulkieran:
Looks like all Travis tests succeeded, but it never properly responded.

src/bin/stratisd.rs
```diff
@@ -118,7 +118,10 @@ fn trylock_pid_file() -> StratisResult<File> {
                     "Must be running as root in order to start daemon.".to_string(),
                 ));
             }
-            return Err(e.into());
+            return Err(StratisError::Error(format!(
```
mulkieran: Could you just drop the special handling of ```PermissionDenied``` ErrorKind, while you're here? It's not a permanently accurate statement, since requirements might be modified from exactly "running as root" in some future, so it has potential to mislead. And the newly added error should allow users to figure out what went wrong from the representation of the IO error in it.

src/bin/stratisd.rs
```diff
@@ -118,7 +118,10 @@ fn trylock_pid_file() -> StratisResult<File> {
...
+                "Failed to create the stratisd PID file at {}: {}",
```
mulkieran: ```create``` -> ```create or open```?

src/bin/stratisd.rs
jbaublitz: Resolved.

src/bin/stratisd.rs
jbaublitz: Resolved.

PR #1632, opened 2019-09-20
Change pid file path and improve failure error message

### LINKED ISSUE #1630: Use /run instead of /var/run for PID file

jbaublitz:
Closes #1630

### DIFF
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -11,7 +11,7 @@ use std::{
     cell::RefCell,
     env,
     fs::{File, OpenOptions},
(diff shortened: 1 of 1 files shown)


mulkieran:
@jbaublitz Please rebase and we should be able to merge this.

jbaublitz:
Rebased.

Issue #1494, opened 2019-04-26
stratisd fails to start when daemon experiences EIO when trying to read up super blocks


tasleson:
If any block device is experiencing errors, stratisd will exit on startup, potentially preventing other pools and file systems from being brought up.

```
DEBUG libstratis::stratis::buff_log: BuffLogger: pass_through: true hold time: none
 INFO stratisd: Using StratEngine
Engine error: Unable to read data at sigblock locations.
```

Needed to run strace to determine what specifically was happening, error message wasn't helpful.
```
...
stat("/dev/nbd0", {st_mode=S_IFBLK|0660, st_rdev=makedev(0x2b, 0), ...}) = 0
openat(AT_FDCWD, "/dev/nbd0", O_RDONLY|O_CLOEXEC) = 6
lseek(6, 512, SEEK_SET)                 = 512
read(6, 0x7ffff07a9cf8, 512)            = -1 EIO (Input/output error)
lseek(6, 4608, SEEK_SET)                = 4608
read(6, 0x7ffff07a9ef8, 512)            = -1 EIO (Input/output error)
close(6)                                = 0
close(4)                                = 0
write(2, "Engine error: ", 14Engine error: )          = 14
write(2, "Unable to read data at sigblock "..., 42Unable to read data at sigblock locations.) = 42
write(2, "\n", 1
...
```

mulkieran:
```StratEngine::initialize``` invokes ```backstore::find_all``` -> ```BDA::device_identifiers```. See #1637, for analysis.

PR #1637, opened 2019-09-26
taint

mulkieran:


mulkieran:
This was exploratory work for #1494 which will be resolved by #1736.

PR #1638, opened 2019-09-26
Update issue templates

mulkieran:
An experiment to see how issue templates actually work. It seems like we create them in a .github directory in our own repo.


mulkieran:
I'm not sure that this is worth doing at this time. We can always reconsider.

PR #1643, opened 2019-09-26
Add a test to verify that only one instance of stratisd can be running

### LINKED ISSUE #1644: stratisd has just enough action going on during startup that we need to put in some tests for that
These are not unit tests, and they are not dbus-tests, so we need to figure out how to construct them and where to place them.

mulkieran:
Closes #1644.

### DIFF
--- a/tests/client-dbus/tests/misc/test_startup.py
+++ b/tests/client-dbus/tests/misc/test_startup.py
@@ -1,0 +1,62 @@
+# Copyright 2019 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


mulkieran:
Had to allow an "unnecessary lambda"

tests/client-dbus/tests/misc/test_startup.py
```diff
@@ -0,0 +1,63 @@
...
+        self.first_process = subprocess.Popen(
+            self.command_line,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            close_fds=True,
+            env=os.environ,
+        )
+
+        # pylint: disable=unnecessary-lambda
+        self.addCleanup(lambda: self.first_process.terminate())
```
jbaublitz: Is the reason that this is:
```self.addCleanup(lambda: self.first_process.terminate())```
and not:
```self.addCleanup(self.first_process.terminate)```
just for readability purposes?

mulkieran: No, I think it's a hold over from a previous edit. I should be able to get rid of it. Let me give that a try.

mulkieran:
rebased, and squashed removal of the pylint directive, see comment: https://github.com/stratis-storage/stratisd/pull/1643#pullrequestreview-294430733.

PR #1646, opened 2019-09-27
Multithreading support for structures at the pool level

jbaublitz:
This is a preliminary PR for exploring what providing thread safety at the pool level would look like if we decide to support multithreading in stratisd. I've chosen locking at the pool level because we do not want to allow modification at lower levels of the stack (such as filesystems) at the same time as higher level operations (such as pool operations) as we cannot currently guarantee no corruption if we do allow that, but we should explore if we can provide locking at lower levels as well. We can say that pools as a locking unit does guarantee this consistency. This would allow multiple pools to be operated on independently such as taking a snapshots of multiple filesystems in different pools in parallel. Related to stratis-storage/project#86.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,6 +14,14 @@ version = "0.10.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "arrayvec"
+version = "0.4.11"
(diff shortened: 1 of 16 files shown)


jbaublitz:
Closing as this will need significant updates and the PR is beyond saving. I'll keep the branch so that we can pull out the parts that are still useful.

PR #1653, opened 2019-10-09
Add an inspect target

mulkieran:


mulkieran:
```cargo inspect``` with no arguments seems to produce nothing. Let's take a look at some later time.

PR #1662, opened 2019-10-16
Fetch some DBus properties through DBus method

jbaublitz:
This is a preliminary PR to resolve stratis-storage/project#52.

### DIFF
--- a/src/dbus_api/blockdev.rs
+++ b/src/dbus_api/blockdev.rs
@@ -2,15 +2,18 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+use std::collections::HashMap;
+
 use dbus::{
     self,
-    arg::IterAppend,
+    arg::{Array, IterAppend, RefArg, Variant},
     tree::{
         Access, EmitsChangedSignal, Factory, MTFn, MethodErr, MethodInfo, MethodResult, PropInfo,
+        Tree,
     },
-    Message,
+    Message, Path,
 };
-
+use itertools::Itertools;
 use uuid::Uuid;
 
 use crate::{
@@ -68,18 +71,6 @@ pub fn create_dbus_blockdev<'a>(
         .emits_changed(EmitsChangedSignal::Const)
         .on_get(get_blockdev_initialization_time);
 
-    let total_physical_size_property = f
-        .property::<&str, _>("TotalPhysicalSize", ())
-        .access(Access::Read)
-        .emits_changed(EmitsChangedSignal::False)
-        .on_get(get_blockdev_physical_size);
-
-    let state_property = f
-        .property::<u16, _>(consts::BLOCKDEV_STATE_PROP, ())
-        .access(Access::Read)
-        .emits_changed(EmitsChangedSignal::True)
-        .on_get(get_blockdev_state);
-
     let pool_property = f
         .property::<&dbus::Path, _>("Pool", ())
         .access(Access::Read)
@@ -98,6 +89,25 @@ pub fn create_dbus_blockdev<'a>(
         .emits_changed(EmitsChangedSignal::False)
         .on_get(get_blockdev_tier);
 
+    let get_all_properties_method = f
+        .method("GetAllProperties", (), get_all_properties)
+        // a{s(bv)}: Dictionary of property names to tuples
+        // In the tuple:
+        // b: Indicates whether the property value fetched was successful
+        // v: If b is true, represents the value for the given property
+        //    If b is false, represents the error returned when fetching the property
+        .out_arg(("results", "a{s(bv)}"));
+
+    let get_properties_method = f
+        .method("GetProperties", (), get_properties)
+        .in_arg(("properties", "as"))
+        // a{s(bv)}: Dictionary of property names to tuples
+        // In the tuple:
+        // b: Indicates whether the property value fetched was successful
+        // v: If b is true, represents the value for the given property
+        //    If b is false, represents the error returned when fetching the property
+        .out_arg(("results", "a{s(bv)}"));
+
     let object_name = make_object_path(dbus_context);
 
     let object_path = f
@@ -109,12 +119,15 @@ pub fn create_dbus_blockdev<'a>(
                 .add_p(devnode_property)
                 .add_p(hardware_info_property)
                 .add_p(initialization_time_property)
-                .add_p(total_physical_size_property)
                 .add_p(pool_property)
-                .add_p(state_property)
                 .add_p(tier_property)
                 .add_p(user_info_property)
                 .add_p(uuid_property),
+        )
+        .add(
+            f.interface(consts::PROPERTY_FETCH_INTERFACE_NAME, ())
+                .add_m(get_all_properties_method)
+                .add_m(get_properties_method),
         );
 
     let path = object_path.get_name().to_owned();
@@ -177,23 +190,77 @@ fn set_user_info(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
     Ok(vec![msg])
 }
 
-/// Get a blockdev property and place it on the D-Bus. The property is
-/// found by means of the getter method which takes a reference to a
-/// blockdev and obtains the property from the blockdev.
-fn get_blockdev_property<F, R>(
-    i: &mut IterAppend,
-    p: &PropInfo<MTFn<TData>, TData>,
-    getter: F,
-) -> Result<(), MethodErr>
+fn get_all_properties(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
+    let message: &Message = m.msg;
+    let object_path = &m.path;
+
+    let return_message = message.method_return();
+
+    let mut return_value: HashMap<String, (bool, Variant<Box<dyn RefArg>>)> = HashMap::new();
+
+    let bd_size_result = blockdev_operation(m.tree, object_path.get_name(), |_, bd| Ok(*bd.size()));
+    let (bd_size_success, bd_size_prop) = match bd_size_result {
+        Ok(bd_size) => (true, Variant(Box::new(bd_size) as Box<dyn RefArg>)),
+        Err(e) => (
+            false,
+            Variant(Box::new(format!("{}: {}", e.errorname(), e.description())) as Box<dyn RefArg>),
+        ),
+    };
+
+    return_value.insert(
+        "TotalPhysicalSize".to_string(),
+        (bd_size_success, bd_size_prop),
+    );
+
+    Ok(vec![return_message.append1(return_value)])
+}
+
+fn get_properties(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
+    let message: &Message = m.msg;
+    let mut iter = message.iter_init();
+    let object_path = &m.path;
+    let properties: Array<String, _> = get_next_arg(&mut iter, 0)?;
+
+    let return_message = message.method_return();
+
+    let return_value: HashMap<String, (bool, Variant<Box<dyn RefArg>>)> = properties
+        .unique()
+        .filter_map(|prop| match prop.as_str() {
+            "TotalPhysicalSize" => {
+                let bd_size_result =
+                    blockdev_operation(m.tree, object_path.get_name(), |_, bd| Ok(*bd.size()));
+                let (bd_size_success, bd_size_prop) = match bd_size_result {
+                    Ok(bd_size) => (true, Variant(Box::new(bd_size) as Box<dyn RefArg>)),
+                    Err(e) => (
+                        false,
+                        Variant(Box::new(format!("{}: {}", e.errorname(), e.description()))
+                            as Box<dyn RefArg>),
+                    ),
+                };
+
+                Some((prop, (bd_size_success, bd_size_prop)))
+            }
+            _ => None,
+        })
+        .collect();
+
+    Ok(vec![return_message.append1(return_value)])
+}
+
+/// Perform an operation on a `BlockDev` object for a given
+/// DBus implicit argument that is a block device
+fn blockdev_operation<F, R>(
+    tree: &Tree<MTFn<TData>, TData>,
+    object_path: &Path<'static>,
+    closure: F,
+) -> Result<R, MethodErr>
 where
     F: Fn(BlockDevTier, &dyn BlockDev) -> Result<R, MethodErr>,
     R: dbus::arg::Append,
 {
-    let dbus_context = p.tree.get_data();
-    let object_path = p.path.get_name();
+    let dbus_context = tree.get_data();
 
-    let blockdev_path = p
-        .tree
+    let blockdev_path = tree
         .get(object_path)
         .expect("tree must contain implicit argument");
 
@@ -202,7 +269,7 @@ where
         .as_ref()
         .ok_or_else(|| MethodErr::failed(&format!("no data for object path {}", object_path)))?;
 
-    let pool_path = p.tree.get(&blockdev_data.parent).ok_or_else(|| {
+    let pool_path = tree.get(&blockdev_data.parent).ok_or_else(|| {
         MethodErr::failed(&format!(
             "no path for parent object path {}",
             &blockdev_data.parent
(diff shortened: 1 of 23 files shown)


src/dbus_api/pool.rs
```diff
@@ -485,41 +490,21 @@ pub fn create_dbus_pool<'a>(
...
-        .on_get(get_pool_extend_state);
-
-    let space_state_property = f
-        .property::<u16, _>(consts::POOL_SPACE_STATE_PROP, ())
-        .access(Access::Read)
-        .emits_changed(EmitsChangedSignal::True)
-        .on_get(get_space_state);
+    let get_all_properties_method = f
+        .method("GetAllProperties", (), get_all_properties)
+        // b: true if filesystems were created
```
mulkieran: Copy/paste errors in comments and return type which I'm ignoring for now.

src/dbus_api/pool.rs
```diff
@@ -360,48 +360,53 @@ fn get_pool_name(i: &mut IterAppend, p: &PropInfo<MTFn<TData>, TData>) -> Result
...
-        Ok(format!("{}", *p.total_physical_size()))
-    })
-}
+    let uuid = get_data!(object_path; default_return; return_message).uuid;
+    let (_, pool) = get_pool!(engine; uuid; default_return; return_message);
 
-fn get_pool_state(i: &mut IterAppend, p: &PropInfo<MTFn<TData>, TData>) -> Result<(), MethodErr> {
-    get_pool_property(i, p, |(_, _, pool)| Ok(pool.state() as u16))
-}
+    let (tpu_success, total_physical_used) = match pool.total_physical_used() {
```
mulkieran: We'll drop total physical used from the stats, as it is ill-defined.

src/dbus_api/pool.rs
```diff
@@ -360,48 +360,53 @@ fn get_pool_name(i: &mut IterAppend, p: &PropInfo<MTFn<TData>, TData>) -> Result
...
-    p: &PropInfo<MTFn<TData>, TData>,
-) -> Result<(), MethodErr> {
-    get_pool_property(i, p, |(_, _, pool)| Ok(pool.extend_state() as u16))
-}
+    let mut return_value: HashMap<String, (bool, Variant<Box<dyn RefArg>>)> = HashMap::new();
+    return_value.insert(
+        consts::POOL_TOTAL_SIZE_PROP.to_string(),
+        (
+            true,
+            Variant(Box::new(format!("{}", pool.total_physical_size()))),
```
mulkieran: Use ```to_string()``` here.

src/dbus_api/pool.rs
```diff
@@ -360,48 +360,53 @@ fn get_pool_name(i: &mut IterAppend, p: &PropInfo<MTFn<TData>, TData>) -> Result
...
+        ),
+    );
+    return_value.insert(
+        consts::POOL_TOTAL_USED_PROP.to_string(),
+        (tpu_success, Variant(Box::new(total_physical_used))),
+    );
+    return_value.insert(
```
mulkieran: I think we will want to drop some or all of these states as well as they relate to internal details that may be changed in future, rendering them meaningless.

src/dbus_api/pool.rs
jbaublitz: Oops! Sorry, misunderstood the issue you assigned me. I'll remove this.

src/dbus_api/pool.rs
jbaublitz: Resolved.

src/dbus_api/pool.rs
jbaublitz: Resolved.

src/dbus_api/pool.rs
jbaublitz: Resolved.

jbaublitz:
The failure in stratisd-nonrust is due to the removal of properties in this PR. I'll fix those up.

jbaublitz:
Now the udev tests are failing. It's unclear to me whether or not this is related to my PR. My guess is that it is due to the failures just having started. However, I don't see anywhere that the udev tests depend on the properties that I removed.

jbaublitz:
This may actually not be due to this PR. See [here](https://github.com/stratis-storage/stratisd/pull/1664#issuecomment-543196897).

mulkieran:
Ok, one possibility for the table type is ```String -> (bool, (bool, var))```. If the outermost bool is false, then that means "unrecognized key", and everything else can be ignored. The code ought to easily supply a default value. If the outer bool is true, but the inner is false, that means value could not be obtained. In that case, since the type is variant, we can freely put whatever we want as the value. Currently, let us put a default value, for simplicity, but in future we could put an error string or some other useful information, and call that an enhancement. Finallly, if both bools are true, we've got the value and we return that.

I think this would make it fairly easy for a client to consume. In the CLI something like:

```
(key_exists, (value_exists, value)) = map[key]
if !key_exists:
     # raise some exception, we have a bug OR
     # do fallback, we have a too-early version of stratisd OR ...
if !value_exists:
     # do reasonable action. If printing out, perhaps substitute something like "<unkown>"
else:
    # use the actual value as appropriate
```

That seems workable to me, and better than anything else I have come up with so far.

jbaublitz:
I've been thinking about it too this morning, and this suggestion seems far more reasonable and expressive than anything I came up with. I'll go with that for now.

src/dbus_api/pool.rs
```diff
@@ -360,48 +360,75 @@ fn get_pool_name(i: &mut IterAppend, p: &PropInfo<MTFn<TData>, TData>) -> Result
...
+    let dbus_context = m.tree.get_data();
+    let engine = dbus_context.engine.borrow();
+    let properties: Array<String, _> = get_next_arg(&mut iter, 0)?;
+
+    let default_return: HashMap<String, PropertyReturn> = HashMap::new();
+    let return_message = message.method_return();
+
+    let uuid = get_data!(object_path; default_return; return_message).uuid;
+    let (_, pool) = get_pool!(engine; uuid; default_return; return_message);
+
```
mulkieran: Please turn the array of properties into a set before filling the map. My goal is just to prevent some frivolously obnoxious person from requesting the same expensive value to the limit of the size of the D-Bus array. If there's a way when filling the HashMap to do this as well, so that a request for a value can be guaranteed to be satisfied exactly once, that's great to.

jbaublitz:
@mulkieran I've been poking around more and I have a prototype for reusing existing code across properties and methods to perform operations on a filesystem object. Take a look and see whether this seems reasonable. I made a few tweaks to the error type of the closure so that we can adequately capture the error message and return it as a `Variant` value in `get_all_properties` as opposed to bubbling up a `MethodErr`.

jbaublitz:
Fixing copy/paste error.

src/dbus_api/pool.rs
jbaublitz: I'm a little confused about the benefit of this. This still doesn't stop the user from supplying a bunch of random values which could still cause problems. Is the rationale that this would at least prevent the engine from running the code to fetch actual properties multiple times?

mulkieran: Thinking further, I don't think it's necessary to return any entry for an unrecognized key. Let the fact that there is no entry for the key in the table be the indicator that the key was unrecognized.

Then the client does:
```
result = map.get(key)
if result is None:
   # raise some exception, we have a bug OR ...
(value_exists, value) = result
if !value_exists:
   # do reasonable action
else:
  # got the value, so use it
```

That makes the type of the table simpler, guarantees that it is no larger than the size of the set of keys that the daemon does recognize, and makes the type of the table the same as that for ```get_all_properties```.

Regarding your previous question, 2^20 keys will require only the computation necessary to realize that they are unrecognized, and in my new formulation won't lead to a very large table, which is also good. But, if the same recognized key is requested 2^20 times, I'ld like to avoid computing the value for that key 2^20 times, and just compute it once.

mulkieran:
> @mulkieran I've been poking around more and I have a prototype for reusing existing code across properties and methods to perform operations on a filesystem object. Take a look and see whether this seems reasonable. I made a few tweaks to the error type of the closure so that we can adequately capture the error message and return it as a `Variant` value in `get_all_properties` as opposed to bubbling up a `MethodErr`.

@jbaublitz We are trying to avoid having any ```StratisResult``` or ```StratisError``` generated in the D-Bus layer, we are trying to stick to only generating D-Bus errors there. What about making the return type of the ```filesystem_operation``` method just a ```Result```?

tests/client-dbus/tests/dbus/filesystem/test_properties.py
```diff
@@ -78,9 +79,12 @@ def testProps(self):
...
         self.assertEqual(len(created), 20)
 
-        used = Filesystem.Properties.Used.Get(filesystem)
+        (used_success, used) = FetchProperties.Methods.GetProperties(
+            filesystem, {"properties": ["Used"]}
+        )["Used"]
 
-        self.assertEqual(used, "12345678")
+        self.assertEqual(used_success, True)
+        self.assertEqual(used, "12345678 bytes")
```
mulkieran: I don't like that the result contains "bytes", we should be sending the bare number over the D-Bus. To fix this, you have to dereference the value, so it isn't Bytes but just a number and _then_ stringify it. We should make sure to do that consistently everywhere.

jbaublitz: In that case, I'd argue it should be an integer instead of a string. If we're not including units, I don't see a reason to encode it as a string.

mulkieran: The reason it was a string was that the integers only go up to u64. With dedup and compression certain sizes that we report could be ridiculously large, larger than would fit in a 64 bit integer. So it is safer long-term to use the string.

jbaublitz: Resolved.

src/dbus_api/pool.rs
jbaublitz: Resolved using `itertools`.

jbaublitz:
@mulkieran Everything you've highlighted is resolved and I've pushed the `BlockDev` implementation for the properties. Like with pool, I've removed the `BlockDev` state property and moved size over to `FetchProperties`. This should be ready for final review.

jbaublitz:
Whoops! Resolved.

src/dbus_api/event_handler.rs
```diff
@@ -63,23 +63,6 @@ impl EngineListener for EventHandler {
                     });
                 }
             }
-            EngineEvent::PoolExtendStateChanged { dbus_path, state } => {
```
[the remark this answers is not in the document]
jbaublitz: Resolved.

src/dbus_api/blockdev.rs
```diff
@@ -177,52 +190,114 @@ fn set_user_info(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
...
-) -> Result<(), MethodErr>
+fn get_all_properties(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
+    let message: &Message = m.msg;
+    let object_path = &m.path;
+
+    let return_message = message.method_return();
+
+    let mut return_value: HashMap<String, (bool, Variant<Box<dyn RefArg>>)> = HashMap::new();
+
+    let bd_size_result = blockdev_operation(m.tree, object_path.get_name(), |_, bd| Ok(*bd.size()));
```
mulkieran: Ideally, we would abstract the shared computation between these two methods into a single computation. The only difference in behavior is that the properties are unknown in the second, but known in the first. So, maybe it would be possible to come up w/ something where the properties get set up in ```get_all_properties``` but just used in ```get_properties``` and the rest is shared.

Seems like: ```vec!["TotalPhysicalSize"].iter()``` would get you to almost the same state in ```get_all_properties``` as ```properties.unique()``` gets you in ```get_properties```, and I bet ```filter_map``` isn't terribly expensive.

src/dbus_api/blockdev.rs
```diff
@@ -177,52 +190,114 @@ fn set_user_info(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
...
+fn get_blockdev_property<F, R>(
+    i: &mut IterAppend,
+    p: &PropInfo<MTFn<TData>, TData>,
+    getter: F,
+) -> Result<(), MethodErr>
+where
+    F: Fn(BlockDevTier, &dyn BlockDev) -> Result<R, String>,
+    R: dbus::arg::Append,
+{
+    let object_path = p.path.get_name();
```
mulkieran: Minor: I like to have as few identifiers as possible, so I avoid them if they're only used once; here you could remove the ```object_path``` identifier altogether and probably get nicer indentation.

NOTE: Applies to ```filesystem.rs``` and ```pool.rs``` as well.

src/dbus_api/pool.rs
```diff
@@ -6,14 +6,15 @@ use std::{collections::HashMap, path::Path, vec::Vec};
...
     self,
-    arg::{Array, IterAppend},
+    arg::{Array, IterAppend, RefArg, Variant},
     tree::{
         Access, EmitsChangedSignal, Factory, MTFn, MethodErr, MethodInfo, MethodResult, PropInfo,
+        Tree,
     },
     Message,
 };
-
```
mulkieran: This looks like a stray line add, I hope its not rustfmt again.

src/engine/engine.rs
```diff
@@ -16,8 +16,7 @@ use devicemapper::{Bytes, Device, Sectors};
 use crate::{
     engine::types::{
         BlockDevState, BlockDevTier, CreateAction, DeleteAction, DevUuid, FilesystemUuid,
-        FreeSpaceState, MaybeDbusPath, Name, PoolExtendState, PoolState, PoolUuid, RenameAction,
-        SetCreateAction, SetDeleteAction,
+        MaybeDbusPath, Name, PoolUuid, RenameAction, SetCreateAction, SetDeleteAction,
```
mulkieran: You should be able to remove ```BlockDevState``` and ```BlockDev::state```, I think.

src/engine/strat_engine/tests/real.rs
```diff
@@ -181,7 +181,7 @@ fn make_linear_test_dev(devnode: &Path, start: Sectors, length: Sectors) -> Line
         get_dm(),
         DmName::new(&format!("stratis_test_{}", Uuid::new_v4())).expect("valid format"),
         None,
-        table.clone(),
```
mulkieran: I think this is a stray edit, although it must be correct. Can you take it out of this PR and push it to a PR for November?

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -621,23 +615,13 @@ impl ThinPool {
         } else if meta_extend_failed {
             new_state = PoolExtendState::MetaFailed;
         }
-        if self.extend_state() != new_state {
+        if self.pool_extend_state != new_state {
```
mulkieran: Please condense this to ```self.pool_extend_state = new_state``` as you did w/, e.g., ```free_space_state```.

src/engine/strat_engine/tests/real.rs
jbaublitz: I added this due to accidentally using nightly for clippy and finding out we'll get an error soon. Do you still want me to separate this out?

src/dbus_api/pool.rs
jbaublitz: This may be due to a nightly rustfmt, potentially the same as the `.clone()` change I made.

src/engine/engine.rs

src/dbus_api/blockdev.rs
jbaublitz: Resolved.

src/dbus_api/pool.rs
jbaublitz: Should I change this back?

jbaublitz:
I found one lingering issue of `.append3()` instead of `.append1()` in the return value of `.get_all_properties()` for pool. I'm fixing that now.

src/engine/strat_engine/tests/real.rs
mulkieran: Go clippy! Yes please, toss it into November.

src/dbus_api/pool.rs
mulkieran: Yes, please do.

src/engine/strat_engine/thinpool/thinpool.rs
mulkieran: Exactly!

src/dbus_api/pool.rs
jbaublitz: Oh! I just looked at this more closely. The grouping of imports was inconsistent here so I was grouping all of the crate imports as one unit. I'll change it back if you don't like that.

src/dbus_api/blockdev.rs
jbaublitz: Resolved.

src/engine/strat_engine/tests/real.rs
jbaublitz: Reverted.

src/engine/strat_engine/thinpool/thinpool.rs
jbaublitz: Resolved.

jbaublitz:
Shared code has now been collapsed. I've avoiding collecting before passing it to `get_properties_shared` methods as we don't want to iterate through before calling `.unique()`. That feels like a waste.

src/dbus_api/pool.rs
```diff
@@ -6,14 +6,15 @@ use std::{collections::HashMap, path::Path, vec::Vec};
...
 use devicemapper::Sectors;
+use itertools::Itertools;
```
mulkieran: We have a special ordering for imports, first from standard library, then external crates furnished by others, then external crates furnished by ourselves, then local imports. So itertools import should be above devicemapper import.

Issue #1659, opened 2019-10-15
Fix and extend device discovery mechanism


mulkieran:

mulkieran:
Moving this to April, we're focusing on #1782 only right now.

PR #1666, opened 2019-10-21
Remove ext3 use from tests

mulkieran:
Related: #1659.


mulkieran:
There was no justification in the original PR that introduced the use of ext3 in the tests. Later, these same tests were having spurious failures, so the wipefs command was inserted into the method that invoked mkfs.ext3. I'ld like to get rid of all of that, if possible. Also, note that now the code matches the test header comments, which claim that XFS is being used.

mulkieran:
Whoops, sorry. I have to fix the test a bit more.

mulkieran:
I update ```test_other_ownership``` so that string match on "ext3" became a string match on "xfs".

mulkieran:
rebased.

PR #1668, opened 2019-10-21
Fix a comment

mulkieran:


mulkieran:
This is so trivial that I plan to just merge it when it passes

PR #1669, opened 2019-10-22
Import StratBlockDev from a sibling module wherever possible

mulkieran:
Related: #1659.


mulkieran:
So trivial I'll merge it as soon as it passes.

PR #1671, opened 2019-10-23
Fix a TODO

mulkieran:


mulkieran:
So trivial that I'll just be merging this once it passes.

PR #1672, opened 2019-10-23
Refactor udev- and other-based device identification mechanism

mulkieran:
Related: https://github.com/stratis-storage/stratisd/issues/1659.

Contributions of this PR:
* Combines different implementations of the same functionality into a single implementation. Previously, udev properties were obtained in one of two ways, either by finding a udev device, making a hash table of all udev property keys and values (decoded lossily) belonging to that device and later inspecting the contents of the hash table OR by directly reading and interpreting the properties when in possession of the device. The current approach uses exclusively the second way and all code that built or searched a hashtable of udev properties has been removed.
* Defines some clearly documented functions which encode interpretations of udev properties. These are is_stratis, is_multipath_member, and is_unclaimed. Previously, the expressions that encoded these interpretations were embedded in larger functions in multiple places, and hence their meaning was undocumented and their code was duplicated.
* Separates udev-based device determination from device determination based on other sources by introducing a UdevOwnership enum. This is an obvious design improvement but will also help to address some existing concerns where Stratis may overwrite a device that is actually owned, but that udev hasn't determined to be owned.
* Defines a general functiion, get_udev_property(), for getting a property from a udev device based on its string. The method is complete, and its type is ```Option<Result<String>>```, encoding the fact that the key might not be present in the udev database and also that there might be a failure in decoding the value. Previously, there was no such function. Instead the code relied on a mechanism of converting the entire set of properties belonging to a udev device to hash table, where the values were converted to lossy strings and then reading items from that hash table.
* Documents an ordering constraint in the identification of devices by udev. It seems that a device can be both a multipath member and a Stratis device. If the device is a multipath member, Stratis must ignore it when trying to reconstruct its pools, so the check for multipath membership must be made first, so that the device is not determined to be a Stratis device. I have no independent confirmation of this fact at this time. However, in ```get_stratis_block_devices``` the existing code base filters out multipath members. This would only be necessary if a device could be both a Stratis device and a multipath member device according to udev. This documentation is in the udev method ```device_ownership()```.
* Removes the file src/engine/strat_engine/backstore/util.rs, but adds the file src/engine/strat_engine/udev.rs. All the methods in util.rs have been removed, except ```get_stratis_block_devices``` which has been moved to ```src/engine/strat_engine/backstore/setup.rs```, which was the only place where the method was used. udev.rs contains solely udev-specific code, and since it is not used exclusively in the backstore, it was placed in the ```strat_engine``` module.
* Changes the ```get_stratis_block_devices``` fallback from quadratic to linear in the number of block devices.
* Iterates once instead of twice through udev block devices in ```blockdevmgr::initialize()```. Does this by getting the ID_WWN udev property in the same place that it determines ownership based on udev. This requires changing the return type of internally defined method ```dev_info()``` to indicate that it returns an additional value.
* Gets rid of a TODO by logging in the somewhat unlikely event of a failure to properly decode the udev ```ID_WWN``` value. ```dev_info``` returns an error if the udev device can not be located.
* Makes the error message returned by ```blockdevmgr::initialize()``` in the case that Stratis thinks that the block device is already owned somewhat less misleading.
* Adds a log entry at the info level to indicate if initial reading of the Stratis devices required the more time consuming fallback method as well as an info level message indicating that device discovery has been started.


mulkieran:
rebased.

mulkieran:
Previous stratisd test failure might have been of interested. Added a log message to this PR which should also help to identify the problem if it shows up again.

mulkieran:
The stratisd failure repeats, and seems to occur even though the device discovery mechanism does not use the fallback, but simply enumerates stratis devices using udev. In the first test, there is some preserved metadata that indicates that one device is owned by the blockdevmgr, but it turns out that 0 devices are found.

So, it seems that udev is doing the wrong thing, but only for this test...since many other tests rely on the device discovery mechanism working at least up to some point.

mulkieran:
And this time it passed, very strange. I'm going to make just one more change.

mulkieran:
Assuming the tests pass, this is up for review again.

mulkieran:
Same two tests failed again, so I think I'll do more work with ```find_all``` after all.

mulkieran:
Because the failure is intermiittent, and because the fallback mechanism is not being used, at least as far as can be told from log messages, the best hypothesis is that udev is identifying some, but not all, of the Stratis block devices via the ID_FS_TYPE property.

mulkieran:
Can not get this thing to fail, or to execute the new code that I put in.

mulkieran:
@bgurney-rh I think that stratisd-nonrust machine has an extra device on it.

bgurney-rh:
@mulkieran The stratisd and stratisd-nonrust tests are executing from the same pool of test servers; all of them have 5 test devices configured in test_config.json.

bgurney-rh:
I'm going to retry the stratisd-nonrust test; it started on the same test system as before, but I didn't see any leftover signatures when I last checked on it.

bgurney-rh:
Sorry; it'd help if I ran "sudo blkid" instead of just "blkid".  I found a stray stratis signature on /dev/vdf.

bgurney-rh:
Now it works; it looks like 2 test systems had stray disk signatures.

mulkieran:
Ok. This is back in final review. Note that the addition of one info log message adds about 20 visible log messages in the stratisd tests. I think that the message is justfied to be at the info level, though. It records the occurrence of a significant action which at this time, only transpires when stratisd starts up. So, it makes our test logs a little busier, but would not make user logs nearly so busy.

mulkieran:
Whoops! I need to close this and re-open against develop branch.

PR #1673, opened 2019-10-24
Remove the sudo commands in every Makefile target

mulkieran:
Related: https://github.com/stratis-storage/ci/issues/38.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,8 +66,8 @@ matrix:
           env: TASK=test
         # destructive tests that can be run on Travis
         - rust: stable
-          env: TASK=test-travis
           sudo: required
+          script: sudo PATH=${TRAVIS_HOME}/.cargo/bin:$PATH make -f Makefile test-travis
         # release
         - rust: stable
           env: TASK=release
--- a/Makefile
+++ b/Makefile
@@ -65,13 +65,13 @@ release:
 	RUSTFLAGS="${DENY}" cargo build --release
 
 test-loop:
-	sudo env "PATH=${PATH}" RUSTFLAGS="${DENY}" RUST_BACKTRACE=1 RUST_TEST_THREADS=1 cargo test loop_
+	RUSTFLAGS="${DENY}" RUST_BACKTRACE=1 RUST_TEST_THREADS=1 cargo test loop_
 
 test-real:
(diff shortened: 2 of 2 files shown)


mulkieran:
What does Travis do? Jenkins tests should fail reasonably quickly.

mulkieran:
I cancelled all uninteresting Travis jobs...I just want to see the behavior for test-travis.

mulkieran:
Travis test failed, because of insufficient permissions, so sudoing is required, the ```sudo: required``` annotation at the top just permits it.

mulkieran:
Travis works! You'll have to click through to see the travis-test run. stratisd test works too! This needs more investigation.

mulkieran:
@bgurney-rh, @jbaublitz, @dkeefe: I'm suggesting this as an alternative fix for our running-the-wrong-toolchain problem. But I don't know if it really works. All I know is that the changes in the PR work for Travis, still work for stratisd, and should work for stratisd non-rust (the failure is that spurious udev one).

Are you able to somehow verify which toolchain is being run, or why it is possible to run these tests without sudo on Jenkins and succeed?

I'll rerun the whole Travis test to show that it succeeds.

mulkieran:
Spurious failures on Travis.

mulkieran:
sudo was required on Travis in order to make the loopbacked devices.

jbaublitz:
I don't think this will solve the problem. It seems like a good maintenance PR, but I think we still need to go with Brian's PR to actually solve the path issue in Jenkins.

mulkieran:
@jbaublitz I'm double-checking what toolchain is being run, by running your commit on top of these changes. If it fails, we can be sure that it is running the old (wrong) toolchain belonging to root. If it succeeds then the new (correct) toolchain established by rustup is being run. I think that should make ti certain. (We can ignore stratisd-nonrust, fedora-server-3 needs a reboot, but not just yet).

jbaublitz:
It appears it does solve the problem! It's finding the right toolchain. I've looked into a little bit more and I think I see how this helps. Sorry about that!

bgurney-rh:
The stratisd-nonrust test was failing due to a stray stratisd signature on a test device.  After a re-run, the test passes.

mulkieran:
We should do this in devicemapper as well.

PR #1677, opened 2019-10-28
stratis-storage/project#52 including changes to devicemapper-rs

jbaublitz:
stratis-storage/project#52 including changes to devicemapper-rs to support a `u128` internally for `Bytes`.


mulkieran:
Deferring consideration until November...we'll go with the simple approach that doesn't rely on devicemapper changes, but does a little arithmetic in or near the D-Bus layer.

PR #1682, opened 2019-10-31
Change DBus path stratis1 to stratis2 in preparation for new version

jbaublitz:
Related to stratis-storage/project#103

### DIFF
--- a/org.storage.stratis2.service
+++ b/org.storage.stratis2.service
@@ -1,4 +1,4 @@
 [D-BUS Service]
-Name=org.storage.stratis1
+Name=org.storage.stratis2
 Exec=/usr/libexec/stratisd
 User=root
--- a/src/dbus_api/consts.rs
+++ b/src/dbus_api/consts.rs
@@ -2,19 +2,19 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-pub const STRATIS_BASE_PATH: &str = "/org/storage/stratis1";
-pub const STRATIS_BASE_SERVICE: &str = "org.storage.stratis1";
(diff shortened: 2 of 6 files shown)


jbaublitz:
@bgurney-rh @mulkieran `cli-with-stratisd` test failure is expected until stratisd and stratis-cli match. `stratisd-nonrust` is a little more concerning as I've updated the `.service` file in my PR but the build doesn't seem to be using it.

mulkieran:
I don't think stratisd-nonrust acknowedges the .service file, although I think that blackbox tests will. It's probably something in the test infrastructure itself that is causing the failure.

jbaublitz:
The error message in `stratisd-nonrust` is specifically `dbus.exceptions.DBusException: org.freedesktop.DBus.Error.ServiceUnknown: The name org.storage.stratis2 was not provided by any .service files` so it does appear that this is a problem with service files. I think we're either going to have to merge trusting my local tests or have PRs use the provided `.service` file in the repo. Does that sound reasonable?

jbaublitz:
I've done a little more investigation. Tests fail the same way locally. I'm going to see if I can find a way to fix it local either by updating tests or changing something about the DBus environment. Will report back once I have more information.

jbaublitz:
@mulkieran @bgurney-rh Through #1685, I was able to determine that this is indeed related to a file on the system needing updating, specifically the stratisd DBus policy file located at `/usr/share/dbus-1/system.d/stratis.conf` in Fedora. I have also confirmed that this file is provided by the stratisd package in Fedora so the CI failure not a code issue, it is a packaging issue. When the contents are changed appropriately from `stratis1` to `stratis2`, stratisd is again able to bring up the DBus API.

jbaublitz:
Also, the changes I made locally are available in this PR under the file `stratisd.conf`. I merely applied these changes on my machine and that fixed it.

Issue #1686, opened 2019-11-01
Divide coinciding name action in create_pool implementations into two possibilities


mulkieran:
Preserve the Identity response if the name and the devices are the same, but if the devices are different, respond with an error. I don't think this needs any change in the CLI, because the error will simply be returned in the appropriate fashion on the D-Bus, so I think that this is a stratisd change only.

Note that when encryption options are added, it will be necessary to distinguish and return Identity only when encryption specification coincides as well.

PR #1687, opened 2019-11-01
Error reporting if arguments to CreatePool differ

jbaublitz:
Putting this up for preliminary review. Only contains sim engine work so far.

Related to #1686

### DIFF
--- a/src/engine/macros.rs
+++ b/src/engine/macros.rs
@@ -82,3 +82,45 @@ macro_rules! set_blockdev_user_info {
         }
     };
 }
+
+macro_rules! create_pool_check_num {
+    ($vec:ident, ($is_one:tt, $is_many:tt)) => {{
+        let joined_string = $vec.join(", ");
+        if $vec.len() == 1 {
+            format!($is_one, joined_string)
+        } else if $vec.len() > 1 {
+            format!($is_many, joined_string)
+        } else {
+            String::new()
+        }
+    }};
+}
+
+macro_rules! create_pool_generate_error_string {
+    ($pool_name:ident, $input:ident, $exists:ident) => {
+        format!(
+            "There was a difference in the blockdevs associated with \
+             the existing pool named {} and the input requesting creation \
+             of a pool by the same name{}{}",
+            $pool_name,
+            create_pool_check_num!(
+                $input,
+                (
+                    " - the input requests blockdev {} \
+                     which does not exist in the current pool",
+                    " - the input requests blockdevs {} \
+                     which do not exist in the current pool"
+                )
+            ),
+            create_pool_check_num!(
+                $exists,
+                (
+                    " - the existing pool contains blockdev {} which was \
+                     not requested by the input",
+                    " - the existing pool contains blockdevs {} which were \
+                     not requested by the input"
+                )
+            ),
+        )
+    };
+}
--- a/src/engine/sim_engine/engine.rs
+++ b/src/engine/sim_engine/engine.rs
@@ -15,6 +15,7 @@ use devicemapper::Device;
 use crate::{
     engine::{
         engine::{Engine, Eventable, Pool},
+        shared::create_pool_idempotent_or_err,
         sim_engine::{pool::SimPool, randomization::Randomizer},
         structures::Table,
         types::{CreateAction, DeleteAction, Name, PoolUuid, Redundancy, RenameAction},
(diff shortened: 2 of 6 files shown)


mulkieran:
I've only had a chance to look at the test failure so far, and I agree that that's looking good.

src/engine/sim_engine/engine.rs
```diff
@@ -40,7 +47,50 @@ impl Engine for SimEngine {
...
 
         match self.pools.get_by_name(name) {
-            Some(_) => Ok(CreateAction::Identity),
+            Some((_, pool)) => {
+                let input_devices = blockdev_paths
+                    .iter()
+                    .cloned()
+                    .collect::<HashSet<&Path, RandomState>>();
+
+                let blockdev_paths: Vec<_> = pool
```
mulkieran: Re-use of ```blockdev_paths``` is somewhat confusing. Try something like ```existing_devices``` by analogy to ```input_devices``` instead.

mulkieran: If you can fuse construction of the currently used devices set into a single expression, that would be great.

src/engine/sim_engine/engine.rs
```diff
@@ -40,7 +47,50 @@ impl Engine for SimEngine {
         let redundancy = calculate_redundancy!(redundancy);
 
         match self.pools.get_by_name(name) {
-            Some(_) => Ok(CreateAction::Identity),
+            Some((_, pool)) => {
+                let input_devices = blockdev_paths
+                    .iter()
+                    .cloned()
+                    .collect::<HashSet<&Path, RandomState>>();
```
mulkieran: We've always let compiler infer the random state in other contexts, maybe that is possible here.

src/engine/macros.rs
```diff
@@ -82,3 +82,44 @@ macro_rules! set_blockdev_user_info {
         }
     };
 }
+
+macro_rules! create_pool_check_num {
```
mulkieran: I generally prefer to avoid macros, as they are slightly more confusing than source. So, if it were possible to make all these functions rather than macros that would be great. The motivation for making them macros would be to share them between sim and strat engine within the existing infrastructure. But we've used the macros only when constrained to, i.e., when functions would not be possible. It would be fine to start a new module in ```src/engine``` to hold any shared functions.

mulkieran: Looks like we need macros to handle the literal format string, which is not really surprising, so sticking with that.

jbaublitz: I'll make the edit for clarity on the `.join()` calls.

src/engine/sim_engine/engine.rs
jbaublitz: Oops! Missed this.

src/engine/sim_engine/engine.rs
jbaublitz: Okay, it existed as an explicit declaration previously in the sim engine so I'll switch over to inference where I find it and see what happens.

src/engine/sim_engine/engine.rs
jbaublitz: This was a change made due to lifetimes. We have two options here. We can either keep it as is, or we can switch from `Path`s to `PathBuf`s as they are owned. I'll try with `PathBuf`s and see if I can get it into one expression. I also disliked having one variable just for the sake of lifetimes so I'll see what I can do.

src/engine/sim_engine/engine.rs
jbaublitz: One of the `HashSet`s is unable to have `RandomState` removed. The compiler can't infer that type parameter. I'm leaving that one as is.

src/engine/sim_engine/engine.rs
jbaublitz: The switch to `PathBuf` was successful.

src/engine/macros.rs
jbaublitz: Resolved.

src/engine/sim_engine/engine.rs
jbaublitz: Resolved.

src/engine/sim_engine/engine.rs
jbaublitz: Resolved.

jbaublitz: Resolved.

mulkieran:
I'm going to take a stab at an edit on one of the shared.rs functions.

PR #1689, opened 2019-11-04
Update changelog for next release

mulkieran:


mulkieran:
I'm hoping that we will not have to add any more PRs to the changelog before the release...but just in case I've left the "do not merge" label on.

PR #1691, opened 2019-11-04
Update policy to allow non-root list

### LINKED ISSUE #1688: Update stratisd.conf to allow non-root user to use methods on FetchProperties interface
Otherwise, obtaining various properties by means of FetchProperties, and hence calling ```stratis pool list``` or ```stratis filesystem list``` which uses these properties will require root permissions, which it did not preivously.

jbaublitz:
Closes #1688 

This should give just enough permission to perform all methods (all of them simply permit reads) in the `FetchProperties` interface in stratisd.

### DIFF
--- a/stratisd.conf
+++ b/stratisd.conf
@@ -27,30 +27,6 @@
          send_interface="org.freedesktop.DBus.Properties"
          send_member="GetAll"/>
 
-  <allow send_destination="org.storage.stratis2.pool"
-         send_interface="org.freedesktop.DBus.Properties"
-         send_member="Get"/>
-
-  <allow send_destination="org.storage.stratis2.pool"
-         send_interface="org.freedesktop.DBus.Properties"
-         send_member="GetAll"/>
-
-  <allow send_destination="org.storage.stratis2.filesystem"
-         send_interface="org.freedesktop.DBus.Properties"
-         send_member="Get"/>
-
-  <allow send_destination="org.storage.stratis2.filesystem"
-         send_interface="org.freedesktop.DBus.Properties"
-         send_member="GetAll"/>
-
-  <allow send_destination="org.storage.stratis2.blockdev"
-         send_interface="org.freedesktop.DBus.Properties"
-         send_member="Get"/>
-
-  <allow send_destination="org.storage.stratis2.blockdev"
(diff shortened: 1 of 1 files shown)


jbaublitz:
I've verified that this permits all non-root lists available in stratis-cli.

mulkieran:
My reasoning about the extra "allows" in the default policy was a bit ill-formed. Since the CLI gets those properties via ```GetManagedObjects``` accessibility of the properties individually seems irrelevant. But, I think the question of whether those ```allows``` are still necessary is questionable, so I'm going to try a few things.

mulkieran:
@jbaublitz: So, I think the important distinction between ```FetchProperties``` and ```Properties``` interfaces is exactly as we were discussing, and that is the reason that specifying allows for particular interfaces for ```FetchProperties``` does not allow the methods to work for those particular interfaces. It is all about the dispatching. The ```Properties``` interface assumes that the particular object may implement multiple interfaces, so it requires the specific interface name to disambiguate among different interfaces. ```FetchProperties``` is interested in actual facts about the pool, filesystem, etc.,  is less interested in the D-Bus, and does not require an interface name. Hence, using ```send_destination``` with an interface name for ```FetchProperties``` in allows in the configuration file has no effect, but for ```Properties``` it does.

But, it is also clear from testing, that simply specifying the service as the destination in the allows subsumes all specifications that specify interfaces as their ```send_destination```. So I am now reasonably confident that we can remove six lines from this configuration file.

jbaublitz:
@mulkieran Ok, that seems to make sense. Removed.

jbaublitz:
I was getting test failures and want to verify that this is not due to the policy change before I give this the okay.

jbaublitz:
Verified to be okay.

stratisd.conf
```diff
@@ -19,38 +19,6 @@
   <allow send_destination="org.storage.stratis2"
 	 send_interface="org.storage.stratis2.FetchProperties"/>
 
-  <allow send_destination="org.storage.stratis2"
```
mulkieran: I think I wasn't clear...please leave the ones for the service in, just remove all the interface-specific ones.

jbaublitz: Resolved.

PR #1693, opened 2019-11-04
Jbaublitz issue stratisd 1686

mulkieran:


mulkieran:
Nope, that didn't jiggle Travis at all, closing. Travis is getting way too smart about what it thinks are the same PRs but actually aren't.

PR #1694, opened 2019-11-04
Add tests for FetchProperties

### LINKED ISSUE #1681: Run up some tests of our properties obtained by FetchProperties
We should definitely make sure of the types. Python is so forgiving, and the justbytes library is fairly forgiving, too, so the CLI tests are sometimes not going to pick up wrongly typed D-Bus values.

jbaublitz:
Closes #1681


mulkieran:
Due to shortness of time, and because the tests are passing, we'll push this back to November.

PR #1695, opened 2019-11-04
Fix bug in pool creation idempotence

### LINKED ISSUE #1686: Divide coinciding name action in create_pool implementations into two possibilities

jbaublitz:
Closes #1686 
Supersedes #1687

### DIFF
--- a/src/engine/macros.rs
+++ b/src/engine/macros.rs
@@ -82,3 +82,45 @@ macro_rules! set_blockdev_user_info {
         }
     };
 }
+
+macro_rules! create_pool_check_num {
+    ($vec:ident, ($is_one:tt, $is_many:tt)) => {{
+        let joined_string = $vec.join(", ");
+        if $vec.len() == 1 {
+            format!($is_one, joined_string)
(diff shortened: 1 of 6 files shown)


mulkieran:
That cli-with-stratisd is failing is exactly correct. The log for the test shows an example of the error that would be returned if the problem were encountered.

mulkieran:
Travis is having test setup errors, I restarted the whole test.

mulkieran:
Whoops! Wrong button.

mulkieran:
It looks to me like Travis is down and doesn't know it yet.

mulkieran:
We've had this problem before, it's trying to test PR #1687.

PR #1697, opened 2019-11-06
Restore TotalPhysicalUsed property on the D-Bus

mulkieran:
This PR does the following:

* Restores the TotalPhysicalUsed property for a pool via the new FetchProperties interface.
* Slightly modifies how total physical used is calculated for a ThinPool abstraction.
* Smooths out some actions in the D-Bus layer in the FetchProperties implementation so that actions
are more uniform. There is a bit more of this available, to remove duplicate code, I think, but I'm leaving that until later.
* Adds a TODO as a reminder about the brittleness of this approach of calculating total physical size.


mulkieran:
rebased.

PR #1701, opened 2019-11-14
Fix for unnecessary clone as checked by clippy

jbaublitz:
Related: https://github.com/stratis-storage/project/issues/106.
Supersedes #1678

### DIFF
--- a/src/engine/strat_engine/tests/real.rs
+++ b/src/engine/strat_engine/tests/real.rs
@@ -181,7 +181,7 @@ fn make_linear_test_dev(devnode: &Path, start: Sectors, length: Sectors) -> Line
         get_dm(),
         DmName::new(&format!("stratis_test_{}", Uuid::new_v4())).expect("valid format"),
         None,
-        table.clone(),
+        table,
     )
     .unwrap()
 }


mulkieran:
There's a few beta failures in devicemapper-rs as well. I'll go ahead and open an issue.

jbaublitz:
@mulkieran This was discovered by accident (I was using rustup's nightly toolchain thinking I was using stable) so I'm mostly in agreement that nightly should not be used for this purpose, particularly because it appears that the current nightly is not complaining about this problem anymore. I do however think that this is a reasonable thing to fix in this PR because we have an objectively unnecessary clone here even if it was discovered by accident. I'm happy to create a new PR for devicemapper clippy errors and @bgurney-rh already addressed the current clippy error for stratisd in #1704. It seems like this might still be a good thing to fix even if it is just in testing just from a code cleanliness perspective.

jbaublitz:
@mulkieran I actually discovered something interesting. We're not currently running clippy at all on tests currently. The way to do this is by using `cargo clippy --tests`. Beta appears to have this failure still too. We should probably add the `--tests` flag to the beta build for clippy.

jbaublitz:
Just verified that this does not cause a clippy error on stable.

mulkieran:
jbaublitz: We are running clippy on tests; the error that was fixed by https://github.com/stratis-storage/stratisd/pull/1704 masked the error being fixed here. One can only speculate, but since clippy is tied to the compiler, the most likely reason is that the https://github.com/stratis-storage/stratisd/pull/1704 error was discovered by the parser, while this error is deeper, and involves a deeper layer.

jbaublitz:
Interesting, is that included in `--all-targets` and `--all-features`? I'll close stratis-storage/project#107.

mulkieran:
> Interesting, is that included in `--all-targets` and `--all-features`? I'll close [stratis-storage/project#107](https://github.com/stratis-storage/project/issues/107).

Yes, it must be ```--all-targets```. There was a time when we were not running clippy on tests, because for a time you couldn't, but that was fixed in https://github.com/stratis-storage/stratisd/pull/1375. We'ld built up quite some cruft by then.

Issue #1690, opened 2019-11-04
devlinks module is exposed at the engine level and really shouldn't be


mulkieran:
The reason it is so exposed is that it is used to supply a ```filesystem_mount_path``` for a filesystem property called, inappropriately, ```Devnode```. In retrospect, ```FilesystemMountPath``` would have been a better name. The filesystem mount path is simply synthesized from certain string values, so it is available even if the filesystem actually can't be mounted. That's probably Ok, because if it could be mounted, that's where it would be mounted.

We should, if possible, change the property name to something like ```FilesystemMountPath``` instead of something generic like ```Devnode```...there are lots of devnodes representing lots of things.

mulkieran:
It's not clear what the sim engine should really be reporting. It's filesystems can never be mounted, so it would kind of be legitimate for it to report nothing or something odd. But we should not make this value optional, because that would make the D-Bus API unnecessarily complicate merely in order to support the idiosyncracies of the sim engine.

mulkieran:
We may decide that the Devnode property name is "good enough". Maybe it is the case that ```FilesystemMountPath``` has a certain ambiguity which ```Devnode``` avoids. Most likely pushing to November and turning this into a simple refactoring job.

mulkieran:
We decided to keep the ```Devnode``` property name. This will be a mildly tricky refactoring job, pushing out to December.

PR #1724, opened 2019-12-03
Move devlinks module down into the strat_engine module

mulkieran:
Resolves: #1690.

Creates a new Filesystem trait method, path_to_mount_filesystem, and uses that method to get the necessary path. Previously, the method used was one that belonged to the devlinks module, and so the devlinks module was hoisted up to the engine level so that it could be used by the D-Bus layer. This broke encapsulation, because the D-Bus layer directly computed the path and because it used a method from the devlinks module, which properly belongs only to the strat engine.

Supply a bogus path for the Sim engine Filesystem implementation. These filesystems are imaginary, so can not be mounted anyway.


src/engine/sim_engine/filesystem.rs
```diff
@@ -43,6 +43,10 @@ impl Filesystem for SimFilesystem {
         self.created
     }
 
+    fn path_to_mount_filesystem(&self, pool_name: &str, fs_name: &str) -> PathBuf {
+        vec!["/somepath", pool_name, fs_name].iter().collect()
```
bgurney-rh: I'm not sure if this line was the root cause, but I saw this test failure in stratisd-nonrust:
```
tests/dbus/filesystem/test_properties.py::SetNameTestCase::testProps FAILED [  9%]
...
>       self.assertEqual(devnode, "/stratis/deadpool/fs")
E       AssertionError: dbus.String('/somepath/deadpool/fs', variant_level=1) != '/stratis/deadpool/fs'

tests/dbus/filesystem/test_properties.py:83: AssertionError
```

mulkieran: Yes, I'll fix this.

mulkieran:
rebased.

Issue #1727, opened 2019-12-05
Write some test for setup::get_blockdevs to explore how it handles devices that can not be found in its metadata


mulkieran:
We had an error for a while in another PR that is not entirely explained when running ```test_setup```.

The error message was:

```"UUIDs of devices found ({}) did not correspond with UUIDs specified in the metadata for this group of devices ({})"``` where the first set of UUIDs was completely empty and the second contained one element.

The metadata had been cached by the test from the pool built by the test. The theory is that the problem arose when a device hanging around from a previous test was found before the devices actually built by the test and passed to the setup() method which eventually called ```get_blockdevs```. The question is, why wasn't the error discovered earlier when trying to assigned the block device to a data or a cache tier, and the message seen instead:

```                        "Stratis device with device number {}, devnode {}, pool UUID {} and device UUID {} had no record in pool metadata",```

mulkieran:
This problem is recurring in our tests, maybe in a different way, so I'm taking an active interest.

mulkieran:
We have a new hypothesis. One or more devices are added to the backstore. the udev database is slow, and is updated with information about some, but not all the devices. Only a subset of the devices are passed to the backstore's setup() method, which then fails. This happens only intemittently, sometimes udev is fast enough.

After ```test_setup()``` fails the next test too, because udev has finally caught up with the device that was made in the previous test. Now that device is passed to ```blockdevmgr::initialize```. Because the udev info is still on it, but the Stratis header has been wiped, the following message is received:

"thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Engine(Invalid, "Device /dev/vdb appears to be already claimed by another, reason: Udev db says stratis, disk meta says no")', src/libcore/result.rs:1084:5"

This story, also, hangs together very nicely.

mulkieran:
Perhaps this is a legitimate behavior of udev. If we decide that that is the case, then it is our test which must be fixed; by removing the dependency on ```find_all```. That would not be too hard.

mulkieran:
This is obsoleted by the complete rewrite of udev-based device discovery and by the removal of all tests that explicitly use find_all() as well as by numerous improvements in Python-based udev tests.

PR #1731, opened 2019-12-06
Make a test a bit more diagnostic

mulkieran:
Related: #1727


mulkieran:
Hoping this will fail...

mulkieran:
rebased, probably it won't fail, out of spite.

mulkieran:
The intermittent error that we were experiencing hasn't recurred for quite a while. We can revive this PR or similar if it comes back.

PR #1733, opened 2019-12-06
Consolidate find_all and get_stratis_block_devices

mulkieran:
Resolves: #1494.
Related: #1659.
Related: #1727.

This PR consolidates the two methods into a single one. The advantages of this are:
* Devices are only iterated through once, rather than being collected and re-iterated through
* The fallback path is easier to understand
* If the fallback path is followed, only one attempt is made to read the Stratis header from each block device found, which is better than two attempts for each
* There are now some useful warning messages and these messages are appropriate for the context in which they are emitted
* Gets rid of stratis_enumerator method; it made the code a bit busier without any real advantage
* Logs at the warn level instead of just returning with an error from the find_all method if there is an error reading the Stratis header on the standard path

Follows: #1723.

NOTE: If this is approved I'm going to put together a final PR with most or all commits consolidated; together they are too messy.

### DIFF
--- a/src/engine/strat_engine/backstore/setup.rs
+++ b/src/engine/strat_engine/backstore/setup.rs
@@ -117,9 +117,21 @@ pub fn find_all() -> StratisResult<HashMap<PoolUuid, HashMap<Device, PathBuf>>> 
             })
         {
             if let Some(devno) = devnode_to_devno(&devnode)? {
-                if let Some((pool_uuid, _)) =
-                    device_identifiers(&mut OpenOptions::new().read(true).open(&devnode)?)?
-                {
+                if let Some((pool_uuid, _)) = match device_identifiers(
+                    &mut OpenOptions::new().read(true).open(&devnode)?,
+                ) {
+                    Ok(ids) => ids,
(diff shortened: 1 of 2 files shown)


mulkieran:
Interesting test failure on Travis...some effort to read Stratis metadata from device totally even to read the bytes, it looks like.

mulkieran:
Seeing something similar on stratisd-nonrust:

```
 INFO libstratis::engine::strat_engine::backstore::setup: Beginning initial search for Stratis block devices
 INFO libstratis::engine::strat_engine::backstore::setup: Could not identify any Stratis devices by a udev search for devices with ID_FS_TYPE="stratis"; using fallback search mechanism
Engine error: Unable to read data at sigblock locations.
```

mulkieran:
And on stratisd tests, similar error, but in ```test_pool_rename```.

mulkieran:
In retrospect these failures were to be expected; previously the fallback path just ignored errors on trying to read Stratis metadata the first time it attempted it, and ignored the devices. I'll just do the same.

mulkieran:
The intermittent error struck again.

mulkieran:
rebased.

PR #1739, opened 2019-12-10
Tidy up udev-based discovery of existing Stratis devices a bit more

mulkieran:
Regularizes device discovery in the find_all() and the block_evaluate() methods, both of which are used to retrieve block devices in order to build Stratis pools.

Leaves blockdevmgr::initialize() device discovery strictly alone, as it has a different purpose from the other methods.

Related: #1659.

### DIFF
--- a/src/engine/strat_engine/backstore/blockdevmgr.rs
+++ b/src/engine/strat_engine/backstore/blockdevmgr.rs
@@ -547,7 +547,7 @@ mod tests {
         backstore::{find_all, get_metadata},
         cmd,
         tests::{loopbacked, real},
-        udev::{block_device_apply, decide_ownership},
+        udev::block_device_ownership,
     };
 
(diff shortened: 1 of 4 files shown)


mulkieran:
This leaves a mess in ```blockdevmgr::initialize()```. But we have already raised the point the this method is doing something different from what ```find_all()``` and ```block_evaluate``` are doing. Those methods are trying to figure out whether a device already belongs to Stratis or not, ```initialize``` is trying to decide whether or not it can take these devices, including writing out new metadata, etc. So maybe now is the time to consider diverging in ```initialize```.

mulkieran:
rebased.

mulkieran:
Our usual intermittent error, logging confirms that it is none of the things that we warn and log for; so we suspect that some of the devices we want are either uninitialized or lack devnodes.

Issue #1732, opened 2019-12-06
udev-based part of device discovery mechanism silently skips uninitialized devices


mulkieran:
It might be wiser to log these at the warning level if the devices appear to be otherwise significant. I have some information on the meaning of this uninitialized marker:

```
Udevd keeps records in udev database when devices were initialized - this is
the "USEC_INITIALIZED" variable that you can find for each device. If this
variable is not present, it means either:

  -- device has not been processed anytime before (based on udev database
records - so if you erase these records, it's like the device has not
processed before)

  -- the recent udev rule processing for the device has timed out (180s by
default I think) and so it was incomplete and hence we can't trust the records
about this device - we either need to replay the uevent (by triggering the
synthetic uevent) or we need to wait for genuine uevent so the rules are
reevaluated.

When you ask udev about records in its database, e.g. through libudev, there's
the "...is_initialized" function. This simply checks for the existence of the
"USEC_INITIALIZED" variable in the udev database and returns 0/1 based on that.

Note also that the records in udev database are refreshed only at the end of
udev rule processing for that device or after the timeout.
```

mulkieran:
This is all handled in the initial read of devices, i.e., ```find_all```, the question is what to do in the event-based situation.

PR #1747, opened 2019-12-13
Log unusual udev occurrences in find_all

mulkieran:
Related: #1732

### DIFF
--- a/src/engine/strat_engine/backstore/setup.rs
+++ b/src/engine/strat_engine/backstore/setup.rs
@@ -53,7 +53,13 @@ pub fn find_all() -> StratisResult<HashMap<PoolUuid, HashMap<Device, PathBuf>>> 
         enumerator.match_property("ID_FS_TYPE", "stratis")?;
         for devnode in enumerator
             .scan_devices()?
-            .filter(|dev| dev.is_initialized())
+            .filter(|dev| {
+                let initialized = dev.is_initialized();
+                if !initialized {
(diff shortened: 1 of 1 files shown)


src/engine/strat_engine/backstore/setup.rs
```diff
@@ -53,14 +53,25 @@ pub fn find_all() -> StratisResult<HashMap<PoolUuid, HashMap<Device, PathBuf>>>
         enumerator.match_property("ID_FS_TYPE", "stratis")?;
         for devnode in enumerator
             .scan_devices()?
-            .filter(|dev| dev.is_initialized())
+            .filter(|dev| {
+                let initialized = dev.is_initialized();
+                if !initialized {
+                    warn!("Found a udev entry for a device identified as a Stratis device, but udev also identified it as uninitialized, discarded the device from the set of devices to process, for safety");
```
bgurney-rh: The phrase "discarded the device" might not be the best choice here; it could be interpreted as "sending discards to the device", which I don't believe is happening in this context.

bgurney-rh: The phrase "discarded the device" might not be the best choice here; it could be interpreted as "sending discards to the device", which I don't believe is happening in this context.

bgurney-rh: The phrase "discarded the device" might not be the best choice here; it could be interpreted as "sending discards to the device", which I don't believe is happening in this context.

mulkieran: it sounds too active...I'll go with omit.

mulkieran:
It would be so nice if the stratisd test would fail this time...

Issue #1753, opened 2019-12-13
Eliminate all non-libudev error returns from find_all()


mulkieran:
There are still some remaining when opening a file or finding a device number. In every case, we can argue that, since we support all this incomplete pool maitenance that the best thing to do is to ignore the problematic device and keep finding all the devices that can be found. We still allow an error on basic failure to communicate with udev at all, because if that happens, no devices can be found and something is seriously wrong.

PR #1754, opened 2019-12-16
Do not return any errors from find_all() while processing devices

mulkieran:
Resolves: #1753.

Instead, for any possible error, log a warning and omit the device from the devices returned.


mulkieran:
rebased.

mulkieran:
rebased.

mulkieran:
I feel like there is a significantly better way that uses just one internal method, so I'm going to try this again. Moving back to in progress.

mulkieran:
Superceded by #1759 .

PR #1759, opened 2019-12-18
Do not return any errors from find_all() while processing devices

mulkieran:
Resolves: #1753.

Supercedes: #1754.

Instead, for any possible error, log a warning and omit the device from the devices returned.


mulkieran:
Note that this supercedes a previous approved version of this PR. It should be semantically equivalent, but it decomposes the problem in a different way than #1754.

mulkieran:
Sorry, adding one more commit principally for better indentation.

mulkieran:
Ok, not touching it again.

PR #1761, opened 2019-12-20
Make setup::get_metadata more robust and more appropriate for its actual use

mulkieran:
Makes get_metadata more robust by adding useful logging when something strange happens,
and generally improving error messages to be as complete as possible given their context.

Changes the specification of get_metadata so it doesn't just ignore devices that don't seem to belong (because they have the wrong pool UUID) but instead returns an error.


mulkieran:
Superceded by #2059 and previous PRs.

Issue #1737, opened 2019-12-09
If a user specifies a multipath member when adding blockdevs or creating a pool, they should get a clear indication of why what they specified is rejected


mulkieran:
Right now, any multipath member is just classified as type DevOwnership::Other (UdevMembership indicates multipath member device).

PR #1764, opened 2019-12-23
Rewrite device identification and initialization

mulkieran:
Resolves: #1737.

Related: #1659.

Contributions are:

* The initialize + resolve_devices functions are re-assembled and decomposed into two different functions, process_devices, which finds all necessary information about devices, and intialize_devices, which does the initialization. We expect this to allow some improvements to the implementation of idempotency in creating pools and adding devices, currently the code must do a complicated analysis of the results of calling BlockDevMgr::initialize(), with the new interface, implementing the idempotency should be much more simple and durable.
* There are better error messages in many contexts.
* There is no case where two file descriptors are simultaneously in use on the same device.
* The code is less eager to do expensive operations, including opening files.
* resolve_devices, along with its dependency on a devicemapper method is removed entirely and the necessary information is obtained from udev.
* Some tests are tidied up and refined a bit. One new test is added.
* The code is more modular. blockdevmgr.rs used to be the fourth largest module, at 851 lines. Since the split it is just over 500 lines, and devices.rs, which contains the device identification and initialization activities is also just over 500.


mulkieran:
Removed a bunch of already merged commits.

mulkieran:
Should we use a closure to gather up errors in ```decide_ownership``` in order to summarize the error there?

mulkieran:
rebased.

mulkieran:
I guess I'll take out the dependency on matches crate for now.

mulkieran:
matches crate dependency unneeded, but now all tests should pass again.

mulkieran:
Tests are good!

mulkieran:
I am again anticipating test success.

mulkieran:
This is that intermittent test error that completely went away for a while. See #1725 for some discussion.

mulkieran:
Note that I changed my mind, and decided to return an error if udev says Stratis, but the device does not present as a Stratis device via its metadata. The user can always try again.

mulkieran:
@bgurney-rh: Please check stratis-tester-3 for old Stratis metadata.

bgurney-rh:
@mulkieran: stratis-tester-3 looks clean; I don't see anything on the test devices.

mulkieran:
@bgurney-rh: Thanks for looking. It's possible that we will want to reboot.

bgurney-rh:
Just for kicks, I'm trying the stratisd test again, on stratis-tester-3.  I made sure there were no stray headers on the test device before the test run.

bgurney-rh:
The test was successful.

mulkieran:
> The test was successful.

Thanks!

mulkieran:
rebased.

mulkieran:
Urgg, getting these spurious errors w/ find_all() not reading the devices properly, even after a udev_settle. I'm going to harden up the tests a bit more.

mulkieran:
Whoops, a formatting error!

mulkieran:
rebased.

mulkieran:
I pulled the tests that were flakey due to udev behavior out of the Jenkins CI in the last two commits, which are the only change. Skipping tests that are flakey is a bit of a step, though note that they both get run on Travis every time, so they're not _really_ skipped, so I'm re-requesting review.

mulkieran:
I had one more thought...if these tests still have the error then our understanding of what causes the error is wrong.

mulkieran:
If this passes, I'll put it up for review again. I won't leave in the reverted commit, but will just remove them both before merging.

mulkieran:
Travis failed w/ a formatting error.

mulkieran:
Since I got this to pass by using the fallback ```find_all()``` method instead of the regular ```find_all()``` method, I'm putting it back up for review. I think this is probably the better course, since it avoids the possible future problem where these tests on Travis also start to fail.

mulkieran:
Superceded.

PR #1765, opened 2019-12-23
Get the device number from udev in find_all

mulkieran:


mulkieran:
I noticed this discrepancy while working on #1764, but I didn't want to mix this change in, so I've pulled it out into a separate PR.

PR #1767, opened 2020-01-03
Unify device discovery on initial start up and event-driven paths

mulkieran:
The major contributions of this PR are:

* Make device identification via udev on event-driven path constant time, rather than linear in the number of block devices.
* Establish a separate module for device identification and put all code for this, whether used on initial start up or event-driven in this module.
* Unify code for event-driven and initial setup discovery. The overall behavior was more or less the same, but now the shared behavior has shared code and shared logging.
* Move some stray bits of engine actions out of bin/stratisd.rs. This is just generally good design; it also enabled the change from linear to constant time.
* Make all device discovery intended to discover Stratis devices for the purpose of setting up pools
more or less uniform.

Resolves: #1732.
Related: #1659.


mulkieran:
rebased.

mulkieran:
There was a tiny clippy error that was in 1.40, but not in beta. I just fixed it since it was obviously correct. I filed a bug against clippy:
https://github.com/rust-lang/rust-clippy/issues/5014.

PR #1769, opened 2020-01-03
Check block device inclusion in the pool not the engine

mulkieran:


mulkieran:
The check for a device in an already existing pool is already handle in liminal.rs, so this can be closed.

PR #1773, opened 2020-01-07
Add back in to_path_buf() that illustrates clippy regression

mulkieran:


mulkieran:
Expect failure with ```clippy::redundant-clone``` on 1.40, but not on beta.

mulkieran:
We can close this too since the bug it illustrates is fixed.

PR #1779, opened 2020-01-13
Unify pool setup on initial setup and udev-event driven paths

mulkieran:
Related: #1659.

The two main contributions of this PR is that unifies the code for setting up the pool into a single
method rather than leaving it as two blocks of duplicated but not identical code, and that it distinguishes between failures to set up a pool that indicate that something might be wrong, and failures to set up a pool which are actually normal conditions.

It also includes a bunch of minor tidies within the code affected.

It does not include a text representation of every device involved in the error messages logged about failure to set up a pool. Doing this is definitely a convenience during development, but could seem far too verbose in the presence of production pools with possibly many devices and potentially numerous udev events.

I added #1780 as future work, and removed some related code in this PR that did not fully cover the problem described in the issue and wouldn't prevent it.

Also related: https://github.com/stratis-storage/project/issues/29, since the PR removes at least three unnecessary expects.

### DIFF
--- a/src/engine/strat_engine/engine.rs
+++ b/src/engine/strat_engine/engine.rs
@@ -187,8 +187,7 @@ impl StratEngine {
                 let mut devices = self
                     .incomplete_pools
                     .remove(&pool_uuid)
-                    .or_else(|| Some(HashMap::new()))
-                    .expect("We just retrieved or created a HashMap");
+                    .unwrap_or_else(HashMap::new);
                 devices.insert(device, dev_node);
                 match setup_pool(pool_uuid, &devices, &self.pools) {
                     Ok((pool_name, pool)) => {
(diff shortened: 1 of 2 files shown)


mulkieran:
Note that this completes the rehabilitation work on device discovery for setup, but leave all of device identification when claiming a device, i.e., for pool create or add blockdevs or cachedevs still quite undone.

jbaublitz:
This look good to me. The `error!()` invocations that got removed [here](https://github.com/stratis-storage/stratisd/pull/1779/files#diff-b7d6bdd1abd0b2d1c789599f8c165beeL199-L216) are the ones related to #1780, is that correct? If so, it looks like this is good to go on my end.

mulkieran:
> This look good to me. The `error!()` invocations that got removed [here](https://github.com/stratis-storage/stratisd/pull/1779/files#diff-b7d6bdd1abd0b2d1c789599f8c165beeL199-L216) are the ones related to #1780, is that correct? If so, it looks like this is good to go on my end.

Exactly those.

PR #1801, opened 2020-02-05
Expose whether or not the pool has a cache in its interface

mulkieran:

PR #1804, opened 2020-02-07
Refactor FetchProperties interface implementation a bit

mulkieran:
Two tidies to the implementation of the FetchProperties interface.


mulkieran:
Note that this is the first, generic part of #1801, doesn't need to be post-release.

mulkieran:
Whoops! Requires a slight revision.

mulkieran:
It's less baroque now.

mulkieran:
I don't think I'll think of anything else.

mulkieran:
rebased.

mulkieran:
Err, one more thing. That macro should be using fully qualifed names.

mulkieran:
macros now use fully qualified names.

PR #1809, opened 2020-02-11
Rewrite device identification and initialization

mulkieran:
Resolves: #1737.

Related: #1659.

Contributions are:

* The initialize + resolve_devices functions are re-assembled and decomposed into two different functions, process_devices, which finds all necessary information about devices, and intialize_devices, which does the initialization. We expect this to allow some improvements to the implementation of idempotency in creating pools and adding devices, currently the code must do a complicated analysis of the results of calling BlockDevMgr::initialize(), with the new interface, implementing the idempotency should be much more simple and durable.
* There are better error messages in many contexts.
* There is no case where two file descriptors are simultaneously in use on the same device.
* The code is less eager to do expensive operations, including opening files.
* resolve_devices, along with its dependency on a devicemapper method is removed entirely and the necessary information is obtained from udev.
* Some tests are tidied up and refined a bit. One new test is added.
* The code is more modular. blockdevmgr.rs used to be the fourth largest module, at 851 lines. Since the split it is just over 500 lines, and devices.rs, which contains the device identification and initialization activities is also just over 500.

Supercedes #1764.


mulkieran:
rebased.

mulkieran:
rebased.

PR #1815, opened 2020-02-12
Require libcryptsetup bindings

mulkieran:
closes https://github.com/stratis-storage/project/issues/141

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 os: linux
+# use most recent available distro to support installation of crypsetup 2.2.0
+# or above with as little fuss as possible.
+dist: bionic
 
 addons:
     apt:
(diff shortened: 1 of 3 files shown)


mulkieran:
Jenkins may fail, as it is likely not to have the libcryptsetup bindings installed yet.

jbaublitz:
> ~Jenkins may fail, as it is likely not to have the libcryptsetup bindings installed yet.~ Err, the dependencies for the libcryptsetup bindings. Travis may fail also.

We probably will need to get a newer version of libcryptsetup. See [this issue](https://github.com/stratis-storage/libcryptsetup-rs/issues/5) and [this comment](https://github.com/stratis-storage/libcryptsetup-rs/pull/4#issuecomment-585260142).

mulkieran:
With these changes we expect Travis to succeed, but Jenkins still to fail.

mulkieran:
Nope. We need to be more sophisticated.

jbaublitz:
I think this is due to how Travis is parsed. If I recall correctly, Travis dedups keys in hashes so this would make sense that dbus is not found as it's probably using the most recent `apt` definition which only specifies the newer version of libcryptsetup.

mulkieran:
Previous change seems to be working!

mulkieran:
Travis problem is solved.

mulkieran:
rebased.

mulkieran:
rebased.

PR #1819, opened 2020-02-12
Use the fallback method for finding Stratis block devices in all tests

mulkieran:


jbaublitz:
It looks like `udevadm settle` is failing now.

PR #1820, opened 2020-02-12
Remove an entire test that relies on udev

mulkieran:


mulkieran:
@bgurney-rh Maybe there is some left over headers or something on stratis-tester-2.

PR #1817, opened 2020-02-12
Preliminary work on encryption

jbaublitz:
Supersedes #1807

PR #1807, opened 2020-02-11
Preliminary work on encryption

jbaublitz:
Supersedes #1771

PR #1823, opened 2020-02-14
Pull out encryption property from encryption PR

jbaublitz:
Resolves [this comment](https://github.com/stratis-storage/stratisd/pull/1817#issuecomment-586056086)
Related to #1807

### DIFF
--- a/src/dbus_api/pool/pool_2_1/api.rs
+++ b/src/dbus_api/pool/pool_2_1/api.rs
@@ -1,0 +1,20 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use dbus::tree::{Access, EmitsChangedSignal, Factory, MTFn, Property};
+
+use crate::dbus_api::{
+    consts,
+    pool::pool_2_1::{
+        props::get_pool_encrypted,
+    },
+    types::TData,
+};
+
+pub fn encrypted_property(f: &Factory<MTFn<TData>, TData>) -> Property<MTFn<TData>, TData> {
+    f.property::<bool, _>(consts::POOL_ENCRYPTED_PROP, ())
+        .access(Access::Read)
+        .emits_changed(EmitsChangedSignal::True)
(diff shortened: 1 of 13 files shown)


jbaublitz:
Rebased.

jbaublitz:
@bgurney-rh Jenkins CI failure is due to stale state in udev. The mentioned disk does not have stratis metadata but udev still thinks it is a stratis device.

tests/client-dbus/tests/dbus/pool/test_properties.py
```diff
@@ -0,0 +1,56 @@
...
+            {
+                "name": self._POOLNAME,
+                "redundancy": (True, 0),
+                "devices": _DEVICE_STRATEGY(),
+            },
+        )
+
+    def testProps(self):
+        """
+        Test reading some filesystem properties.
```
mulkieran: One copy-paste error, might as well get it now...should be "pool" not filesystem.

src/dbus_api/pool/pool_2_1/api.rs
```diff
@@ -0,0 +1,14 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use dbus::tree::{Access, EmitsChangedSignal, Factory, MTFn, Property};
+
+use crate::dbus_api::{consts, pool::pool_2_1::props::get_pool_encrypted, types::TData};
+
+pub fn encrypted_property(f: &Factory<MTFn<TData>, TData>) -> Property<MTFn<TData>, TData> {
+    f.property::<bool, _>(consts::POOL_ENCRYPTED_PROP, ())
+        .access(Access::Read)
+        .emits_changed(EmitsChangedSignal::True)
```
mulkieran: I'ld like to go with ```EmitChangedSignal::Const```, as opposed to ```True```. Otherwise, to make our code consistent we would have to make a change in ```EventHandler::notify``` to handle possible changes. And I don't think we really expect to be doing encryption of a non-encrypted pool any time soon. If we ever end up doing that, we'ld want to bump the D-Bus version number, but that will be a small effort compared to the other work we would have to do to make that happen.

jbaublitz:
Rebased and PR revisions

PR #1810, opened 2020-02-11
Make engines require at least one blockdev for blockdev operations

jbaublitz:
Blocked by #1807 
Supersedes #1740

PR #1824, opened 2020-02-14
Make engines require at least one blockdev for blockdev operations

### LINKED ISSUE #1304: Need to guard against a situation where cache or data device are constructed from an empty list of devices
This can happen for a number of reasons, but it is certain to lead to bad behavior.

jbaublitz:
Supersedes #1810 
Resolved #1304

### DIFF
--- a/src/engine/sim_engine/pool.rs
+++ b/src/engine/sim_engine/pool.rs
@@ -116,7 +116,7 @@ impl Pool for SimPool {
             return if !self.has_cache() && tier == BlockDevTier::Cache {
                 Err(StratisError::Engine(
                     ErrorEnum::Invalid,
-                    "At least one blockdev path is required to add blockdevs".to_string(),
+                    "At least one blockdev path is required to initialize a cache.".to_string(),
                 ))
             } else {
                 Ok(SetCreateAction::new(vec![]))
--- a/src/engine/sim_engine/pool.rs
+++ b/src/engine/sim_engine/pool.rs
@@ -112,6 +112,12 @@ impl Pool for SimPool {
         paths: &[&Path],
         tier: BlockDevTier,
     ) -> StratisResult<SetCreateAction<DevUuid>> {
+        if paths.is_empty() {
+            return Err(StratisError::Engine(
+                ErrorEnum::Invalid,
+                "At least one blockdev path is required to add blockdevs".to_string(),
+            ));
+        }
         let devices: HashSet<_, RandomState> = HashSet::from_iter(paths);
 
         let device_pairs: Vec<_> = devices
@@ -332,7 +338,11 @@ mod tests {
         let mut engine = SimEngine::default();
         let pool_name = "pool_name";
         let uuid = engine
-            .create_pool(pool_name, &[], None)
+            .create_pool(
+                pool_name,
+                strs_to_paths!(["/dev/one", "/dev/two", "/dev/three"]),
+                None,
+            )
             .unwrap()
             .changed()
             .unwrap();
@@ -351,7 +361,11 @@ mod tests {
         let mut engine = SimEngine::default();
         let pool_name = "pool_name";
         let uuid = engine
-            .create_pool(pool_name, &[], None)
+            .create_pool(
+                pool_name,
+                strs_to_paths!(["/dev/one", "/dev/two", "/dev/three"]),
+                None,
+            )
             .unwrap()
             .changed()
             .unwrap();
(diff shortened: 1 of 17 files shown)


src/engine/strat_engine/pool.rs
```diff
@@ -309,6 +309,12 @@ impl Pool for StratPool {
         paths: &[&Path],
         tier: BlockDevTier,
     ) -> StratisResult<SetCreateAction<DevUuid>> {
+        if paths.is_empty() {
```
mulkieran: The rest of the semantic changes seem right, but this one seems like it's too broad. In the general case, I think the empty list should be viewed as a set that is already contained in the blockdevs already belonging to the tier, whether it is cache or data. The error case should be restricted to where this would kick off making a new cache, and that can now be checked w/ the has_cache() function on ```Pool```. So something like:

```
if paths.is_empty() {
  return if !self.has_cache() && tier == Cache {
    Err(...)  
  } else {
    Ok(Identity)
  }
}

Comment (jbaublitz) [reply src/engine/strat_engine/pool.rs]: I see your point, and I am still concerned that this will be a very confusing thing for people to navigate. From a mathematical perspective, this makes sense, but from a user standpoint, I'm not sure this will be clear to those who are not thinking in terms of set theory. I don't think everyone using Stratis will have that background. If we do decide to go with this option, I think we need to be very clear in the error message about why it's failing because up until this point, we have not introduced any idempotent behavior that is accepted in one context where the action has already been done and fails when the action is requested and could be acted upon. This seems inconsistent with the line we've perhaps accidentally drawn and I'd like to be very clear with the user about why we're making this exception if you're set on this being the way forward.

Comment (mulkieran) [reply src/engine/strat_engine/pool.rs]: > I see your point, and I am still concerned that this will be a very confusing thing for people to navigate. From a mathematical perspective, this makes sense, but from a user standpoint, I'm not sure this will be clear to those who are not thinking in terms of set theory. I don't think everyone using Stratis will have that background. If we do decide to go with this option, I think we need to be very clear in the error message about why it's failing because up until this point, we have not introduced any idempotent behavior that is accepted in one context where the action has already been done and fails when the action is requested and could be acted upon. This seems inconsistent with the line we've perhaps accidentally drawn and I'd like to be very clear with the user about why we're making this exception if you're set on this being the way forward.

I'm not sure about the meaning one of " we have not introduced any dempotent behavior that is accepted in one context where the action has already been done and fails when the action is requested and could be acted upon". But remember, we're not asking users of Stratis to understand set theory, only people who write clients of the D-Bus interface. Defining users of Stratis as people who use the CLI, they'll never see this because the parser will reject commands that specify 0 blockdevs.

**Comment by jbaublitz** [comment]
Taking out of review. The requested changes from @mulkieran mean I need to rework some of the tests to specify the distinction between `add_blockdevs` initializing the cache or just adding to it.

**Comment by jbaublitz** [comment]
Back in review.

File: src/engine/sim_engine/pool.rs [hunk 2]
```diff
@@ -112,6 +112,16 @@ impl Pool for SimPool {
         paths: &[&Path],
         tier: BlockDevTier,
     ) -> StratisResult<SetCreateAction<DevUuid>> {
+        if paths.is_empty() {
+            return if !self.has_cache() && tier == BlockDevTier::Cache {
+                Err(StratisError::Engine(
+                    ErrorEnum::Invalid,
+                    "At least one blockdev path is required to add blockdevs".to_string(),
```
mulkieran: "At least one blockdev path is required to initialize a cache." would be better, and matches pool_create error. This also applies to ```StratEngine::add_blockdevs``` below.

tests/client-dbus/tests/dbus/pool/test_add_cache_devs.py
```diff
@@ -186,22 +90,26 @@ def testSomeDevs(self):
...
+        blockdevs3 = blockdevs(props={"Tier": BlockDevTiers.Cache}).search(
+            managed_objects
+        )
+        self.assertEqual(len(list(blockdevs3)), num_devices_added)
 
-        # The number of datadevs has remained the same
-        blockdevs5 = blockdevs(props={"Pool": pool, "Tier": 0}).search(managed_objects)
-        self.assertEqual(frozenset(op for (op, _) in blockdevs5), self._devpaths)
+        # There must be datadevs belonging to this pool as it was created
+        blockdevs4 = blockdevs(props={"Pool": pool, "Tier": 0}).search(managed_objects)
```
mulkieran: Might as well use your enum type for different tiers here.

mulkieran:
Just some Python formatting and what not

jbaublitz:
Whoops! Missed that. I fixed it locally and will run it by CI on Travis.

PR #1703, opened 2019-11-14
Changes corresponding to u128 changes to Bytes

jbaublitz:
Supersedes #1677 
Related stratis-storage/project#52

PR #1830, opened 2020-02-20
Changes corresponding to u128 changes to Bytes

jbaublitz:
Supersedes #1703 
Related stratis-storage/project#52

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -186,7 +186,7 @@ dependencies = [
 [[package]]
 name = "devicemapper"
 version = "0.28.0"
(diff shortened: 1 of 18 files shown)


jbaublitz:
I'm attaching this as April but you can move it back as desired. The reason I'm bringing this up again is that I looked into the `u128` stabilization issue on Github and it seems they've completed lowering `u128`s when they are not natively supported (see [here](https://github.com/rust-lang/rust/issues/35118)). I'm not sure if you had additional concerns, but I just want to make sure we don't forget about this because of the code we had to introduce to work around this.

mulkieran:
Pushing back to May...too little time too review...the little hack in the D-Bus layer is just too localized to be worrisome.

jbaublitz:
Okay not a problem! Just wanted to make sure this doesn't get lost.

Issue #1656, opened 2019-10-10
Require libblkid 2.32


mulkieran:
This contains the additions which recognize Stratis metadata. If we do this, we can drop the fallback mechanism that exists in the device discovery portion that is intended to cover the situation where libblkid 2.32 is not available.

mulkieran:
Removing the sub-issue from the board, just leaving the issue.

mulkieran:
We'll remove anything that exists solely to support lower versions of libblkid, document it in the changelog and elsewhere, and expect downstream to package appropriately.

mulkieran:
bz: https://bugzilla.redhat.com/show_bug.cgi?id=1785746.

mulkieran:
Even Fedora 30's default version of libblkid suppports recognizing stratis devices, so we can, at least, drop a bunch of comments.

mulkieran:
We might as well close this, as any actions remaining in the code base are unrelated to the libblkid version. The bz can persist independently, as it is really now just a downstream issue.

PR #1833, opened 2020-02-20
Update comments about libblkid versions

mulkieran:
Related: #1656.


mulkieran:
Failure in stratisd-nonrust looked goofy, so I just restarted.

Issue #1782, opened 2020-01-15
Extend device discovery to identify encrypted devices


mulkieran:
We need to extend our existing device discovery framework to identify encrypted devices.

If we think purely of the organization of the data structures returned by ```find_all()``` or an analogous function which is specific to encrypted devices:
* There must be an enumeration for encrypted block devices (not Stratis ones, because that will not be visible). We expect that ```ID_FS_TYPE``` will have value "luks2", I believe.
* Each device presented by the udev enumeration must be examined using libcryptsetup to take a peak at the correct token which must identify the Stratis pool and device UUID. This information is
enough to group the devices by pool UUID. As with the existing code, we ignore the possibility of duplicate Stratis device UUIDs.

An important consideration is that we must handle correctly a situation where some pools are encrypted and some pools are not.

What might conceivably make this trickier is that some devices might be already opened by some outside agencies and thus available as Stratis devices and some, belonging to the same pool, might be still unopened.

jbaublitz:
A couple notes as I get started on this:
1. `TYPE=crypto_LUKS` will be the exact value returned via udev for encrypted devices.
2. I'd like to discuss a little bit more what we plan to store in the LUKS token. For the purpose of this PR, it only matters for the return type of the method I'm stubbing out so it should be relatively easy to work around. My current thought is that we need at the very least keyfile path, pool UUID or name (depending on which makes more sense in the context), and device UUID.

> An important consideration is that we must handle correctly a situation where some pools are encrypted and some pools are not.

I'll investigate this and keep this in mind, but based on what I know of that part of the code base, I think this should actually work relatively nicely with my solution below.

> What might conceivably make this trickier is that some devices might be already opened by some outside agencies and thus available as Stratis devices and some, belonging to the same pool, might be still unopened.

This is thankfully relatively simple to achieve due to the libcryptsetup API. My current thought is that we should iterate through all LUKS devices and confirm that they belong to stratisd and then through all active LUKS devices. For each active device, libcryptsetup allows us to retrieve the path of the physical device on which it exists. We can then compare the set of physical devices that have been activated to the set of physical devices that have not been activated and belong to stratisd. From there we can take two steps: first we open any remaining devices that are inactive. Once we've done this, we construct pools from all of the resulting devicemapper nodes that will be associated with their respective pool UUIDs.

mulkieran:
[...]

PR #1814, opened 2020-02-11
Extend identify module for identification of encrypted devices

mulkieran:
related: #1782

PR #1834, opened 2020-02-21
Encrypted device discovery

jbaublitz:
Related #1782 

PR for initial work on encrypted device discovery. I'm putting this up as a PR so I don't lose track of it in my git history.

### DIFF
--- a/src/engine/strat_engine/backstore/identify.rs
+++ b/src/engine/strat_engine/backstore/identify.rs
@@ -172,11 +172,6 @@ fn find_all_stratis_devices() -> libudev::Result<HashMap<PoolUuid, HashMap<Devic
     Ok(pool_map)
 }
 
-/// Returns the physical path of the device
-fn identify_encrypted_stratis_device(_dev: &libudev::Device) -> Option<PathBuf> {
-    None
-}
-
 /// Get the physical device path from the logical activated decrypted device
 fn logical_path_to_physical_path(_path: &Path) -> Option<PathBuf> {
     None
@@ -199,7 +194,7 @@ fn find_all_closed_encrypted_stratis_devices(
 
     let closed_encrypted_device_map = encrypted_devices
         .iter()
-        .filter_map(|dev| identify_encrypted_stratis_device(dev))
+        .filter_map(|dev| dev.devnode().map(|path| path.to_path_buf()))
         .fold(HashSet::new(), |mut acc, phy_path| {
             if !opened_encrypted_devices.contains(&phy_path) {
                 acc.insert(phy_path);
--- a/src/engine/strat_engine/backstore/identify.rs
+++ b/src/engine/strat_engine/backstore/identify.rs
@@ -172,6 +172,11 @@ fn find_all_stratis_devices() -> libudev::Result<HashMap<PoolUuid, HashMap<Devic
     Ok(pool_map)
 }
 
+/// Returns the physical path of the device
+fn identify_encrypted_stratis_device(_dev: &libudev::Device) -> Option<PathBuf> {
+    None
+}
+
 /// Get the physical device path from the logical activated decrypted device
 fn logical_path_to_physical_path(_path: &Path) -> Option<PathBuf> {
     None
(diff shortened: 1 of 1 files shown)


src/engine/strat_engine/backstore/identify.rs
```diff
@@ -171,6 +172,43 @@ fn find_all_stratis_devices() -> libudev::Result<HashMap<PoolUuid, HashMap<Devic
...
+        .collect::<HashSet<_>>();
+
+    let mut enumerator = block_enumerator(&context)?;
+    enumerator.match_property("ID_FS_TYPE", "crypto_LUKS")?;
+
+    let encrypted_devices: Vec<_> = enumerator.scan_devices()?.collect();
+
+    let closed_encrypted_device_map = encrypted_devices
+        .iter()
+        .filter_map(|dev| identify_encrypted_stratis_device(dev))
```
mulkieran: Unless ```identify_encrypted_stratis_device``` does something that I don't understand you can use existing udev method ```devnode``` here.

jbaublitz: Per stand up, I think this may be a misunderstanding as I realize my comments in the PR are a bit vague.

The algorithm takes the following steps:
1. It takes all available Stratis devices as an input. This could be unencrypted devices or encrypted devices that have already been opened.
2. It filters out all unencrypted devices using `logical_path_to_physical_path` which will return `None` for unencrypted devices and the path of the physical device by using the libcryptsetup bindings for all logical `/dev/mapper` devices corresponding to an encrypted physical device that has been opened.
3. It searches again, this time for all encrypted physical volumes using the `crypto_LUKS` udev type.
4. It uses `identify_encrypted_stratis_device` to probe for our Stratis token to determine whether or not the LUKS2 volume we are looking at is indeed a Stratis encrypted volume. If it is not, it returns `None`. If it is, it returns the same thing as `devnode` as you suggested. My guess is that I will end up using `devnode` internally in `identify_encrypted_stratis_device` but the two are not currently interchangeable as it stands in my design.
5. Filter out all of the devices that have already been determined to be opened in step 2 by checking if `opened_encrypted_devices` contains the physical path that we mapped to in step 4 using `identify_encrypted_stratis_device`. If it is not already opened, pass it along to the next step.
6. Return all unopened devices from the function to be opened [here](https://github.com/stratis-storage/stratisd/pull/1834/files#diff-5434ab93d532c0d2d6f44f725907f5e6R268).

Does this make a little more sense? There may be room for improvement still but hopefully this explains it a little bit better.

mulkieran: > Per stand up, I think this may be a misunderstanding as I realize my comments in the PR are a bit vague.
> 
> The algorithm takes the following steps:
> 
> 1. It takes all available Stratis devices as an input. This could be unencrypted devices or encrypted devices that have already been opened.
> 2. It filters out all unencrypted devices using `logical_path_to_physical_path` which will return `None` for unencrypted devices and the path of the physical device by using the libcryptsetup bindings for all logical `/dev/mapper` devices corresponding to an encrypted physical device that has been opened.
> 3. It searches again, this time for all encrypted physical volumes using the `crypto_LUKS` udev type.
> 4. It uses `identify_encrypted_stratis_device` to probe for our Stratis token to determine whether or not the LUKS2 volume we are looking at is indeed a Stratis encrypted volume. If it is not, it returns `None`. If it is, it returns the same thing as `devnode` as you suggested. My guess is that I will end up using `devnode` internally in `identify_encrypted_stratis_device` but the two are not currently interchangeable as it stands in my design.
> 5. Filter out all of the devices that have already been determined to be opened in step 2 by checking if `opened_encrypted_devices` contains the physical path that we mapped to in step 4 using `identify_encrypted_stratis_device`. If it is not already opened, pass it along to the next step.
> 6. Return all unopened devices from the function to be opened [here](https://github.com/stratis-storage/stratisd/pull/1834/files#diff-5434ab93d532c0d2d6f44f725907f5e6R268).
> 
> Does this make a little more sense? There may be room for improvement still but hopefully this explains it a little bit better.

Yes. That is quite clear. You might as well close this now; I've adopted it into https://github.com/stratis-storage/stratisd/pull/1814 and will try to proceed from there, probably stubbing out some code about the event-based portion of the device discovery in the same spirit.

PR #1837, opened 2020-02-22
Upgrade dbus requirement to 0.8

### LINKED ISSUE #1755: dbus-rs 0.7 came out in September; we should consider moving up at some time
We're currently pinned to 0.6.4 which came out in January 2019. We need to evaluate what the new version brings, as it has been under very active development; it is possible that it will bring us advantages which we can take advantage of on the upgrade.

mulkieran:
Resolves #1755

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -175,11 +175,11 @@ dependencies = [
 
 [[package]]
 name = "dbus"
-version = "0.6.4"
+version = "0.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "libc 0.2.66 (registry+https://github.com/rust-lang/crates.io-index)",
- "libdbus-sys 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


jbaublitz:
One note before I forget is that we may be able to get [this constant](https://github.com/stratis-storage/stratisd/pull/1837/files#diff-514a2718520e983db27476ed65de5994R13) from `dbus` instead of the corresponding `-sys` package so that we don't have to add an explicit crate dependency. This is the recommended way to do it if we can use the export from `dbus` like I think we can.

mulkieran:
> One note before I forget is that we may be able to get [this constant](https://github.com/stratis-storage/stratisd/pull/1837/files#diff-514a2718520e983db27476ed65de5994R13) from `dbus` instead of the corresponding `-sys` package so that we don't have to add an explicit crate dependency. This is the recommended way to do it if we can use the export from `dbus` like I think we can.

I think we can!

mulkieran:
I force-pushed in a moment of absent-mindedness...

mulkieran:
rebased.

PR #1841, opened 2020-02-25
Add ability to recognize LUKS devices to udev implementation

mulkieran:
Related: #1782


mulkieran:
Note that we'll need the ability to recognize a LUKS device given a single udev device for event-based device discovery. The LUKS designation in UdevOwnership is not absolutely necessary, but it may aid users attempting to initialize an encrypted device. Without it, they will be told that the device does not appear unowned, which is a little less helpful.

Issue #1822, opened 2020-02-14
Design an interface for the explicit part of reading and writing luks2 metadata


mulkieran:
* You might as well call the new module something like "crypt.rs", and
I believe it should be a sibling of the "udev.rs" module.

* I believe that it should be testable using our standard infrastructure,
which supplies a bunch of devices to work with. You should be able to
write a bunch of tests that ensure that the appropriate token can be written
for a designated keyslot, and also read back out. If you can take the tests down
a level, and write them like we write our Stratis metadata tests, which don't require
an actual device, but rather just an open file descriptor, that would also be
nice, but I suspect that it is made impossible by the libcryptsetup interface.
Essentially, your tests should be able to encrypt a device completely, setting
a Stratis token in the selected keyslot, and then reopen the device by
locating the keyslot that Stratis indicates that it owns with the token,
opening the device, and reading some
data back out from the device. For these tests, you should not put Stratis
metadata on the device. Just put any kind of value that can't be easily
duplicated, this will be an easier and more direct way to do the test.

* From a device discovery point of view the following would be ideal. It should be
possible to, given a device identified as luks2, scan all keyslots for any token
which identifies the device as a Stratis device. It should return some collection
of results, rather than just a single result, in case there are more than one
keyslot which seems to belong to Stratis. For each keyslot identified as belonging
to Stratis it should return both the keyslot number and the token, either in
its original JSON form, or in some useful representation of that form.
You will find the serde crate, which we already make use of for our Stratis
variable length metadata, helpful here, most likely. For device discovery,
I think identifying the token and opening the device should be separate actions.

* For device initialization, it will be necessary to offer facilities to
write the token. Since reading tokens is inexpensive, Stratis does not need to have
a designated reserved keyslot, so you may want to design the
facility to write to the next open keyslot. You may want to combine writing the
token and setting the key into a single action, as you do not want one with out
the other.

mulkieran:
We had some discussion about initialization, which I'll just make notes on here:

1. Does ```device::process_devices```, which collects device information, and rejects unsuitable devices need to change at all? I think that it may be that the answer is "no".
[...]

PR #1847, opened 2020-02-27
Prototype of methods in crypt module

jbaublitz:
Related to #1822

### DIFF
--- a/src/engine/strat_engine/backstore/blockdev.rs
+++ b/src/engine/strat_engine/backstore/blockdev.rs
@@ -100,7 +100,8 @@ impl StratBlockDev {
             disown_device(&mut OpenOptions::new().write(true).open(&self.devnode)?)?;
         } else {
             let name = name_from_uuids(&self.bda.pool_uuid(), &self.bda.dev_uuid());
-            destroy_encrypted_stratis_device(&get_physical_device_path_from_name(&name)?)?;
+            let physical_path = get_physical_device_path_from_name(&name)?;
+            destroy_encrypted_stratis_device(&physical_path)?;
         }
         Ok(())
     }
--- a/src/engine/strat_engine/backstore/blockdev.rs
+++ b/src/engine/strat_engine/backstore/blockdev.rs
@@ -15,7 +15,10 @@ use crate::{
         engine::BlockDev,
         strat_engine::{
             backstore::{
-                crypt::destroy_encrypted_stratis_device,
+                crypt::{
+                    destroy_encrypted_stratis_device, get_physical_device_path_from_name,
+                    name_from_uuids,
+                },
                 metadata::{disown_device, BDAExtendedSize, MDADataSize, BDA},
                 range_alloc::RangeAllocator,
             },
@@ -96,7 +99,8 @@ impl StratBlockDev {
         if !self.is_encrypted() {
             disown_device(&mut OpenOptions::new().write(true).open(&self.devnode)?)?;
         } else {
-            destroy_encrypted_stratis_device(&self.devnode)?;
+            let name = name_from_uuids(&self.bda.pool_uuid(), &self.bda.dev_uuid());
+            destroy_encrypted_stratis_device(&get_physical_device_path_from_name(&name)?)?;
         }
         Ok(())
     }
--- a/src/engine/strat_engine/backstore/blockdev.rs
+++ b/src/engine/strat_engine/backstore/blockdev.rs
@@ -15,7 +15,7 @@ use crate::{
         engine::BlockDev,
         strat_engine::{
             backstore::{
-                crypt::{destroy_encrypted_stratis_device, get_device_name_from_metadata},
+                crypt::destroy_encrypted_stratis_device,
                 metadata::{disown_device, BDAExtendedSize, MDADataSize, BDA},
                 range_alloc::RangeAllocator,
             },
@@ -96,8 +96,7 @@ impl StratBlockDev {
         if !self.is_encrypted() {
             disown_device(&mut OpenOptions::new().write(true).open(&self.devnode)?)?;
         } else {
-            let device_name = get_device_name_from_metadata(&self.devnode)?;
-            destroy_encrypted_stratis_device(&self.devnode, device_name.as_str())?;
+            destroy_encrypted_stratis_device(&self.devnode)?;
         }
         Ok(())
     }
--- a/src/engine/strat_engine/backstore/blockdev.rs
+++ b/src/engine/strat_engine/backstore/blockdev.rs
@@ -9,6 +9,7 @@ use std::{fs::OpenOptions, path::PathBuf};
 use chrono::{DateTime, TimeZone, Utc};
 
 use devicemapper::{Device, Sectors};
+use libcryptsetup_rs::{CryptInit, CryptKeyslot, CryptWipePattern, EncryptionFormat};
 
 use crate::{
     engine::{
@@ -81,8 +82,40 @@ impl StratBlockDev {
     }
 
     /// Remove information that identifies this device as belonging to Stratis
+    ///
+    /// If self.is_encrypted() is true, destroy all keyslots and wipe the LUKS2 header.
+    /// If self.is_encrypted() is false, wipe the Stratis metadata on the device.
+    /// Both of these actions will destroy the Stratis metadata so that it is no longer
+    /// accessible by stratisd or visible to blkid.
     pub fn disown(&self) -> StratisResult<()> {
-        disown_device(&mut OpenOptions::new().write(true).open(&self.devnode)?)
+        if !self.is_encrypted() {
+            disown_device(&mut OpenOptions::new().write(true).open(&self.devnode)?)?;
+        } else {
+            let mut device = CryptInit::init(&self.devnode())?;
+            device
+                .context_handle()
+                .load::<()>(EncryptionFormat::Luks2, None)?;
+
+            let max_keyslots = CryptKeyslot::max_keyslots(EncryptionFormat::Luks2)?;
+            for i in 0..max_keyslots {
+                device.keyslot_handle().destroy(i)?;
+            }
+
+            let (md_size, ks_size) = device.settings_handle().get_metadata_size()?;
+            let total_luks2_metadata_size = *md_size + *ks_size;
+            device.wipe_handle().wipe(
+                &self.devnode(),
+                CryptWipePattern::Zero,
+                0,
+                total_luks2_metadata_size,
+                4096,
+                false,
+                None,
+                // Hack until libcryptsetup-rs#43 is merged
+                &mut (),
+            )?;
+        }
+        Ok(())
     }
 
     pub fn save_state(&mut self, time: &DateTime<Utc>, metadata: &[u8]) -> StratisResult<()> {
@@ -159,6 +192,10 @@ impl BlockDev for StratBlockDev {
 
     fn get_dbus_path(&self) -> &MaybeDbusPath {
         &self.dbus_path
+    }
+
+    fn is_encrypted(&self) -> bool {
+        false
     }
 }
 
--- a/src/engine/strat_engine/backstore/blockdev.rs
+++ b/src/engine/strat_engine/backstore/blockdev.rs
@@ -47,6 +47,8 @@ impl StratBlockDev {
     /// - other_segments: segments claimed for non-Stratis metadata use
     /// - user_info: user settable identifying information
     /// - hardware_info: identifying information in the hardware
+    /// - key_description: optional argument enabling encryption using
+    ///                    the specified key in the kernel keyring
     /// Returns an error if it is impossible to allocate all segments on the
     /// device.
     /// NOTE: It is possible that the actual device size is greater than
--- a/src/engine/strat_engine/backstore/blockdev.rs
+++ b/src/engine/strat_engine/backstore/blockdev.rs
@@ -9,13 +9,16 @@ use std::{fs::OpenOptions, path::PathBuf};
 use chrono::{DateTime, TimeZone, Utc};
 
 use devicemapper::{Device, Sectors};
-use libcryptsetup_rs::{CryptInit, CryptKeyslot, CryptWipePattern, EncryptionFormat};
 
 use crate::{
     engine::{
         engine::BlockDev,
         strat_engine::{
             backstore::{
+                crypt::{
+                    destroy_encrypted_stratis_device, encrypted_device_is_active,
+                    get_device_name_from_metadata, wipe_encrypted_stratis_device,
+                },
                 metadata::{disown_device, BDAExtendedSize, MDADataSize, BDA},
                 range_alloc::RangeAllocator,
             },
@@ -96,28 +99,12 @@ impl StratBlockDev {
         if !self.is_encrypted() {
             disown_device(&mut OpenOptions::new().write(true).open(&self.devnode)?)?;
         } else {
-            let mut device = CryptInit::init(&self.devnode())?;
-            device
-                .context_handle()
-                .load::<()>(EncryptionFormat::Luks2, None)?;
-
-            let max_keyslots = CryptKeyslot::max_keyslots(EncryptionFormat::Luks2)?;
-            for i in 0..max_keyslots {
-                device.keyslot_handle().destroy(i)?;
+            let device_name = get_device_name_from_metadata(&self.devnode)?;
+            if encrypted_device_is_active(&device_name) {
+                destroy_encrypted_stratis_device(&self.devnode, device_name.as_str())?;
+            } else {
+                wipe_encrypted_stratis_device(&self.devnode)?;
             }
-
-            let (md_size, ks_size) = device.settings_handle().get_metadata_size()?;
-            let total_luks2_metadata_size = *md_size + *ks_size;
-            device.wipe_handle().wipe::<()>(
-                &self.devnode(),
-                CryptWipePattern::Zero,
-                0,
-                total_luks2_metadata_size,
-                4096,
-                false,
-                None,
-                None,
-            )?;
         }
         Ok(())
     }
--- a/src/engine/strat_engine/backstore/blockdev.rs
+++ b/src/engine/strat_engine/backstore/blockdev.rs
@@ -15,10 +15,7 @@ use crate::{
         engine::BlockDev,
         strat_engine::{
             backstore::{
-                crypt::{
-                    destroy_encrypted_stratis_device, encrypted_device_is_active,
-                    get_device_name_from_metadata, wipe_encrypted_stratis_device,
-                },
+                crypt::{destroy_encrypted_stratis_device, get_device_name_from_metadata},
                 metadata::{disown_device, BDAExtendedSize, MDADataSize, BDA},
                 range_alloc::RangeAllocator,
             },
@@ -100,11 +97,7 @@ impl StratBlockDev {
             disown_device(&mut OpenOptions::new().write(true).open(&self.devnode)?)?;
         } else {
             let device_name = get_device_name_from_metadata(&self.devnode)?;
-            if encrypted_device_is_active(&device_name) {
-                destroy_encrypted_stratis_device(&self.devnode, device_name.as_str())?;
-            } else {
-                wipe_encrypted_stratis_device(&self.devnode)?;
-            }
+            destroy_encrypted_stratis_device(&self.devnode, device_name.as_str())?;
         }
         Ok(())
     }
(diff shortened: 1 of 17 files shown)


jbaublitz:
Blocked by stratis-storage/libcryptsetup-rs#27

jbaublitz:
Blocked by stratis-storage/libcryptsetup#29. I'll add a workaround here for now.

jbaublitz:
@mulkieran If you're interested in looking at the prototype of the syscall implementation, it is in a "preview" state at the moment.

It compiles and is organized in a way that I think will eventually go up for review without many changes.

TODO:
* Add tests to make sure the functionality works as expected.
* Add token initialization in as well so that it works with the existing activation method in this PR.
* Figure out how to deal with the current CI failures as Travis does not have cryptsetup 2.3. This means that the `MemSafeHandler::alloc` method does not exist yet. There are ways around this, but I'm not quite happy with any of them so I'll have to think on this some more.

jbaublitz:
Rebased.

src/engine/strat_engine/backstore/crypt.rs
```diff
@@ -0,0 +1,274 @@
...
+
+    let stratis_device_name = get_stratis_device_name(&mut crypt_device)?;
+    crypt_device.token_handle().activate_by_token::<()>(
+        Some(stratis_device_name.as_str()),
+        Some(LUKS2_TOKEN_ID),
+        None,
+        CryptActivateFlags::empty(),
+    )?;
+
+    // Assert that physical device path is equal to the path of the provided
```
mulkieran: In the final version, this should return an error, because whatever bad things happen, we don't really want to crash here. But it is definitely a reasonable thing to verify, i.e., we shouldn't go on if it is false.

jbaublitz: Agreed, I intend to clean that up when I submit this for review.

jbaublitz:
> Figure out how to deal with the current CI failures as Travis does not have cryptsetup 2.3. This means that the MemSafeHandler::alloc method does not exist yet. There are ways around this, but I'm not quite happy with any of them so I'll have to think on this some more.

@mulkieran @bgurney-rh There is currently no way to set this up that both cleans up keys and works with 2.2 in CI. What do we want to do? Do we want to try to get 2.3 into Travis? Do we want to disable Travis for now?

mulkieran:
> > Figure out how to deal with the current CI failures as Travis does not have cryptsetup 2.3. This means that the MemSafeHandler::alloc method does not exist yet. There are ways around this, but I'm not quite happy with any of them so I'll have to think on this some more.
> 
> @mulkieran @bgurney-rh There is currently no way to set this up that both cleans up keys and works with 2.2 in CI. What do we want to do? Do we want to try to get 2.3 into Travis? Do we want to disable Travis for now?

@jbaublitz Can you delay all this by letting ```read_key``` be unsafe for now?

jbaublitz:
> > > Figure out how to deal with the current CI failures as Travis does not have cryptsetup 2.3. This means that the MemSafeHandler::alloc method does not exist yet. There are ways around this, but I'm not quite happy with any of them so I'll have to think on this some more.
> > 
> > 
> > @mulkieran @bgurney-rh There is currently no way to set this up that both cleans up keys and works with 2.2 in CI. What do we want to do? Do we want to try to get 2.3 into Travis? Do we want to disable Travis for now?
> 
> @jbaublitz Can you delay all this by letting `read_key` be unsafe for now?

I can't. The problem here is that we have to get memory allocated and cleared after dropping it somehow. For 2.2 we have no allocation or memzero options, only a safe free and as we don't have the safe allocation method, we can't use the safe free. For 2.3 we have both allocation and memzero. The only reason safe free works with our API for 2.2 is that the safe allocation is used internally for reading the keyfile so we can reliably and safely free that memory. It is not exposed externally in 2.2 so any memory we would allocate with publicly available methods would cause memory corruption with the safe free.

mulkieran:
We agreed to put in an a litle bit of unsafety to get this passing CI again.

jbaublitz:
Blocked by #1861

jbaublitz:
Travis is now fixed. See stratis-storage/project#152 for the status of Jenkins infrastructure.

bgurney-rh:
Re-running Jenkins tests, now that each of the test nodes has cryptsetup 2.3 installed.

bgurney-rh:
@jbaublitz all tests passed.

jbaublitz:
@bgurney-rh Great! Thank you!

jbaublitz:
Rebased PR onto the final version of #1862 and pushed a new commit calling encrypted device initialization with `None` for the key description which will disable encryption initialization. We can eventually pass in the parameter from the D-Bus method for this parameter. For now I'll leave this as `None` as that is in the scope of #1817.

jbaublitz:
Rebased to test the fix in #1862

mulkieran:
@jbaublitz: The last commit changes to ```initialize_devices``` make sense. However, ```initialize_devices``` has an existing problem, which is that the cleanup actions, in case one blockdev is not properly initialized, don't get started reliably. This will get more troublesome w/ encryption additions. I'm going to take another pass at ```initialize_devices``` to see if that can be improved.

jbaublitz:
> @jbaublitz: The last commit changes to `initialize_devices` make sense. However, `initialize_devices` has an existing problem, which is that the cleanup actions, in case one blockdev is not properly initialized, don't get started reliably. This will get more troublesome w/ encryption additions. I'm going to take another pass at `initialize_devices` to see if that can be improved.

How are you planning to do this? Are you going to put up a PR before we merge this one? I can keep working on this and rebase once that's merged if that's the plan.

mulkieran:
> > @jbaublitz: The last commit changes to `initialize_devices` make sense. However, `initialize_devices` has an existing problem, which is that the cleanup actions, in case one blockdev is not properly initialized, don't get started reliably. This will get more troublesome w/ encryption additions. I'm going to take another pass at `initialize_devices` to see if that can be improved.
> 
> How are you planning to do this? Are you going to put up a PR before we merge this one? I can keep working on this and rebase once that's merged if that's the plan.

I hope the PR will be up sometime this morning. Keep working where you are and plan on a rebase.

jbaublitz:
TODO: Clean up LUKS2 headers if `key_description` is not `None`.

jbaublitz:
TODO: `.to_simple_ref()` for UUIDs in name.

jbaublitz:
Based on the CI failure message, it looks like we're making progress. It's no longer a bug in the logic of the test but now more of a problem around the size of the real device used. If I restrict the device size to 1GB, it may resolve the issue. If I'm understanding the log message correctly, it looks like it tried to read 10GB of disk into memory which would probably result in an OOM kill signal being sent which is what happened. I'll restrict it to 1GB and see if that fixes it.

bgurney-rh:
Yeah, I can see an OOM event on one of the test nodes:
```
kernel: Out of memory: Killed process 990 (libstratis-0994) total-vm:4331736kB, anon-rss:3590808kB, file-rss:4kB, shmem-rss:0kB, UID:0 pgtables:7156kB oom_score_adj:0
```

That system has 4 GB of memory.  I don't know if we have a memory requirement for running the stratisd tests, but I had previously been able to run them on a system with 2 GB of memory.

bgurney-rh:
However, on the most recent run, I did not see an oom event, so it's not 100% (though it did still fail, and it left a crypt device that triggered a failed test on my PR.)

bgurney-rh:
I checked after the last 2 stratisd test runs, and it looks like there's no longer a crypt device being left behind after the failure.

jbaublitz:
I fixed the compilation errors (@bgurney-rh that's why there was no lingering crypt device).

jbaublitz:
I've completed both TODOs that came out of the last standup where we discussed the crypt module. The most recent addition is a change to `.disown()` and an `.is_encrypted()` method on `StratBlockDev`s that will trigger whether we wipe the LUKS2 metadata or Stratis metadata.

jbaublitz:
The final commit I'm about to push should implement encryption once the `key_description` parameter is not hardcoded to be `None` when invoking `initialize_devices`. For now encryption will always be disabled as no key description is passed in. This should meet with #1817 and allow the user to specify encryption over D-Bus.

Blockdevs are now aware of whether they are encrypted or not and paths and devnodes should match up to the unencrypted logical device as opposed to the encrypted physical device.

src/engine/strat_engine/backstore/blockdev.rs
```diff
@@ -59,6 +61,7 @@ impl StratBlockDev {
         upper_segments: &[(Sectors, Sectors)],
         user_info: Option<String>,
         hardware_info: Option<String>,
+        key_description: Option<String>,
```
mulkieran: Please add some docs for the new parameter.

src/engine/strat_engine/backstore/devices.rs
```diff
@@ -237,9 +239,32 @@ pub fn initialize_devices(
...
+        let dev_uuid = Uuid::new_v4();
+        let (path, deviceno) = match key_description {
+            Some(desc) => {
+                let path = initialize_encrypted_stratis_device(
+                    &dev_info.devnode,
+                    pool_uuid,
+                    dev_uuid,
+                    desc,
+                )?;
+                let dm = DM::new()?;
```
mulkieran: We have a perpetual devicemapper context defined in ```src/engine/strat_engine/dm.rs``` which you should use here.

src/engine/strat_engine/backstore/crypt.rs
```diff
@@ -0,0 +1,476 @@
...
+use libcryptsetup_rs::{
+    c_uint, CryptActivateFlags, CryptDevice, CryptInit, CryptStatusInfo, CryptVolumeKeyFlags,
+    EncryptionFormat, LibcryptErr, SafeMemHandle,
+};
+
+use crate::engine::{DevUuid, PoolUuid};
+
+type Result<T> = std::result::Result<T, LibcryptErr>;
+
+// Stratis token JSON keys
```
mulkieran: Please make constants private as much as possible. We can always make them public as we find we need to.

src/engine/strat_engine/backstore/devices.rs
```diff
@@ -237,9 +239,32 @@ pub fn initialize_devices(
...
+                let info = dm.device_info(&DevId::Name(DmName::new(&name_from_uuids(
```
mulkieran: The method ```name_from_uuids``` should possibly be defined in ```src/engine/strat_engine/names.rs```.

src/engine/strat_engine/backstore/devices.rs
```diff
@@ -237,9 +239,32 @@ pub fn initialize_devices(
...
+                    &pool_uuid, &dev_uuid,
+                ))?))?;
+                (path, info.device())
+            }
+            None => (dev_info.devnode.clone(), dev_info.devno),
+        };
+
+        let mut f = OpenOptions::new().write(true).open(&path)?;
 
+        // NOTE: May need to rethink hw_id in the context of an encrypted device -
```
mulkieran: Let's put the ```hw_id``` question on the standup agenda.

src/engine/strat_engine/backstore/devices.rs
```diff
@@ -237,9 +239,32 @@ pub fn initialize_devices(
         dev_info: &DeviceInfo,
         pool_uuid: PoolUuid,
         mda_data_size: MDADataSize,
+        key_description: Option<&str>,
     ) -> StratisResult<StratBlockDev> {
-        let mut f = OpenOptions::new().write(true).open(&dev_info.devnode)?;
+        let dev_uuid = Uuid::new_v4();
+        let (path, deviceno) = match key_description {
+            Some(desc) => {
+                let path = initialize_encrypted_stratis_device(
```
mulkieran: Does the call to ```initialize_encrypted_stratis_devices```, if it fails, leave anything that can or should be cleaned up? If not, can you add a comment indicating why not?

src/engine/strat_engine/backstore/devices.rs
```diff
@@ -228,6 +229,7 @@ pub fn initialize_devices(
     devices: Vec<DeviceInfo>,
     pool_uuid: PoolUuid,
     mda_data_size: MDADataSize,
+    key_description: Option<&str>,
```
mulkieran: Is it possible to add a test that can make use of the ```key_description``` parameter _within_ this module?

src/engine/strat_engine/backstore/crypt.rs
jbaublitz: Now that everything has an appropriately scoped API, I think they can all be private. Fixing.

src/engine/strat_engine/backstore/devices.rs
jbaublitz: Let's talk about it in stand up. This is only related to crypt devices so I could see it going either way based on what is already in the `names.rs` module.

src/engine/strat_engine/backstore/devices.rs
jbaublitz: Sounds good.

src/engine/strat_engine/backstore/devices.rs
jbaublitz: This depends on what you want to test. If we're testing anything like whether or not a device gets encrypted, we do already have a test for that in the `crypt` module and we will introduce issues around setting up the keyring. We can probably do that but it seems slightly redundant if we already do that elsewhere. If you'd like to test something like teardown, we may still have to set up the keyring for that but that seems less redundant to me. What are your thoughts on this?

src/engine/strat_engine/backstore/devices.rs
jbaublitz: I'll take a closer look at this as we may leave just the devicemapper device around and I could clean that up. The `.disown()` method should take care of that for all previously activated devices so I'll double check that that happens.

src/engine/strat_engine/backstore/blockdev.rs
jbaublitz: Resolved.

src/engine/strat_engine/backstore/devices.rs
mulkieran: Feel free to leave it here for now.

src/engine/strat_engine/backstore/devices.rs
mulkieran: We decided in standup that encrypted devices never have a hw_id.

src/engine/strat_engine/backstore/devices.rs
jbaublitz: Okay, will do.

jbaublitz:
This next commit is not ready for review yet, but it passes my git hooks and it's a move in the right direction for cleanup.

jbaublitz:
TODO: Potentially add method that only deactivates if active wrapping `wipe` and `deactivate`.

src/engine/strat_engine/backstore/devices.rs
jbaublitz: This should be mostly resolved.

src/engine/strat_engine/backstore/devices.rs
jbaublitz: Resolved.

src/engine/strat_engine/backstore/devices.rs
jbaublitz: Resolved.

jbaublitz:
> Comment in test...I usually prefer non-closures where it works just the same, as then the reader doesn't have to reason about the environment. Other than that the code looks good. I didn't dig super deep into the one dead function...I guess there are a few more that are really dead but do not need annotations, but they all looked fine.

The dead function and all of the ones that it depends on will ultimately be what you use in device discovery. If the top level function is allowed to be dead code, so are all of the ones it depends on. This is why annotations are not needed for those.

> About testing...I have some questions:
> 
>     * Would there be a way to set the key once for multiple tests, so that one key-setting operation would work for tests in multiple different modules?

I'm not sure if there's a way to clean this up reliably. Even though we only execute tests one at a time, what we'd end up having to do is ensure that all tests that rely on a key have executed before cleaning it up and this introduces and ordering problem. While it would be possible to artificially introduce ordering with a single thread, this is very error prone when we add new tests. The other option would be to introduce some kind of reference counter and after all tests have run, clean up the key. This also has its own issues as we would need to store that reference count somewhere and I'm not sure if there's a way to do that without global mutable state which would introduce unsafe code. Either solution seems more error prone that adding a reusable test utility to add and clean up the key which we can use if we decide against mocking.

>     * I looked at the existing tests in devices and think that some capture the spirit of what we would also want to do for encryption:
>       
>       * test_ownership - that would be reasonable to do for encryption, too. Note that the checks as currently written would fail on encrypted devices.

This seems reasonable and like a good case for using actual LUKS2 metadata.
 
>     * test_initialize - not ready yet, because it relies on find_all which is not yet prepared
> 
>     * test_failure_cleanup - as you suggested, would be good to do for encrypted too

I think this is doable too. This also seems like a good case for actually using LUKS2 metadata instead of mocking.

>     * The other two, test_nonexistent_paths and test_duplicate_devnodes ought to work the same for both
> 
> 
> We also discussed the possibility of mocking. We should certainly keep that in mind.

One final note is that if we decide against keyfiles in certain scenarios except where absolutely needed, we can always use the passthrough option for encryption to simply test metadata operations and cleanup if we never need to activate the device. This may work for both ownership and cleanup as neither one really requires activating the device. I'll look into this more today.

mulkieran:
>> I didn't dig super deep into the one dead function...I guess there are a few more that are really dead but do not need annotations, but they all looked fine.
> 
> The dead function and all of the ones that it depends on will ultimately be what you use in device discovery. If the top level function is allowed to be dead code, so are all of the ones it depends on. This is why annotations are not needed for those.

Exactly. Just letting you know that the code that is not being exercised yet was also not really scrutinized yet by me. So I may come up with more thoughts once I start exercising it.

mulkieran:
> I'm not sure if there's a way to clean this up reliably. Even though we only execute tests one at a time, what we'd end up having to do is ensure that all tests that rely on a key have executed before cleaning it up and this introduces and ordering problem. While it would be possible to artificially introduce ordering with a single thread, this is very error prone when we add new tests. The other option would be to introduce some kind of reference counter and after all tests have run, clean up the key. This also has its own issues as we would need to store that reference count somewhere and I'm not sure if there's a way to do that without global mutable state which would introduce unsafe code. Either solution seems more error prone that adding a reusable test utility to add and clean up the key which we can use if we decide against mocking.

My thought was that we could give them all a distinct name prefix, like ```crypt_test...``` and exercise them all in a separate test task. Let's talk about this more in standup.

jbaublitz:
> > > I didn't dig super deep into the one dead function...I guess there are a few more that are really dead but do not need annotations, but they all looked fine.
> > 
> > 
> > The dead function and all of the ones that it depends on will ultimately be what you use in device discovery. If the top level function is allowed to be dead code, so are all of the ones it depends on. This is why annotations are not needed for those.
> 
> Exactly. Just letting you know that the code that is not being exercised yet was also not really scrutinized yet by me. So I may come up with more thoughts once I start exercising it.

Not a problem! Let me know if anything needs to get altered.

I'm about to put this PR back up for review.

src/engine/strat_engine/backstore/crypt.rs
```diff
@@ -0,0 +1,476 @@
...
+            let lossy_disk_string = String::from_utf8_lossy(&disk_buffer);
+            if lossy_disk_string.contains(test_string) {
+                return Err(Box::new(io::Error::new(
+                    io::ErrorKind::Other,
+                    "Disk was not encrypted!",
+                )));
+            }
+            std::mem::drop(devicenode);
+
+            let close_active = |name: &str| -> Result<()> {
```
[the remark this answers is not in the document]
jbaublitz: Resolved.

jbaublitz:
Putting up for review while I work on more testing as discussed above.

mulkieran:
@jbaublitz For ```test_ownership``` you can generalize the test like this.

* initialize the devices, get a list of StratBlockDevs as before
* For each StratBlockDev, verify that the correct Stratis identifiers are on whatever devnode that blockdev has AND verify that the number of StratBlockDevs returned by ```initialize_devices``` is the same as the number of DeviceInfo's passed in the ```devices``` argument
* Wipe the devices as before, (but this time your rewritten disown method is called for each blockdev)
* Go through the same list of StratBlockdevs, and verify that Stratis identifiers can not be found

The only significant change is that the iteration which checks for presence of Stratis metadata, checks the devnode corresponding to the given blockdev, not the devnodes originally passed to the test.

jbaublitz:
The latest commit is a significant one. It fixes a number of issues that were latent in the crypt module prior to testing.
* Wipe size when cleaning up Stratis devices must be a multiple of `SECTOR_SIZE` in libcryptsetup.
* We cannot destroy keyslots that do not exist. We must query the LUKS2 token for the value of active keyslots to destroy
* Checking the format of a LUKS2 device triggered a bug in libcryptsetup-rs. The fix should be merged shortly.

Other than this, there's some general clean up that I did, improved logging, and above all a working ownership test for both encrypted and unencrypted devices.

jbaublitz:
I'm closing this PR, squashing it down to one commit, leaving the branch as is, and adding to another branch based off of this one to try to fix CI. I think this is a result of #1864.

Issue #1840, opened 2020-02-25
Make Stratis metadata relocatable


mulkieran:
We need the ability to write the data to arbitrarily chosen locations on a device, for backup or for other uses. Right now, the metadata is implicitly offset from 0. Change that by parameterizing the appropriate methods on an offset.

mulkieran:
We have other more pressing metadata tasks. This will go on a board when we need it.

jbaublitz:
I'm actually interested in reviving this one. My current thoughts are that if we can find a way to do this, this would allow adding support for initializing the Stratis metadata with room for a LUKS2 header and then converting the pool to an encrypted pool at a later time.

Considerations for this include how we would have to modify blkid to detect the superblock if it is at an offset. We need to make sure that superblocks at an offset would still be picked up by blkid.

PR #1849, opened 2020-02-27
Static header offset

lleshchi:
Related #1840 

The index variable has been added to BDA's `load` and `initialize` methods. Currently, the tests in `bda.rs `and `static_header.rs` use an index of 0, but I will expand this to test other offsets, if this looks good so far.

### DIFF
--- a/src/engine/strat_engine/backstore/metadata/bda.rs
+++ b/src/engine/strat_engine/backstore/metadata/bda.rs
@@ -31,7 +31,7 @@ impl BDA {
     /// Initialize a blockdev with a Stratis BDA.
     pub fn initialize<F>(
         f: &mut F,
-        index: usize,
+        offset: usize,
         identifiers: StratisIdentifiers,
         mda_data_size: MDADataSize,
         blkdev_size: BlockdevSize,
@@ -47,7 +47,7 @@ impl BDA {
             initialization_time,
         );
 
-        header.write(f, index, MetadataLocation::Both)?;
+        header.write(f, offset, MetadataLocation::Both)?;
 
         let regions =
             mda::MDARegions::initialize(STATIC_HEADER_SIZE.sectors().bytes(), header.mda_size, f)?;
@@ -57,11 +57,11 @@ impl BDA {
 
     /// Load a BDA on initial setup of a device.
     /// Returns None if no BDA appears to exist.
-    pub fn load<F>(f: &mut F, index: usize) -> StratisResult<Option<BDA>>
+    pub fn load<F>(f: &mut F, offset: usize) -> StratisResult<Option<BDA>>
     where
         F: Read + Seek + SyncAll,
     {
-        let header = match StaticHeader::setup(f, index)? {
+        let header = match StaticHeader::setup(f, offset)? {
             Some(header) => header,
             None => return Ok(None),
         };
--- a/src/engine/strat_engine/backstore/metadata/bda.rs
+++ b/src/engine/strat_engine/backstore/metadata/bda.rs
@@ -60,7 +60,7 @@ impl BDA {
     where
         F: Read + Seek + SyncAll,
     {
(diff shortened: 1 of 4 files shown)


src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -31,6 +31,7 @@ impl BDA {
     /// Initialize a blockdev with a Stratis BDA.
     pub fn initialize<F>(
         f: &mut F,
+        index: usize,
```
mulkieran: Try renaming ```index``` to ```offset```. You'll also want to document the meaning of this parameter. There is no particular format that we enforce with documentation in our Rust projects, unlike w/ our Python projects, where we use the sphinx markup.

src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -56,11 +57,11 @@ impl BDA {
 
     /// Load a BDA on initial setup of a device.
     /// Returns None if no BDA appears to exist.
-    pub fn load<F>(f: &mut F) -> StratisResult<Option<BDA>>
+    pub fn load<F>(f: &mut F, index: usize) -> StratisResult<Option<BDA>>
     where
         F: Read + Seek + SyncAll,
     {
-        let header = match StaticHeader::setup(f)? {
+        let header = match StaticHeader::setup(f, index)? {
```
mulkieran: If the static header is offset, then I think that the MDA header below will need to be offset as well.

src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -46,7 +47,7 @@ impl BDA {
             initialization_time,
         );
 
-        header.write(f, MetadataLocation::Both)?;
+        header.write(f, index, MetadataLocation::Both)?;
```
mulkieran: If the static header is offset, then the MDA will have to be offset as well.

src/engine/strat_engine/backstore/metadata/static_header.rs
```diff
@@ -183,7 +183,7 @@ impl StaticHeader {
             write_region(f, &signature_block, &zeroed)?;
         } else {
             f.seek(SeekFrom::Start(
-                bytes!(static_header_size::SIGBLOCK_REGION_SECTORS) as u64,
+                (index + bytes!(static_header_size::SIGBLOCK_REGION_SECTORS)) as u64,
```
mulkieran: I think using ```SeekFrom::Current``` will make this simpler.

mulkieran:
@lleshchi Another option, and this one has just occurred to me, is not to pass an index to the methods at all, but to require as a precondition that the file "pointer" when passed, be at the correct location. This alternative approach is worth looking at, but you should go ahead and  modify the tests as I suggested first.

mulkieran:
@lleshchi Another thought...we identified one place in the ```static_header``` tests where the test was allocating more space than necessary... probably because it was adapted from a test that used BDA methods that were required the additional space. You might as well make a separate PR to fix that; it should really have gone in with one of the refactoring PRs we cleared months ago.

mulkieran:
@lleshchi: Actually, the test failure indicates a bug in our implementation; ```StaticHeader::wipe``` should also be relocatable.

I think that it's time to revise our implementation plan, assuming that we're going to try the alternative approach that doesn't take an offset. Try the following:

* Hold onto this PR as a reference.
* Open a new PR.
  - In the new PR, begin by just working on the ```static_header``` module.
    - Modify ```StaticHeader::write```, ```setup```, and ```wipe``` so that they all relocate, but use the new approach where the file is just queued up to the correct location.
    - Modify the tests in ```static_header``` so that they all take an offset argument and so that they all succeed.
  - Then begin work on the BDA implementation.

PR #1850, opened 2020-02-27
More refinements to device discovery

mulkieran:


mulkieran:
rebased.

mulkieran:
My udev testing experiment didn't work out so well, but this logging might as well go in immediately. You can see the effects in the Jenkins stratisd tests; look for "set up". You can't see any log messages for discovered devices, though. I should think that these would be being generated when the Python-based udev tests are run, but it's more work to get the stratisd logs to be displayed in those tests.

PR #1854, opened 2020-02-27
Use pyudev 0.22.0 in tests

mulkieran:


mulkieran:
Note that pylint doesn't actually show the deprecation warnings on the lint test. Where they are visible is in the Jenkins non-rust tests. But we won't need to do anything for the CI, because when we update to Fedora 32 for the Jenkins machines we'll get the new version of pyudev that comes along with that distro. In the meantime, there is no reason to let the deprecation warnings bother us.

mulkieran:
This is too trivial to bother anyone for review, merging.

PR #1859, opened 2020-03-02
Tidy up some tests in the sim engine module

mulkieran:


src/engine/sim_engine/engine.rs
```diff
@@ -231,32 +231,15 @@ mod tests {
...
-    /// Creating a new pool with the same name and arguments should return identity
+    /// Creating a new pool with the same name and arguments should return
+    /// identity.
     fn create_pool_name_collision() {
         let name = "name";
         let mut engine = SimEngine::default();
-        engine
-            .create_pool(name, &[Path::new("/s/d")], None)
-            .unwrap();
+        let devices = [Path::new("/s/d")];
```
jbaublitz: I created a testing macro that you could use here if that's of interest. Since it's already merged, it might make sense to use that instead. It's located [here](https://github.com/stratis-storage/stratisd/blob/develop-2.0.1/src/engine/macros.rs#L157-L162).

mulkieran: > I created a testing macro that you could use here if that's of interest. Since it's already merged, it might make sense to use that instead. It's located [here](https://github.com/stratis-storage/stratisd/blob/develop-2.0.1/src/engine/macros.rs#L157-L162).

Probably best to hit a few more missed spots...

mulkieran:
I'll pick up a few more fixes while here...

mulkieran:
This has new bits now, so I'm re-requesting review.

PR #1860, opened 2020-03-04
Do not include stratis_identifiers field in DeviceInfo struct

mulkieran:


mulkieran:
The failure is real, copy-paste error in test changes caused it.

PR #1861, opened 2020-03-04
Use cryptsetup-2.3 instead of 2.2

jbaublitz:
Use our packaged version of cryptsetup 2.3 in Travis CI.

Initial PR related to stratis-storage/project#152

PR #1862, opened 2020-03-04
Use cryptsetup 2.3 instead of 2.2 in Travis

jbaublitz:
Supersedes #1861 
Related to stratis-storage/project#152

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,9 @@ before_install:
   - sudo apt-get -q update
   # cryptsetup-bin conflicts with custom built cryptsetup
   - sudo apt-get remove cryptsetup-bin
+  # Linking fails if libcryptsetup 2.2 is present - must force
+  # remove due to system dependencies
+  - sudo dpkg --purge --force-all libcryptsetup12
   - sudo apt-get install -y libargon2-0 libjson-c3
   - wget "https://github.com/jbaublitz/stratisd/raw/deb/cryptsetup_2.3.0-1_amd64.deb"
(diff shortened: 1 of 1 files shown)


jbaublitz:
I found a linking bug if libcryptsetup 2.2 is installed. Because most system packages in Ubuntu depend on this, I need to add one more step to purge the system library or linking fails due to two present libcryptsetups of the same major version, one without `crypt_safe_free`. I'll rerequest review once I've confirmed that this `dpkg` command that I've found works locally also works for Travis based on how it behaves with #1847.

mulkieran:
ok!

jbaublitz:
Reproduced the issue and confirmed that this fixes it on my Ubuntu `bionic` VM. Will confirm on Travis.

jbaublitz:
My best guess at why this was not visible before is that prior to today, the `crypt` module added in #1847 was all dead code and was probably pruned prior to linking. Now that I've begun using it in the `devices` module in #1847, linking was required.

mulkieran:
Great! Glad we caught it here.

jbaublitz:
Tests are now running in #1847. Back up for review.

mulkieran:
Do you want to do anything about the dpkg warning ```dpkg: warning: unable to delete old directory '/etc/cryptsetup-initramfs': Directory not empty``` on the install command ```sudo dpkg -i ./cryptsetup_2.3.0-1_amd64.deb```?

jbaublitz:
This doesn't concern me as we will never reboot and that is the only time initramfs would be necessary.

PR #1863, opened 2020-03-04
Develop 2.0.1 purgatory

### LINKED ISSUE #1122: Add a new interface for reports; use it to shown liminal block devices
Depends on #1121. Add Incomplete to PoolState pool property. Incomplete pools should be shown in the D-Bus API, using the new pool state. Blockdevs should also be shown in D-Bus, using the Blockdev State field to indicate which of the 5 blockdev states they are (missing, bad, spare, notInUse, In-Use) with missing devices as Missing. Devices that are present in the incomplete pool, I guess list them as NotInUse, would make the most sense?

mulkieran:
closes #1122


mulkieran:
We decided to go with a separate report interface instead.

PR #1865, opened 2020-03-05
Restructure initialize_devices to eliminate two bugs

mulkieran:


mulkieran:
There should really be a test to verify the correct behavior of ```initialize_devices``` if initialization of one device fails after others have been initialized, but there has never been such. Since this PR is sort of blocking #1847, I'm putting up the changes now, and I'll add the test in a subsequent PR.

@jbaublitz Note that the encryption changes would affect ```initialize_one()``` both in initialization and for cleanup. After a cursory inspection, I _think_ that ```wipe_devices``` would just have to be extended to wipe both the Stratis metadata and the LUKS info, but I'm not certain.

jbaublitz:
> There should really be a test to verify the correct behavior of `initialize_devices` if initialization of one device fails after others have been initialized, but there has never been such. Since this PR is sort of blocking #1847, I'm putting up the changes now, and I'll add the test in a subsequent PR.

Okay sounds good. I'll look this over soon. So far it seems good, but I want to examine a little more.

> @jbaublitz Note that the encryption changes would affect `initialize_one()` both in initialization and for cleanup. After a cursory inspection, I _think_ that `wipe_devices` would just have to be extended to wipe both the Stratis metadata and the LUKS info, but I'm not certain.

If you mean wipe LUKS2 metadata for encrypted devices and Stratis metadata for unencrypted devices, I agree that seems like it should be sufficient. If you mean wipe Stratis and then LUKS2 metadata for encrypted devices, I don't think that's necessary. I'm happy to discuss more though if we're thinking different things.

mulkieran:
rebased.

PR #1867, opened 2020-03-05
Simplify nested match expression

lleshchi:


src/engine/strat_engine/backstore/metadata/static_header.rs
```diff
@@ -262,6 +262,31 @@ impl StaticHeader {
             }
         }
 
+        fn compare_static_headers<F>(
```
mulkieran: The code looks fine. See if you can remove all the inline comments in the ```compare_static_headers``` method, and instead have a doc comment on the method which captures everything the method is responsible for.

PR #1868, opened 2020-03-05
Add a test to verify cleanup of devices

mulkieran:


mulkieran:
An uninteresing failure in the non-rust tests; the startup test very occasionally hangs.

PR #1889, opened 2020-03-16
Tidy up Python-based udev tests

mulkieran:
There are two important parts of this PR:

* General tidies and simplifications
* Omitting test code that was specific to an older version of libblkid. Since we require the newer version in the new release, and our test infrastructure uses the newer version, we don't need support in the test code for the older version.


mulkieran:
The ultimate goal is to capture stratisd logging in the tests. But this is enough changes that it seems right to put it up for review now.

mulkieran:
Well, a further ultimate goal is to determine if the tests can be modified to be useful when detecting encrypted devices. So actually logging is a preliminary goal, but still a bit far.

PR #1891, opened 2020-03-17
Tidy up udev tests further

mulkieran:
This PR simplifies and clarifies a bunch of code and allows the stratisd log level to be determined by the environment variable RUST_LOG set in the CI script that runs the tests. It also makes all tests have the uniform structure most suitable for tests, simplifying the one test that is a loop of repeated tests into a single test.

It does not yet make the stratisd logs actually available; there are at least three ways to go about this:
1. Tackle the problem with the tests as is.
2. Convert to a pure unittest setup, and then tackle the problem.
3. Convert to a pure pytest setup, and then tackle the problem.

There are a bunch of other things that the PR doesn't address:
1. Any individual test may be very long, resulting in voluminous logs.
2. Some tests lack realism, for example the order in which devices are made to come online is dependent on the order in which they occur in their pools.
3. etc.

### DIFF
--- a/tests/client-dbus/tests/udev/_loopback.py
+++ b/tests/client-dbus/tests/udev/_loopback.py
@@ -23,6 +23,8 @@ import uuid
 
 _LOSETUP_BIN = os.getenv("STRATIS_LOSETUP_BIN", "/usr/sbin/losetup")
 
+_SIZE_OF_DEVICE = 1024 ** 4  # 1 TiB
+
 
 class LoopBackDevices:
     """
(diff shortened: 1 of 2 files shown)


mulkieran:
Still not displaying. The test infrastructure is capturing the output because the test has not failed. We need to convince it not to.

jbaublitz:
I've worked on this issue before. There may be a way but it's a little tricky because of the layering of pytest and unittest. I know unittest replaces `sys.stdout` with something else and pytest has an explicit modification of the underlying file descriptors, not just the Python objects. These two things together make it a little bit challenging.

mulkieran:
> I've worked on this issue before. There may be a way but it's a little tricky because of the layering of pytest and unittest. I know unittest replaces `sys.stdout` with something else and pytest has an explicit modification of the underlying file descriptors, not just the Python objects. These two things together make it a little bit challenging.

I recall you looking into that a bit, late last year, and I remember a bit. We might take the extreme route, and drop the use of pytest for these tests. But right now, I'm gonna try to make an error happen, and perhaps that will allow the logs to be displayed.

mulkieran:
nonrust tests are expected to fail. I want to see what the logs entries from stratisd look like for a failed test. They may be voluminous.

mulkieran:
whoops! Included some unintended junk.

mulkieran:
> > I've worked on this issue before. There may be a way but it's a little tricky because of the layering of pytest and unittest. I know unittest replaces `sys.stdout` with something else and pytest has an explicit modification of the underlying file descriptors, not just the Python objects. These two things together make it a little bit challenging.
> 
> I recall you looking into that a bit, late last year, and I remember a bit. We might take the extreme route, and drop the use of pytest for these tests. But right now, I'm gonna try to make an error happen, and perhaps that will allow the logs to be displayed.

Another route is to make it pure pytest, rather than pure unittest. It is possible, even likely, that pytest has more help for this kind of problem, there are only about six actual tests, and a huge amount of infrastructure that wouldn't have to be changed at all, and Hypothesis works better with pytest.

Issue #1855, opened 2020-02-28
Would like to capture output of stratisd logs for udev tests


mulkieran:
Right now, the logs could give us some information of how the devices are picked up, if they were available.

PR #1893, opened 2020-03-19
More udev test tidies

mulkieran:
This does not yet display stratisd log entries, because most likely some of the work for that needs to be done on stratisd itself, not just the tests:

The contributions of this PR are:
* It moves support code out of the test class, and into the containing module. This means that fewer methods take self as an argument unnecessarily, and that only test methods are in the test class.
* It changes the behavior of LoopBackDevices.device_file so that it raises some kind of exception if there is no device file. Previously, it returned None, in both cases. In both cases, this was a mistake.
* It removes some deprecated pyudev usage. This usage was deprecated just one release, but also 4 years ago. High time to drop it, especially since it floods the test output with deprecation warnings whenever there is a failure.
* It uses a context manager for starting and stopping the daemon. This helps clarify the test code; and is likely to make the eventual displaying of the logs easier.

Related: #1855.

### DIFF
--- a/tests/client-dbus/tests/udev/_loopback.py
+++ b/tests/client-dbus/tests/udev/_loopback.py
@@ -118,8 +118,11 @@ class LoopBackDevices:
         :param token: Opaque representation of some loop back device
         :return: Full file path or None if not currently attached
         :raises: KeyError if token is unknown
+        :raises: AssertionError if devnode is None
         """
-        return self.devices[token][0]
+        devnode, _ = self.devices[token]
+        assert devnode is not None
+        return devnode
 
     def destroy_devices(self):
         """
--- a/tests/client-dbus/tests/udev/_loopback.py
+++ b/tests/client-dbus/tests/udev/_loopback.py
@@ -117,8 +117,9 @@ class LoopBackDevices:
         Return the block device full name for a loopback token
         :param token: Opaque representation of some loop back device
         :return: Full file path or None if not currently attached
+        :raises: KeyError if token is unknown
         """
-        return self.devices[token][0] if token in self.devices else None
+        return self.devices[token][0]
 
     def destroy_devices(self):
         """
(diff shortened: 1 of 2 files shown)


mulkieran:
Stdout and stderr are both displayed when -s flag is set.

mulkieran:
Try again with "-s" not set.

mulkieran:
Ok. We do need that "-s".

mulkieran:
Capturing the output, it is not None, but it shows up as empty. This is probably a problem on the stratisd side. Perhaps the buffered logging is causing nothing to be sent to stderr.

mulkieran:
Yeah. If we use the "--debug" flag, the log's aren't buffered and _everything_ comes spewing out.

mulkieran:
Yeah, on SIGINT the buffered logs get printed out. On SIGTERM, nothing, it seems.

```
[root@localhost stratisd]# ./target/debug/stratisd --sim
^CLog handle going out of scope; writing contents of log buffer to log
DEBUG libstratis::stratis::buff_log: 2020-03-23 22:42:47.373385359 UTC BuffLogger: pass_through: false hold time: PT1800S
 INFO stratisd: 2020-03-23 22:42:47.374507456 UTC stratis daemon version 2.0.1 started
 INFO stratisd: 2020-03-23 22:42:47.374541657 UTC Using SimEngine
DEBUG stratisd: 2020-03-23 22:42:47.374887638 UTC Engine state: 
SimEngine {
    pools: {},
    rdm: RefCell {
        value: {Randomizer 0,
    },
}
 INFO stratisd: 2020-03-23 22:42:47.376598284 UTC D-Bus API is available
 INFO stratisd: 2020-03-23 22:44:01.844717646 UTC SIGINT received, exiting
stratisd is quitting!
[root@localhost stratisd]# ./target/debug/stratisd --sim
Terminated
```

I think we need to backup and improve stratisd's handling of command-line arguments.

PR #1897, opened 2020-03-24
Sleep one second between calls to now()

### LINKED ISSUE #1896: test_early_times_err fails
```rust
[  661s] ---- engine::strat_engine::backstore::metadata::bda::tests::test_early_times_err stdout ----
[  661s] thread 'engine::strat_engine::backstore::metadata::bda::tests::test_early_times_err' panicked at 'assertion failed: `(left != right)`
[  661s]   left: `2020-03-24T06:09:36.646479850Z`,
[  661s]  right: `2020-03-24T06:09:36.646479850Z`', src/engine/strat_engine/backstore/metadata/bda.rs:197:9
[  661s] note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Just tried to build latest fedora package again... Any ideas what could be the reason?

mulkieran:
closes #1896

### DIFF
--- a/src/engine/strat_engine/backstore/metadata/bda.rs
+++ b/src/engine/strat_engine/backstore/metadata/bda.rs
@@ -139,7 +139,7 @@ impl BDA {
 
 #[cfg(test)]
 mod tests {
-    use std::io::Cursor;
+    use std::{io::Cursor, thread, time};
 
     use proptest::{collection::vec, num};
 
@@ -173,6 +173,7 @@ mod tests {
     /// of saved data is older than timestamp of most recently written data.
     fn test_early_times_err() {
         let data = [0u8; 3];
+        let sleep_time = time::Duration::from_secs(1);
 
         // Construct a BDA.
         let sh = random_static_header(0, 0);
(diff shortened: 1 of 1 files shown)


mulkieran:
rebased.

src/engine/strat_engine/backstore/metadata/bda.rs
```diff
@@ -187,8 +188,8 @@ mod tests {
         .unwrap();
 
         let timestamp0 = Utc::now();
+        thread::sleep(sleep_time);
         let timestamp1 = Utc::now();
-        assert_ne!(timestamp0, timestamp1);
```
jbaublitz: Just one question before I approve this. As I recall from reading the metadata document, if both times are the same, one timestamp is incremented, correct? So is this why we don't need to assert that they're different?

mulkieran: I have decided to trust the sleep() guarantee implicitly, that 1 second must elapse between calls. The assertion was there before because I had a hunch that the problem we're encountering now might sometime occur, that it wasn't absolutely guaranteed not to occur...and since the test wouldn't work w/out it, an assert was appropriate.

PR #1898, opened 2020-03-24
Simplify nested match expression

lleshchi:
Signed-off-by: Leah Leshchinsky <lleshchi@redhat.com>

### DIFF
--- a/src/engine/strat_engine/backstore/metadata/static_header.rs
+++ b/src/engine/strat_engine/backstore/metadata/static_header.rs
@@ -235,6 +235,33 @@ impl StaticHeader {
             }
         }
 
+        // Action taken when both sigblocks are interpreted as valid.
+        //
+        // If both sigblocks are interpreted as a Stratis headers,
+        // compare contents of static headers.
+        //
+        // If only a single sigblock is interpreted as a Stratis header,
+        // overwrite the other sigblock with the contents of the valid
+        // Stratis header sigblock.
+        //
+        // If neither sigblocks are valid Stratis headers,
+        // return Ok(None)
+        fn ok_ok_static_header_handling<F>(
+            f: &mut F,
+            maybe_sh1: Option<StaticHeader>,
+            maybe_sh2: Option<StaticHeader>,
+        ) -> StratisResult<Option<StaticHeader>>
+        where
+            F: Seek + SyncAll,
+        {
+            match (maybe_sh1, maybe_sh2) {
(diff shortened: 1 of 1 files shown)


src/engine/strat_engine/backstore/metadata/static_header.rs
```diff
@@ -235,6 +235,33 @@ impl StaticHeader {
             }
         }
 
+        // Action taken when both sigblocks are interpreted as valid.
```
mulkieran: This is not quite correct. It's more like: ```Action taken when both sigblock locations are analyzed without encountering an error.``` The important distinction is that if an error were encountered, then it would be uncertain whether their was Stratis metadatad there or not. In this method, it is completely known whether there is a sigblock there.

src/engine/strat_engine/backstore/metadata/static_header.rs
```diff
@@ -235,6 +235,33 @@ impl StaticHeader {
...
+        //
+        // If both sigblocks are interpreted as a Stratis headers,
+        // compare contents of static headers.
+        //
+        // If only a single sigblock is interpreted as a Stratis header,
+        // overwrite the other sigblock with the contents of the valid
+        // Stratis header sigblock.
+        //
+        // If neither sigblocks are valid Stratis headers,
```
mulkieran: Number of verb "are" does not agree with subject "neither".

src/engine/strat_engine/backstore/metadata/static_header.rs
```diff
@@ -309,12 +337,7 @@ impl StaticHeader {
         ) {
             // We read both copies without an IO error.
             (Ok(buf_loc_1), Ok(buf_loc_2)) => match (buf_loc_1, buf_loc_2) {
-                (Ok(loc_1), Ok(loc_2)) => match (loc_1, loc_2) {
-                    (Some(loc_1), Some(loc_2)) => compare_static_headers(f, loc_1, loc_2),
-                    (None, None) => Ok(None),
-                    (Some(loc_1), None) => write_header(f, loc_1, MetadataLocation::Second),
-                    (None, Some(loc_2)) => write_header(f, loc_2, MetadataLocation::First),
-                },
+                (Ok(loc_1), Ok(loc_2)) => ok_ok_static_header_handling(f, loc_1, loc_2),
```
mulkieran: Please get rid of the inline comments just above and also the few below. They were put in when the body of the method was about five times as long as it is ow, to serve as some sort of sign posts. But since all the improvements have been made to the method, they are a bit worse than redundant.

PR #1899, opened 2020-03-25
Capture logs from stratisd on udev tests

### LINKED ISSUE #1855: Would like to capture output of stratisd logs for udev tests

mulkieran:
closes #1855


mulkieran:
The logs are a bit voluminous, and the test output is harder to read now, but I think that for now it is very valuable to see them. If we keep this logging output for a while, I'll take additional steps to make the log chunks a bit more navigatable.

mulkieran:
Travis CI build verified.

Issue #1901, opened 2020-03-25
No longer dump the entire engine state into the logs


mulkieran:
The engine state is potentially quite useful for debugging, but it is extremely inconvenient when it ends up in the logs. Our ultimate plan is to make it available via some other mechanism, possibly a D-Bus report interface. But we want remove it from the logs now, as, in its current form, it is mainly a burden to users trying to read the logs.

mulkieran:
Note that the management of this logging seems to be confined to ```bin/stratisd.rs```.

PR #1904, opened 2020-03-25
Do not dump state on timer expiration

bgurney-rh:
Related: #1901

### DIFF
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -57,11 +57,6 @@ fn print_err(err: &StratisError) {
     eprintln!("{}", err);
 }
 
-/// Log the engine state in a formatted way.
-fn log_engine_state(engine: &dyn Engine) {
-    debug!("Engine state: \n{:#?}", engine);
-}
-
 /// Configure the env_logger as necessary in order to allow the buffered
 /// logger to work correctly. Return a Handle to the underlying env_logger.
 pub fn from_env_logger(
@@ -425,8 +420,6 @@ fn run(matches: &ArgMatches, buff_log: &buff_log::Handle<env_logger::Logger>) ->
     } else {
         FD_INDEX_ENGINE
     };
-
-    log_engine_state(&*engine.borrow());
 
     loop {
         if fds[FD_INDEX_UDEV].revents != 0 {
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -410,12 +410,6 @@ fn run(matches: &ArgMatches, buff_log: &buff_log::Handle<env_logger::Logger>) ->
         SetTimeFlags::Default,
     );
 
-    fds.push(libc::pollfd {
-        fd: tfd.as_raw_fd(),
-        revents: 0,
-        events: libc::POLLIN,
-    });
-
     let eventable = engine.borrow().get_eventable();
 
     if let Some(evt) = eventable {
--- a/src/bin/stratisd.rs
+++ b/src/bin/stratisd.rs
@@ -362,7 +362,7 @@ fn run(matches: &ArgMatches, buff_log: &buff_log::Handle<env_logger::Logger>) ->
     */
     const FD_INDEX_UDEV: usize = 0;
     const FD_INDEX_SIGNALFD: usize = 1;
-    const FD_INDEX_ENGINE: usize = 2 ;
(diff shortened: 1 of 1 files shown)


src/bin/stratisd.rs
```diff
@@ -363,7 +363,6 @@ fn run(matches: &ArgMatches, buff_log: &buff_log::Handle<env_logger::Logger>) ->
     */
     const FD_INDEX_UDEV: usize = 0;
     const FD_INDEX_SIGNALFD: usize = 1;
-    const FD_INDEX_DUMP_TIMERFD: usize = 2;
```
mulkieran: Mind the gap! You're now missing a value in the order, and that's probably going to cause trouble. Pay close attention to how the indices for ```fds``` are managed.

bgurney-rh: I also need to update the comments above.  Hopefully that should be done now.

src/bin/stratisd.rs
```diff
@@ -355,16 +355,14 @@ fn run(matches: &ArgMatches, buff_log: &buff_log::Handle<env_logger::Logger>) ->
...
     1   == SIGNAL FD index
-    2   == TIMER FD for periodic dump index
-    3   == engine index if eventable
-    3/4 == Start of dbus client file descriptor(s)
-            * 3 if engine is not eventable
-            * else 4
+    2   == engine index if eventable
+    2/3 == Start of dbus client file descriptor(s)
+            * 2 if engine is not eventable
+            * else 3
```
jbaublitz: I'm a little confused by the documentation here. This seems like it was here before you touched it, but from my reading, the first line seems to indicate that 2 is eventable and the second bullet point says 2 if not eventable. @mulkieran can you potentially shed a little light on what this might be trying to communicate?

mulkieran:
> We are still logging engine state [here](https://github.com/stratis-storage/stratisd/pull/1904/files#diff-7435f55ae52c4ef2caf7d613cf4793e1L437) (on line 437). Do we want to do an initial dump or remove all dumps all together?

I would like to remove all dumps, including the initial one, due to the fact that any dump is going to clutter the log pretty badly.

mulkieran:
https://github.com/stratis-storage/stratisd/issues/1906 is just slightly related, in that by removing all unnecessary file descriptor handling, we make the move to any other mechanism just a bit easier.

src/bin/stratisd.rs
mulkieran: > I'm a little confused by the documentation here. This seems like it was here before you touched it, but from my reading, the first line seems to indicate that 2 is eventable and the second bullet point says 2 if not eventable. @mulkieran can you potentially shed a little light on what this might be trying to communicate?

The dbus index start gets shifted left to 2 if the engine is not eventable; the sim engine actually isn't eventable.

jbaublitz: Oh okay! It was a little bit hard for me to parse at first, but this makes a lot more sense.

bgurney-rh:
If I delete that last log_engine_state() call, I get a build error of "error: function is never used: 'log_engine_state'".

We probably do want to be able to log engine state at an arbitrary time, just not on every single startup.  Could we add a call that's triggered by a signal?

jbaublitz:
> If I delete that last log_engine_state() call, I get a build error of "error: function is never used: 'log_engine_state'".
> 
> We probably do want to be able to log engine state at an arbitrary time, just not on every single startup. Could we add a call that's triggered by a signal?

Yesterday we discussed in stand up that we intend to expose that over the D-Bus through a different mechanism that does not involve logging at a later time. For now, I think you can remove this method.

src/bin/stratisd.rs
```diff
@@ -412,12 +405,6 @@ fn run(matches: &ArgMatches, buff_log: &buff_log::Handle<env_logger::Logger>) ->
         SetTimeFlags::Default,
     );
 
-    fds.push(libc::pollfd {
```
mulkieran: Keep on removing! Since ```tfd``` is no longer put in the ```fds``` array, it doesn't need to exist at all. So you can remove all the code that sets it up as well, and once you've removed that, you should find some other things that only exist because they are necessary for setting up ```tfd```, and then you can remove those as well.

bgurney-rh:
OK; I'm closing this, and creating a final version.

Issue #1892, opened 2020-03-18
Remove "--invalid-name" flags from pylint calls in check.py


mulkieran:
We should make the push to make everything consistent. These invalid names are grandfathered in unnecessarily.

mulkieran:
@lleshchi Now that I've pushed #1909, you can get started whenever is convenient.

PR #1909, opened 2020-03-27
Do not dump lots and lots of text on certain failures in tests

mulkieran:
Related: #1656.
Related: #1892


mulkieran:
This is related to #1892, in the sense that I was looking to see if the volume of code that has pylint errors could be reduced. I decided to be more conservative than I had planned, so I haven't reduced the number of pylint errors, just split them across more directories.

mulkieran:
rebased.

PR #1908, opened 2020-03-26
Crypt module without cryptsetup-2.2 compatibility

jbaublitz:
Related to #1888

PR #1910, opened 2020-03-27
Add crypt module and use for device initialization

### LINKED ISSUE #1822: Design an interface for the explicit part of reading and writing luks2 metadata

jbaublitz:
Supersedes #1908 
Closes #1822

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -329,25 +329,27 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "libcryptsetup-rs"
-version = "0.2.0"
+version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "either 1.5.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.66 (registry+https://github.com/rust-lang/crates.io-index)",
- "libcryptsetup-rs-sys 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libcryptsetup-rs-sys 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "pkg-config 0.3.9 (registry+https://github.com/rust-lang/crates.io-index)",
+ "semver 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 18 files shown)


jbaublitz:
@mulkieran One question before we merge. Do we want to release a new version of libcryptsetup-rs first so we can fix Cargo.toml?

mulkieran:
I missed the question!

mulkieran:
@jbaublitz I think it would be better if you cut a new release of cryptsetup; it would make our Cargo.toml a little more standard and be a kind of checkpoint for the cryptsetup bindings.

jbaublitz:
Travis is not responding. Is it sufficient to rely on Jenkins here or should I open a new PR?

PR #1895, opened 2020-03-23
Add macro log events to pool create, rename, and destroy operations

bgurney-rh:
Supersedes #1874

Issue #1877, opened 2020-03-10
Need to distinguish PoolUuid, DevUuid, FilesystemUuid by something more than aliases


mulkieran:
Currently, all these are just aliases for Uuid. This will make some definitions necessary for https://github.com/stratis-storage/stratisd/pull/1874 to have conflicts like:
```
error[E0119]: conflicting implementations of trait `std::fmt::Display` for type `engine::types::actions::RenameAction<uuid::Uuid>`:
  --> src/engine/types/actions.rs:33:1
   |
27 | impl fmt::Display for RenameAction<PoolUuid> {
   | -------------------------------------------- first implementation here
...
33 | impl fmt::Display for RenameAction<DevUuid> {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `engine::types::actions::RenameAction<uuid::Uuid>`

error: aborting due to previous error
```

jbaublitz:
@mulkieran I forgot to mention that I looked at this in stand up.

I see two potential paths forward:
1. Wrapping `Uuid` in three distinct structs and exposing the same methods provided by `Uuid` for `DevUuid`, etc. (so `DevUuid` would be a wrapping struct also with the method `::nil()` and `.to_simple_ref()` available.
2. Wrapping `Uuid` in three distinct structs and relying on modifying the `Uuid` code to rely on a `Deref` implementation.

Despite the fact that 2 will probably result in more manual wrangling of dereferences throughout the code, I prefer this solution as it will leave to less of what seems to me like a bandaid fix and will instead allow us to keep `Uuid` completely separate logically from `DevUuid` and friends.

mulkieran:
Thanks for looking at this. Let's talk about it in standup tomorrow. I don't understand the distinction between the two that you are drawing.

jbaublitz:
Sounds good!

mulkieran:
@jbaublitz I'ld like to restrict the impl to the three Ds, Deref, Debug, and Display, as below, as I'ld like the Display info to include the particular category of UUID. What do you think? (Ultimately, of course, we'ld define the type for each category via a single macro, as they would be identical modulo the name of the type and the Display string, so this really constitutes an example that we'ld follow exactly for PoolUuid and DevUuid. The macro would also help enforce the restriction that these types don't implement any other methods.)

```
use std::fmt;
use std::ops::Deref;

use uuid::Uuid;

[...]

PR #1912, opened 2020-03-27
Add macro log events to pool create, rename, and destroy operations

bgurney-rh:
Supersedes #1895


bgurney-rh:
Rebased.

mulkieran:
I want to hold this until #1817 gets merged, as it changes the behavior of CreatePool a bit. Not guaranteed that there will be a necessary change to this PR, but waiting until it gets merged is less work than figuring out what will happen.

mulkieran:
Might be blocked by #1877 as well, depending on the implementation choice.

bgurney-rh:
> Please take a look at the changes suggested in #1915. Two are cosmetic, but the last commit is fairly significant, since it swaps the order of the action string and the other post-operation strings.

Looks good; I just applied them.

mulkieran:
Superceded by #2288

PR #1914, opened 2020-03-30
Add stratisd build checks for external dependency versions

### LINKED ISSUE #1913: Consider enforcing non-Rust dependencies via a build script
stratisd requires cryptsetup version 2.3.0. We expect that downstream packagers will enforce this via packaging. However, when building from source in the presence of cryptsetup version 2.2.0, the result is a compiler error which is shown below. Although we document the mandatory dependency, we should also consider a build script which enforces it, to avoid any possible confusion among users building from source. But, if we enforce this dependency via a build script it makes sense to also enforce our libblkid dependency since running with too low a version of libblkid would be expected to result in stratisd failing to locate existing Stratis devices on startup.

```
 Compiling libstratis v2.0.1 (/home/mulhern/stratis2/stratisd)
error[E0599]: no function or associated item named `alloc` found for struct `libcryptsetup_rs::mem::SafeMemHandle` in the current scope
   --> src/engine/strat_engine/backstore/crypt.rs:693:45
    |
693 |         let mut key_buffer = SafeMemHandle::alloc(STRATIS_KEY_SIZE)?;
    |                                             ^^^^^ function or associated item not found in `libcryptsetup_rs::mem::SafeMemHandle`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `libstratis`.

To learn more, run the command again with --verbose.
make: *** [Makefile:50: build] Error 101
```

jbaublitz:
Closes #1913

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,10 @@ language: rust
 # Use a package which supplies cryptsetup 2.3.0 which is required
 # by features added in libcryptsetup-rs that are used in stratisd.
 before_install:
+  - sudo add-apt-repository -y "deb http://us.archive.ubuntu.com/ubuntu/ eoan main" 
   - sudo apt-get -q update
+  # Update to a more recent version of blkid
+  - sudo apt-get -y install blkid libblkid-dev
   # cryptsetup-bin conflicts with custom built cryptsetup
   - sudo apt-get remove cryptsetup-bin
   - sudo apt-get install -y libargon2-0 libjson-c3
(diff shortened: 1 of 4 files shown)


mulkieran:
@jbaublitz: I'll let you look into the errors...

jbaublitz:
@bgurney-rh It looks like stratisd tests have a lingering device. Other than that, Travis is fixed.

bgurney-rh:
I cleaned it up; this test seemed to leave a device behind when it was interrupted:
```
test engine::strat_engine::backstore::devices::tests::real_test_crypt_ownership ... Build was aborted
```

jbaublitz:
Thanks @bgurney-rh! I was hoping it wasn't related to the crypt module.

PR #1922, opened 2020-03-31
Fix bug on initialization failure and add test

### LINKED ISSUE #1920: Crypt module device wiping bug
Current method `ensure_inactive` in the crypt module does not actually have idempotent behavior as the name and documentation suggests. Due to the libcryptsetup deactivation methods not being idempotent and rather returning an error if an inactive device is deactivated, our wiping method (which relies on idempotently ensuring that devices are inactive) will return an error if the device is already inactive. This is triggered in the case of rolling back an initialization if activation could not be accomplished but the LUKS2 header has been written to the device. The crypt module will attempt to deactivate a device that has never been activated and will return early prior to wiping the LUKS2 metadata from the device. This leaves the encrypted device in a state where it cannot be initialized again without first wiping the device with a utility outside of Stratis.

@mulkieran This is a bug, not a problem with udev after all. I did some digging down to the superblock level and discovered this issue.

jbaublitz:
Closes #1920


mulkieran:
@jbaublitz Please rebase, thanks.

jbaublitz:
Rebased.

PR #1917, opened 2020-03-31
Conform to pylint naming requirements

lleshchi:
Related #1892

PR #1928, opened 2020-04-01
Conform to pylint naming requirements

### LINKED ISSUE #1892: Remove "--invalid-name" flags from pylint calls in check.py

lleshchi:
Related #1917 
Closes #1892

### DIFF
--- a/scripts/stratis_id.py
+++ b/scripts/stratis_id.py
@@ -114,7 +114,10 @@ def dump_stratis_signature_area(block_device):
         )
 
 
-if __name__ == "__main__":
+def main():
+    """
+    The main method
+    """
     if len(sys.argv) != 2:
         print("syntax: stratis_signature.py <block device>")
         sys.exit(2)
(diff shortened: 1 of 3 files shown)


scripts/stratis_id.py
```diff
@@ -122,9 +121,9 @@ def dump_stratis_signature_area(block_device):
 
     dump_stratis_signature_area(sys.argv[1])
 
-    sig = stratis_signature(sys.argv[1])
```
mulkieran, line 125: For the code in the ```if __name__ == "__main__":``` define a main method and just invoke that, as

```
if __name__ == "__main__":
  main()
```

That way, the code is better and pylint doesn't insist that those identifiers need to be capitalized like they're constants. (They _are_ constants, according to Python, which is why pylint wants them capitalized. But that's not what they're _supposed_ to be, and writing a main method will fix that.)

PR #1929, opened 2020-04-02
Remove process_unowned_devices

mulkieran:
Related: #1656
Related: #1782

Also removes find_all_block_devices_with_stratis_signatures, a method used only for testing. In one case, a test was removed. For all other tests, a different mechanism for getting the necessary information was used. find_all_block_devices_with_stratis_signatures depended transitively on process_unowned_devices, and failed to find all devices if the use in identify_block_device was removed. The use in tests was the only reason to retain process_unowned_devices; at this point it can be removed entirely.


mulkieran:
rebased.

PR #1930, opened 2020-04-02
Change allow to cfg(test)

mulkieran:


mulkieran:
Trivial, merging as soon as is convenient.

PR #1931, opened 2020-04-02
Fixes for devices.rs and crypt.rs existing implementation

jbaublitz:
Fixes for #1910 

Adds fixes for mapping physical device size to logical device size and an early return error in the `crypt.rs` module where querying the metadata for keyslot information should not result in an early return.


jbaublitz:
Rebased.

Issue #1870, opened 2020-03-06
Tracking issue for proposed libc changes


mulkieran:
https://github.com/rust-lang/libc/pull/1678

jbaublitz:
Merged.

mulkieran:
Let's hang onto this until they release the new version.

mulkieran:
We'll check back on the release next month.

jbaublitz:
As a last ditch effort, we can submit a PR to bump the version if it comes down to that. They've been very helpful when I've done that in the past and only have you wait if there's something specific holding it up.

mulkieran:
Ok. Depends on what you think we need.

jbaublitz:
I think we can hold off until we're ready to release, but we can try that at that point if there's been no new release.

mulkieran:
Excellent. Feel free to do that as soon as is convenient.

PR #1933, opened 2020-04-02
Use constants exposed by libc version 0.2.68

mulkieran:
Related: #1870


mulkieran:
This was missed previously because this code is in the testing submodule.

I ran across it while reviewing #1932 

Unimportant detail: libc's ```KEYCTL_UNLINK``` is u32 instead of i32. Obviously the representation is the same, since 9 is quite small.

PR #1936, opened 2020-04-03
Verify correct behavior of pre-initialization device processing in the presence of encryption

mulkieran:
* Improves encapsulation by moving one function into the devices module
* Extends one test that is already running in the presence of encryption to verify correct operation
* Fixes one bug in the processing of devices specified for initialization


mulkieran:
On Travis, just a formatting error.

PR #1938, opened 2020-04-03
Remove unnecessary cfg directives

jbaublitz:
Related #1908 

When removing the theoretical cryptsetup 2.2 support, this seems to have
made it into the codebase anyway. Remove as it is now unnecessary.


jbaublitz:
Rebased.

PR #1939, opened 2020-04-03
Rework crypt device naming

### LINKED ISSUE #1935: Refine names of dm-crypt devices synthesized during initialization
Currently, the names we use have format ```<pool uuid in simple form>-<device uuid in simple form>```.

These names are public enough that we should disambiguate them with a "stratis" prefix of some sort, as we do with other names.

Given that they can't be viewed as a private detail of the encryption mechanism, we should define the names more formally, in our names.rs module.

We have a countervailing consideration: the DM UUIDs synthesized by cryptsetup from the names specified are 111 characters, the hard limit on UUID length is 129. If cryptsetup changes its DM UUID synthesis algorithm, and makes the UUID longer, it could overflow the allowed length of the UUID.

If we add a prefix, like "stratis-", that's 8 characters eating up the remaining free space available for the DM UUID.

Should we shorten our name in order to avoid this? ~~There is an obvious way to cut the number of characters used to represent a UUID by half; just use the string representation of the raw number, rather than the string representation of the hexadecimal representation of the number.~~ This does not work, because it omits considerations of the representation. We can _store_ the UUIDs in half the bits; but representing them in the names is a whole different thing.

This is a precaution against future changes in cryptsetup, but it might be a prudent choice.

jbaublitz:
Closes #1935

### DIFF
--- a/src/engine/strat_engine/names.rs
+++ b/src/engine/strat_engine/names.rs
@@ -18,7 +18,17 @@ use crate::{
 
 const FORMAT_VERSION: u16 = 1;
 
-/// Get a devicemapper name from the device UUIDs.
+/// Get a devicemapper name from the device UUID.
+///
+/// Prerequisite: len(format!("{}", FORMAT_VERSION)
+///             + len("stratis")                         7
+///             + len("private")                         7
+///             + len("crypt")                           5
+///             + num_dashes                             4
+///             + len(dev uuid)                          32
+///             < 128
+///
+/// which is equivalent to len(format!("{}", FORMAT_VERSION) < 73
 pub fn format_crypt_name(dev_uuid: &DevUuid) -> String {
     format!(
         "stratis-{}-private-{}-crypt",
--- a/src/engine/strat_engine/names.rs
+++ b/src/engine/strat_engine/names.rs
@@ -12,11 +12,20 @@ use std::{
 use devicemapper::{DmNameBuf, DmUuidBuf};
 
 use crate::{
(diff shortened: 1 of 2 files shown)


bgurney-rh:
A couple of example device names after this change are below:

```
stratis-1-private-2960f6d0670344b29799fc9f6d70b848-crypt
stratis-1-private-56d881a5fb874e41863eb59983b49fc1-crypt
```

src/engine/strat_engine/names.rs
```diff
@@ -12,12 +12,21 @@ use std::{
...
 
 use crate::{
-    engine::types::{FilesystemUuid, PoolUuid},
+    engine::types::{DevUuid, FilesystemUuid, PoolUuid},
     stratis::{ErrorEnum, StratisError, StratisResult},
 };
 
 const FORMAT_VERSION: u16 = 1;
 
+/// Get a devicemapper name from the device UUIDs.
```
mulkieran, line 21: Please change to singular.

Also, put in proof the name is short enough as:

```
Prerequisite: len(format!("{}", FORMAT_VERSION) 
                     + len("stratis")         7
                     + len("private")        7
                     + len("crypt")            5
                     + num_dashes        4
                     + len(dev_uuid)     32
                     < 128
which is equivalent to len(format!("{}", FORMAT_VERSION) < 73
```

see examples below.

jbaublitz: Resolved.

PR #1944, opened 2020-04-07
Jbaublitz encryption me 3

mulkieran:
The two additional commits do two very different things.

* The first drops encryption from the cache. We decided that we would be more minimalist and not encrypt the cache, and we should stick with that.
* The second is more of a venture. My first thought was just to drop the key_desc() method, since it had been rendered dead by the first commit. But then it occurred to me that it would be a great convenience to us to make it available on the D-Bus. The question then comes up;  how secure is it if the key description information is available to any user, not just root? I'm currently seeking further information on that subject.


jbaublitz:
I'm a little confused about these changes. While we are planning to make init_cache mutually exclusive with encryption for this release, the reason for this design was to enable this to be added in the future if we eventually decide to add an encrypted cache at some point. This will not change that the D-Bus API does not expose encryption for the cache so this feels to me like we're removing work that may come up again in the future. Can you give me a little more insight into what is driving this?

mulkieran:
> I'm a little confused about these changes. While we are planning to make init_cache mutually exclusive with encryption for this release, the reason for this design was to enable this to be added in the future if we eventually decide to add an encrypted cache at some point. This will not change that the D-Bus API does not expose encryption for the cache so this feels to me like we're removing work that may come up again in the future. Can you give me a little more insight into what is driving this?

My first objection to leaving the key description as an argument to ```init_cache``` in at this time is that it requires global reasoning about the program for the programmer to be able to figure out that at the point in the D-Bus layer where the key description is extracted from the pool and passed to the ```init_cache``` method it must always be ```None``` and therefore the key description argument to ```init_cache``` is always ```None```. Once they figure out it must always be ```None``` they will wonder why it wasn't just ```None``` in the first place to save them a bunch of hard work figuring out what is going on. And once they get there, they'll wonder why, if it is always ```None```,  it is an argument at all. I want to elide all that, and save them the trouble, by removing it now.

The idea is not fundamentally to change our design or our expectations about eventually encrypting the cache, just to remove  a small bit of very confusing code. Because of the difficulties of supporting multiple versioned interfaces and changing the semantic of ```add_cachdevs``` between those two interfaces that area of code already is  difficult to follow. The other confusing code is necessary for proper operation, but this code is not, so it is best to remove it.

jbaublitz:
> > I'm a little confused about these changes. While we are planning to make init_cache mutually exclusive with encryption for this release, the reason for this design was to enable this to be added in the future if we eventually decide to add an encrypted cache at some point. This will not change that the D-Bus API does not expose encryption for the cache so this feels to me like we're removing work that may come up again in the future. Can you give me a little more insight into what is driving this?
> 
> My first objection to leaving the key description as an argument to `init_cache` in at this time is that it requires global reasoning about the program for the programmer to be able to figure out that at the point in the D-Bus layer where the key description is extracted from the pool and passed to the `init_cache` method it must always be `None` and therefore the key description argument to `init_cache` is always `None`. Once they figure out it must always be `None` they will wonder why it wasn't just `None` in the first place to save them a bunch of hard work figuring out what is going on. And once they get there, they'll wonder why, if it is always `None`, it is an argument at all. I want to elide all that, and save them the trouble, by removing it now.
> 
> The idea is not fundamentally to change our design or our expectations about eventually encrypting the cache, just to remove a small bit of very confusing code. Because of the difficulties of supporting multiple versioned interfaces and changing the semantic of `add_cachdevs` between those two interfaces that area of code already is difficult to follow. The other confusing code is necessary for proper operation, but this code is not, so it is best to remove it.

Okay this makes a little more sense to me. Thanks for explaining!

mulkieran:
Merged into #1950.

PR #1945, opened 2020-04-07
Revert "Require error-chain 0.12.2"

bgurney-rh:


mulkieran:
We can merge as soon as we confirm that the blackbox tests are back to passing.

bgurney-rh:
I just hand-applied this to a retrofitted ci repository where a previous attempt failed.  The retrofit would preserve the existing "stratisd" directory, but clone a new stratis-cli directory, since the stratis-blackbox-test script failed before that point.

All blackbox tests passed.

PR #1947, opened 2020-04-07
Relax size requirement on Stratis passphrases

### LINKED ISSUE #1921: Relax restrictions on encryption passphrase
Our current encryption implementation relies on the user providing a 512-bit randomly generated passphrase in the kernel keyring. After discussing how the KEK is generated from the passphrase supplied by users in cryptsetup, it seems that we should be able to relax this restriction on the passphrase with no adverse effects.

Things to consider:
* Password checking library to check for passphrase strength.
* Allowing variable-length passphrases and clearly document the maximum length.
* We should still support randomly generated passphrases even with non-printable characters if the user would prefer to use a byte-based key as opposed to a traditional passphrase with printable characters only.

jbaublitz:
Closes #1921


jbaublitz:
Rebased.

PR #1954, opened 2020-04-09
Rework disk scanning strategy in tests

jbaublitz:
See the commit message for an explanation of the changes.


jbaublitz:
@mulkieran Sounds good. I will rebase now. I tried to see if there was any event driven way to wait until a device was no longer busy, but I was unable to find any.

jbaublitz:
Rebased.

mulkieran:
In standup, we decided to throw in a few FIXME's to mark the spots were we might want to make changes later. The one about using "nix" instead of "libc" is more of a TODO than a FIXME; it needs a closer look to absolutely decide if it can be done.

PR #1957, opened 2020-04-13
Log at the info level if skipping initialization on a stratis device

mulkieran:


mulkieran:
@bgurney-rh: Once this is merged, please run the blackbox tests again and see if this info message shows up in the stratisd logs. For the CLI tests, I believe the message should never show up.

PR #1958, opened 2020-04-14
Rework disk scanning strategy in tests

jbaublitz:
Supersedes #1954

### DIFF
--- a/src/engine/strat_engine/backstore/crypt.rs
+++ b/src/engine/strat_engine/backstore/crypt.rs
@@ -506,16 +506,16 @@ fn ensure_inactive(device: &mut CryptDevice, name: &str) -> Result<()> {
             name
         );
     }
-    if log_on_failure!(
+    let status = log_on_failure!(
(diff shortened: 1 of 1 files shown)


jbaublitz:
@mulkieran I can't believe I didn't think of this before! Moving back to in progress because we can use libcryptsetup-rs to check if it's busy as opposed to sleeping. We should definitely do this instead.

mulkieran:
Ok!

jbaublitz:
@mulkieran I'm going to leave this PR in progress until CI passes but it does pass locally on real devices so I'm going to wager a guess that this works. I was thinking about the problem incorrectly because I wanted to check if the physical device is busy where that's not necessarily the problem. We actually want to verify that the devicemapper crypt device is not busy which we can do easily through libcryptsetup-rs. This should make this test reliable and we should be able to keep it moving forward.

jbaublitz:
Also removing the FIXME now.

jbaublitz:
Ready for review.

PR #1960, opened 2020-04-15
Fix bug in generation of error messages in check_device_ids

mulkieran:


mulkieran:
Related: #1877, since distinguishing between PoolUuid's and plain old Uuid was what uncovered the bug. The empty device lists were caused by checking whether a pool UUID could be found in a set containing only device UUIDs.

PR #1962, opened 2020-04-15
Rework crypt API to be more granular

### LINKED ISSUE #1961: Encrypted stratis blockdevs require the key to be in the keyring longer than necessary
While writing encryption blackbox tests, I realized that the key is required to be available in the keyring to disown encrypted devices which has no logical purpose. It boils down to the `CryptHandle::setup()` method which checks if a device is an encrypted stratis device. Part of this check actually verifies that the correct key is in the kernel keyring which is out of scope for checking whether the device is an encrypted Stratis device.

My recommendation is that we make this check available in some public `crypt.rs` method but perform it neither in `CryptHandle::setup()` nor `CryptHandle::can_setup` as both of these methods were designed to only make a decision at a metadata level, not a machine environment level and we will bump into unintended consequences like the one I outlined above if we do not distinguish between them.

jbaublitz:
Closes #1961

### DIFF
--- a/src/engine/strat_engine/backstore/crypt.rs
+++ b/src/engine/strat_engine/backstore/crypt.rs
@@ -880,9 +880,9 @@ mod tests {
             }
 
             for path in paths {
-                if !CryptHandle::can_unlock(path) {
+                if CryptHandle::can_unlock(path) {
                     return Err(Box::new(StratisError::Error(
-                        "All devices should not longer be able to be unlocked".to_string(),
+                        "All devices should no longer be able to be unlocked".to_string(),
                     )));
                 }
             }
--- a/src/engine/strat_engine/backstore/crypt.rs
+++ b/src/engine/strat_engine/backstore/crypt.rs
@@ -875,6 +875,18 @@ mod tests {
                 }
             }
 
+            for handle in handles.iter_mut() {
+                handle.wipe()?;
+            }
+
+            for path in paths {
+                if !CryptHandle::can_unlock(path) {
+                    return Err(Box::new(StratisError::Error(
+                        "All devices should not longer be able to be unlocked".to_string(),
+                    )));
+                }
+            }
+
             Ok(())
         }
 
--- a/src/engine/strat_engine/backstore/crypt.rs
+++ b/src/engine/strat_engine/backstore/crypt.rs
@@ -238,7 +238,6 @@ impl CryptHandle {
             .load::<()>(EncryptionFormat::Luks2, None)
             .is_err()
         {
-            info!("Device {} is not a LUKS2 device", physical_path.display());
             Ok(None)
         } else {
             Ok(Some(device))
@@ -250,7 +249,7 @@ impl CryptHandle {
     /// is formatted as a LUKS2 device, etc.)
     // TODO: Use this method when adding blockdevs to a pool to verify that
     // the same key is being used to encrypt new devices as already encrypted devices.
-    #[allow(dead_code)]
+    #[cfg(test)]
     pub fn can_unlock(physical_path: &Path) -> bool {
         fn can_unlock_with_failures(physical_path: &Path) -> Result<bool> {
             let device_result = CryptHandle::device_from_physical_path(physical_path);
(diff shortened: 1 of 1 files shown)


src/engine/strat_engine/backstore/crypt.rs
```diff
@@ -226,22 +225,73 @@ impl CryptHandle {
...
+            Ok(Some(device))
+        }
+    }
+
+    /// Check whether the given physical device can be unlocked with the current
+    /// environment (e.g. the proper key is in the kernel keyring, the device
+    /// is formatted as a LUKS2 device, etc.)
+    // TODO: Use this method when adding blockdevs to a pool to verify that
+    // the same key is being used to encrypt new devices as already encrypted devices.
+    #[allow(dead_code)]
```
mulkieran, line 253: I _think_ you should be able to test this in a basic way. Its intended use is in the circumstances where a device has already been opened with a given KEK.

So, a valid test would be to:
1. Activate a device. You have now a physical path and a logical path.

3. Test simply by invoking this method on on the physical path you got in step 1. It should return true.

4. Deactivate the device, and test again. I believe this method should still return true on an inactive device.

5. Wipe the device. I believe it should now return false.

Obviously, this is a basic test, since it doesn't check the thing that we'll use it to guard against, which is that the key corresponding to the key description has somehow changed underneath us. Setting that test up would be pretty hard to do at this level in the code.
But the suggested test would exercise the basic functionality, which is still useful.

src/engine/strat_engine/backstore/crypt.rs
```diff
@@ -257,20 +307,22 @@ impl CryptHandle {
...
-        );
-        if device
-            .context_handle()
-            .load::<()>(EncryptionFormat::Luks2, None)
-            .is_err()
-            || !is_encrypted_stratis_device(&mut device)
-        {
-            return Ok(None);
-        }
+        let device_result = CryptHandle::device_from_physical_path(physical_path);
```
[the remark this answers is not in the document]
jbaublitz: It's a habit in my personal projects. I'll fix it here.

src/engine/strat_engine/backstore/crypt.rs
jbaublitz: > I _think_ you should be able to test this in a basic way. Its intended use is in the circumstances where a device has already been opened with a given KEK.

That's not entirely correct. I'm happy to update the documentation but this has nothing to do with the device having been activated as much as it checking if the correct key is available and it _could_ be activated. This would work even on a device that has been initialized but not yet activated as long as the correct key is available in the keyring. It runs a test to see if it could be activated with the available keys. Do you still want me to test the same way?

> 
> So, a valid test would be to:
> 
>     1. Activate a device. You have now a physical path and a logical path.
> 
>     2. Test simply by invoking this method on on the physical path you got in step 1. It should return true.
> 
>     3. Deactivate the device, and test again. I believe this method should still return true on an inactive device.
> 
>     4. Wipe the device. I believe it should now return false.
> 
> 
> Obviously, this is a basic test, since it doesn't check the thing that we'll use it to guard against, which is that the key corresponding to the key description has somehow changed underneath us. Setting that test up would be pretty hard to do at this level in the code.
> But the suggested test would exercise the basic functionality, which is still useful.

It will actually test if the key has changed. It is not simply checking for the existence of a key with the correct description but whether or not the given key could actually unlock the device without ever unlocking it.

src/engine/strat_engine/backstore/crypt.rs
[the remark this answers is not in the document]
jbaublitz: Resolved.

src/engine/strat_engine/backstore/crypt.rs
mulkieran: Exactly!

When actually using ```can_unlock``` during initialization, for #1964 it will have to be used on a block device that is already activated, since when adding a device to a pool, it is expected that all the other devices that are worth checking are activated and in use for that pool. So that's probably the most important condition under which to test the function.

jbaublitz: Ok, now I think I understand a little more now what you're getting at. I think the point of confusion was more that I could see us theoretically using this on a set of closed block devices during device discovery to make sure that all of the devices that are locked can be opened or something similar. This would allow us to check that all are openable before beginning to open some and realizing that others cannot be opened. This may or may not be the way we want to go but it could help from an error reporting standpoint and would also mean that we wouldn't have to roll anything back. For now, I'll test what you've outlined because that will certainly help us either way.

jbaublitz:
Waiting to put into review until CI passes.

src/engine/strat_engine/backstore/crypt.rs
jbaublitz: Resolved.

PR #1967, opened 2020-04-17
Make BlockDevMgr::add handle the key being switched after initialization

### LINKED ISSUE #1964: In encryption, guard against key description being bound to different KEK in keyring
Stratis maintains information about the _key description_, which can be public, and not about its corresponding key in the keyring.

The threat is that the ```key description``` which every encrypted block device in a given pool must have in common, will, during stratisd operation, have its corresponding key changed. If stratisd is called upon to add block devices, then it could add its new block devices encrypted with a different KEK than the old ones.

The way to check this is to verify that it is possible to activate a device that is already claimed by the blockdevmgr using the stored key description. If this is the case, then this is proof that the KEK is the same as was used to initially encrypt the devices.

The appropriate place to check this is at the blockdevmgr level or below, as it is the blockdevmgr that tracks the key description.

It should always be only necessary to check one block dev that the blockdevmgr already owns, these must be all equivalent in that regard; or else our code has already failed.

If the blockdevmgr has not blockdevs, then no check is necessary. It might be reasonable to take this possiblity into account, even though I believe that it is impossible at this time to create an empty blockdevmgr.

A possible method, defined in the BlockDevMgr impl should not have to take any arguments, something like:

```
impl BlockDevMgr {
  fn verify_valid_KEK(&self) -> bool {
     /// If 0 blockdevs belonging to BlockDevMgr, true
     /// else use existing crypt method in #1962 to verify KEK
  } 
}
```

This method should be invoked in ```BlockDevMgr::add``` immediately before ```devices::initialize``` is invoked.

An interesting question is how to test this functionality.

This is a very rough design, and subject to change.

jbaublitz:
Closes #1964

### DIFF
--- a/src/engine/strat_engine/backstore/blockdevmgr.rs
+++ b/src/engine/strat_engine/backstore/blockdevmgr.rs
@@ -20,6 +20,7 @@ use crate::{
         strat_engine::{
             backstore::{
                 blockdev::StratBlockDev,
+                crypt::CryptHandle,
                 devices::{initialize_devices, process_and_verify_devices, wipe_blockdevs},
                 metadata::MDADataSize,
             },
@@ -161,6 +162,19 @@ impl BlockDevMgr {
             .collect()
     }
 
+    /// Check that the registered key description for these block devices can
+    /// unlock at least one of the existing block devices registered.
+    /// Precondition: self.block_devs Must have at least one device.
+    pub fn has_valid_passphrase(&self) -> bool {
+        CryptHandle::can_unlock(
+            self.block_devs
+                .iter()
+                .next()
+                .expect("Must have at least one blockdev")
+                .physical_path(),
+        )
+    }
+
     /// Add paths to self.
     /// Return the uuids of all blockdevs corresponding to paths that were
     /// added.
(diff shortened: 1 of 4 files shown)


jbaublitz:
Waiting to put into review until CI passes.

src/engine/strat_engine/backstore/blockdevmgr.rs
```diff
@@ -427,6 +454,67 @@ mod tests {
...
+            key_desc: &str,
+            _: Option<()>,
+        ) -> Result<(PoolUuid, BlockDevMgr), Box<dyn Error>> {
+            let pool_uuid = Uuid::new_v4();
+            let bdm = BlockDevMgr::initialize(
+                pool_uuid,
+                &paths[..2],
+                MDADataSize::default(),
+                Some(key_desc.to_string()),
+            )?;
```
mulkieran, line 472: Please verify add with same key works here,  to get both paths covered.

src/engine/strat_engine/backstore/blockdevmgr.rs
```diff
@@ -161,6 +162,19 @@ impl BlockDevMgr {
             .collect()
     }
 
+    /// Check that the registered key description for these block devices can
+    /// unlock at least one of the existing block devices registered.
+    /// Precondition: self.block_devs Must have at least one device.
```
mulkieran, line 167: ^Must^must

src/engine/strat_engine/backstore/blockdevmgr.rs
jbaublitz: Resolved.

src/engine/strat_engine/backstore/blockdevmgr.rs
jbaublitz: Resolved.

PR #1969, opened 2020-04-20
Return error in SimPool::add_blockdevs() when no cache initialized for Cache Tier device

lleshchi:
Related stratis-storage/project#166 
Related https://github.com/stratis-storage/stratisd/issues/1973

Expect CLI tests to fail

### DIFF
--- a/src/engine/sim_engine/pool.rs
+++ b/src/engine/sim_engine/pool.rs
@@ -162,20 +162,19 @@ impl Pool for SimPool {
         paths: &[&Path],
         tier: BlockDevTier,
     ) -> StratisResult<SetCreateAction<DevUuid>> {
-        if tier == BlockDevTier::Cache {
-            if self.has_cache() {
-                if paths.is_empty() {
-                    // If the cache has been initialized and we are adding zero blockdevs,
-                    // treat adding no new blockdev as the empty st.
-                    return Ok(SetCreateAction::new(vec![]));
-                }
-            } else {
-                return Err(StratisError::Engine(
+        if tier == BlockDevTier::Cache && !self.has_cache() {
+               return Err(StratisError::Engine(
                     ErrorEnum::Invalid,
                     "The cache has not been initialized; you must use init_cache first to initialize the cache.".to_string(),
                 ));
-            };
         }
+
+        if paths.is_empty(){
+            // If the cache has been initialized and we are adding zero blockdevs,
+            // treat adding no new blockdev as the empty set.
+            return Ok(SetCreateAction::new(vec![]));
+        }
+
         let devices: HashSet<_, RandomState> = HashSet::from_iter(paths);
 
         let device_pairs: Vec<_> = devices
--- a/src/engine/sim_engine/pool.rs
+++ b/src/engine/sim_engine/pool.rs
@@ -163,15 +163,14 @@ impl Pool for SimPool {
         tier: BlockDevTier,
     ) -> StratisResult<SetCreateAction<DevUuid>> {
         if tier == BlockDevTier::Cache && !self.has_cache() {
-               return Err(StratisError::Engine(
+            return Err(StratisError::Engine(
                     ErrorEnum::Invalid,
                     "The cache has not been initialized; you must use init_cache first to initialize the cache.".to_string(),
                 ));
         }
 
-        if paths.is_empty(){
-            // If the cache has been initialized and we are adding zero blockdevs,
-            // treat adding no new blockdev as the empty set.
+        if paths.is_empty() {
+            // Treat adding no new blockdev as the empty set.
             return Ok(SetCreateAction::new(vec![]));
         }
 
--- a/src/engine/sim_engine/pool.rs
+++ b/src/engine/sim_engine/pool.rs
@@ -162,18 +162,18 @@ impl Pool for SimPool {
         paths: &[&Path],
         tier: BlockDevTier,
     ) -> StratisResult<SetCreateAction<DevUuid>> {
-        if paths.is_empty() {
-            // Pool must be initialized with data devices so only check
-            // if the cache has been initialized.
-            return if !self.has_cache() && tier == BlockDevTier::Cache {
-                Err(StratisError::Engine(
+        if tier == BlockDevTier::Cache {
+            if self.has_cache() {
+                if paths.is_empty() {
+                    // If the cache has been initialized and we are adding zero blockdevs,
+                    // treat adding no new blockdev as the empty st.
+                    return Ok(SetCreateAction::new(vec![]));
+                }
+            } else {
+                return Err(StratisError::Engine(
                     ErrorEnum::Invalid,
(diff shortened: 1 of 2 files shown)


src/engine/sim_engine/pool.rs
```diff
@@ -162,18 +162,18 @@ impl Pool for SimPool {
...
-        if paths.is_empty() {
-            // Pool must be initialized with data devices so only check
-            // if the cache has been initialized.
-            return if !self.has_cache() && tier == BlockDevTier::Cache {
-                Err(StratisError::Engine(
+        if tier == BlockDevTier::Cache {
+            if self.has_cache() {
+                if paths.is_empty() {
+                    // If the cache has been initialized and we are adding zero blockdevs,
+                    // treat adding no new blockdev as the empty st.
```
jbaublitz, line 169: I think this line has a typo.

src/engine/sim_engine/pool.rs
```diff
@@ -162,18 +162,18 @@ impl Pool for SimPool {
...
+            } else {
+                return Err(StratisError::Engine(
                     ErrorEnum::Invalid,
                     "The cache has not been initialized; you must use init_cache first to initialize the cache.".to_string(),
-                ))
-            } else {
-                // If the cache has been initialized or we are adding zero blockdevs,
-                // treat adding no new blockdevs as the empty set.
-                Ok(SetCreateAction::new(vec![]))
+                ));
```
jbaublitz, line 176: This block could potentially be condensed to only checking for an uninitialized cache and returning an error.

src/engine/sim_engine/pool.rs
```diff
@@ -162,18 +162,18 @@ impl Pool for SimPool {
...
             };
         }
         let devices: HashSet<_, RandomState> = HashSet::from_iter(paths);
```
jbaublitz, line 179: We could potentially branch here and if `paths` is empty, return an unchanged idempotent success (`Ok(SetCreateAction::new(vec![]))`). If it is not empty, proceed with the addition to the sim engine.

src/engine/sim_engine/pool.rs
```diff
@@ -162,20 +162,19 @@ impl Pool for SimPool {
...
-                // If the cache has been initialized or we are adding zero blockdevs,
-                // treat adding no new blockdevs as the empty set.
-                Ok(SetCreateAction::new(vec![]))
-            };
+                ));
+        }
+
+        if paths.is_empty(){
+            // If the cache has been initialized and we are adding zero blockdevs,
+            // treat adding no new blockdev as the empty set.
```
jbaublitz, line 174: Update this comment and we should be good to go!

mulkieran:
> @mulkieran Clippy seems to be unhappy with the nested if statements. I took a look and I'm debating whether we want to restructure this quite a bit:
> 
>     * First check `!self.has_cache() && tier == BlockDevTier::Cache` and if this evaluates to `true`, return an error.
> 
>     * Next check `paths.is_empty()` and if this is `true`, regardless of tier, return `Ok(SetCreateAction::new(vec![]))`.
> 
>     * Once we've established that it is not an empty list of devices and that the cache is initialized, proceed with branching on the `BlockDevTier` with a match statement and do the actual actions in either case.
> 
> 
> What are your thoughts?
> 
> EDIT: I'm referring to the `strat_engine` here.

That seems perfectly reasonable to me. If we manage to hoist the ```process_devices``` method up in future we can put it in the spot where we check for an empty list of devices, and that will work out well.

There is only the question of idempotency. Are we happy to error on attempting to construct a cache even if there are no devices? I think that's a reasonable decision and does not conflict with idempotency requirements. It is certainly clearer that way.

jbaublitz:
> > @mulkieran Clippy seems to be unhappy with the nested if statements. I took a look and I'm debating whether we want to restructure this quite a bit:
> > ```
> > * First check `!self.has_cache() && tier == BlockDevTier::Cache` and if this evaluates to `true`, return an error.
> > 
> > * Next check `paths.is_empty()` and if this is `true`, regardless of tier, return `Ok(SetCreateAction::new(vec![]))`.
> > 
> > * Once we've established that it is not an empty list of devices and that the cache is initialized, proceed with branching on the `BlockDevTier` with a match statement and do the actual actions in either case.
> > ```
> > 
> > 
> > What are your thoughts?
> > EDIT: I'm referring to the `strat_engine` here.
> 
> That seems perfectly reasonable to me. If we manage to hoist the `process_devices` method up in future we can put it in the spot where we check for an empty list of devices, and that will work out well.
> 
> There is only the question of idempotency. Are we happy to error on attempting to construct a cache even if there are no devices? I think that's a reasonable decision and does not conflict with idempotency requirements. It is certainly clearer that way.

I would prefer to always error if the cache has not been constructed, even for empty lists, for clarity.

src/engine/strat_engine/pool.rs
```diff
@@ -354,42 +354,38 @@ impl Pool for StratPool {
...
-                                    .and(Ok(bdi))
-                            });
-                    self.thin_pool.resume()?;
-                    let bdev_info = bdev_info_res?;
-                    Ok(SetCreateAction::new(bdev_info))
-                }
-            } else {
-                Err(StratisError::Error(format!(
+        let bdev_info = if tier == BlockDevTier::Cache && !self.has_cache() {
+            Err(StratisError::Error(format!(
```
jbaublitz, line 358: This should return early.

src/engine/strat_engine/pool.rs
```diff
@@ -354,42 +354,38 @@ impl Pool for StratPool {
...
                             "No cache has been initialized for pool with UUID {} and name {}; it is therefore impossible to add additional devices to the cache",
                             pool_uuid.to_simple_ref(),
                             pool_name)
                 ))
-            }
+        } else if paths.is_empty() {
+            // If the cache has been initialized and we are adding zero blockdevs,
+            // treat adding no new blockdevs as the empty set.
+            Ok(SetCreateAction::new(vec![]))
```
jbaublitz, line 366: This should return early.

src/engine/strat_engine/pool.rs
```diff
@@ -354,42 +354,38 @@ impl Pool for StratPool {
...
-                ))
-            }
+                ));
+        } else if paths.is_empty() {
+            // If the cache has been initialized and we are adding zero blockdevs,
+            // treat adding no new blockdevs as the empty set.
+            return Ok(SetCreateAction::new(vec![]));
+        } else if tier == BlockDevTier::Cache {
+            // If adding cache devices, must suspend the pool; the cache
+            // must be augmented with the new devices. Note that this
```
mulkieran, line 369: Please drop all the comments that start with ```Note``` up to the empty check, and instead put in a TODO about ```process_and_verify_devices``` next to the empty check. You can also drop the comments in the empty check, they've become kind of obsolete, since the empty check is now for both cache and data devs.

PR #1975, opened 2020-04-22
Preliminary prototype of Report interface

jbaublitz:
Related to #1122 

@mulkieran Happy to change the name from `partial_pool_devices` to something better but this was the best I could think of at the moment.

TODOs:
* Testing
  * Test that sim engine responds with a successful mock result on valid input and error on invalid input
  * Generate a situation with an incomplete pool in the Rust unit tests and check that the value reported by this interface is correct
* Discuss providing an engine state report as we used to do in our logging
  * Should this be provided by a different method like "get_unstable_report"?
  * Is there another way to indicate that the interface may change?
* Do we want to allow unprivileged users to get reports or just root?

### DIFF
--- a/src/dbus_api/api/report_2_1/methods.rs
+++ b/src/dbus_api/api/report_2_1/methods.rs
@@ -15,7 +15,6 @@ use crate::{
         util::{engine_to_dbus_err_tuple, get_next_arg, msg_code_ok, msg_string_ok},
     },
     engine::ReportType,
-    stratis::StratisError,
 };
 
 pub fn get_report(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
@@ -40,7 +39,7 @@ pub fn get_report(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
     let msg = match serde_json::to_string(&engine.get_report(report_type)) {
         Ok(string) => return_message.append3((true, string), msg_code_ok(), msg_string_ok()),
         Err(e) => {
-            let (rc, rs) = engine_to_dbus_err_tuple(&StratisError::Serde(e));
+            let (rc, rs) = engine_to_dbus_err_tuple(&e.into());
             return_message.append3(default_return, rc, rs)
         }
     };
--- a/src/dbus_api/api/report_2_1/methods.rs
+++ b/src/dbus_api/api/report_2_1/methods.rs
@@ -1,0 +1,49 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use std::convert::TryFrom;
+
+use dbus::{
+    tree::{MTFn, MethodInfo, MethodResult},
+    Message,
+};
+
+use crate::{
+    dbus_api::{
+        types::TData,
+        util::{engine_to_dbus_err_tuple, get_next_arg, msg_code_ok, msg_string_ok},
+    },
+    engine::ReportType,
+    stratis::StratisError,
+};
+
+pub fn get_report(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
+    let message: &Message = m.msg;
+    let mut iter = message.iter_init();
+    let report_name: &str = get_next_arg(&mut iter, 0)?;
+
+    let return_message = message.method_return();
+    let default_return = (false, String::new());
+
+    let report_type = match ReportType::try_from(report_name) {
+        Ok(rt) => rt,
+        Err(e) => {
+            let (rc, rs) = engine_to_dbus_err_tuple(&e);
+            return Ok(vec![return_message.append3(default_return, rc, rs)]);
+        }
+    };
+
+    let dbus_context = m.tree.get_data();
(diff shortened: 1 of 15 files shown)


jbaublitz:
> Related to #1122
> 
> @mulkieran Happy to change the name from `partial_pool_devices` to something better but this was the best I could think of at the moment.

Per standup, I'll change this to `errored_pool_devices`.

> 
> TODOs:
> 
>     * Testing
>       
>       * Test that sim engine responds with a successful mock result on valid input and error on invalid input
>       * Generate a situation with an incomplete pool in the Rust unit tests and check that the value reported by this interface is correct
> 
>     * Discuss providing an engine state report as we used to do in our logging
>       
>       * Should this be provided by a different method like "get_unstable_report"?
>       * Is there another way to indicate that the interface may change?

We'll be documenting this in the D-Bus API docs.

> 
>     * Do we want to allow unprivileged users to get reports or just root?

We will allow unprivileged users to query this.

mulkieran:
We discussed a bunch of things in standup;

1. We'll configure it so all users can have access to the interface, we can change that later, since we're making it officially unstable. We'll guarantee valid JSON.
2. If the whole interface is unstable it doesn't matter how ugly the engine report is; fixing it to be a bit better won't break anything.
3. We'll document the report interface at a high level in the design doc and at a lower-level in the D-Bus API doc.
4. For testing, for this PR, we can be basic and just do what can be done against the SimEngine. We can also add a test to the blackbox tests, and should probably do that. But anything else would probably be overkill.

I would expect to use this interface in the udev tests, and that would allow further testing.

mulkieran:
@jbaublitz We just had a comment race condition, but we seem to be in agreement...

jbaublitz:
> @jbaublitz We just had a comment race condition, but we seem to be in agreement...

Haha yes, this looks good to me.

src/engine/engine.rs
```diff
@@ -18,11 +20,38 @@ use crate::{
         BlockDevTier, CreateAction, DeleteAction, DevUuid, FilesystemUuid, MaybeDbusPath, Name,
         PoolUuid, RenameAction, SetCreateAction, SetDeleteAction,
     },
-    stratis::StratisResult,
+    stratis::{ErrorEnum, StratisError, StratisResult},
 };
 
 pub const DEV_PATH: &str = "/stratis";
 
+/// The type of report for which to query.
```
mulkieran, line 28: Can you move ```ReportType``` into ```types.rs```  (and update the header comment)?

src/engine/engine.rs
```diff
@@ -18,11 +20,38 @@ use crate::{
...
+
+impl<'a> TryFrom<&'a str> for ReportType {
+    type Error = StratisError;
+
+    fn try_from(name: &str) -> StratisResult<ReportType> {
+        match name {
+            "errored_pool_report" => Ok(ReportType::ErroredPoolDevices),
+            _ => Err(StratisError::Engine(
+                ErrorEnum::NotFound,
+                format!("Report {} not found", name),
```
mulkieran, line 44: For here say something more formal like: ```Report name {} not understood```.

src/dbus_api/api/report_2_1/methods.rs
```diff
@@ -0,0 +1,49 @@
...
+        }
+    };
+
+    let dbus_context = m.tree.get_data();
+    let engine = dbus_context.engine.borrow();
+
+    let msg = match serde_json::to_string(&engine.get_report(report_type)) {
+        Ok(string) => return_message.append3((true, string), msg_code_ok(), msg_string_ok()),
+        Err(e) => {
+            let (rc, rs) = engine_to_dbus_err_tuple(&StratisError::Serde(e));
```
mulkieran, line 43: Might as well use ```&e.into()``` here instead of being explicit w/ ```StratisError::Serde```

mulkieran:
@jbaublitz Or we could just defer worrying about the engine state report until later. This establishes the interface and adds the report that I'm immediately interested in, so it would be fine to consider this PR complete in itself.

jbaublitz:
@mulkieran I started preliminary work on it so I'm happy to continue. My idea was to define:
```
impl<'a> Into<Value> for &'a [DATA_TYPE]
```
at each layer of StratEngine that we want to expose some information. Does this make sense to you?

mulkieran:
> @mulkieran I started preliminary work on it so I'm happy to continue. My idea was to define:
> 
> ```
> impl<'a> Into<Value> for &'a [DATA_TYPE]
> ```
> 
> at each layer of StratEngine that we want to expose some information. Does this make sense to you?

That makes sense, but as I work on the event-based device discovery I become even more eager to get this in without the engine state report, as I think we'll be making changes to the structure formerly known as ```incomplete_pools``` very soon.

Can you hold off on the engine state report; get this ready for final; and then begin work on the key-handling part of https://github.com/stratis-storage/project/issues/171?

jbaublitz:
Sounds good to me. I'm going to split this into two PRs. The work I've already done on engine state can hang around until we want to add that.

src/engine/engine.rs

src/dbus_api/api/report_2_1/methods.rs
jbaublitz: Resolved.

jbaublitz:
CI is failing due to Github downtime.

Issue #1784, opened 2020-01-15
Internal SimEngine state debug log expresses uuid with individual bytes


lleshchi:
As seen below, the debug log is expressing the uuid as a decimal byte-by-byte representation on multiple lines, rather than the standard hex single-line representation. This is inefficient to read or for any sort of uuid comparison, since a conversion to hex would be required. 

```
INFO stratisd: Dump timer expired, dumping state
DEBUG stratisd: Engine state: 
SimEngine {
    pools: {
        (
            "test_pool",
            SimpleRef(
                Uuid(
                    [
                        232,
                        190,
                        118,
                        78,
                        189,
                        175,
                        73,
                        180,
                        179,
                        149,
                        173,
                        18,
                        144,
                        210,
                        29,
                        255,
                    ],
                ),
            ),
        ): SimPool {
            block_devs: {
                Uuid(
                    [
                        185,
                        195,
                        66,
                        136,
                        180,
                        230,
                        65,
                        56,
                        139,
                        68,
                        190,
                        164,
                        27,
                        252,
                        190,
                        236,
                    ],
                ): SimDev {
                    devnode: "junk",
                    rdm: RefCell {
                        value: {Randomizer 0,
                    },
                    user_info: None,
                    hardware_info: None,
                    initialization_time: 1579119473,
                    dbus_path: MaybeDbusPath(
                        Some(
                            Path(
                                "/org/storage/stratis2/6",
                            ),
                        ),
                    ),
                },
            },
```

mulkieran:
The reason this happens is that the output of the log is simply a pretty print of the Debug output of the whole engine. The Debug format for UUIDs is defined to show each byte individually. Hence the verbose output.

mulkieran:
Related: https://github.com/stratis-storage/project/issues/93.

mulkieran:
This was resolved by #1907, #1979 will not suffer from the same defect.

PR #1979, opened 2020-04-23
Work on engine state reporting

jbaublitz:
Related to #1122
Related to #1784.
Resolves: https://github.com/stratis-storage/project/issues/93.

### DIFF
--- a/src/engine/sim_engine/engine.rs
+++ b/src/engine/sim_engine/engine.rs
@@ -30,15 +30,23 @@ pub struct SimEngine {
 }
 
 impl<'a> Into<Value> for &'a SimEngine {
+    // Precondition: SimPool Into<Value> impl return value always pattern matches
+    // Value::Object(_)
     fn into(self) -> Value {
         json!({
             "pools": Value::Array(
-                // FIXME: Add reporting for devices in pools
-                self.pools.iter().map(|(name, uuid, _)| {
-                    json!({
+                self.pools.iter().map(|(name, uuid, pool)| {
+                    let json = json!({
                         "pool_uuid": uuid.to_simple_ref().to_string(),
                         "name": name.to_string(),
-                    })
+                    });
+                    let pool_json = pool.into();
+                    if let (Value::Object(mut map), Value::Object(submap)) = (json, pool_json) {
+                        map.extend(submap.into_iter());
+                        Value::Object(map)
+                    } else {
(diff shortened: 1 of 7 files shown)


jbaublitz:
Blocked by #1975

jbaublitz:
CI is failing due to Github issues.

bgurney-rh:
I sent a retry command to TravisCI and the cli-with-stratisd test; hopefully that allows the tests to re-run cleanly.

jbaublitz:
Thank you, it looks like that worked!

mulkieran:

mulkieran:
The structure of this is fine. I do not know if there is a compelling case for the mock objects on the SimEngine, though. The internal structures of the SimEngine will vary a good deal from those of the StratEngine. The intended use is to allow person wanting to debug remotely or for sos reports to take a look at the StratEngine. So, having the SimEngine just return an empty report would be fine with me.

For the StratEngine, I would be happy merging just this outer skeleton as is; and filling in the remaining bits incrementally.

mulkieran:
We decided in standup to defer this a bit longer. The reason is that it might make sense to implement json crate's Serialize and Deserialize tratis, rather than take the approach here.

We also discussed a bit about the SimEngine's implementation. We noted that there is already a mock-up of incomplete pools that went in with a previous commit. This gets moved around in this PR but basically remains the same. The key thing is not to spend too much effort on the sim engine.

Should the D-Bus always return its representation of ```None``` for the SimEngine? If we decide to do that, we could have it return ```None``` for the errored devices report as well. I'ld be happy with that. We should only spend time on the SimEngine reports if we can demonstrate a real advantage.

jbaublitz:
Per our discussion in #1995, should I resume work on this?

mulkieran:
> Per our discussion in #1995, should I resume work on this?

We should talk about it again in standup.

jbaublitz:
Sounds good!

jbaublitz:
I accidentally applied the most recent changes during a rebase so the first commit is a combination of the original PR and changes that I just made.

jbaublitz:
TODO: Make sim_engine consistent with the strat_engine.

jbaublitz:
I agree that we may have to diverge as we've done in other areas surrounding encryption in the sim_engine. I'm not determined to keep it identical, but it just seemed to make sense here as it was possible.

PR #1980, opened 2020-04-23
Add D-Bus Report interface

### LINKED ISSUE #1122: Add a new interface for reports; use it to shown liminal block devices

jbaublitz:
Closes #1122 

Supercedes #1975.

### DIFF
--- a/src/dbus_api/api/mod.rs
+++ b/src/dbus_api/api/mod.rs
@@ -14,6 +14,7 @@ use crate::dbus_api::{
 
 pub mod manager_2_0;
 pub mod manager_2_1;
+pub mod report_2_1;
 pub mod shared;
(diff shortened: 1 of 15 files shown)


mulkieran:
I'm opening a separate issue for the engine state report, since I'ld like the attached issue to  clear.

PR #1983, opened 2020-04-24
Prototype of key handling API

jbaublitz:
Related to stratis-storage/project#171


jbaublitz:
CI failures are problems with how the tests handle key descriptions. Fixing.

src/engine/engine.rs
```diff
@@ -17,12 +17,25 @@ use devicemapper::{Bytes, Sectors};
...
     },
     stratis::StratisResult,
 };
 
 pub const DEV_PATH: &str = "/stratis";
+/// The maximum size of pool passphrases stored in the kernel keyring
+pub const MAX_STRATIS_PASS_SIZE: usize = 512 / 8;
+
+pub trait KeyActions {
+    /// Add a key to the kernel keyring. The output is an idempotent return type
```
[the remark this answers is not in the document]
jbaublitz: Resolved.

mulkieran:
@jbaublitz: Please rebase.

jbaublitz:
@mulkieran The reason CI is failing here is that stratisd is looking for a key with the prefix added in this PR but the key added in the current version of the CLI tests does not add it with this prefix. How would you like me to proceed? This seems like it might be a circular dependency for tests passing. The stratis-cli PR requires the new stratisd key operations to succeed while stratisd requires a change to the format of the key description set in the CLI tests.

mulkieran:
@jbaublitz I was expecting the CLI tests to fail, so I didn't take a close look at them. Are you saying theirs a failure even when running against stratis CLI #548 code?

jbaublitz:
I think the problem here is more that we're pulling from `develop-2.0.1` for `cli-with-stratisd` tests so while stratis-storage/stratis-cli#548 works, `develop-2.0.1` does not. If you were already expecting the `cli-with-stratisd` tests to fail, this is a non-issue as we can just merge this even if they are failing. I just wasn't sure if you were going to require those to pass before merging.

src/dbus_api/api/mod.rs
```diff
@@ -36,10 +37,17 @@ pub fn get_base_tree<'a>(dbus_context: DbusContext) -> (Tree<MTFn<TData>, TData>
         .add(
             f.interface(consts::MANAGER_INTERFACE_NAME_2_1, ())
                 .add_m(manager_2_1::create_pool_method(&f))
+                .add_m(manager_2_1::add_key_method(&f))
```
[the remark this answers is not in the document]
jbaublitz: The XML for the D-Bus introspection data is currently set to `AddKey` so are you requesting making all instances of "add" across the D-Bus API "set" (same for "reset", "unset", etc.) instead?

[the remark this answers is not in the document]
jbaublitz: One additional note: this would also require one addition commit on the corresponding stratis-cli PR.

mulkieran:
@jbaublitz: Please rebase and resolve conflicts when you get the chance.

src/engine/strat_engine/keys.rs
```diff
@@ -0,0 +1,428 @@
...
+        Err(io::Error::last_os_error().into())
+    } else {
+        Ok(())
+    }
+}
+
+/// Perform an idempotent add of the given key data with the given key description.
+///
+/// Successful return values:
+/// * `Ok(CreateAction::Identity)`: The key was already in the keyring with the
```
mulkieran, line 192: Why can't you use ```RenameAction``` instead of ```CreateAction``` here? You could rename ```RenameAction``` to something more general, but it seems like it captures the sematics better. I recall that we discussed a possibility of improving ```RenameAction``` so it may seem a bit unsatisfactory to use, but it exists and it does capture all the cases.

src/engine/strat_engine/mod.rs
```diff
@@ -10,13 +10,14 @@ mod device;
...
 mod engine;
+mod keys;
 mod liminal;
 mod names;
 mod pool;
 mod serde_structs;
 mod thinpool;
 
-pub use self::engine::StratEngine;
+pub use self::{engine::StratEngine, keys::StratKeyActions};
```
mulkieran, line 20: ```StratKeyActions``` shouldn't need to be exported here, either.

src/engine/strat_engine/names.rs
```diff
@@ -18,6 +18,53 @@ use crate::{
...
+    }
+
+    /// Return the application-level key description (the key description with no
+    /// Stratis prefix added).
+    pub fn as_application_str(&self) -> &str {
+        &self.0
+    }
+}
+
+impl Display for KeyDescription {
```
mulkieran, line 56: You might want not to implement Display since the format is not truley canonica. Maybe a ```to_system_key_desc``` method would be a better choice.

src/engine/strat_engine/keys.rs
```diff
@@ -0,0 +1,428 @@
...
+    }
+}
+
+/// Handle for kernel keyring interaction.
+#[derive(Debug)]
+pub struct StratKeyActions;
+
+#[cfg(test)]
+impl StratKeyActions {
+    pub fn set_no_fd(
```
mulkieran, line 355: Please put a header comment on this.

src/engine/strat_engine/keys.rs
jbaublitz: Structurally, I think `Rename` provides what we want here, but these actions are very specific to what action we're performing. There's no condition where there's no source for the create action because create actions don't have a source and target, only a target. If we rename `NoSource` to something else, we lose the meaning of what it's supposed to indicate which is very specific to renaming. I worry that we'll bump eventually introduce another bug related to  overloading types if we try to make it too general.

If you don't think the current solution is workable, I see the two most workable solutions here being creating a new type for `Create` actions that we only use here (something like `CreateUpdateAction`, and we may end up using it in the future as this kind of pattern is common in idempotent frameworks like Ansible) or extending the `CreateAction` type and reworking the other places that it's used. In either solution, we could potentially add a `Updated`/`Changed` variant to indicate that the identifier remained the same but the linked data was changed.

mulkieran: > Structurally, I think `Rename` provides what we want here, but these actions are very specific to what action we're performing. There's no condition where there's no source for the create action because create actions don't have a source and target, only a target. If we rename `NoSource` to something else, we lose the meaning of what it's supposed to indicate which is very specific to renaming. I worry that we'll eventually introduce another bug related to overloading types if we try to make it too general.
> 
> If you don't think the current solution is workable, I see the two most workable solutions here being creating a new type for `Create` actions that we only use here (something like `CreateUpdateAction`, and we may end up using it in the future as this kind of pattern is common in idempotent frameworks like Ansible) or extending the `CreateAction` type and reworking the other places that it's used. In either solution, we could potentially add a `Updated`/`Changed` variant to indicate that the identifier remained the same but the linked data was changed.

You have a ```SetCreateAction```, why not make a ```MappingCreateAction``` with constructors:
```Identity```, ```ValueChanged```, ```Created``` or something like that?

File: src/engine/types/actions.rs [hunk 7]
```diff
@@ -47,6 +47,33 @@ impl<T> EngineAction for CreateAction<T> {
     }
 }
 
+#[derive(Debug, PartialEq, Eq)]
+pub enum MappingCreateAction<T> {
+    Created(T),
+    Identity,
+    ValueChanged,
```
mulkieran, line 54: Maybe ```ValueChanged``` should also be parameterized on ```T```. I know it makes no difference in this commit because ```T``` is bound to ```()``` in the only place it is used, for security reasons, I presume, but it makes more sense to me.

src/engine/engine.rs
```diff
@@ -36,17 +36,18 @@ pub trait KeyActions {
...
     /// Successful return values:
-    /// * `Ok(CreateAction::Identity)`: The key was already in the keyring with the
-    /// appropriate key description and key data.
-    /// * `Ok(CreateAction::Created(false)`: The key was newly added to the keyring.
-    /// * `Ok(CreateAction::Created(true)`: The key description was already present
+    /// * `Ok(MappingCreateAction::Identity)`: The key was already in the keyring
+    /// with the appropriate key description and key data.
+    /// * `Ok(MappingCreateAction::Created(()))`: The key was newly added to the
+    /// keyring.
+    /// * `Ok(MappingCreateAction::Changed)`: The key description was already present
```
mulkieran, line 43: You might want to make a note here the parameterizing on the ```()``` type is part of a security protocol.

src/engine/types/actions.rs
```diff
@@ -47,6 +47,33 @@ impl<T> EngineAction for CreateAction<T> {
     }
 }
 
+#[derive(Debug, PartialEq, Eq)]
+pub enum MappingCreateAction<T> {
```
mulkieran, line 51: Please add doc comments here as appropriate.

tests/client-dbus/tests/udev/test_udev.py
```diff
@@ -266,7 +266,7 @@ def _simple_initial_discovery_test(self, *, key_description=None):
 
         settle()
 
-        with ServiceContextManager():
+        with OptionalKeyServiceContextManager(key_data) as key_description:
```
mulkieran, line 269: We want both services to be ```OptionalKeyServiceContextManagers``` in this test, although it won't change the behavior for the second at this time.

src/engine/engine.rs
jbaublitz: I'm not sure quite what you mean. Are you suggesting a security protocol section in the documentation?

src/engine/types/actions.rs
jbaublitz: I had also considered that so if you're also thinking that might be a good idea, I'll add it.

tests/client-dbus/tests/udev/test_udev.py
jbaublitz: If you're looking to repopulate the key, we'll have to make changes to the `KernelKey` structure because it randomly generates a key description so the key description of the key available will not be the same as what's stored in the metadata.

mulkieran: Whoops! Yes, that is what we really need to do.

jbaublitz: This may not be relevant for this PR but it may be in future tests so should we modify this now?

src/engine/engine.rs
jbaublitz: Disregard the previous comment, I misread your suggestion.

mulkieran: So, I'm worried about somebody coming along and saying to themselves that the method would be so much more useful if it returned the mapping, not just ```()``` and making it so. So, a comment that says that this is a deliberate choice would be helpful here. I haven't dug into the rest of the code yet deeply, but I think part of the point is that it should be impossible. So, I think it should be worth a note right here.

tests/client-dbus/tests/udev/test_udev.py
mulkieran: @jbaublitz Please give that a try, we will definitely need it.

tests/client-dbus/src/stratisd_client_dbus/_data.py
```diff
@@ -62,6 +62,20 @@
 <arg name="return_code" type="q" direction="out"/>
 <arg name="return_string" type="s" direction="out"/>
 </method>
+<method name="SetKey">
```
mulkieran, line 65: This chunk of introspection data found its way into the wrong Manager interface.

tests/client-dbus/tests/udev/_utils.py
```diff
@@ -346,3 +351,32 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
 
         return False
+
+
+class OptionalKeyServiceContextManager:
+    """
+    A service context manager that accepts an optional key
+    """
+
+    def __init__(self, key_spec=None):
```
mulkieran, line 361: Please make ```key_spec``` a mandatory keyword, i.e, use the ```*```.

tests/client-dbus/tests/udev/test_udev.py
```diff
@@ -252,7 +252,7 @@ class UdevTest3(UdevTest):
     daemon, brings it up again, and allows it to discover the existing pool.
     """
 
-    def _simple_initial_discovery_test(self, *, key_description=None):
+    def _simple_initial_discovery_test(self, *, key_spec=None):
```
mulkieran, line 255: Probably time to add a param doc comment for ```key_spec```.

PR #1989, opened 2020-04-28
Future-proof Python code

jbaublitz:
~~As long as CI passes with these changes, we should merge this to avoid problems when we upgrade to Fedora 31 on Jenkins machines.~~

Related: https://github.com/stratis-storage/project/issues/175.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -117,6 +117,32 @@ jobs:
           (cd scripts; make fmt-travis);
           (cd tests/client-dbus; make fmt-travis)
 
+    # MANDATORY PYTHON CHECKS ON FUTURE SUPPORTED INTERPRETER
+    - name: "lint Python code on future supported interpreter"
+      language: python
+      python: "3.7.7"
+      install:
+        - >
+          pip
+          install
+          pylint==2.4.4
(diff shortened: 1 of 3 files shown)


mulkieran:
What are the problems?

jbaublitz:
We discussed in stand up that this is related to later versions of pylint and isort in Fedora 31.

mulkieran:
Given the conflicts in the equivalent stratis CLI PR, we'll have to do a little planning.

mulkieran:
> As long as CI passes with these changes, we should merge this to avoid problems when we upgrade to Fedora 31 on Jenkins machines.

We wouldn't expect to have problems on the Jenkins machines, but we do have to manage our Travis CI. If we wanted to merge this now, under the expectation that we'll move up to Fedora 31, we could, but we should also add a set of Travis tasks, matching the ones for the recommended development interpreter, but with the default versions of things from Fedora 31. Then when we have all succesfully moved up, we'ld make those the recommended development interpreter tasks and drop the old recommended development interpreter tasks.

jbaublitz:
Conflicts resolved and .travis.yml is updated.

PR #1993, opened 2020-04-30
Make can_setup return device identifiers

### LINKED ISSUE #1992: Make it possible to obtain the Stratis device identifers from the physical device
We'll need this in order to track discovered LUKS devices that belong to Stratis.

I think this should involve changing ```can_setup``` so that it returns an ```Option``` type which contains the Stratis identifiers, and changing the rest accordingly.

jbaublitz:
Closes #1992


mulkieran:
The stratisd-nonrust tests timed out on the unique instance test, restarting.

Issue #1968, opened 2020-04-18
Hoist use of devices::process_and_verify_devices out of blockdevmgr to backstore and then to StratPool


mulkieran:
This would do two things.

1. Move idempotency checks up several layers.
2. Ensure more fully that the invariant that a BlockDevMgr has at least one device holds.

This is not a certain solution, but it looks promising at this time.

If it is moved up to StratPool, it can substitute for the check for 0 devices passed in ```init_cache``` and ```add_blockdevs``` methods, changing the check from a check about the number of devices passes as arguments to a check about the number of devices that actually get added.

It could also improve the error messages, since it would be possible to pass both the cache device and the data devices UUIDs to ```process_and_verify_devices```.

PR #1994, opened 2020-05-01
Rework sim and strat engine logic for simplified handling of devices

### LINKED ISSUE #1973: In StratPool::add_blockdevs(), we should handle the case for adding zero devices to the data tier
When zero devices are added to the Cache Tier, the case is handled separately. https://github.com/stratis-storage/stratisd/blob/develop-2.0.1/src/engine/strat_engine/pool.rs#L359

A similar handling of this case for the Data Tier might be reasonable, rather than calling `backstore.add_datadevs` on an empty list. 
https://github.com/stratis-storage/stratisd/blob/develop-2.0.1/src/engine/strat_engine/pool.rs#L396

lleshchi:
Closes stratis-storage/project#166
Closes #1973

Expect CLI tests to fail

Note: This PR leaves out the `process_and_verify_devices` action, with the expectation that the `is_empty()` check will be replaced with this method in the future. #1968

### DIFF
--- a/src/engine/sim_engine/pool.rs
+++ b/src/engine/sim_engine/pool.rs
@@ -162,20 +162,18 @@ impl Pool for SimPool {
         paths: &[&Path],
         tier: BlockDevTier,
     ) -> StratisResult<SetCreateAction<DevUuid>> {
-        if paths.is_empty() {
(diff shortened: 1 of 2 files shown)


mulkieran:
@lleshchi Please rebase.

PR #2003, opened 2020-05-05
Return the whole result from _create_pool

mulkieran:


mulkieran:
This is virtually trivial.

PR #2005, opened 2020-05-05
Add TryFrom/Into for Stratis LUKS2 token data type

### LINKED ISSUE #1995: Consider making a struct for Stratis LUKS tokens and implementing Serialize and Deserialize traits
This would be tidier if it were possible. I do not see any obvious lifetime-related counter indications. It would be desirable to verify that the keyslots value is empty, which might make things less simple.

jbaublitz:
Closes #1995

### DIFF
--- a/src/engine/strat_engine/backstore/crypt.rs
+++ b/src/engine/strat_engine/backstore/crypt.rs
@@ -797,7 +797,7 @@ fn stratis_token_is_valid(json: &Value) -> bool {
     if let Err(ref e) = result {
         debug!(
             "LUKS2 token in the Stratis token slot does not appear \
-            to be a Stratis token: {}. Reporting as not a Stratis device.",
+            to be a Stratis token: {}.",
             e,
         );
     }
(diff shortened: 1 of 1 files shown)


mulkieran:
I think the errors may be real, pushing this back to "in progress".

jbaublitz:
The bug should be fixed.

src/engine/strat_engine/backstore/crypt.rs
```diff
@@ -674,72 +782,22 @@ fn check_luks2_token(device: &mut CryptDevice) -> Result<()> {
...
-    let keyslots_valid = json
-        .get(TOKEN_KEYSLOTS_KEY)
-        .and_then(|arr| arr.as_array())
-        .map(|arr| arr.is_empty())
-        .unwrap_or(false);
-    if !keyslots_valid {
-        debug!("Keyslots value is invalid");
+    let result = StratisLuks2Token::try_from(json);
+    if let Err(ref e) = result {
+        debug!("Conversion of Stratis JSON token failed with error: {}", e);
```
mulkieran, line 798: Please reword slightly:

```"Interpretation of LUKS2 token in JSON format as Stratis token failed with error: {}", e```

Otherwise it looks like it's reporting an unexpected error, rather than something that could likely happen under normal operation.

jbaublitz: I've reworded slightly from what you had here but I think I've addressed the concern you had.

PR #2011, opened 2020-05-06
Remove can_setup and expose more information in CryptHandle

### LINKED ISSUE #2000: Make key description available from crypt methods
This value needs to be extracted from the encrypted device metadata in order to set it in the StratBlockDev struct. But it would also be desirable to extract it earlier via the ```can_setup``` method. The reason is that if the devices are grouped by pool UUID then a pre-check of the key_descriptions for a single pool can be done. If they are not all the same, then the devices can be shuffled into the not-yet-constructed ```rejected_devices``` data structure, because in that case something will be unrecoverably wrong. It would also be desirable for the ```activate``` method to return the same result values: the stratis identifiers and the key description. This allows Stratis to check that the values at activation are the same as the values at discovery if that ever becomes necessary.

jbaublitz:
Closes #2000

### DIFF
--- a/src/engine/strat_engine/backstore/crypt.rs
+++ b/src/engine/strat_engine/backstore/crypt.rs
@@ -18,9 +18,10 @@ use libcryptsetup_rs::{
     CryptVolumeKeyFlags, CryptWipePattern, EncryptionFormat, LibcryptErr, SafeMemHandle,
 };
 
-#[cfg(test)]
-use crate::engine::strat_engine::backstore::metadata::StratisIdentifiers;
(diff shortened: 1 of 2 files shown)


src/engine/strat_engine/backstore/crypt.rs
```diff
@@ -487,8 +482,16 @@ impl CryptHandle {
         Ok(name)
     }
 
+    /// Query the Stratis metadata for the device activation name.
```
mulkieran, line 485: This method looks fine, except for the header comment and error message, which look like copy/paste errors.

PR #2014, opened 2020-05-07
Get rid of unnecessary self uses

mulkieran:


mulkieran:
These probably got left behind when we fixed up macros to use fully qualified names.

PR #2015, opened 2020-05-07
Require blockdev() invocations to specify the path type

### LINKED ISSUE #2007: Make blockdev's Devnode D-Bus property the actual devnode
Currently, for encrypted devices, the value is the "/dev/mapper" link to the devnode. We want to return the actual devnode.

We would like to do this within the engine, rather than adjusting it within the D-Bus layer. We hope that not much internal code is making use of ```Blockdev::devnode()``` when it should be using a more internal value.

jbaublitz:
Closes #2007 

@mulkieran I have taken a different approach from what we discussed.

I decided to make `.devnode()` allowed internally as well as externally. This is necessary because of some of the places we operate on `&dyn BlockDev` types where we cannot use any methods other than the ones in the `BlockDev` trait. My way of mitigating our issue of not know what type of path we are getting back is converting `BlockDevPath` into a struct and making the fields private with two different methods to initialize a `BlockDevPath` type that should be easily extensible if we decide to add additional paths that we want to keep track of in the future.

`.devnode()` now returns a `BlockDevPath` reference which implements a number of different path getters. This then requires the user calling `devnode()` to explicitly state the type of path that they want which may actually help with code clarity. A typical invocation for the metadata_path has changed from:
```rust
bd.devnode()
```
to:
```rust
bd.devnode().metadata_path()
```
for example. Let me know what you think of this direction.

### DIFF
--- a/src/engine/types/mod.rs
+++ b/src/engine/types/mod.rs
@@ -2,7 +2,14 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-use std::{borrow::Borrow, convert::TryFrom, fmt, ops::Deref, rc::Rc};
+use std::{
+    borrow::Borrow,
+    convert::TryFrom,
+    fmt, io,
+    ops::Deref,
+    path::{Path, PathBuf},
+    rc::Rc,
+};
 
 mod actions;
 pub use crate::engine::types::actions::{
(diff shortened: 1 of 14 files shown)


jbaublitz:
Updating to move `blockdev()` implementation to the `impl StratBlockDev` and then call this implementation from the `BlockDev` trait.

jbaublitz:
Putting back in review.

src/engine/types/mod.rs
```diff
@@ -145,3 +152,60 @@ impl<'a> TryFrom<&'a str> for ReportType {
...
+            logical_path,
+        })
+    }
+
+    /// Path to the physical device storing the data.
+    pub fn physical_path(&self) -> &Path {
+        &self.physical_path
+    }
+
+    /// Path to the physical or logical device where the metadata should be written.
```
mulkieran, line 190: ^metadata^Stratis metadata

PR #2017, opened 2020-05-08
Manage encrypted devices in setup and device discovery

### LINKED ISSUE #1780: Handle duplicate Stratis device UUIDs for the same pool
Ideally, this would be checked for on both the initial startup path and on the udev-event drive part. For the udev-event driven part, this should be checked both for presence in existing pools, i.e., when a new device shows up but its doppelganger has already been incorporated into the pool and also that it does not introduce duplicate devices into the list of devices for "incomplete" pools.

Note that #748 and this issue intersect, although the scenario actually described in #748 does not mention duplicate Stratis UUIDs, the problem would be the same.

mulkieran:
Closes #1780
Related #1659
Related stratis-storage/project#171

### DIFF
--- a/src/engine/strat_engine/backstore/backstore.rs
+++ b/src/engine/strat_engine/backstore/backstore.rs
@@ -114,7 +114,7 @@ impl Backstore {
     /// Precondition:
     ///   * key_description.is_some() -> every StratBlockDev in datadevs has a
     ///   key description and that key description == key_description
-    ///   *key_description.is_none() -> no StratBlockDev in datadevs has a
+    ///   * key_description.is_none() -> no StratBlockDev in datadevs has a
     ///   key description.
     ///   * no StratBlockDev in cachedevs has a key description
     ///
--- a/src/engine/strat_engine/backstore/backstore.rs
+++ b/src/engine/strat_engine/backstore/backstore.rs
@@ -678,7 +678,7 @@ impl Recordable<BackstoreSave> for Backstore {
 
 #[cfg(test)]
 mod tests {
-    use std::{collections::HashMap, fs::OpenOptions, path::PathBuf};
+    use std::{collections::HashMap, fs::OpenOptions};
 
     use uuid::Uuid;
 
@@ -686,9 +686,12 @@ mod tests {
 
     use crate::engine::{
         strat_engine::{
-            backstore::metadata::device_identifiers,
+            backstore::{
+                identify::StratisInfo,
+                metadata::{device_identifiers, StratisIdentifiers},
+            },
             cmd,
-            liminal::{convert_to_infos, get_bdas, get_blockdevs},
+            liminal::{get_bdas, get_blockdevs, LStratisInfo},
             tests::{loopbacked, real},
         },
         types::DevUuid,
@@ -898,7 +901,7 @@ mod tests {
 
         let pool_uuid = Uuid::new_v4();
 
-        let (backstore_save, devices): (_, HashMap<Device, (DevUuid, PathBuf)>) = {
+        let (backstore_save, infos): (_, HashMap<DevUuid, LStratisInfo>) = {
             let mut backstore =
                 Backstore::initialize(pool_uuid, paths1, MDADataSize::default(), None).unwrap();
 
@@ -944,8 +947,18 @@ mod tests {
                     .iter()
                     .map(|(device_uuid, blockdev)| {
                         (
-                            *blockdev.device(),
-                            (*device_uuid, blockdev.devnode().physical_path().to_owned()),
+                            *device_uuid,
+                            LStratisInfo {
+                                ids: StratisInfo {
+                                    identifiers: StratisIdentifiers {
+                                        pool_uuid,
+                                        device_uuid: *device_uuid,
+                                    },
+                                    device_number: *blockdev.device(),
+                                    devnode: blockdev.devnode().metadata_path().to_owned(),
+                                },
+                                luks: None,
+                            },
                         )
                     })
                     .collect(),
@@ -953,7 +966,6 @@ mod tests {
         };
 
         {
-            let infos = convert_to_infos(pool_uuid, &devices);
             let bdas = get_bdas(&infos).unwrap();
             let (datadevs, cachedevs) = get_blockdevs(&backstore_save, &infos, bdas).unwrap();
             let mut backstore = Backstore::setup(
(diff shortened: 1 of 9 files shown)


mulkieran:
rebased, conflicts resolved.

mulkieran:
Prior failure looked quite spurious, there were LUKS devices found in the tests which must have been sitting around from a previous test.

mulkieran:
Encountered a lingering LUKS device again...

mulkieran:
rebased, still expecting all ```stratisd_nonrust``` tests to pass.

mulkieran:
And they did!

mulkieran:
Same spurious error about LUKS devices sitting around...

mulkieran:
@jbaublitz: Great! I've developed one worry since I wrote this, though. I think that moving to hopeless is a bit too aggressive when duplicate device UUIDs are found. I think it would be better to check the device number of the older info for existence, and if it doesn't exist replace the older info with the newer one. I think this would handle the real possiblity that a device gets removed and then plugged back in and thereby receives a new device number. We can't duplicate that behavior with our udev tests setup unfortunately, but I think it's a realistic possibility.

jbaublitz:
@mulkieran That makes sense to me because we don't register "remove" events, do we?

mulkieran:
@jbaublitz No, we don't register those events, we only accumulate info. Which is a decision we inherited from the previous implementation; it's good you brought that up. We should actually discuss the implications of that in standup.

mulkieran:
rebased.

mulkieran:
rebased.

src/engine/strat_engine/backstore/backstore.rs
```diff
@@ -102,12 +102,22 @@ impl Backstore {
...
     /// to the data tier cap device.
+    ///
     /// Precondition: backstore_save.data_segments is not empty. This is a
     /// consequence of the fact that metadata is saved by the pool, and if
     /// a pool exists, data has been allocated to the cap device.
+    ///
+    /// Precondition:
+    ///   * key_description.is_some() -> every StratBlockDev in datadevs has a
+    ///   key description and that key description == key_description
+    ///   *key_description.is_none() -> no StratBlockDev in datadevs has a
```
drckeefe, line 117: should there be a space between * and key_description.is_none()?  If so, it is also in pool.rs

src/engine/strat_engine/backstore/identify.rs
```diff
@@ -118,6 +188,55 @@ fn device_identifiers_wrapper(
...
+                Err(err) => {
+                    warn!(
+                            "udev identified device {} as a LUKS device, but could not read LUKS header from the device, disregarding the device: {}",
+                            devnode.display(),
+                            err,
+                            );
+                    None
+                }
+                Ok(Some(handle)) => {
+                    match KeyDescription::from_system_key_desc(handle.key_description()) {
```
jbaublitz, line 214: I looked into it and when I reviewed your changes to my key management PR, I thought you had also changed `CryptHandle` to use `KeyDescription`. It seems that I missed that and I think it really should. Based on this, I think it would be good to change this to `KeyDescription` instead of a `String` so that either can be accessed.

src/engine/strat_engine/backstore/metadata/static_header.rs
```diff
@@ -66,6 +67,15 @@ impl fmt::Display for StratisIdentifiers {
     }
 }
 
+impl<'a> Into<Value> for &'a StratisIdentifiers {
+    fn into(self) -> Value {
+        json!({
+            "pool_UUID": Value::from(self.pool_uuid.to_simple_ref().to_string()),
+            "device_UUID": Value::from(self.device_uuid.to_simple_ref().to_string())
```
jbaublitz, line 74: These should be `pool_uuid` and `device_uuid` if we're following conventions for JSON keys.

src/engine/strat_engine/liminal.rs
```diff
@@ -316,35 +312,308 @@ pub fn get_blockdevs(
...
+            key_description: info.key_description,
+        }
+    }
+}
+
+impl<'a> Into<Value> for &'a LLuksInfo {
+    // Precondition: (&StratisInfo).into() pattern matches Value::Object()
+    fn into(self) -> Value {
+        let mut json = json!({
+            "key_description": Value::from(self.key_description.to_system_string())
```
jbaublitz, line 347: Is the reason you've chosen `to_system_string()` for debugging purposes?

src/engine/strat_engine/backstore/backstore.rs
mulkieran: Yes. It's meant to be a bullet. I can fix that.

src/engine/strat_engine/liminal.rs
```diff
@@ -422,13 +761,208 @@ impl LiminalDevices {
...
+        // device attached to the event.
+        fn combine_remove_devices(info_1: LInfo, info_2: LInfo) -> Option<LInfo> {
+            match (info_1, info_2) {
+                (LInfo::Luks(luks_info), LInfo::Stratis(_)) => Some(LInfo::Luks(luks_info)),
+                (LInfo::Stratis(strat_info), LInfo::Luks(_)) => {
+                    Some(LInfo::Stratis(LStratisInfo {
+                        ids: strat_info.ids,
+                        luks: None,
+                    }))
+                }
```
jbaublitz, line 930: I'm not sure what to make of this handling mainly because it seems like this would only occur if a closed encrypted device has the same identifiers as a physical device that's been registered with Stratis. Can you explain what the motivation was for this?

src/engine/strat_engine/backstore/identify.rs
mulkieran: I changed the ```CryptInitializer``` methods to take a ```KeyDescription``` as a parameter. I was more hesitant with ```CryptHandle``` for what are probably trifling semantic reasons. I'll go ahead and see how that goes.

src/engine/strat_engine/liminal.rs
mulkieran: > Is the reason you've chosen `to_system_string()` for debugging purposes?

I was of two minds about this. I'll change it to ```as_application_str```, at present it doesn't really make any difference to me.

src/engine/strat_engine/liminal.rs
mulkieran: > I'm not sure what to make of this handling mainly because it seems like this would only occur if a closed encrypted device has the same identifiers as a physical device that's been registered with Stratis. Can you explain what the motivation was for this?

It doesn't have any good physical explanation. I'll plug in some warnings, and we can see if they ever turn up.

mulkieran:
@jbaublitz Ok, I think this is ready for another look.

jbaublitz:
This all looks good to me!

Issue #1987, opened 2020-04-27
devlinks module uses Pool, Filesystem traits, but shouldn't


mulkieran:
Since it is internal to the strat engine, it should use the concrete strat engine types. This issue really follows #1690 which wasn't entirely completed.

PR #2020, opened 2020-05-08
Move most of devlinks functionality into a udev rule

jbaublitz:
Closes stratis-storage/project#95

@mulkieran One important note is that I left the devlinks module around for two specific cases:
* Testing
* Triggering synthetic uevents when filesystems and pools are renamed to indicate to udev that it should make the symlinks consistent

Other than that, I've deleted a good deal of the code in devlinks and moved it into the udev rule and the script it calls.


jbaublitz:
We'll see how CI fares with these changes.

jbaublitz:
@mulkieran All of the tests should pass except for the udev test with duplicate pool names. That one should pass once we get the udev rule on the Jenkins nodes but until then will not.

jbaublitz:
I am very confused as to why that test passed...

jbaublitz:
Ok, the logs indicate that the tests passing is valid. stratisd fails to rename the symlinks as expected, but it's simply that the tests don't depend on the symlink, only that udev picks up the pools being renamed.

mulkieran:
This might also close #1987.

mulkieran:
@jbaublitz Please rebase.

PR #2021, opened 2020-05-11
Update version to 2.1.0 in Cargo.toml

lleshchi:
Related https://github.com/stratis-storage/project/issues/176

Precedes https://github.com/stratis-storage/stratis-cli/pull/552

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "libstratis"
-version = "2.0.1"
+version = "2.1.0"
 authors = ["Stratis Developers <stratis-devel@lists.fedorahosted.com>"]
 edition = "2018"
 build = "build.rs"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "libstratis"
-version = "2.0.1"
+version = "2.1.0"
 authors = ["Stratis Developers <stratis-devel@lists.fedorahosted.com>"]
 edition = "2018"
```
lleshchi, line 5: Should the edition be updated? If so, Jenkins checks currently fail.

jbaublitz: There is currently no Rust 2020 edition. It's a little bit confusing but they only periodically release editions. The latest one that I'm aware of is 2018.

mulkieran: There are rumours of a 2020 or 2021 edition, with better support for error management, but I haven't heard anything definite yet.

PR #2026, opened 2020-05-12
Move functionality from stratisd.rs into libstratis

mulkieran:
To enable testing of behavior in rust unit or integration tests instead of just across the D-Bus.


mulkieran:
The wierd thing is that the no-defaults Travis task failed for me when I had an error in the optional dbus implementation where I import from dbus_api without protecting the import with a cfg...dbus_enabled flag, but not on Travis.

mulkieran:
Superceded by #2085.

PR #2029, opened 2020-05-13
More udev test tidies

mulkieran:
Continue to tidy up udev tests. This is partly a general tidy, but another important thing it does is it jettisons all tests marked as expected failures that have to do with encryption, and it removes the code that was introduced to support these tests. It preserves any improvements that were introduced while introducing those tests. The purpose of all this is to omit tests that require key management, but are expected to fail from the tests that have to be fixed up in #1983, and make the work  on that easier.


mulkieran:
> I think the errors in Jenkins are residual from a failed run so I'm approving this.

Yes, they've been cropping up every so often.

PR #2034, opened 2020-05-16
Jbaublitz issue project 171

mulkieran:


mulkieran:
Well, the kernel doesn't seem to have any problems with key descriptions with ";"s in them, or, at least, it seems like the key was successfully created. This is really a sideshow, but I guess we should see if it is possible to set keys with ";"s in the key descriptions and thereby cause ourselves some trouble. I guess the only real trouble that we know about is that the list of Stratis keys would turn up empty, because the identifying prefix would never be read.

mulkieran:
Another slight concern related to listing is number of keys. It looks like the root user is allowed a million these days.

mulkieran:
stratisd error was due to LUKS devices hanging around.

jbaublitz:
> Well, the kernel doesn't seem to have any problems with key descriptions with ";"s in them, or, at least, it seems like the key was successfully created. This is really a sideshow, but I guess we should see if it is possible to set keys with ";"s in the key descriptions and thereby cause ourselves some trouble. I guess the only real trouble that we know about is that the list of Stratis keys would turn up empty, because the identifying prefix would never be read.

I think you should take a look at the documentation for [`keyctl()`](http://man7.org/linux/man-pages/man2/keyctl.2.html) under the `KEYCTL_DESCRIBE` section. In particular, it explicitly states the following:
> In particular, the description field will not contain semicolons; it should be parsed by working backwards from the end of the string to find the last semicolon.

It's interesting to hear that you seem to be able to set a semicolon and we should probably figure out if the documentation is out of date or not.

jbaublitz:
> Another slight concern related to listing is number of keys. It looks like the root user is allowed a million these days.

We can very easily update the buffer for key IDs so I'm happy to just put it to the maximum number of keys and allocate it to the heap so we don't cause stack problems if that's preferable. I didn't think my original value would necessarily be the final option and mainly just didn't want to allocate a very large memory segment if we didn't need to. However, now that I think of it, we need all of the keys under and outside of Stratis's management to be able to reliably filter and list all of Stratis's keys so this might make sense.

jbaublitz:
One additional comment on your allocation strategy for commit 0e7fc99ad83ba1ffb5fce4ef9d59b2c9508d1062: allocating a `Vec` and passing a pointer to a vector into a syscall doesn't grow the length dynamically. If you want to dynamically grow the length, you would have to unsafely set the length to the desired length after reading it into the pointer to the vector with enough capacity. As it stands, your commit would iterator through an empty iterator as iter is going to use length, not capacity to determine how many elements are in the iterator and I don't see an explicit setting of the length after you read to the buffer. My typical pattern here is initializing a vector with the same length as capacity and then truncating the buffer to the desired size as this does not reduce capacity, only length, and can be done entirely in safe code unlike the alternative I mentioned.

mulkieran:
> One additional comment on your allocation strategy for commit [0e7fc99](https://github.com/stratis-storage/stratisd/commit/0e7fc99ad83ba1ffb5fce4ef9d59b2c9508d1062): allocating a `Vec` and passing a pointer to a vector into a syscall doesn't grow the length dynamically. If you want to dynamically grow the length, you would have to unsafely set the length to the desired length after reading it into the pointer to the vector with enough capacity. As it stands, your commit would iterator through an empty iterator as iter is going to use length, not capacity to determine how many elements are in the iterator and I don't see an explicit setting of the length after you read to the buffer. My typical pattern here is initializing a vector with the same length as capacity and then truncating the buffer to the desired size as this does not reduce capacity, only length, and can be done entirely in safe code unlike the alternative I mentioned.

I avoid that problem by using then ```len()``` method that you defined in the ```KeyIdList``` impl that returns the value of ```num_key_ids```. But I think you're right that there could be a better way that drops the ```num_key_ids``` field entirely, using ```Vec::resize```. I'll give that a try.

mulkieran:
> > Well, the kernel doesn't seem to have any problems with key descriptions with ";"s in them, or, at least, it seems like the key was successfully created. This is really a sideshow, but I guess we should see if it is possible to set keys with ";"s in the key descriptions and thereby cause ourselves some trouble. I guess the only real trouble that we know about is that the list of Stratis keys would turn up empty, because the identifying prefix would never be read.
> 
> I think you should take a look at the documentation for [`keyctl()`](http://man7.org/linux/man-pages/man2/keyctl.2.html) under the `KEYCTL_DESCRIBE` section. In particular, it explicitly states the following:
> 
> > In particular, the description field will not contain semicolons; it should be parsed by working backwards from the end of the string to find the last semicolon.
> 
> It's interesting to hear that you seem to be able to set a semicolon and we should probably figure out if the documentation is out of date or not.

Yes, I looked at that documentation and took it as the gospel truth originally, and even added a comment based on that to the ```parse``` method explaining why it parsed from the right as recommended. But then I remembered that you can't always trust the kernel docs, so I tried the experiment of setting a key with ";" in the udev test, and it seemed to be set just fine. My guess is that somewhere else, there is an instruction never to put a semi-colon in a key description, but that the kernel itself does no checking on that. So, after some additional thought, I added a semi-colon check to ```KeyDescription``` and at that point, I consider that Stratis has done all it can do.

mulkieran:
> > > Well, the kernel doesn't seem to have any problems with key descriptions with ";"s in them, or, at least, it seems like the key was successfully created. This is really a sideshow, but I guess we should see if it is possible to set keys with ";"s in the key descriptions and thereby cause ourselves some trouble. I guess the only real trouble that we know about is that the list of Stratis keys would turn up empty, because the identifying prefix would never be read.
> > 
> > 
> > I think you should take a look at the documentation for [`keyctl()`](http://man7.org/linux/man-pages/man2/keyctl.2.html) under the `KEYCTL_DESCRIBE` section. In particular, it explicitly states the following:
> > > In particular, the description field will not contain semicolons; it should be parsed by working backwards from the end of the string to find the last semicolon.
> > 
> > 
> > It's interesting to hear that you seem to be able to set a semicolon and we should probably figure out if the documentation is out of date or not.
> 
> Yes, I looked at that documentation and took it as the gospel truth originally, and even added a comment based on that to the `parse` method explaining why it parsed from the right as recommended. But then I remembered that you can't always trust the kernel docs, so I tried the experiment of setting a key with ";" in the udev test, and it seemed to be set just fine. My guess is that somewhere else, there is an instruction never to put a semi-colon in a key description, but that the kernel itself does no checking on that. So, after some additional thought, I added a semi-colon check to `KeyDescription` and at that point, I consider that Stratis has done all it can do.

Yes, confirmed more directly:

```
> keyctl list 213798219
...
708354073: --alswrv  1000  1000 user: with;with
```

jbaublitz:
Should I consider the commits ready to be applied?

mulkieran:
@jbaublitz: Yes, please do. I think I'm done.

mulkieran:
Merged into #1983.

PR #2044, opened 2020-05-21
Modify device discovery and setup to support encryption

### LINKED ISSUE #1780: Handle duplicate Stratis device UUIDs for the same pool

mulkieran:
Closes #1780
Related #1659
Related stratis-storage/project#171

Supercedes #2017


mulkieran:
@jbaublitz Authors can't request themselves for review. Ready to sign off?

mulkieran:
Travis is done...just not reporting back.

PR #2045, opened 2020-05-21
Make all Python based tests pure unittest

mulkieran:
Related: https://github.com/stratis-storage/project/issues/105


mulkieran:
Once this is merged we can remove pytest from fedora-dependencies.sh

mulkieran:
Note also that there is something interesting now being reported from the ```unique_instance``` test case.

bgurney-rh:
> Note also that there is something interesting now being reported from the `unique_instance` test case.

Interesting; "unclosed file" may explain some of the hangs that I saw in rare instances of the stratisd-nonrust test.

Issue #1741, opened 2019-12-10
Modify *Action types slightly


mulkieran:
I think it might be useful to make them more uniform. There's a lot of binariness about them, and I'ld kind of prefer them to be a bit more binary. It's a classic to separate list into empty and other, and I'ld kind of like to do that, even though the component of a Set*Action is actually a Vec, and not a list.

So, change *Action to have two constructors, Identity and Changed.

So, change Set*Action so that it has two constructors, Identity and Changed, and which one is made can be decided on by the ```new``` function based on whether the Vec argument is empty or not.

For RenameAction, have two constructors, Source and NoSource, and let the argument of Source be itself an enumerated type w/ two constructors, Identity and Changed. NoSource is a nullary constructor, as it is now.

I think that this would increase uniformity in a useful way.

mulkieran:
@jbaublitz: Please look into this and see if it is doable. (I was going to let this sit for a while, but I noticed that #1705 has some additions to one action type, and so it seemed best to move this up a bit).

jbaublitz:
@mulkieran I've been attempting to implement both of these changes and I'm seeing potential problems with both. While my original implementation does not have these issues, I'm still open to other suggestions that may be more uniform and not have the problems I'll outline below.

### `Set*Action` as an enum
Making this a two-variant enum becomes an issue because of the inner `Vec`. While this at first glance provides more uniform handling, `Vec`s in an enum can easily be constructed as a zero length vector. This arises from the lack of public/private controls on enums. As a result, we will be forced to check two aspects of the data structure to see if anything changed. One is a newtype variant with an empty `Vec`. The other is the unit variant. Either of these indicates no changes and we cannot force those interacting with the API to choose one over the other, only suggest. This seems to add more complexity any checking that is done on this enum and will bubble up into the D-Bus layer. My current solution only has one possible condition in which an action with no state change was carried out. If we can keep it that way, I think it will make things simpler regardless of what we choose.

### Variants indicating failure to carry out a requested action
[...]

PR #2046, opened 2020-05-22
Unlock discovered encrypted devices

jbaublitz:
This is up for preliminary review. So far, it only contains the engine code.

I've followed the pattern of existing `Action` types in the `actions.rs` module despite what we've discussed in #1741 because of the complexity it adds to handling the types (checking both for the empty variant and an empty list). I feel like this may cause code clarity problems.

Related to stratis-storage/project#171

### DIFF
--- a/src/engine/strat_engine/engine.rs
+++ b/src/engine/strat_engine/engine.rs
@@ -17,17 +17,17 @@ use crate::{
         event::get_engine_listener_list,
         shared::create_pool_idempotent_or_err,
         strat_engine::{
-            backstore::find_all,
+            backstore::{find_all, CryptHandle},
             cmd::verify_binaries,
             devlinks,
             dm::{get_dm, get_dm_init},
             keys::StratKeyActions,
-            liminal::LiminalDevices,
+            liminal::{LInfo, LLuksInfo, LiminalDevices},
             names::{validate_name, KeyDescription},
             pool::StratPool,
         },
         structures::Table,
-        types::{CreateAction, DeleteAction, RenameAction, ReportType},
+        types::{CreateAction, DeleteAction, DevUuid, RenameAction, ReportType, SetUnlockAction},
         Engine, EngineEvent, Name, Pool, PoolUuid, Report,
     },
     stratis::{ErrorEnum, StratisError, StratisResult},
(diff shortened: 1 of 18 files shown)


jbaublitz:
I'll keep on with the D-Bus portion.

src/engine/strat_engine/engine.rs
```diff
@@ -253,6 +253,65 @@ impl Engine for StratEngine {
...
+                        probably due to a missing key: {}",
+                        luks_info.ids.devnode.display(),
+                        luks_info.ids.identifiers.device_uuid.to_simple_ref(),
+                        e,
+                    );
+                    false
+                } else {
+                    true
+                }
+            } else {
```
jbaublitz, line 271: One note here is that we may not want to report this as "still locked" as this case would be hit more due to disk formatting issues. What do you think, @mulkieran?

mulkieran:
In standup we decided to split up the actions on the D-Bus side so that each pool is activated separately, rather than have a single call to activate all pools.

jbaublitz:
Changed have been made to enable the engine to unlock one pool at a time. Starting on the D-Bus interface.

src/engine/strat_engine/engine.rs
```diff
@@ -253,6 +253,50 @@ impl Engine for StratEngine {
         }
     }
 
+    fn unlock_pool(&mut self, pool_uuid: PoolUuid) -> StratisResult<SetUnlockAction<DevUuid>> {
```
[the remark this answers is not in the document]
jbaublitz: Sure thing!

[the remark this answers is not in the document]
jbaublitz: Resolved.

jbaublitz:
The D-Bus API is now implemented too. The handling of the input is slightly different from usual because typically we receive object paths but because these pools don't yet have an associated object path, we need to use the UUID as the input.

src/engine/strat_engine/liminal.rs
```diff
@@ -529,6 +529,74 @@ impl LiminalDevices {
...
+        Ok(unlocked)
+    }
+
+    /// Get a list of pool UUIDs from all of the LUKS2 devices that are currently
+    /// locked in the set of pools that are not yet set up.
+    pub fn locked_pool_uuids(&self) -> Vec<PoolUuid> {
+        self.errored_pool_devices
+            .iter()
+            .filter_map(|(pool_uuid, map)| {
+                let has_locked = map.iter().fold(false, |has_locked, (_, info)| {
```
mulkieran, line 583: ```fold``` works here, but ```any``` is more direct and a bit easier to understand. It's used elsewhere, probably in ```setup_pool```.

mulkieran: I think that at some point we might want to establish a purpose-built data structure for the values of ```errored_pool_devices```, but that's not something we need to worry about now.

mulkieran:
In standup, we decided to stick w/ the error return if any device in the pool could not be activated.

src/engine/strat_engine/liminal.rs
jbaublitz: Resolved.

mulkieran:
@jbaublitz Sorry, I accidentally hit that button somehow, and now GitHub has a merge commit. You can force push to remove it.

jbaublitz:
@mulkieran Resolved.

jbaublitz:
@mulkieran I think the boolean handling should be fixed for this PR now.

mulkieran:
For the D-Bus API, UnlockPool method, we decided that a request to unlock the devices in a pool that had not previously been encrypted would return an error, if the devices were still unset up and if the pool had been setup. For an encrypted pool, there will be no error if the pool is already set up, this is considered the identity.

jbaublitz:
Change made to error on unencrypted pools and return the identity on set up encrypted pools.

jbaublitz:
@mulkieran Changes applied.

mulkieran:
@jbaublitz If you're comfortable with this, you can close it and put up a new PR for final review. The log entry that perplexed me is exactly what should be expected, although superficially it seems odd. I'm thinking about filing an issue, but the issue is not about behavior, but only about the when log entries are generated.

PR #2048, opened 2020-05-22
Do not return an error if key description could not be read

mulkieran:


mulkieran:
There are actually more deep errors, so closing this.

PR #2049, opened 2020-05-22
Fix some bugs in key list methods

mulkieran:


mulkieran:
Problem still unsolved, the description syscall still fails, even though the key id is certainly valid.

mulkieran:
I think it will be best to merge this, in order to restore the blackbox tests to normal functioning.

PR #2052, opened 2020-05-27
Expose hopeless_pools in engine state report

### LINKED ISSUE #2050: Extend Report of engine state to include hopeless sets
Right now it only includes the sets of devices that have a chance of becoming pools. It should include the sets that have been assigned to the hopeless category.

jbaublitz:
Closes #2050


jbaublitz:
Okay, I think this should be fixed.

PR #2053, opened 2020-05-27
Leave the LUKS device behind when a Stratis device is removed

mulkieran:


mulkieran:
I believe this change is obviously correct, so we might as well make it, regardless of whether we "need" it to be correct for device activation.

PR #2055, opened 2020-05-28
Jbaublitz issue project 171 pt2

mulkieran:


mulkieran:
Test was a bit broken adding fixes to test and resubmitting.

mulkieran:
Whoops. Another test error.

mulkieran:
There is an error in the test, but it's interesting. I can't seem to get the D-Bus object for the object path of the newly created pool. It may be that the D-Bus object just hasn't been made available yet. The important thing is that both encrypted and non-encrypted are failing consistently in the same way.

mulkieran:
Nope. Uninteresting error.

mulkieran:
We have a real passing test!

jbaublitz:
For the idempotency tweak, this would also return a successful value for unencrypted pools. I would think that we'd at least want to guard against unlocking an unencrypted pool if we're going to add this code.

mulkieran:
> For the idempotency tweak, this would also return a successful value for unencrypted pools. I would think that we'd at least want to guard against unlocking an unencrypted pool if we're going to add this code.

We already go ahead and unlock sets of unencrypted devices.

I _think_ I would be more comfortable if we were consistent about that if we decided to do that. We can do a check of sets of devices. If they contain StratisInfos w luks field None, then they are unencrypted, and an error is returned. If already a setup pool which is not encrypted, then an error is returned. That leaves the non-error situation being only for pools that are encrypted or sets of devices where Stratis is managing the encryption.

But this is more complicated then the approach of unlocking anything that doesn't fail to unock.

mulkieran:
@jbaublitz: The stratisd test had a kind of odd error in a crypt ownership test. It's hard to trace where it came from because of the nature of the test, but you might want to take a look. I don't think I've ever seen it before.

mulkieran:
@jbaublitz: Ok, thanks. I'll keep my out for it, and maybe we can build up the test to make it easier to trace the source of that error if it happens again.

mulkieran:
Last commit is not a lot of code added, but it is a significant new test. I have pretty strong expectations of it passing.

mulkieran:
Test mistake.

mulkieran:
And passing now.

mulkieran:
Passing!

mulkieran:
Merged  into other PRs.

PR #2063, opened 2020-05-30
Improvements to handling of loopbacked devices in udev testing infrastructure

mulkieran:


mulkieran:
Yes!

mulkieran:
So, if the udev tests are fixed, so that they better simulate reality, they all pass even if stratisd does not catch the Change event. See #2064 .

mulkieran:
Extending my dictatorial powers even further; however, it is all udev test infrastructure code changes.

PR #2083, opened 2020-06-11
Remove a module that has lost its purpose

mulkieran:


mulkieran:
Think I'ld better add a comment as well.

PR #2084, opened 2020-06-11
Add more context for some errors

mulkieran:
Supercedes #2080 .


mulkieran:
Wierd non-rust test error. udev probably was just too slow.

Issue #1977, opened 2020-04-22
Organized scheme for testing the multiple versioned D-Bus interfaces


mulkieran:
There should be some way to do this efficiently without too much code duplication. Right now the tests do their work and they pass, but it's all a bit disorganized.

mulkieran:
Closing in favor of https://github.com/stratis-storage/ci/issues/97.

Issue #1319, opened 2018-11-12
test_udev.py functionality maybe more properly belongs in stratisd unit tests


mulkieran:
Related: #1189.

mulkieran:
The reason for this supposition is that these tests can not interact directly with the udev functionality or directly test behavior and state of engine, but must instead interrogate the D-Bus. Obviously, the D-Bus exports vastly less information about actual behavior than could be obtained if access to actual engine data structures were available to the tests.

mulkieran:
This is becoming obsolete. The external approach that we are using has been much improved and we plan to expose more information on the D-Bus, so the tests will be able to do better verification by means of the  D-Bus: see #2194.

PR #2085, opened 2020-06-17
Drop buffered logging and dbus_interface tests

### LINKED ISSUE #1055: Hang in event loop will prevent log dump via signal
With https://github.com/stratis-storage/stratisd/pull/1026 we will have the ability to dump debug information.  However, this requires the event loop to be operational, which depending on the type of error encountered may not be true (eg. stalled external command).

### LINKED ISSUE #1262: Document log buffering and use of SIGUSR1
Document in the stratisd man page our buffering behavior if stratisd is not run with --debug (by default it is not) and how to ensure all buffered log output is dumped using SIGUSR1.

### LINKED ISSUE #1433: Log entries are expected to include timestamp, but do not
See related discussion #1430.

### LINKED ISSUE #2060: Check the return code from D-Bus calls in unittest.setUp methods
Otherwise, the setUp method may not set the test up correctly, but this will not be clear, leading to some confusion on test failure. See stratis-storage/stratisd#1852 for an example.

### LINKED ISSUE #2086: Get rid of the buffering in the logger
I think this was introduced to mitigate the problems causing by dumping the Debug representation of the engine. But we don't do that any more, see #1901, and we do give access to a representation of the state via the D-Bus, see #1981, so we can let the buffering go away.

mulkieran:
This PR does the following:

* Ports a bunch of code previously in stratisd.rs into the library. This increases order, allows better linting, and will allow better testing via the Rust testing interface.
* Drops the use of the buffered logger, reverting to the standard env_logger interface. Reverts to the standard timestamp for the env_logger, which is UTC. Drops the --debug option from stratisd. It was overloaded oddly. If we ever change to a different logging impelemtation, it will make even less sense. This means that the logger will log at the default level, which is ERROR, unless the  the RUST_LOG environment variable is set. Buffered logs always logged at the trace level, unless RUST_LOG was set.
* Drops the dbus_interface tests. As the project matures, these are less and less useful.

Closes #2086 
Closes #1433 
Closes #2060
Closes #1262
Closes #1055
Closes https://github.com/stratis-storage/project/issues/105
Related #1977
Related #1319


mulkieran:
Things to do:
* Extend timestamp to have greater than 1 second precision?
* Add a test within the Rust test framework that takes advantage of the newly exposed code in the library.

mulkieran:
Preliminary review only.

mulkieran:
Ok! Seems like everybody's on board with this generally.

mulkieran:
@bgurney-rh If you find the time, can you verify that the installed version of stratisd from this PR is logging at the debug level? This is not super-urgent, and it ought to be true, but it would be good to confirm it.

mulkieran:
Now, running with the ```--debug``` flag set is unsupported.

```
[amulhern@mulhern stratisd]$ ./target/debug/stratisd  --debug --sim
error: Found argument '--debug' which wasn't expected, or isn't valid in this context

USAGE:
    stratisd [FLAGS]

For more information try --help
```

mulkieran:
In order to log at the debug level, not the error level, must set RUST_LOG environment variable:

```
# RUST_LOG=stratisd=debug,libstratis=debug ./target/debug/stratisd --sim
 INFO 2020-06-22T21:06:38Z: libstratis::stratis::run: stratis daemon version 2.1.0 started
 INFO 2020-06-22T21:06:38Z: libstratis::stratis::run: Using SimEngine
 INFO 2020-06-22T21:06:38Z: libstratis::stratis::dbus_support: D-Bus API is available
^C INFO 2020-06-22T21:06:41Z: libstratis::stratis::run: SIGINT received, exiti
```

mulkieran:
I think this is ready for final review. We won't merge it until we've confirmed that the systemd service file is doing the right thing.

mulkieran:
Re-requesting review. The only thing new is the last commit.

mulkieran:
This'll be blocked by #2093, which should go in first, so "pending".

mulkieran:
rebased. conflicts resolved.

mulkieran:
We decided to add a ```--log-level``` flag w/ the normal semantics.

mulkieran:
Up for review with a ```--log-level``` option added. Here's the help:

```
# RUST_LOG=semver ./target/debug/stratisd --help
stratis 2.1.0
Stratis storage management

USAGE:
    stratisd [FLAGS] [OPTIONS]

FLAGS:
    -h, --help       Prints help information
        --sim        Use simulator engine
    -V, --version    Prints version information

OPTIONS:
        --log-level <log-level>    Sets level for generation of log messages. [values: trace, debug, info, warn, error]
```

mulkieran:
@bgurney-rh If you could run a blackbox test and check that this version of the PR also has logging at the debug level, but not the trace level, that would be great.

bgurney-rh:
In "journalctl -f" on the machine running the blackbox test, I see "WARN", "INFO" and "DEBUG" messages scrolling by, but not "TRACE".

PR #2088, opened 2020-06-18
Demote debug log entry to trace log entry

mulkieran:


mulkieran:
This is absolutely a trace level log entry. If it is at the trace level, we can keep it around, to remind us that that method is being executed way too often, and, at the same time, we can print all debug level and above logging in our Jenkins tests. I did a quick check, and this should double the number of log entries that we currently see in the udev tests, which is not so insupportable.

PR #2091, opened 2020-06-22
Switch to udev for symlink handling

jbaublitz:
Supersedes #2020 
Closes stratis-storage/project#95
Related #1987


scripts/stratis_udev/stratis_uuids_to_names.py
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/python3
```
mulkieran, line 1: Right now the ```scripts``` directory is holding one Python script that we would not choose to distribute. But this one we would. So, either this should go somewhere else, or more likely the existing script should be moved somewhere else (or possibly removed).

mulkieran: We would want to run all our usual linting and formatting facilities on this script once we decided where to put it. I ran the linter and saw a couple actually interesting errors, as well as some minor things. I'm just going to do a high-level review for right now.

scripts/stratis_udev/stratis_uuids_to_names.py
```diff
@@ -0,0 +1,72 @@
...
+#
```
mulkieran, line 2: I have this recollection that somebody told us that it would be just as good, probably even better, to write this in Rust and install the executable, but I can't remember quite who said that. I'm not sure how doing that would influence our decision about where to put it.

mulkieran:
Looks like we have a bunch of valid complaints from the beta Rust linter.

jbaublitz:
I'll take a look. Last time I ran this the linter passed so I'll investigate this.

scripts/stratis_udev/stratis_uuids_to_names.py
jbaublitz: Yes, the recommendation is that udev shell out as little as possible. I was thinking of rewriting this in Rust so does that change your thought on where we should put it in the repository?

jbaublitz:
> Where is the standard place for any script that a udev rule will invoke?

Per the udev maintainer, this is `/usr/lib/udev`.

src/engine/strat_engine/devlinks.rs
```diff
@@ -2,170 +2,266 @@
...
+    use crate::engine::{
+        engine::Pool,
+        strat_engine::pool::StratPool,
+        types::{Name, PoolUuid},
+    };
 
-        for leftover in existing_dirs {
-            pool_removed(&Name::new(leftover));
+    /// Set up the root Stratis directory, where dev links as well as temporary
+    /// MDV mounts will be created. This must occur before any pools are setup.
```
mulkieran, line 121: I believe that the statement about temporary MDV mounts will become wholely false with this change, so it might as well be removed.

scripts/stratis_udev/stratis_uuids_to_names.py
mulkieran: Right now I'm inclining toward moving the existing Python script to somewhere less canonical, since it's for internal use only. ```scripts``` is a pretty usual designation for things that might go in a release. Maybe ```developer_toolbox``` would be a better choice for a place to stow the existing script; that makes it pretty obvious that its not for regular distribution.

But that doesn't exactly solve the problem of where this should go if it were in Rust.

mulkieran: A quick look says that our bin directory might not be such a bad idea.

mulkieran:
beta clippy still complaining about some minor things...

jbaublitz:
> beta clippy still complaining about some minor things...

No changes were made, I just rebased.

src/engine/strat_engine/devlinks.rs
```diff
@@ -2,170 +2,266 @@
...
+                "Synthetic udev events were not able to be triggered: {}. Migration of \
+                filesystem links associated with renamed pool {} failed",
+                e, pool_name,
+            );
         }
+    }
+
+    pub fn filesystem_renamed(pool_name: &str, fs_name: &str, _: &str) {
+        fn trigger_udev(pool_name: &str, fs_name: &str) -> StratisResult<()> {
+            let path: PathBuf = [DEV_PATH, pool_name, fs_name].iter().collect();
```
mulkieran, line 72: Might as well use ```filesystem_mount_path()``` method here.

udev/99-stratisd.rules
```diff
@@ -0,0 +1 @@
+ENV{DM_NAME}=="stratis-1-[0-9a-f]*-thin-fs-[0-9a-f]*", PROGRAM="/usr/bin/stratis_uuids_to_names.py $env{DM_NAME}", SYMLINK+="stratis/%c{1}/%c{2}"
```
mulkieran, line 1: How does using the ```+=``` operator result in the old symlink, on a rename of, say, a filesystem, being removed? It sure looks like it works, so what is the implicit rule that causes this to happen?

jbaublitz: My understanding here is that it is because of the type of synthetic event that stratisd is generating. I have not tested what would happen if we generated an `add` event, but the `change` event certainly seems to work. This made sense to me as the best option to try when I wrote it because a `change` event is responsible for updating an existing entry. As such, if udev has any awareness of which symlinks are associated with a rule (and it appears from the behavior that it may), `+=` would mean that we don't clobber existing symlinks generated from other rules (like `/dev/mapper` symlinks for example) and udev would be able to clean up the old symlink associated with this rule if it differs from the new one and then add the new one that we've specified. I'm not sure if this is exactly how it works, but it doesn't seem like an unreasonable guess based on the behavior and the limited digging I've done into udev rules.

mulkieran: > My understanding here is that it is because of the type of synthetic event that stratisd is generating. I have not tested what would happen if we generated an `add` event, but the `change` event certainly seems to work. This made sense to me as the best option to try when I wrote it because a `change` event is responsible for updating an existing entry. As such, if udev has any awareness of which symlinks are associated with a rule (and it appears from the behavior that it may), `+=` would mean that we don't clobber existing symlinks generated from other rules (like `/dev/mapper` symlinks for example) and udev would be able to clean up the old symlink associated with this rule if it differs from the new one and then add the new one that we've specified. I'm not sure if this is exactly how it works, but it doesn't seem like an unreasonable guess based on the behavior and the limited digging I've done into udev rules.

Exactly...it seems like we ought to use ```+=``` in order not to clobber the symlinks established by other rules. Note that there is another assignment operator, which implies final, that is, no other rule can assign a new value to the symlink. So that all makes sense.

But, is there some implicit udev semantics that says a single rule can only ever make one symlink? Apparently not, as there is a syntax for specifying multiple symlinks at one time. Maybe the semantics for a rule is that every time the rule gets run, all the effects of previously running the rule get erased?

My current hypothesis is that the udev database keeps track of what rule caused what thing to be set, and when the rule is triggered it undoes the previous effects of the rule. I have no evidence for this, beyond observation.

jbaublitz: I'll leave this as unresolved until we have a better understanding of this.

jbaublitz:
The most recent commit should fix CI.

jbaublitz:
Closed in favor of #2098

Issue #2035, opened 2020-05-18
Figure out why test_unique_instance is stalling so frequently these days


mulkieran:

mulkieran:
With unittest framework instead of pytest we see some interesting output previously suppressed:

```
tests.misc.test_startup.TestUniqueInstance.test_unique_instance
test_unique_instance (tests.misc.test_startup.TestUniqueInstance)
Verify that a second stratisd instance can not be started. ... ok
/usr/lib64/python3.7/subprocess.py:883: ResourceWarning: subprocess 260717 is still running
  ResourceWarning, source=self)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
/usr/lib64/python3.7/unittest/suite.py:84: ResourceWarning: unclosed file <_io.BufferedReader name=3>
  return self.run(*args, **kwds)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
/usr/lib64/python3.7/unittest/suite.py:84: ResourceWarning: unclosed file <_io.BufferedReader name=5>
  return self.run(*args, **kwds)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```

PR #2093, opened 2020-06-22
Avoid over-parsing the contents of the PID file

mulkieran:
Closes: #2035.


mulkieran:
Whoops. One last thing I want to try.

mulkieran:
I'll claim that this closes #2035, because it does getting rid of all the subprocess originating warnings that were previously visible. The trick was to wait for the process to terminate, mostly, although not sending stdout and stderr to a pipe also clearly helped.

PR #2095, opened 2020-06-24
Revise logging approach, drop dbus_interface tests, do some refactoring

### LINKED ISSUE #1055: Hang in event loop will prevent log dump via signal

### LINKED ISSUE #1262: Document log buffering and use of SIGUSR1

### LINKED ISSUE #1433: Log entries are expected to include timestamp, but do not

### LINKED ISSUE #2060: Check the return code from D-Bus calls in unittest.setUp methods

### LINKED ISSUE #2086: Get rid of the buffering in the logger

mulkieran:
This PR does the following:

* Ports a bunch of code previously in stratisd.rs into the library. This increases order, allows better linting, and will allow better testing via the Rust testing interface.
* Drops the use of the buffered logger, reverting to the standard env_logger interface. Reverts to the standard timestamp for the env_logger, which is UTC. Drops the --debug option from stratisd. It was overloaded oddly. If we ever change to a different logging impelemtation, it will make even less sense. This means that the logger will log at the default level, which is ERROR, unless the  the RUST_LOG environment variable is set. Buffered logs always logged at the trace level, unless RUST_LOG was set.
* Drops the dbus_interface tests. As the project matures, these are less and less useful.
* Adds a new option, --log-level which is a shorthand for setting the log level of the stratisd components.

Closes #2086 
Closes #1433 
Closes #2060
Closes #1262
Closes #1055
Closes https://github.com/stratis-storage/project/issues/105
Related #1977
Related #1319

Supercedes: #2085


mulkieran:
Changes identical to #2085, does not require new review.

PR #2098, opened 2020-06-25
Use udev rule instead of devlinks code

jbaublitz:
Supersedes #2091 

This is not quite ready yet. I want to edit `stratis_uuids_to_names.rs` and run CI before putting it up for review.

This PR takes the legacy testing code out of #2091


jbaublitz:
TODO: resolve [this revision request](https://github.com/stratis-storage/stratisd/pull/2091#discussion_r443702615).

mulkieran:
I've been experimenting a bit...when the pool comes up, if the links are gone, but the pool is still intact it does not re-create them. Ultimately, we should deal with this too, probably by sending change events as with a rename. This test is an odd case, though, because some external entity would have to remove the links.

More interesting is the question of whether the links are set up properly when the pool is reconstructed from devices. I would expect that they would be set up correctly, but I'll be interested in confirming this at some point.

udev/99-stratisd.rules
```diff
@@ -0,0 +1 @@
+ENV{DM_NAME}=="stratis-1-[0-9a-f]*-thin-fs-[0-9a-f]*", PROGRAM="/usr/bin/stratis_uuids_to_names.py $env{DM_NAME}", SYMLINK+="stratis/%c{1}/%c{2}"
```
mulkieran, line 1: Please update PROGRAM entry for new executable name.

jbaublitz:
@mulkieran I'll remove you from the requested review until I put the PR in review. Sorry about the confusion!

PR #2101, opened 2020-06-29
Use udev rule instead of devlinks code

### LINKED ISSUE #1987: devlinks module uses Pool, Filesystem traits, but shouldn't

jbaublitz:
Closes https://github.com/stratis-storage/project/issues/95.
Closes #1987.

Supersedes #2098


jbaublitz:
@mulkieran This is still not quite ready for review, but I've made significant progress.

1. I found a way to log to journald directly from the executable so error reporting should show up there now.
2. I squashed the bug we discussed in stand up using a `filter_map` approach.
3. The udev rule is now up to date with the new expected path.
4. I've determined that udev must track which symlinks are associated with which devices because I removed the udev rule before destroying the Stratis filesystem I was testing on and udev still knew to clean up the symlinks.

Overall from a testing perspective, this seems to be working but I'm going to put this on hold for now to work with libcryptsetup-rs.

jbaublitz:
TODO: Resolve [this comment](https://github.com/stratis-storage/stratisd/pull/2091#discussion_r443702615).

jbaublitz:
Rebased.

jbaublitz:
I found a way to remove the `systemd` crate dependency which is causing the build to fail. This would pull in `libsystemd` which I really don't think we need to require. Working on updates to use `/dev/log` directly instead.

jbaublitz:
Rebased.

jbaublitz:
Rebased.

jbaublitz:
Rebased.

src/engine/strat_engine/devlinks.rs
```diff
@@ -3,169 +3,97 @@
...
-            .map(|dir_e| dir_e.map(|d| d.file_name().into_string().expect("Unix is utf-8")))
-            .collect::<Result<HashSet<_>, _>>()?;
-
-        for (pool_name, _, _) in pools {
-            existing_dirs.remove(&pool_name.to_owned());
-        }
-
-        for leftover in existing_dirs {
-            pool_removed(&Name::new(leftover));
+pub fn pool_renamed(pool_name: &str, _: &str) {
```
mulkieran, line 28: At this point, the second parameter _can_ be dropped altogether.

src/engine/strat_engine/devlinks.rs
```diff
@@ -3,169 +3,97 @@
...
+            e, pool_name,
         );
     }
 }
 
-/// Create a symlink to the new filesystem's block device within its pool's
-/// directory.
-pub fn filesystem_added(pool_name: &str, fs_name: &str, devnode: &Path) {
-    let p = filesystem_mount_path(pool_name, fs_name);
+pub fn filesystem_renamed(pool_name: &str, fs_name: &str, _: &str) {
```
mulkieran, line 64: Must as well drop this unused parameter, also.

mulkieran, line 64: Please put in a header comment, and also for method above.

Cargo.toml
```diff
@@ -28,6 +33,10 @@ timerfd = "1.0.0"
 itertools = "0.8.0"
 libcryptsetup-rs = "0.4.0"
 
+[dependencies.regex]
```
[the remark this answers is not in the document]
jbaublitz: Resolved.

src/engine/strat_engine/devlinks.rs
jbaublitz: Resolved.

src/engine/strat_engine/devlinks.rs
jbaublitz: Resolved.

jbaublitz: Resolved.

Cargo.toml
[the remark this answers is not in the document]
mulkieran: > Just so you're aware, this won't change the behavior at all, but I agree it's more readable. Fixing.

Yeah. It's just programmer behavior that I'm trying to affect. Anything that makes it a clearer read.

jbaublitz:
Rebased.

jbaublitz:
@mulkieran Take a special look at the most recent commit as it's removing cruft that is no longer needed in the engine API with the udev script.

mulkieran:
> @mulkieran Take a special look at the most recent commit as it's removing cruft that is no longer needed in the engine API with the udev script.

Great!

mulkieran:
@jbaublitz Please rebase, thanks.

jbaublitz:
@mulkieran Rebased.

jbaublitz:
@mulkieran Ok, that should be resolved!

src/bin/stratis_uuids_to_names.rs
```diff
@@ -0,0 +1,201 @@
...
+        Some(dm_name) => dm_name,
+        None => {
+            return Err(StratisUdevError::new(
+                "Thinly provisioned filesystem devicemapper name required as argument.",
+            ));
+        }
+    };
+    match args.next().as_deref() {
+        Some(action) => {
+            if action == "remove" {
```
mulkieran, line 153: I get this from udevadm:

```
$ udevadm test --action=help
add
remove
change
move
online
offline
bind
unbind
```

I _think_ we'll probably want to exclude them all except for ```add``` and ```change```.

src/bin/stratis_uuids_to_names.rs
```diff
@@ -0,0 +1,201 @@
...
+    arg::{RefArg, Variant},
+    blocking::Connection,
+    Path,
+};
+use lazy_static::lazy_static;
+use regex::Regex;
+use uuid::Uuid;
+
+pub const DEV_LOG: &str = "/dev/log";
+pub const SYSTEM_DAEMON_INFO: &str = "<30>";
```
mulkieran, line 34: I couldn't figure out where these two numbers come from or why they're in brackets.

src/bin/stratis_uuids_to_names.rs
```diff
@@ -0,0 +1,201 @@
...
+    };
+
+    let managed_objects = get_managed_objects()?;
+
+    if let Some((pool_uuid, fs_uuid)) = udev_name_to_uuids(&dm_name)? {
+        let pool_name = uuid_to_stratis_name(&managed_objects, STRATIS_POOL_IFACE, pool_uuid)?
+            .ok_or_else(|| StratisUdevError::new("Could not get pool name from UUID."))?;
+        let fs_name = uuid_to_stratis_name(&managed_objects, STRATIS_FS_IFACE, fs_uuid)?
+            .ok_or_else(|| StratisUdevError::new("Could not get filesystem name from UUID."))?;
+        println!("{} {}", pool_name, fs_name);
```
mulkieran, line 169: Can you swap the println! into the ```main``` method? It seems a bit tidier.

src/bin/stratis_uuids_to_names.rs
```diff
@@ -0,0 +1,201 @@
...
+    }
+}
+
+fn main() -> Result<(), StratisUdevError> {
+    let sock = UnixDatagram::unbound().map_err(StratisUdevError::new)?;
+    match main_report_error() {
+        Ok(Some((pool_name, fs_name))) => {
+            sock.send_to(
+                format!(
+                    "{}Symlink /dev/stratis/{}/{} succesfully created.",
```
mulkieran, line 182: Can you drop the "succesfully"? "created" should be enough.

jbaublitz:
@mulkieran Should be ready for another pass.

src/bin/stratis_uuids_to_names.rs
```diff
@@ -0,0 +1,206 @@
...
+
+fn uuid_to_stratis_name(
+    managed_objects: &GMORet,
+    iface_name: &'static str,
+    uuid: Uuid,
+) -> Result<Option<String>, StratisUdevError> {
+    let mut names: Vec<_> = managed_objects
+        .values()
+        .filter_map(|map| {
+            if map.contains_key(iface_name)
```
mulkieran, line 118: Should be able to drop ```contains_key``` here, since using ```get```.

src/bin/stratis_uuids_to_names.rs
```diff
@@ -0,0 +1,206 @@
...
+        .name("fs")
+        .map(|pu| Uuid::parse_str(pu.as_str()).expect("Format validated by regex"));
+    Ok(pool_uuid.and_then(|pu| fs_uuid.map(|fu| (pu, fu))))
+}
+
+fn uuid_to_stratis_name(
+    managed_objects: &GMORet,
+    iface_name: &'static str,
+    uuid: Uuid,
+) -> Result<Option<String>, StratisUdevError> {
```
mulkieran, line 114: I think ```Option``` part of return type is unnecessary; either there is an error or the name is found.

src/bin/stratis_uuids_to_names.rs
```diff
@@ -31,7 +31,11 @@ use regex::Regex;
 use uuid::Uuid;
 
 pub const DEV_LOG: &str = "/dev/log";
+/// Syslog priority syntax
+/// (3 (SYSTEM) << 3) | 6 (INFO)
```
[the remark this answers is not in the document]
jbaublitz: That is part of the syntax. It is required.

udev/99-stratisd.rules
```diff
@@ -0,0 +1 @@
+ENV{DM_NAME}=="stratis-1-[0-9a-f]*-thin-fs-[0-9a-f]*", PROGRAM="/usr/lib/udev/stratis_uuids_to_names $env{DM_NAME} $env{ACTION}", SYMLINK+="stratis/%c{1}/%c{2}"
```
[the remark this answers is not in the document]
jbaublitz: What is the concern here? I'm unclear on what potential problems you see this causing. And what kind of indirection would you see being helpful? A symlink? Use of a variable provided by udev internally? I think that because we are providing that executable just like we are providing the udev rule, there is no ambiguity in the absolute path that we're providing so if the package is installed, one should be able to access the other without a problem.

src/bin/stratis_uuids_to_names.rs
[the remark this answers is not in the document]
jbaublitz: (Just to clarify, if the brackets are not provided, the number will be added to the log message instead of being interpreted as a priority)

udev/99-stratisd.rules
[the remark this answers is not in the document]
bgurney-rh: The macro `%{_libdir}` evaluates to `/usr/lib64` on an x86-64 system, yet the correct path to use for this file is `/usr/lib/udev`.  This may be only specific to the spec file, for packaging, but it does call into question why there's a `%{_udevrulesdir}` macro (which resolves to `/usr/lib/udev/rules.d`), and not one for `/usr/lib/udev`.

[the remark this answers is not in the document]
mulkieran: @bgurney-rh has hit on the area of my concern; it's unforeseen interactions between Fedora/RHEL packaging and our hard-coding. Right now there are none; because the spec file also hard codes the path: https://github.com/stratis-storage/ci/pull/106/commits/b82400732e14201f10bf28b30f0cb32b9086a123. So we can proceed.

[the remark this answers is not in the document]
jbaublitz: I'm going to test removing the absolute path and using a relative one instead, and I'll report back.

src/bin/stratis_uuids_to_names.rs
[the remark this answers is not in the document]
mulkieran: ok, thx!

src/engine/strat_engine/engine.rs
```diff
@@ -438,7 +430,6 @@ mod test {
         assert_eq!(engine.liminal_devices, LiminalDevices::default());
 
         engine.teardown().unwrap();
-        remove_dir_all(DEV_PATH).unwrap();
```
mulkieran, line 441: Could you edit the test comment above as well? Where it says remove "/stratis".

udev/99-stratisd.rules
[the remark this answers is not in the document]
jbaublitz: This works.

jbaublitz:
Rebased and PR revisions made. @mulkieran I changed on of the commits slightly to keep the accesses of `submap` all at the same closure level just to make it a little bit more readable but other than that. If that works for you, this should be ready to go.

Issue #1719, opened 2019-11-26
If a pool has failed, stratisd should refuse to perform standard maintenance operations


mulkieran:
Instead it should return an error indicating that the pool has failed and that some repair actions should be made before these routine administrative actions should be attempted.

Certain repair-type commands should be allowed, regardless.

PR #2107, opened 2020-07-01
Improve cacheing, use and logging of changes to thinpool state

mulkieran:
This PR corrects a number of problems.

These problems arose from a decision that was made to expose a PoolState field used and managed by the Thinpool implementation on the D-Bus interface. This was a poor design choice, because it did not allow proper abstraction and separation of concerns. This property was removed in 1.0.0 for those reasons and others.

This removal gives the internal implementation freedom to alter the internal structures of the Thinpool. In particular, it seems reasonable, rather than storing a value that could be used as a D-Bus property to simply store the result of the thinpool status() call since it is a complete representation of the data obtained from that call.

If this decision is made, then the result of ThinPool::total_physical_used can be calculated from the cached result of the thinpool status() call. Previously this was not possible, because the necessary information was discarded from the PoolState enum. This has the advantage that the value is consistent with the rest of the information in the thinpool and that it is not necessary to use an ioctl to get the value, it is now pure computation. Since the main use of this function is to calculate a value to export on the D-Bus, this is a useful gain in efficiency and stability.

A further improvement is the enhancement of the Thinpool::set_state() method to provide useful and informative logging. Previously, if the pool was in a bad state, various error! and warn! messages were emitted. This caused several problems. First, the journal could get inundated with repetitious messages. Second if the thinpool was restored to a good state, this event was not logged; the log reader could only infer from the cessation of the repeated messages that some improvements must have occurred. Third, several of the log entries were at the error level, which was incorrect, since stratisd would still be able to continue operation under these conditions. Fourth, the log entries did not clearly identify the device involved. Now, a single log entry is emitted when a _change_ of state is detected and it contains all the identifying information available. These log entries can be scrutinized in the existing stratisd test output, and are clearly more informative.

This PR does not address the problems with the test_add_datadevs test in pool.rs, but transforms it to be semantically equivalent to its previous version. I plan to do more work to improve the test and on the PoolExtendState type and associated field in a later PR.

Closes: #1255
Related: https://github.com/stratis-storage/stratisd/issues/1719

### DIFF
--- a/src/engine/sim_engine/pool.rs
+++ b/src/engine/sim_engine/pool.rs
@@ -24,7 +24,7 @@ use crate::{
         structures::Table,
         types::{
             BlockDevTier, CreateAction, DevUuid, FilesystemUuid, KeyDescription, MaybeDbusPath,
-            Name, PoolExtendState, PoolState, PoolUuid, Redundancy, RenameAction, SetCreateAction,
+            Name, PoolExtendState, PoolUuid, Redundancy, RenameAction, SetCreateAction,
             SetDeleteAction,
         },
     },
@@ -44,7 +44,6 @@ pub struct SimPool {
     filesystems: Table<SimFilesystem>,
     redundancy: Redundancy,
     rdm: Rc<RefCell<Randomizer>>,
-    pool_state: PoolState,
     pool_extend_state: PoolExtendState,
     dbus_path: MaybeDbusPath,
 }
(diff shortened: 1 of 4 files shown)


mulkieran:
This is running forever on real_test_add_datadevs, which I think is somewhat significant.

mulkieran:
```test_add_datadevs``` is flawed, we should mount the filesystem, rather than writing directly to the devnode.

mulkieran:
And now the test has complete succesfully, because the loop condition also checks the pool state. What this tells us is that the pool extend state never ceases to be Good, which seems odd.

mulkieran:
Travis test has passed, putting up for review.

PR #2109, opened 2020-07-02
Remove tree target from Makefile

bgurney-rh:
Closes: stratis-storage/project#188


mulkieran:
I moved this to June, we can clear it w/in the hour and the attached issue as well.

Issue #1256, opened 2018-10-04
If it was necessary to extend a thinpool device but no attempt was made what should the extend status be?


mulkieran:
This would happen when there was insufficient space left on the backing device.

Currently I think it is Good, but that seems a bit too positive to fit the circumstances.

Maybe extend state needs more values, or maybe there should be another state, or something.

mulkieran:
Related: #1257.

PR #2113, opened 2020-07-07
Remove pool_extend_state field from both engines

mulkieran:
This removes another field from the ThinPool implementation that was originally intended to be exported over the D-Bus but is no longer used. In this way it is similar to #2103.

Closes: #1256.


mulkieran:
This is a bit long-winded for a simple removal, but each commit justifies itself, so I think it makes sense to leave the PR as is.

mulkieran:
rebased.

mulkieran:
Jenkins failure looked spurious, certainly unrelated, re-running.

PR #2115, opened 2020-07-07
Get rid of an obsoleted lint

mulkieran:


mulkieran:
@jbaublitz: This one owes its existence to the discussion we had about narrowing conversions, as you probably guessed.

Issue #2097, opened 2020-06-25
Write a Rust script to read the static header and the variable length metadata and display it


mulkieran:
This would simply use methods in the existing metadata module. The script takes one argument, which is the device to look at. It then invokes ```BDA::load``` and ```BDA::load_state()``` in sequence. It then somehow prints out the result in a human readable format.

The results are the value of a valid signature block, information about the MDA regions, and the most recently written variable length metadata if any.

The variable length metadata is already JSON, anyway, if valid, so it should be easy to format and print out.

The sigblock is represented by the StaticHeader struct, and should be able to be converted to, e.g., JSON pretty easily, as is the MDA header information.

This script can be placed in the src/bin directory.

mulkieran:
We should discuss this in standup before starting.

PR #2116, opened 2020-07-08
Add stratis_dumpmetadata program

bgurney-rh:
Related: #2097

### DIFF
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -11,7 +11,10 @@ use serde_json::Value;
 use libstratis::engine::BDA;
 
 fn run(devpath: String) -> Result<(), String> {
-    let mut devfile = OpenOptions::new().read(true).open(&devpath).unwrap();
+    let mut devfile = OpenOptions::new()
+        .read(true)
+        .open(&devpath)
+        .map_err(|_the_io_error| "Error opening device")?;
 
     let bda = BDA::load(&mut devfile).unwrap().unwrap();
     println!("{:#?}", bda);
@@ -36,8 +39,8 @@ fn main() {
 
     match run(devpath) {
         Ok(()) => {}
-        Err(_e) => {
-            eprintln!("Error encountered");
+        Err(e) => {
+            eprintln!("Error encountered: {}", e);
             process::exit(2);
         }
     }
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -27,11 +27,13 @@ fn run(devpath: &str) -> Result<(), String> {
         .load_state(&mut devfile)
         .map_err(|stateload_err| format!("Error during load state: {}", stateload_err))?;
     println!("State JSON data:");
+
     let state_json: Value = serde_json::from_slice(&loaded_state.unwrap())
         .map_err(|extract_err| format!("Error during state JSON extract: {}", extract_err))?;
     let state_json_pretty: String = serde_json::to_string_pretty(&state_json)
         .map_err(|parse_err| format!("Error during state JSON parse: {}", parse_err))?;
     println!("{}", state_json_pretty);
+
     Ok(())
 }
 
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -14,7 +14,7 @@ fn run(devpath: String) -> Result<(), String> {
     let mut devfile = OpenOptions::new()
         .read(true)
         .open(&devpath)
-        .map_err(|_the_io_error| "Error opening device")?;
+        .map_err(|_the_io_error| format!("Error opening device: {}", _the_io_error))?;
 
     let bda = BDA::load(&mut devfile).unwrap().unwrap();
     println!("{:#?}", bda);
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -10,7 +10,7 @@ use serde_json::Value;
 
 use libstratis::engine::BDA;
 
-fn run(devpath: String) -> Result<(), String> {
+fn run(devpath: &str) -> Result<(), String> {
     let mut devfile = OpenOptions::new()
         .read(true)
         .open(&devpath)
@@ -41,7 +41,7 @@ fn main() {
         eprintln!("Usage: stratis_dumpmetadata <device>");
         process::exit(1);
     }
-    let devpath = args[1].clone();
+    let devpath = &args[1];
 
     match run(devpath) {
         Ok(()) => {}
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -36,6 +36,9 @@ fn main() {
 
     match run(devpath) {
         Ok(()) => {}
-        Err(_e) => println!("Error encountered"),
+        Err(_e) => {
+            eprintln!("Error encountered");
+            process::exit(2);
+        }
     }
 }
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -10,14 +10,7 @@ use serde_json::Value;
 
 use libstratis::engine::BDA;
 
-fn main() {
-    let args: Vec<String> = env::args().collect();
-    if args.len() < 2 {
-        println!("Usage: stratis_dumpmetadata <device>");
-        process::exit(1);
-    }
-    let devpath = args[1].clone();
-
+fn run(devpath: String) {
     let mut devfile = OpenOptions::new().read(true).open(&devpath).unwrap();
 
     let bda = BDA::load(&mut devfile).unwrap().unwrap();
@@ -30,4 +23,15 @@ fn main() {
     let state_json: Value = serde_json::from_slice(&loaded_state.unwrap()).unwrap();
     let state_json_pretty: String = serde_json::to_string_pretty(&state_json).unwrap();
     println!("{}", state_json_pretty);
+}
+
+fn main() {
+    let args: Vec<String> = env::args().collect();
+    if args.len() < 2 {
+        println!("Usage: stratis_dumpmetadata <device>");
+        process::exit(1);
+    }
+    let devpath = args[1].clone();
+
+    run(devpath);
 }
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -21,7 +21,13 @@ fn run(devpath: &str) -> Result<(), String> {
         .ok_or_else(|| "No Stratis BDA metadata found".to_string())?;
     println!("{:#?}", bda);
 
-    devfile.seek(SeekFrom::Start(0)).unwrap();
+    match devfile.seek(SeekFrom::Start(0)) {
+        Ok(_result) => {}
+        Err(e) => {
+            eprintln!("Error during seek: {}", e);
+            process::exit(1);
+        }
+    }
 
     let loaded_state = bda
         .load_state(&mut devfile)
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -14,7 +14,7 @@ fn run(devpath: String) -> Result<(), String> {
     let mut devfile = OpenOptions::new()
         .read(true)
         .open(&devpath)
-        .map_err(|_the_io_error| format!("Error opening device: {}", _the_io_error))?;
+        .map_err(|the_io_error| format!("Error opening device: {}", the_io_error))?;
 
     let bda = BDA::load(&mut devfile)
         .map_err(|bda_load_error| format!("BDA detected but error found: {}", bda_load_error))?
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -1,0 +1,29 @@
+use std::env;
+
+use std::fs::OpenOptions;
+use std::io::{Seek, SeekFrom};
+use std::vec::Vec;
+
+use serde_json::Value;
+
+use libstratis::engine::strat_engine::backstore::metadata::BDA;
+
+fn main() {
+    let args: Vec<String> = env::args().collect();
+    let devpath = args[1].clone();
+    println!("Device path: {}", devpath);
+
+    let mut devfile = OpenOptions::new().read(true).open(&devpath).unwrap();
+
+    let bda = BDA::load(&mut devfile).unwrap().unwrap();
+    println!("{:#?}", bda);
+
+    devfile.seek(SeekFrom::Start(0)).unwrap();
+
+    //let mut buf: Vec<u8> = Vec::new();
+    let loaded_state = bda.load_state(&mut devfile).unwrap();
+    println!("State JSON data:");
+    let state_json: Value = serde_json::from_slice(&loaded_state.unwrap()).unwrap();
+    let state_json_pretty: String = serde_json::to_string_pretty(&state_json).unwrap();
+    println!("{}", state_json_pretty);
+}
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -20,7 +20,6 @@ fn main() {
 
     devfile.seek(SeekFrom::Start(0)).unwrap();
 
-    //let mut buf: Vec<u8> = Vec::new();
     let loaded_state = bda.load_state(&mut devfile).unwrap();
     println!("State JSON data:");
     let state_json: Value = serde_json::from_slice(&loaded_state.unwrap()).unwrap();
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -18,7 +18,7 @@ fn run(devpath: String) -> Result<(), String> {
 
     let bda = BDA::load(&mut devfile)
         .unwrap()
-        .ok_or_else(|| format!("No Stratis BDA metadata found"))?;
+        .ok_or_else(|| "No Stratis BDA metadata found".to_string())?;
     println!("{:#?}", bda);
 
     devfile.seek(SeekFrom::Start(0)).unwrap();
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -7,7 +7,7 @@ use std::{
 
 use serde_json::Value;
 
-use libstratis::engine::strat_engine::backstore::metadata::BDA;
+use libstratis::engine::BDA;
 
 fn main() {
     let args: Vec<String> = env::args().collect();
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -1,8 +1,9 @@
-use std::env;
-
-use std::fs::OpenOptions;
-use std::io::{Seek, SeekFrom};
-use std::vec::Vec;
+use std::{
+    env,
+    fs::OpenOptions,
+    io::{Seek, SeekFrom},
+    vec::Vec,
+};
 
 use serde_json::Value;
 
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -41,7 +41,7 @@ fn main() {
     let args: Vec<String> = env::args().collect();
     if args.len() < 2 {
         eprintln!("Usage: stratis_dumpmetadata <device>");
-        process::exit(1);
+        process::exit(2);
     }
     let devpath = &args[1];
 
@@ -49,7 +49,7 @@ fn main() {
         Ok(()) => {}
         Err(e) => {
             eprintln!("Error encountered: {}", e);
-            process::exit(2);
+            process::exit(1);
         }
     }
 }
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -38,7 +38,7 @@ fn run(devpath: String) -> Result<(), String> {
 fn main() {
     let args: Vec<String> = env::args().collect();
     if args.len() < 2 {
-        println!("Usage: stratis_dumpmetadata <device>");
+        eprintln!("Usage: stratis_dumpmetadata <device>");
         process::exit(1);
     }
     let devpath = args[1].clone();
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -10,7 +10,7 @@ use serde_json::Value;
 
 use libstratis::engine::BDA;
 
-fn run(devpath: String) {
+fn run(devpath: String) -> Result<(), String> {
     let mut devfile = OpenOptions::new().read(true).open(&devpath).unwrap();
 
     let bda = BDA::load(&mut devfile).unwrap().unwrap();
@@ -23,6 +23,7 @@ fn run(devpath: String) {
     let state_json: Value = serde_json::from_slice(&loaded_state.unwrap()).unwrap();
     let state_json_pretty: String = serde_json::to_string_pretty(&state_json).unwrap();
     println!("{}", state_json_pretty);
+    Ok(())
 }
 
 fn main() {
@@ -33,5 +34,8 @@ fn main() {
     }
     let devpath = args[1].clone();
 
-    run(devpath);
+    match run(devpath) {
+        Ok(()) => {}
+        Err(_e) => println!("Error encountered"),
+    }
 }
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -21,13 +21,9 @@ fn run(devpath: &str) -> Result<(), String> {
         .ok_or_else(|| "No Stratis BDA metadata found".to_string())?;
     println!("{:#?}", bda);
 
-    match devfile.seek(SeekFrom::Start(0)) {
-        Ok(_result) => {}
-        Err(e) => {
-            eprintln!("Error during seek: {}", e);
-            process::exit(1);
-        }
-    }
+    devfile
+        .seek(SeekFrom::Start(0))
+        .map_err(|seek_err| format!("Error during seek: {}", seek_err))?;
 
     let loaded_state = bda
         .load_state(&mut devfile)
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -11,7 +11,6 @@ use libstratis::engine::strat_engine::backstore::metadata::BDA;
 fn main() {
     let args: Vec<String> = env::args().collect();
     let devpath = args[1].clone();
-    println!("Device path: {}", devpath);
 
     let mut devfile = OpenOptions::new().read(true).open(&devpath).unwrap();
 
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -2,6 +2,7 @@ use std::{
     env,
     fs::OpenOptions,
     io::{Seek, SeekFrom},
+    process,
     vec::Vec,
 };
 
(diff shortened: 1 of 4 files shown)


src/bin/stratis_dumpmetadata.rs
```diff
@@ -0,0 +1,29 @@
...
+    println!("Device path: {}", devpath);
+
+    let mut devfile = OpenOptions::new().read(true).open(&devpath).unwrap();
+
+    let bda = BDA::load(&mut devfile).unwrap().unwrap();
+    println!("{:#?}", bda);
+
+    devfile.seek(SeekFrom::Start(0)).unwrap();
+
+    //let mut buf: Vec<u8> = Vec::new();
```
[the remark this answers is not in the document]
bgurney-rh: Whoops, thought I removed all of these.  Resolved.

src/bin/stratis_dumpmetadata.rs
```diff
@@ -0,0 +1,29 @@
...
+use std::vec::Vec;
+
+use serde_json::Value;
+
+use libstratis::engine::strat_engine::backstore::metadata::BDA;
+
+fn main() {
+    let args: Vec<String> = env::args().collect();
+    let devpath = args[1].clone();
+    println!("Device path: {}", devpath);
```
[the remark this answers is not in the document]
bgurney-rh: Resolved.

src/bin/stratis_dumpmetadata.rs
```diff
@@ -0,0 +1,29 @@
+use std::env;
+
```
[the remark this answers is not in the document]
bgurney-rh: Resolved.

bgurney-rh:
Ok; my primary goal was to post this in a PR, now that it's starting to become fit for purpose.

I do want to add error handling right after the `BDA::load` call, once I figure out how.

mulkieran:
@bgurney-rh: Please post what, e.g, a failure to read the BDA, because there's no BDA on the device, looks like when you run the code now. Then we can talk about whether we want to be a little more helpful to users in standup.

bgurney-rh:
Here is the outcome of a "failure" (reading a device that doesn't have Stratis metadata):
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/bin/stratis_dumpmetadata.rs:18:15
```

bgurney-rh:
Here is a successful run:
```
BDA {
    header: StaticHeader {
        blkdev_size: BlockdevSize(
            Sectors(16775168),
        ),
        identifiers: StratisIdentifiers {
            pool_uuid: 9f39d1a4-a909-46b7-8c25-6bd1e3929be5,
            device_uuid: 92d035be-7d9c-47ee-b2a1-98c5b0f1b8ad,
        },
        mda_size: MDASize(
            Sectors(2032),
        ),
        reserved_size: ReservedSize(
            Sectors(6144),
        ),
        flags: 0,
        initialization_time: 1594056204,
    },
    regions: MDARegions {
        region_size: MDARegionSize(
            Sectors(508),
        ),
        mda_headers: [
            Some(
                MDAHeader {
                    last_updated: 2020-07-06T17:23:24.523694355Z,
                    used: MetaDataSize(
                        Bytes(443),
                    ),
                    data_crc: 1341058357,
                },
            ),
            None,
        ],
    },
}
State JSON data:
{
  "backstore": {
    "cap": {
      "allocs": [
        [
          0,
          16766976
        ]
      ]
    },
    "data_tier": {
      "blockdev": {
        "allocs": [
          [
            {
              "length": 16766976,
              "parent": "92d035be-7d9c-47ee-b2a1-98c5b0f1b8ad",
              "start": 8192
            }
          ]
        ],
        "devs": [
          {
            "uuid": "92d035be-7d9c-47ee-b2a1-98c5b0f1b8ad"
          }
        ]
      }
    }
  },
  "flex_devs": {
    "meta_dev": [
      [
        1638400,
        32768
      ]
    ],
    "thin_data_dev": [
      [
        65536,
        1572864
      ],
      [
        1671168,
        15095808
      ]
    ],
    "thin_meta_dev": [
      [
        0,
        32768
      ]
    ],
    "thin_meta_dev_spare": [
      [
        32768,
        32768
      ]
    ]
  },
  "name": "spool1",
  "thinpool_dev": {
    "data_block_size": 2048
  }
}
```

mulkieran:
@bgurney-rh: thx. A successful run looks pretty useful and an unsuccessful run is pretty unfriendly.

mulkieran:
There are a few versions, which are verified. If they were wrong, then an error would be returned. It might be helpful to report the expected version, but not really vital. We should skip it for now, because it would probably be sensible to use a command-line options for that, and so for that we would need to use a library, presumably clap like stratisd.rs uses.

bgurney-rh:
I just updated the imports.  If I were to squash this PR, it would reduce it to only modifying one line in `src/engine/mod.rs`, and adding one line to `src/engine/strat_engine/mod.rs`.

mulkieran:
> I just updated the imports. If I were to squash this PR, it would reduce it to only modifying one line in `src/engine/mod.rs`, and adding one line to `src/engine/strat_engine/mod.rs`.

Excellent.

src/bin/stratis_dumpmetadata.rs
```diff
@@ -0,0 +1,28 @@
...
+    vec::Vec,
+};
+
+use serde_json::Value;
+
+use libstratis::engine::BDA;
+
+fn main() {
+    let args: Vec<String> = env::args().collect();
+    let devpath = args[1].clone();
```
mulkieran, line 14: @bgurney-rh: Other than where we see ```unwrap()``` this use of the index operator is the only place where the program can just fail. Try invoking and passing no arguments, and you'll see an immediate failure.

What that means, of course, is that about 1/2 the lines in this ```main()``` method can be reasonably expected to cause a failure. I'll be giving you a plan for how to address this.

bgurney-rh:
I pushed a simple check for zero parameters.

Trying to deconstruct what I need to do for the "run()" method, I looked at this line, which I know looks awkward with the two "unwrap()" calls:
```
let bda = BDA::load(&mut devfile).unwrap().unwrap();
```

If I take one of them away, the build fails with this:

```
error[E0599]: no method named `load_state` found for enum `std::option::Option<libstratis::engine::strat_engine::backstore::metadata::bda::BDA>` in the current scope
  --> src/bin/stratis_dumpmetadata.rs:28:28
   |
28 |     let loaded_state = bda.load_state(&mut devfile).unwrap();
   |                            ^^^^^^^^^^ method not found in `std::option::Option<libstratis::engine::strat_engine::backstore::metadata::bda::BDA>`
```

In other words, bda.load_state() (after the bda.load() call on line 23) will either return something ("Some(x)"), or nothing ("None").  And obviously, if it returns nothing, I want the program to exit with a println!() call stating that nothing was found (but in a way that can catch the failure before that call, if possible).

mulkieran:
> I pushed a simple check for zero parameters.

That looks fine, and should hold up until we need to be more sophisticated w/ our argument handling.

mulkieran:
> Trying to deconstruct what I need to do for the "run()" method, I looked at this line, which I know looks awkward with the two "unwrap()" calls:
> 
> ```
> let bda = BDA::load(&mut devfile).unwrap().unwrap();
> ```
> 
> If I take one of them away, the build fails with this:
> 
> ```
> error[E0599]: no method named `load_state` found for enum `std::option::Option<libstratis::engine::strat_engine::backstore::metadata::bda::BDA>` in the current scope
>   --> src/bin/stratis_dumpmetadata.rs:28:28
>    |
> 28 |     let loaded_state = bda.load_state(&mut devfile).unwrap();
>    |                            ^^^^^^^^^^ method not found in `std::option::Option<libstratis::engine::strat_engine::backstore::metadata::bda::BDA>`
> ```
> 
> In other words, bda.load_state() (after the bda.load() call on line 23) will either return something ("Some(x)"), or nothing ("None"). And obviously, if it returns nothing, I want the program to exit with a println!() call stating that nothing was found (but in a way that can catch the failure before that call, if possible).

Right now, you're overthinking this. You can separate out the contents of the ```run()``` method _without_ modifying the code within. And this will put you in a good position for the next step. So, the challenge here is purely to set up the ```run()``` method so that it compiles and the program does pretty much the same as before. The return type can still be ```()``` like with ```main()```. (Note that the compiler lets you leave out the return type if it's ```()``` and clippy _insists_ that you leave it out.) So the only challenge in this step is getting the parameter(s) to your ```run()``` method correct and passing the arguments properly. Then run it and see that it crashes just as before on a missing BDA (just complains about a different line).

bgurney-rh:
I tried this, and it seems to work the same way:
```
use std::{
    env,
    fs::OpenOptions,
    io::{Seek, SeekFrom},
    process,
    vec::Vec,
};

use serde_json::Value;

use libstratis::engine::BDA;

fn run(devpath: String) {
    let mut devfile = OpenOptions::new().read(true).open(&devpath).unwrap();

    let bda = BDA::load(&mut devfile).unwrap().unwrap();
    println!("{:#?}", bda);

    devfile.seek(SeekFrom::Start(0)).unwrap();

    let loaded_state = bda.load_state(&mut devfile).unwrap();
    println!("State JSON data:");
    let state_json: Value = serde_json::from_slice(&loaded_state.unwrap()).unwrap();
    let state_json_pretty: String = serde_json::to_string_pretty(&state_json).unwrap();
    println!("{}", state_json_pretty);
}

fn main() {
    let args: Vec<String> = env::args().collect();
    if args.len() < 2 {
        println!("Usage: stratis_dumpmetadata <device>");
        process::exit(1);
    }
    let devpath = args[1].clone();

    run(devpath);
}
```

Failure from a device that has no BDA:
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/bin/stratis_dumpmetadata.rs:16:15
```

mulkieran:
Great! Go ahead and push that change, and we can start to modify it.

mulkieran:
By the way, ```()``` is, more or less, Rust's equivalent of Python's ```None```. But whereas the type of ```None``` is ```NoneType``` in Python the type of ```()``` is also ```()``` in Rust, the same symbol for two different, but closely related, things. Just as ```None``` is the only value of type ```NoneType```, ```()``` is the only value of ```()```. (This will be a helpful fact to know, soon.)

src/bin/stratis_dumpmetadata.rs
```diff
@@ -0,0 +1,37 @@
...
+    io::{Seek, SeekFrom},
+    process,
+    vec::Vec,
+};
+
+use serde_json::Value;
+
+use libstratis::engine::BDA;
+
+fn run(devpath: String) {
```
mulkieran, line 13: Now you'll want to change the return type of ```run()``` so it has a ```Result``` type. My suggestion is this for the signature:

```
fn run(devpath: String) -> Result<(), String> {
  ...
}
```

Once you do that, you'll get quite a few build errors, not surprisingly. There will be at least one complaint about the ```run()``` method. No surprise, there. You've changed the return type, but you haven't changed the return value to match the type. To fix it up so it builds (doesn't yet do the correct thing) you'll need to add a line, shown below.

src/bin/stratis_dumpmetadata.rs
```diff
@@ -0,0 +1,37 @@
...
+    let bda = BDA::load(&mut devfile).unwrap().unwrap();
+    println!("{:#?}", bda);
+
+    devfile.seek(SeekFrom::Start(0)).unwrap();
+
+    let loaded_state = bda.load_state(&mut devfile).unwrap();
+    println!("State JSON data:");
+    let state_json: Value = serde_json::from_slice(&loaded_state.unwrap()).unwrap();
+    let state_json_pretty: String = serde_json::to_string_pretty(&state_json).unwrap();
+    println!("{}", state_json_pretty);
```
mulkieran, line 25: Make the very last line ```Ok(())``` and this method should compile.

src/bin/stratis_dumpmetadata.rs
```diff
@@ -0,0 +1,37 @@
...
+
+fn main() {
+    let args: Vec<String> = env::args().collect();
+    if args.len() < 2 {
+        println!("Usage: stratis_dumpmetadata <device>");
+        process::exit(1);
+    }
+    let devpath = args[1].clone();
+
+    run(devpath);
```
mulkieran, line 36: This line here will have a compiler error associated with it. Because ```run(devpath)``` now evaluates to a ```Result``` you'll need to match the ```Result``` and do the right thing depending on whether it is ```Err``` or ```Ok```.

src/bin/stratis_dumpmetadata.rs
```diff
@@ -33,5 +34,8 @@ fn main() {
     }
     let devpath = args[1].clone();
 
-    run(devpath);
+    match run(devpath) {
+        Ok(()) => {}
+        Err(_e) => println!("Error encountered"),
```
mulkieran, line 39: You might as well use ```eprintln!``` here so the text goes out on stderr.

mulkieran: Right here is where you get to decide what the exit code should be be as well. In our simple model, we can just have the usual two exit codes, and refine it as we might need later.

I recall you were using subprocess::exit in an example; that's the way to go.

bgurney-rh: I think you mean "process::exit"; "subprocess" might be from Python.

bgurney-rh:
I feel that, at some point, I'll have to use `Option<T>` somewhere.

src/bin/stratis_dumpmetadata.rs
```diff
@@ -0,0 +1,44 @@
...
+    process,
+    vec::Vec,
+};
+
+use serde_json::Value;
+
+use libstratis::engine::BDA;
+
+fn run(devpath: String) -> Result<(), String> {
+    let mut devfile = OpenOptions::new().read(true).open(&devpath).unwrap();
```
mulkieran, line 14: Step (1)

Ok! Here's the first ```unwrap()``` in the function. We might as well address this one, because we'll quickly see results. We know, for example, that this will fail if the file just does not exist. Easy to test.

Recall that ```unwrap()``` can be called on either an ```Option``` or a ```Result``` type. We need to figure out which this one is. I know it's a ```Result```, so we'll take that as given.

Now, we decide on the message we want, that's up to you.

Our goal is to convert that ```unwrap()``` call to something that will ultimately have the effect we want: print a useful error message to stderr and then exit.

We can do that by converting that ```unwrap()``` call to something that will cause an immediate return from the ```run``` function with a useful error message if there was an error when opening the file.

We want to use the function ```map_err()```, see here: **https://doc.rust-lang.org/std/result/enum.Result.html#method.map_err.__**.

And we want to return immediately from the function if there is an error, so we use the ```?``` operator as well.

So, the call will look something like this:

```
.map_err(|the_io__err| <your chosen error string>)?;
```

instead of 

```
unwrap();
```

mulkieran: Note

```
|the_io_error| <your chosen error string>
```

is just an anonymous function that you write that takes as an argument the io error from the open operation and returns a string. You see them here and there in Python; the keyword that identifies them in Python is ```lambda```. We have a few in the CLI.

src/bin/stratis_dumpmetadata.rs
```diff
@@ -0,0 +1,44 @@
...
+    let args: Vec<String> = env::args().collect();
+    if args.len() < 2 {
+        println!("Usage: stratis_dumpmetadata <device>");
+        process::exit(1);
+    }
+    let devpath = args[1].clone();
+
+    match run(devpath) {
+        Ok(()) => {}
+        Err(_e) => {
```
mulkieran, line 39: Step (2)

With the change above, this branch will be executed if, for example, the specified file is not there.

But it will just print out "Error encountered" which is not as specific as we want.

To print out the actual error string that you decided on in Step (1) will require a little more work.

src/bin/stratis_dumpmetadata.rs
bgurney-rh: I'm getting a mess of errors when I try this.

First try:
```
error: expected type, found `"Device not found"`
  --> src/bin/stratis_dumpmetadata.rs:14:92
   |
14 |     let mut devfile = OpenOptions::new().read(true).open(&devpath).map_err(|the_io_error| <"Device not found">)?;
   |                                                                                            ^^^^^^^^^^^^^^^^^^ expected type
```

What does this mean?  If I remove the unwrap, I see this on the subsequent call of BDA::load:

```
error[E0277]: the trait bound `std::result::Result<std::fs::File, std::io::Error>: std::io::Read` is not satisfied
  --> src/bin/stratis_dumpmetadata.rs:16:25
   |
16 |     let bda = BDA::load(&mut devfile).unwrap().unwrap();
   |                         ^^^^^^^^^^^^ the trait `std::io::Read` is not implemented for `std::result::Result<std::fs::File, std::io::Error>`
```

Trying to enter that type doesn't work:
```
error: comparison operators cannot be chained
  --> src/bin/stratis_dumpmetadata.rs:14:104
   |
14 |     let mut devfile = OpenOptions::new().read(true).open(&devpath).map_err(|the_io_err| std::io::Error <"Device not found">)?;
   |                                                                                                        ^^^^^^^^^^^^^^^^^^^^
   |
   = help: use `::<...>` instead of `<...>` to specify type arguments
   = help: or use `(...)` if you meant to specify fn arguments
```

...at this point, I obviously did something so wrong, that rustc is giving me a non sequitur message, and I should just bail out here.

mulkieran:
Whoops! From this ```<"Device not found">``` remove the angle brackets. A literal string "Device not found" is syntactically correct. Allthough, it's more like "Error opening device", 'cause it could be permissions that caused the problem, or anything really.

bgurney-rh:
I'm still probably heading in the wrong direction:
```
error[E0423]: expected value, found struct `std::io::Error`
  --> src/bin/stratis_dumpmetadata.rs:14:89
   |
14 |     let mut devfile = OpenOptions::new().read(true).open(&devpath).map_err(|the_io_err| std::io::Error, "Error opening device")?;
   |                                                                                         ^^^^^^^^^^^^^^ did you mean `std::io::Error { /* fields */ }`?
   |
help: possible better candidates are found in other modules, you can import them into scope
   |
1  | use chrono::format::Item::Error;
   |
1  | use clap::Format::Error;
   |
1  | use core::fmt::Error;
   |
1  | use dbus::MessageType::Error;
   |
     and 14 other candidates

error[E0061]: this function takes 1 argument but 2 arguments were supplied
  --> src/bin/stratis_dumpmetadata.rs:14:68
   |
14 |     let mut devfile = OpenOptions::new().read(true).open(&devpath).map_err(|the_io_err| std::io::Error, "Error opening device")?;
   |                                                                    ^^^^^^^ ---------------------------  ---------------------- supplied 2 arguments
   |                                                                    |
   |                                                                    expected 1 argument

error: aborting due to 2 previous errors
```

bgurney-rh:
And I was (mis)led to put a comma in there by this previous error:
```
error: expected one of `!`, `)`, `,`, `.`, `::`, `?`, `{`, or an operator, found `"Error opening device"`
  --> src/bin/stratis_dumpmetadata.rs:14:104
   |
14 |     let mut devfile = OpenOptions::new().read(true).open(&devpath).map_err(|the_io_err| std::io::Error "Error opening device")?;
   |                                                                                                       -^^^^^^^^^^^^^^^^^^^^^^ expected one of 8 possible tokens
   |                                                                                                       |
   |                                                                                                       help: missing `,`

error[E0423]: expected value, found struct `std::io::Error`
  --> src/bin/stratis_dumpmetadata.rs:14:89
   |
14 |     let mut devfile = OpenOptions::new().read(true).open(&devpath).map_err(|the_io_err| std::io::Error "Error opening device")?;
   |                                                                                         ^^^^^^^^^^^^^^ did you mean `std::io::Error { /* fields */ }`?
   |
help: possible better candidates are found in other modules, you can import them into scope
   |
1  | use chrono::format::Item::Error;
   |
1  | use clap::Format::Error;
   |
1  | use core::fmt::Error;
   |
1  | use dbus::MessageType::Error;
   |
     and 14 other candidates

```

src/bin/stratis_dumpmetadata.rs
```diff
@@ -0,0 +1,47 @@
...
+
+use serde_json::Value;
+
+use libstratis::engine::BDA;
+
+fn run(devpath: String) -> Result<(), String> {
+    let mut devfile = OpenOptions::new()
+        .read(true)
+        .open(&devpath)
+        .map_err(|_the_io_error| format!("Error opening device: {}", _the_io_error))?;
```
mulkieran, line 17: Go ahead and remove the ```_``` from ```_the_io_error``` since you are now using it.

src/bin/stratis_dumpmetadata.rs
```diff
@@ -0,0 +1,47 @@
...
+
+    let bda = BDA::load(&mut devfile).unwrap().unwrap();
```
mulkieran, line 19: Ok! Now the outermost unwrap() here is our next target. It means that there is definitely no Stratis header on the device, hence no BDA. We're going to treat no BDA as an error for purposes of this script.

src/bin/stratis_dumpmetadata.rs
bgurney-rh: Resolved.

src/bin/stratis_dumpmetadata.rs
bgurney-rh: Resolved.

src/bin/stratis_dumpmetadata.rs
```diff
@@ -0,0 +1,53 @@
...
+    let state_json_pretty: String = serde_json::to_string_pretty(&state_json)
+        .map_err(|parse_err| format!("Error during state JSON parse: {}", parse_err))?;
+    println!("{}", state_json_pretty);
+    Ok(())
+}
+
+fn main() {
+    let args: Vec<String> = env::args().collect();
+    if args.len() < 2 {
+        println!("Usage: stratis_dumpmetadata <device>");
```
mulkieran, line 41: This should be ```eprintln!``` as well

src/bin/stratis_dumpmetadata.rs
```diff
@@ -0,0 +1,53 @@
...
+        .map_err(|bda_load_error| format!("BDA detected but error found: {}", bda_load_error))?
+        .ok_or_else(|| "No Stratis BDA metadata found".to_string())?;
+    println!("{:#?}", bda);
+
+    devfile.seek(SeekFrom::Start(0)).unwrap();
+
+    let loaded_state = bda
+        .load_state(&mut devfile)
+        .map_err(|stateload_err| format!("Error during load state: {}", stateload_err))?;
+    println!("State JSON data:");
```
mulkieran, line 29: Please add a few more newlines, to group the actions better, like just after this ```println!```

src/bin/stratis_dumpmetadata.rs
```diff
@@ -0,0 +1,53 @@
...
+    io::{Seek, SeekFrom},
+    process,
+    vec::Vec,
+};
+
+use serde_json::Value;
+
+use libstratis::engine::BDA;
+
+fn run(devpath: String) -> Result<(), String> {
```
mulkieran, line 13: Step (1): Please change parameter type to &str. We don't need to copy the string, we just needed a reference to it.

src/bin/stratis_dumpmetadata.rs
```diff
@@ -0,0 +1,53 @@
...
+    Ok(())
+}
+
+fn main() {
+    let args: Vec<String> = env::args().collect();
+    if args.len() < 2 {
+        println!("Usage: stratis_dumpmetadata <device>");
+        process::exit(1);
+    }
+    let devpath = args[1].clone();
```
mulkieran, line 44: Step (2). Don't clone here. A reference will do, it isn't necessary to make a copy.

src/bin/stratis_dumpmetadata.rs
bgurney-rh: Resolved.

src/bin/stratis_dumpmetadata.rs
bgurney-rh: Resolved.

src/bin/stratis_dumpmetadata.rs
bgurney-rh: Resolved.

src/bin/stratis_dumpmetadata.rs
bgurney-rh: I'm not entirely sure what you mean (though I also might have missed this since I was editing in an 80-column terminal).

Would it be better to have a newline right before every `let <variable> = <function_call()>.map[_err]()` set of lines?

Here's an example of what it would look like:
```
fn run(devpath: &str) -> Result<(), String> {
    let mut devfile = OpenOptions::new()
        .read(true)
        .open(&devpath)
        .map_err(|the_io_error| format!("Error opening device: {}", the_io_error))?;

    let bda = BDA::load(&mut devfile)
        .map_err(|bda_load_error| format!("BDA detected but error found: {}", bda_load_error))?
        .ok_or_else(|| "No Stratis BDA metadata found".to_string())?;
    println!("{:#?}", bda);

    devfile.seek(SeekFrom::Start(0)).unwrap();

    let loaded_state = bda
        .load_state(&mut devfile)
        .map_err(|stateload_err| format!("Error during load state: {}", stateload_err))?;

    println!("State JSON data:");
    let state_json: Value = serde_json::from_slice(&loaded_state.unwrap())
        .map_err(|extract_err| format!("Error during state JSON extract: {}", extract_err))?;

    let state_json_pretty: String = serde_json::to_string_pretty(&state_json)
        .map_err(|parse_err| format!("Error during state JSON parse: {}", parse_err))?;

    println!("{}", state_json_pretty);
    Ok(())
}
```

mulkieran: Well, the way I'm thinking of it, is that the code takes a few steps to get some data, and then it displays it. Then there's a rinse, repeat sort of behavior. So after any ```println!()```, I'ld put in a newline, indicating that that action was completed and that a new action is commencing, which will result in more metadata being displayed, or an error return, of course.

bgurney-rh: I pushed a commit to add newlines (hopefully in the right places); let me know how it looks.

mulkieran: I think it looks good, I'm resolving it.

src/bin/stratis_dumpmetadata.rs
```diff
@@ -0,0 +1,55 @@
...
+        .read(true)
+        .open(&devpath)
+        .map_err(|the_io_error| format!("Error opening device: {}", the_io_error))?;
+
+    let bda = BDA::load(&mut devfile)
+        .map_err(|bda_load_error| format!("BDA detected but error found: {}", bda_load_error))?
+        .ok_or_else(|| "No Stratis BDA metadata found".to_string())?;
+    println!("{:#?}", bda);
+
+    devfile.seek(SeekFrom::Start(0)).unwrap();
```
mulkieran, line 24: Best to go ahead and deal with this ```unwrap()```.

src/bin/stratis_dumpmetadata.rs
```diff
@@ -21,7 +21,13 @@ fn run(devpath: &str) -> Result<(), String> {
         .ok_or_else(|| "No Stratis BDA metadata found".to_string())?;
     println!("{:#?}", bda);
 
-    devfile.seek(SeekFrom::Start(0)).unwrap();
+    match devfile.seek(SeekFrom::Start(0)) {
```
mulkieran, line 24: It makes much more sense to handle this just the same as all the other ```unwrap()```s.

The things you want are:

* a useful error
* that there will be an immediate exit from this function if the seek fails

and all that will be handled by use of ```map_err()``` and the ```?``` operator.

src/bin/stratis_dumpmetadata.rs
bgurney-rh: Resolved.

src/bin/stratis_dumpmetadata.rs
bgurney-rh: ...oh, that's why it didn't work.  I had abandoned the entire `map_err` statement for a `match` block, because I was getting "unused Result which must be used".

All that over a single `?` character.

mulkieran: Ah. You can imagine that ```?``` character as behaving like a macro. So,

```
<expr>?;
```

is more or less equivalent to:

```
match <expr> {
  Err(err) => return Err(err),
  Ok(thing) => thing,
};
```

there's a bit more to it than that, actually, but that _is_ the basic idea.

bgurney-rh:
I feel comfortable with the messages.  I'm going to close this, and create a new PR with these commits squashed, and rebased onto develop-2.1.0.

PR #2130, opened 2020-07-15
Add stratis_dumpmetadata program

bgurney-rh:
Supersedes: #2116
Closes: #2097

### DIFF
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -27,10 +27,11 @@ fn run(devpath: &str) -> Result<(), String> {
 
     let loaded_state = bda
         .load_state(&mut devfile)
-        .map_err(|stateload_err| format!("Error during load state: {}", stateload_err))?;
+        .map_err(|stateload_err| format!("Error during load state: {}", stateload_err))?
+        .ok_or_else(|| "No BDA state found".to_string())?;
     println!("State JSON data:");
 
-    let state_json: Value = serde_json::from_slice(&loaded_state.unwrap())
+    let state_json: Value = serde_json::from_slice(&loaded_state)
         .map_err(|extract_err| format!("Error during state JSON extract: {}", extract_err))?;
     let state_json_pretty: String = serde_json::to_string_pretty(&state_json)
         .map_err(|parse_err| format!("Error during state JSON parse: {}", parse_err))?;
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -28,7 +28,7 @@ fn run(devpath: &str) -> Result<(), String> {
     let loaded_state = bda
         .load_state(&mut devfile)
         .map_err(|stateload_err| format!("Error during load state: {}", stateload_err))?;
-    println!("State JSON data:");
+    println!("Pool metadata:");
 
     if let Some(loaded_state) = loaded_state {
         let state_json: Value = serde_json::from_slice(&loaded_state)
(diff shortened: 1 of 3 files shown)


src/bin/stratis_dumpmetadata.rs
```diff
@@ -0,0 +1,57 @@
...
+    devfile
+        .seek(SeekFrom::Start(0))
+        .map_err(|seek_err| format!("Error during seek: {}", seek_err))?;
+
+    let loaded_state = bda
+        .load_state(&mut devfile)
+        .map_err(|stateload_err| format!("Error during load state: {}", stateload_err))?;
+    println!("State JSON data:");
+
+    let state_json: Value = serde_json::from_slice(&loaded_state.unwrap())
```
jbaublitz, line 33: @mulkieran Should this `unwrap` be an `.expect()`?

mulkieran: > @mulkieran Should this `unwrap` be an `.expect()`?

Whoops! Thanks for noticing.

@bgurney-rh This ```unwrap()``` is for the ```Option``` part of the return type of ```load_state()```. So you should handle it like you handled the ```Option``` part of the result type of ```load()```. And you should handle it around line 30, where you handle the ```Result``` part of the return type of ```load_state()```.

bgurney-rh:
Take a look at the latest commit; does the message need to change, or do I need to do something different?

src/bin/stratis_dumpmetadata.rs
```diff
@@ -0,0 +1,58 @@
...
+    println!("{:#?}", bda);
+
+    devfile
+        .seek(SeekFrom::Start(0))
+        .map_err(|seek_err| format!("Error during seek: {}", seek_err))?;
+
+    let loaded_state = bda
+        .load_state(&mut devfile)
+        .map_err(|stateload_err| format!("Error during load state: {}", stateload_err))?
+        .ok_or_else(|| "No BDA state found".to_string())?;
```
mulkieran, line 31: Ok! We need to branch right here, depending on whether ```loaded_state``` is ```None``` or ```Some(..)```. So, finish the computation of ```loaded_state``` so that it is an ```Option```, like:

```
let loaded_state = bda
        .load_state(&mut devfile)
        .map_err(|stateload_err| format!("Error during load state: {}", stateload_err))?;
```

now,

```
if let Some(loaded_state) = loaded_state {
  // all the computations you've already constructed for when
  // loaded_state is Some should go on this branch
} else {
  // Different output for when loaded_state is None.
  // Why is this possible? With very many devices in a pool, the
  // metadata may only be written to a selected set, Or, the pool
  // could have gone down before the metadata could be written.
  // Either way, it's a valid possibility.
}
```

Why not use the equivalent

```
match loaded_state {
  Some(loaded_state) => { // if branch stuff },
  None => { // else branch stuff },
}
```

instead of the ```if-let``` expression, as they are completely equivalent? clippy prefers ```if-let``` and the resulting indentation is slightly easier on the eyes.

src/bin/stratis_dumpmetadata.rs
```diff
@@ -27,15 +27,18 @@ fn run(devpath: &str) -> Result<(), String> {
 
     let loaded_state = bda
         .load_state(&mut devfile)
-        .map_err(|stateload_err| format!("Error during load state: {}", stateload_err))?
-        .ok_or_else(|| "No BDA state found".to_string())?;
+        .map_err(|stateload_err| format!("Error during load state: {}", stateload_err))?;
     println!("State JSON data:");
```
mulkieran, line 31: I think a better title than "State JSON data" would be "pool metadata", because that's really what it is. I've been calling it "variable length metadata", because that's how we have to deal with it, but "pool metadata" is a far more descriptive term for the user. Can you change this heading to "Pool metadata:", and where there is none just report "None" or "None found" or something like that?

Note that it will be hard to demonstrate the situation where the pool metadata isn't there w/out making a very large pool or crashing the pool with very precise timing. But it will occur eventually.

Note that you could actually use this script, with some alteration, to detect when the pool metadata gets written to a device. I've been somewhat worried about delays under certain conditions, and there's an issue out for that.

Issue #1165, opened 2018-09-07
Need a testing infrastructure for D-Bus signals


mulkieran:
AFAICT, this infrastructure would be best written in Rust. It seems like the Python implementation of D-Bus wrappers all require some heavy duty event-loop stuff for monitors and bring in more dependencies that way. A Rust implementation would probably have less overhead.

mulkieran:
Currently, the only properties that emit signals are Name on pool and filesystem interfaces.

mulkieran:
It looks like there was a bug in sim engine that caused properties not to be emitted, but we have all the infrastructure to do the emitting, and with the bug fixed I see this:

```
‣ Type=signal  Endian=l  Flags=1  Version=1  Priority=0 Cookie=8
  Sender=:1.3494  Path=/org/storage/stratis2/1  Interface=org.freedesktop.DBus.Properties  Member=PropertiesChanged
  UniqueName=:1.3494
  MESSAGE "sa{sv}as" {
          STRING "org.storage.stratis2.pool";
          ARRAY "{sv}" {
                  DICT_ENTRY "sv" {
                          STRING "Name";
                          VARIANT "s" {
                                  STRING "poolname2";
                          };
                  };
          };
          ARRAY "s" {
          };
  };
```

via "busctl monitor" when a pool is renamed.

mulkieran:
@mvollmer: Can you confirm that this is what you would expect to see if a signal were being emitted when the pool "Name" property is changed?

mvollmer:
> @mvollmer: Can you confirm that this is what you would expect to see if a signal were being emitted when the pool "Name" property is changed?

Yes, and I indeed do get this signals also in my own manual testing. I must have missed them earlier.

mulkieran:
@mvollmer: Great!

I'm continuing to hold this bug open, because figuring out a good way to test that we do put out the signals as appropriate is desirable.

jbaublitz:
@mulkieran I've found some Rust functionality in dbus-rs that should be able to listen for signals and send a notification back to main thread to indicate success. I'll see if I can put together some infrastructure for this. However, I'm unsure where this test would go as all of our Rust tests are currently unit tests and I don't think this would work very well as a unit test.

mulkieran:
[...]

PR #2138, opened 2020-07-20
Have sim engine also emit change signals for renames

mulkieran:
Related: #1165.


mulkieran:
This is just to bring the sim engine and the strat engine into alignment, in case this is useful for testing. There are equivalent actions in two places in the strat engine.

Issue #2117, opened 2020-07-08
Use clippy to find casting and change to From and TryFrom conversions


jbaublitz:
When testing down-casting integers on s390x, a big endian system, going from a u64 to a u32 appeared to truncate the last 32 bits of the integer which, in this case, contain all of the information that we're looking for. Casting between integers of the same width should not be a problem, but in the interest of making stratisd cross-platform and as safe as possible, we should enable the clippy lint about unsafe conversions between integers and try to use the `From` and `TryFrom` traits for integer conversions instead.

~- [ ] Confirm behavior on s390x~
- [ ] Enable clippy lint for integer conversions
- [ ] Change all casts to `From`/`TryFrom`

mulkieran:
Note that the use of ```Try/TryFrom``` is to hand off the management of endianness to Rust. We would still in all places know that a narrowing conversion was safe, and so we would ```expect()``` it to succeed.

mulkieran:
Note that we're talking about a bunch of lints that are in the pedantic category and that are not turned on by default. Currently the only special lint that we turn on is the "needless_borrow" lint, which is actually in the "nursery" category, not the "pedantic" category.

mulkieran:
Related: https://github.com/stratis-storage/project/issues/97.

mulkieran:
First thing is to confirm the s390x behavior.

jbaublitz:
@mulkieran It turns out that this was a sneaky bug in libcryptsetup-rs tests that perhaps was only visible on s390x for architecture for platform specific reasons, but the failure based on the code that I had was indeed a bug. The PR that I put up for the release of 0.4.2 does indeed fix it but it is not related to endianness. Ultimately it was a problem with an expression such as:
```
let ptr = &mut 1 as *mut _ as *mut c_void
```
This is not normally a problem unless you then try to convert the pointer to a `&mut 64` reference as the compiler will automatically assign 32 bits to a bare integer constant without changing the above to:
```
let ptr = &mut 1u64 as *mut _ as *mut c_void
```
When I added a type to the integer constants, the tests passed even when converting from the pointer to a `u64`. This has already been fixed in libcryptsetup-rs but I'm going to put explicit types on the integer constants anyway just so that this doesn't come back in the future.

As for safe conversions, I still think this is a valuable code improvement even though the problem we originally thought we found with endianness is not a concern here. Would you like me to continue with this?

mulkieran:
[...]

PR #2139, opened 2020-07-21
Rework our casting standard

jbaublitz:
Related to #2117

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,9 +10,9 @@ RUST_2018_IDIOMS = -D bare-trait-objects \
 DENY = -D warnings -D future-incompatible -D unused ${RUST_2018_IDIOMS}
 
 CLIPPY_DENY = -D clippy::cast_lossless \
-	      -D clippy::cast_possible_truncation \
-	      -D clippy::needless_borrow \
-	      -D warnings
+              -D clippy::cast_possible_truncation \
+              -D clippy::needless_borrow \
+              -D warnings
 
 ${HOME}/.cargo/bin/cargo-outdated:
 	cargo install cargo-outdated
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,7 @@ stratisd.8.gz: stratisd.8
 	gzip --stdout docs/stratisd.8 > docs/stratisd.8.gz
 
 clippy:
-	cargo clippy --all-targets --all-features -- -D warnings -D clippy::needless_borrow -D clippy::cast_lossless
+	cargo clippy --all-targets --all-features -- -D warnings -D clippy::needless_borrow -D clippy::cast_lossless -D clippy::cast_possible_truncation
 
 .PHONY:
 	audit
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,11 @@ RUST_2018_IDIOMS = -D bare-trait-objects \
 
 DENY = -D warnings -D future-incompatible -D unused ${RUST_2018_IDIOMS}
 
+CLIPPY_DENY = -D clippy::cast_lossless \
+	      -D clippy::cast_possible_truncation \
+	      -D clippy::needless_borrow \
+	      -D warnings
+
 ${HOME}/.cargo/bin/cargo-outdated:
 	cargo install cargo-outdated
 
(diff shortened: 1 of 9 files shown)


src/engine/strat_engine/backstore/metadata/mda.rs
```diff
@@ -406,7 +406,7 @@ impl MDAHeader {
         // size of metadata is greater than 2^32 - 1 bytes, which is
         // unlikely.
         assert!(*self.used.bytes() as u64 <= std::usize::MAX as u64);
-        let mut data_buf = vec![0u8; *self.used.bytes() as usize];
+        let mut data_buf = vec![0u8; convert_int!(*self.used.bytes(), u64, usize)];
```
mulkieran, line 409: You can remove the assert! and the comment above it.

src/engine/macros.rs
```diff
@@ -189,3 +189,29 @@ macro_rules! init_cache_generate_error_string {
...
+
+macro_rules! convert_int {
+    ($expr:expr, $from_type:ty, $to_type:ty) => {
+        <$to_type as std::convert::TryFrom<$from_type>>::try_from($expr).map_err(|_| {
+            StratisError::Error(format!(
+                "Failed to convert from {} to {}",
+                stringify!($from_type),
+                stringify!($to_type)
+            ))
+        })?
```
mulkieran, line 201: Can you take the ```?``` out of the macro and put it at each expansion site? I think it helps a bit in reading comprehension to have it visible in the methods.

jbaublitz: This honestly may be even better from a functionality perspective too as the method the macro is being called from currently must return a `StratisResult`. This would give more flexibility. I'll fix that.

Makefile
```diff
@@ -92,7 +92,7 @@ stratisd.8.gz: stratisd.8
 	gzip --stdout docs/stratisd.8 > docs/stratisd.8.gz
 
 clippy:
-	cargo clippy --all-targets --all-features -- -D warnings -D clippy::needless_borrow
+	cargo clippy --all-targets --all-features -- -D warnings -D clippy::needless_borrow -D clippy::cast_possible_truncation
```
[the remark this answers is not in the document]
jbaublitz: Resolved.

src/engine/macros.rs
```diff
@@ -189,3 +189,29 @@ macro_rules! init_cache_generate_error_string {
         )
     };
 }
+
+macro_rules! convert_int {
+    ($expr:expr, $from_type:ty, $to_type:ty) => {
+        <$to_type as std::convert::TryFrom<$from_type>>::try_from($expr).map_err(|_| {
+            StratisError::Error(format!(
+                "Failed to convert from {} to {}",
```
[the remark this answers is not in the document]
jbaublitz: Resolved.

src/engine/strat_engine/backstore/metadata/mda.rs
jbaublitz: Resolved.

src/engine/macros.rs
jbaublitz: Resolved.

jbaublitz:
One note about the `CLIPPY_DENY` variable in this diff is that they're lined up in vim for me so apparently there's some kind of tab width difference here. I'm not sure how you want to address that. I can definitely make it unaligned for me in vim and make sure it's aligned here if that's desirable.

mulkieran:
> One note about the `CLIPPY_DENY` variable in this diff is that they're lined up in vim for me so apparently there's some kind of tab width difference here. I'm not sure how you want to address that. I can definitely make it unaligned for me in vim and make sure it's aligned here if that's desirable.

@jbaublitz I put up two tiny PRs that are now in the review slot. They both just expand the tabs to spaces in multiline variable assignment. There's no corresponding PR for libcryptsetup-rs, because I had already expanded the tabs to spaces there. I think this is the best solution, and what we should do w/ ```CLIPPY_DENY``` as well.

jbaublitz:
> > One note about the `CLIPPY_DENY` variable in this diff is that they're lined up in vim for me so apparently there's some kind of tab width difference here. I'm not sure how you want to address that. I can definitely make it unaligned for me in vim and make sure it's aligned here if that's desirable.
> 
> @jbaublitz I put up two tiny PRs that are now in the review slot. They both just expand the tabs to spaces in multiline variable assignment. There's no corresponding PR for libcryptsetup-rs, because I had already expanded the tabs to spaces there. I think this is the best solution, and what we should do w/ `CLIPPY_DENY` as well.

Perfect! I'll adjust that here.

jbaublitz:
Rebased and ready for review.

PR #2146, opened 2020-07-24
Emit InterfacesAdded and InterfacesRemoved at D-Bus layer of stratisd

### LINKED ISSUE #1377: Investigate if dbus-rs emits InterfaceChanged signals properly
Even if our objects do not dynamically change interfaces, InterfaceAdded is the signal that should be emitted when an object is added to the set of object available via GetManagedObjects, and vice-versa for InterfaceRemoved.

~~We discussed using multiple interfaces to indicate a pool gaining or losing capabilities based on its state. In order to work, changing interfaces should (automatically) cause InterfaceChanged events to be generated. We need to verify this is the case before going much further in investigating this as part of our stable API plans.~~ We are not yet in a situation where we need to change interfaces on our objects.

jbaublitz:
Closes #1377

### DIFF
--- a/src/dbus_api/api/manager_2_0/methods.rs
+++ b/src/dbus_api/api/manager_2_0/methods.rs
@@ -10,6 +10,7 @@ use dbus::{
 use crate::{
     dbus_api::{
         api::shared::create_pool_shared,
+        consts,
         types::TData,
         util::{engine_to_dbus_err_tuple, get_next_arg, msg_code_ok, msg_string_ok},
     },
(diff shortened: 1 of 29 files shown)


mulkieran:
In ```create_dbus_pool```, ```create_dbus_blockdev```, ```create_dbus_filesystem``` a DeferredAction is constructed and placed in a queue, and similarly on removal. Then, in ```process_deferred_actions``` appropriate D-Bus related actions are taken. I think this must be the place where the methods in this PR should be invoked.

jbaublitz:
@mulkieran Yes, that's what I'm working on now.

jbaublitz:
TODO: Try to factor out more specific property generation into `shared.rs` for each D-Bus object type. This should eliminate having two separate places where the property formats need to be changed.

jbaublitz:
@mvollmer If you'd like to test this PR out before it's merged, my initial tests with `busctl monitor` seem to indicate that the desired signals are now being sent. Let me know if you bump into any problems or have additional feedback!

mvollmer:
> @mvollmer If you'd like to test this PR out before it's merged, my initial tests with `busctl monitor` seem to indicate that the desired signals are now being sent. Let me know if you bump into any problems or have additional feedback!

@jbaublitz I tried this out and it indeed all works as expected. I have updated https://github.com/cockpit-project/cockpit/pull/14109 to depend on this PR.

Thanks, this is excellent!

jbaublitz:
@mvollmer Excellent! Just one note that I will need to keep rebasing this PR until it gets merged so I'm not sure how you're doing dependency management here, but if you're manually updating the git repo, just be aware that rebases will be happening until this gets reviewed.

jbaublitz:
Rebased.

jbaublitz:
@mulkieran Good catch! Resolved.

PR #2152, opened 2020-07-30
Change stratis_uuids_to_names message from INFO to DEBUG level

bgurney-rh:
Related: #704

Issue #704, opened 2018-01-02
/dev/stratis may not be a directory


agrover:
As @trgill mentions in #682, somebody could have been shady and created a /dev/stratis that is not a directory, which will cause us to fail.

BTW, setup_devlinks() is going to some trouble to not just remove whatever is there and recreate, because the user could have a shell sitting in /dev/stratis/foo, and having it suddenly be sitting in an unlinked empty directory would be confusing.

I'm not sure if this is worth fixing or not. Can't we come up with infinite ways that a root user could deliberately cause stratisd to fail to work properly?

mulkieran:
We can use ```fs::metadata``` to check if the directory we just created is a directory. If not it was already there and was a file.

mulkieran:
This is old, but it is related to https://github.com/stratis-storage/project/issues/95.

jbaublitz:
@mulkieran Once we lift this into udev, we will no longer have control over that piece of things. What we can do, however, is check at the very beginning of stratisd execution, fail to start, and provide a good error message to let users know exactly what the problem is. Does that sound reasonable to you?

mulkieran:
> @mulkieran Once we lift this into udev, we will no longer have control over that piece of things. What we can do, however, is check at the very beginning of stratisd execution, fail to start, and provide a good error message to let users know exactly what the problem is. Does that sound reasonable to you?

Sounds good!

mulkieran:
We haven't addressed this yet. An interesting thing would be to find out what happens with current version if this does occur. Reassigning.

bgurney-rh:
I set up a test where a blank file is created, via `echo "" > /dev/stratis`.

Then, I ran these commands: `stratis pool create spool1 /dev/vdb2` and `stratis fs create spool1 sfs1`.

I saw this in journalctl:
```
stratis_uuids_t[44733]: Symlink /dev/stratis/spool1/sfs1 created.
systemd-udevd[44728]: dm-5: Failed to create symlink '/dev/stratis/spool1/sfs1.tmp-b253:5' to '../../dm-5': Not a directory
```

However, the file is still there, and the directory in `/dev/stratis` was not created:

```
# ls -l /dev/stratis
-rw-r--r--. 1 root root 1 Jul 28 12:44 /dev/stratis
```

mulkieran:
@bgurney-rh : thx

mulkieran:
[...]

PR #2153, opened 2020-07-30
Change stratis_uuids_to_names message from INFO to DEBUG level

bgurney-rh:
Supersedes: #2152 
Related: #704


jbaublitz:
@mulkieran I thought this was small enough to merge. I'm happy to revise the debug message to say "reported... to stratis_uuid_to_names" instead of "reported... to udev" if you have a strong preference for that. I understand that users can theoretically run this by hand, but I feel like referring to a component by name that is not necessarily known to the user could result in confusion for systems administrators.

mulkieran:
@jbaublitz: Very happy w/ that wording.

PR #2155, opened 2020-08-03
Add script to migrate legacy symlinks

bgurney-rh:
Related: stratis-storage/project#203

### DIFF
--- a/developer_tools/stratis_migrate_symlinks.sh
+++ b/developer_tools/stratis_migrate_symlinks.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+# Migrate the symlinks of Stratis filesystems from the old
+# /stratis path to /dev/stratis, via synthetically generating
+# udev change events.
+
+if [ ! -d /stratis ]
+then
+	echo "'/stratis' directory does not exist; nothing to do."
+	exit 0
+fi
+
+for i in $(find /stratis)
+do
+	if [ -h $i ] && [ -b $i ]
+	then
+		devname=$i
+		echo "Link name: $devname"
+		linktgt=$(readlink $devname)
+		tgtbase=$(basename $linktgt)
+		echo "Link target: $linktgt"
+		echo "Target base name: $tgtbase"
+		futurename="/dev$devname"
+		echo "Future name: $futurename"
+		if [ -h $futurename ] && [ -b $futurename ]
+		then
+			echo "Link seems to already exist"
+			# Paths from udev are set relative
+			# to /dev/stratis, so use readlink -e.
+			futuretgt=$(readlink -e $futurename)
+			echo "Future link target: $futuretgt"
+			if [ ! $futuretgt -ef $linktgt ]
+			then
+				echo "Targets do not match; sending change event to re-synchronize..."
+				udevadm test --action=change /sys/class/block/$tgtbase 1>/dev/null 2>&1
+				rm -fv $devname
+			else
+				echo "Targets match"
+				rm -fv $devname
+			fi
+		else
+			echo "No future link found; sending change event..."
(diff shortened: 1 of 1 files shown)


bgurney-rh:
Rebased.

developer_tools/stratis_migrate_symlinks.sh
```diff
@@ -0,0 +1,49 @@
...
+				echo "Targets do not match; sending change event to re-synchronize..."
+				udevadm test --action=change /sys/class/block/$tgtbase 1>/dev/null 2>&1
+				rm -fv $devname
+			else
+				echo "Targets match"
+				rm -fv $devname
+			fi
+		else
+			echo "No future link found; sending change event..."
+			udevadm test --action=change /sys/class/block/$tgtbase 1>/dev/null 2>&1
```
mulkieran, line 43: Why can't you use linktgt here instead of ```/sys/class/block/...```.

developer_tools/stratis_migrate_symlinks.sh
```diff
@@ -0,0 +1,49 @@
...
+# /stratis path to /dev/stratis, via synthetically generating
+# udev change events.
+
+if [ ! -d /stratis ]
+then
+	echo "'/stratis' directory does not exist; nothing to do."
+	exit 0
+fi
+
+for i in $(find /stratis)
```
mulkieran, line 13: Can you decorate this loop with some docs about the rationale for this approach?

developer_tools/stratis_migrate_symlinks.sh
```diff
@@ -0,0 +1,49 @@
...
+			echo "Future link target: $futuretgt"
+			if [ ! $futuretgt -ef $linktgt ]
+			then
+				echo "Targets do not match; sending change event to re-synchronize..."
+				udevadm test --action=change /sys/class/block/$tgtbase 1>/dev/null 2>&1
+				rm -fv $devname
+			else
+				echo "Targets match"
+				rm -fv $devname
+			fi
```
mulkieran, line 40: Please combine the ```rm -fv $devname``` lines into one immediately after the ```if-else``` block.

mulkieran:
@bgurney-rh Here's my suggestion for pre-conditions to satisfy before running:

* Verify that udev rule is in correct place.
* Verify that udev script is in correct place.
* Additionally, call udev script and ask it if it is ready to run. Note that this will involve a minor addition to the udev script:
    * An additional flag which allows it to be called to use the D-Bus to verify that stratisd is running and has the correct version number. Since this is a modification to a Rust executable which already uses the D-Bus, that should be a lot easier than trying to do the check in the shell script.

bgurney-rh:
Admittedly, I wanted to get the rest of the "plumbing" done before tackling the version check.

It is possible to run `/usr/libexec/stratisd -V`, but on what condition of the version string do we proceed with the rest of the script?  As it is, it only acts on the existence of the `/stratis` directory, and on symlinks to block files (the cryptic if condition line `if [ -h $i ] && [ -b $i ]`).

mulkieran:
> Admittedly, I wanted to get the rest of the "plumbing" done before tackling the version check.
> 
> It is possible to run `/usr/libexec/stratisd -V`, but on what condition of the version string do we proceed with the rest of the script? As it is, it only acts on the existence of the `/stratis` directory, and on symlinks to block files (the cryptic if condition line `if [ -h $i ] && [ -b $i ]`).

The crucial bit is that the correct version of stratisd has to be up and running and replying on the D-Bus for the udev script to work at all. I think in previous discussions we may have overlooked that. We'll have to discuss that a little bit more.

But, the first two points, checking that the udev rule and the udev script are where they need to be you can work into the script now, I think.

bgurney-rh:
Rebased.

mulkieran:
Let's consider this superceded by #2174.

Issue #2157, opened 2020-08-05
Add Makefile target and Travis task to test the extras feature


jbaublitz:
Currently we are not testing the extras feature in CI which could result in breakage when some of our utilities locatied in `src/bin` fall out of sync with the engine API.

This occurred to me when considering the order of merge between #2156 and #2134.

mulkieran:
Closed by attached PR.

PR #2158, opened 2020-08-05
Add build-extras Makefile target

bgurney-rh:
Related: #2157


jbaublitz:
@bgurney-rh For now, leave the target you removed. This extras target is for all future `extras` executables and we want the ability to be able to make them individually.

bgurney-rh:
OK; I'll drop that last patch (that removes the stratis-dumpmetadata target), and spin this into a new PR.

PR #2136, opened 2020-07-20
stratis-min

jbaublitz:
Prototype for #2134 
Related to stratis-storage/project#209

Issue #2134, opened 2020-07-16
Add a script to set the Stratis keys in the initramfs


jbaublitz:
This process would involve creating another executable in `src/bin/` in the stratisd repo. This executable would use existing code from the engine in stratisd to set a key in the keyring and unlock locked devices.

Motivations
---------------
* This could be used to further develop inspection and repair tools that do not require a complete instance of stratisd to be running.
* This could also be useful to users when attempting to set up something like encryption in a semi-automated way at boot.

Steps
-------

- [x] Write an executable that uses the key infrastructure in stratisd to do key operations.
- [x] Do some investigation into getting user input from dracut-based initramfs booting.
- [x] Write a prototype of a dracut module that can query the user for passphrase input and unlock the locked crypt devices.
- [x] Test that booting from this initramfs does indeed unlock root encrypted filesystems without issue at boot.

alexminder:
Would it be possible to support [Clevis](https://github.com/latchset/clevis) for automated decryption a filesystem and rootfs particular?

mulkieran:
@alexminder We actually have an issue out for NBDE for block devices: https://github.com/stratis-storage/project/issues/192. It's a bit terse right now, but it will be acquiring more detail, probably in the fairly near future. But we don't support encryption of individual filesystems at all. That would be very much a separate issue. rootfs is also a larger story, although we have done some investigation there as well. When we have made a little more progress it will likely receive its own issue.

alexminder:
@mulkieran  My interest is in TPM2 store for encryption keys which supported by Clevis.

alexminder:
TPM2 will be native supported by cryptsetup without Clevis, soon. https://gitlab.com/cryptsetup/cryptsetup/-/merge_requests/51 .
Would you plan to implement TPM2 support in stratir and which way?

mulkieran:
@alexminder Please file an issue for this in the project repo: https://github.com/stratis-storage/project/issues/new. We can track it there much better than in this issue, which has really rather limited scope and will probably close in a week.

Please include in the issue both the link you included in your comment and motivation, i.e., what would make this enhancement desirable.

alexminder:
The issue have been created https://github.com/stratis-storage/project/issues/207 .

mulkieran:
@alexminder Thanks!

PR #2160, opened 2020-08-05
more stratis-min

jbaublitz:
Related to #2134 
~Blocked by #2136~


jbaublitz:
@mulkieran The failures on Jenkins are the same failures that I saw in the initramfs due to no controlling terminal. Given that we may not want to require a controlling terminal for file descriptors that are passed in, we may want to reconsider #2163. If we leave it to the client (stratis-cli or the consumer of our D-Bus API) to handle this, it leaves room for error, but it also will work in more environments.

jbaublitz:
I rebased and I'm going to see what happens with the tests this time.

jbaublitz:
I'm not sure where you're seeing information about the lock file. The message I'm getting is: `ERROR: IO error: Inappropriate ioctl for device (os error 25)`. This would make sense to me because if it's communicating over D-Bus, the terminal settings will be set on the provided file descriptor and that probably means that Jenkins doesn't have a controlling terminal for stdin.

mulkieran:
```
test_duplicate_pool_name (tests.udev.test_udev.UdevTest5)
Create more than one pool with the same name, then dynamically fix it ... ERROR
/usr/lib64/python3.7/subprocess.py:883: ResourceWarning: subprocess 44052 is still running
  ResourceWarning, source=self)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
/usr/lib64/python3.7/unittest/case.py:656: ResourceWarning: unclosed file <_io.TextIOWrapper name=3 encoding='UTF-8'>
  outcome.errors.clear()
ResourceWarning: Enable tracemalloc to get the object allocation traceback
/usr/lib64/python3.7/unittest/case.py:656: ResourceWarning: unclosed file <_io.TextIOWrapper name=6 encoding='UTF-8'>
  outcome.errors.clear()
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```

This _looks_ like the kind of error that we used to get when our test to verify that no 2 stratisd processes could be running at the same time would stall out, and which I got once again sometime last week for the first time in a long time.

jbaublitz:
@mulkieran Ok we're looking at different tests! I was looking at `cli-with-stratisd`. I see that in `stratisd-nonrust` now. I think it may still be due to a bug in my code related to file descriptors. I'm going to try to pull the file descriptor code out into its own PR and see if I can fix it.

mulkieran:
> @mulkieran Ok we're looking at different tests! I was looking at `cli-with-stratisd`. I see that in `stratisd-nonrust` now. I think it may still be due to a bug in my code related to file descriptors. I'm going to try to pull the file descriptor code out into its own PR and see if I can fix it.

Yeah. But now I'm more willing to believe that the errors that I'm looking at in the udev tests are a result rather than a cause of the other errors. You're right that the cli-with-stratisd errors look like just the same problem.

jbaublitz:
@mberndt123 I think you left a message in #stratis-storage on Freenode. This PR might be interesting for you as it will provide the necessary scaffolding for booting from a Stratis-based rootfs with encrypted devices. My guess is your current setup is using stratisd for unencrypted rootfs, but please correct me if I'm wrong.

mberndt123:
Hi @jbaublitz,

yes, I'm using that for an unencrypted rootfs.
Before Stratis I was using a really old-fashioned setup with an MBR partition table, a swap partition and single root partition containing everything. It was working fine, but at some point I needed more space and I bought a new drive. Of course moving the OS and all the data over is a bit of a chore, so I decided that if I have to do it anyway, I might as well do it in a way that makes it easier in the future. Next time, I just want to add a new storage device, remove the old one and be done with it. And since Stratis is the cool new kid on the block (device), I thought I'd give it a try. While I was at it, I also created separate file systems for /home and / so that I'd be able to snapshot them independently. I had considered such a setup before, but ultimately decided against it because I didn't like the idea of having to make a decision about how big those file systems were going to be. With Stratis, this problem just doesn't exist because of thin provisioning. I find all this really very cool.

I'm not going to bother with Stratis' encryption support, because honestly I find it pretty lame that it can only be switched on and off at the pool level rather than the fs level. That means I'd have to copy all my data around *again* to create a new pool, which is exactly what I was trying to avoid in the first place. I don't even have a drive I could put it on, because at this point I've obviously exceeded my old drive's capacity.

jbaublitz:
> Hi @jbaublitz,
> 
> yes, I'm using Stratis for an unencrypted rootfs.
> Before Stratis I was using a really old-fashioned setup with an MBR partition table, a swap partition and single root partition containing everything. It was working fine, but at some point I needed more space and I bought a new drive. Of course moving the OS and all the data over is a bit of a chore, so I decided that if I have to do it anyway, I might as well do it in a way that makes it easier in the future. Next time, I just want to add a new storage device, remove the old one and be done with it. And since Stratis is the cool new kid on the block (device), I thought I'd give it a try. While I was at it, I also created separate file systems for /home and / so that I'd be able to snapshot them independently. I had considered such a setup before, but ultimately decided against it because I didn't like the idea of having to make a decision about how big those file systems were going to be. With Stratis, this problem just doesn't exist because of thin provisioning. I find all this really very cool.
> 
> I'm not going to bother with Stratis' encryption support, because honestly I find it pretty lame that it can only be switched on and off at the pool level rather than the fs level. That means I'd have to backup all my data _again_ to create a new pool, which is exactly what I was trying to avoid in the first place. I don't even have a drive I could put it on, because at this point I've obviously exceeded my old drive's capacity. Thanks, but no thanks.
> 
> Any chance per-fs encryption would ever be implemented? I would probably try it in that case… Or if there were a way to add encryption to a pool after creation.

@mberndt123 Our decision to encrypt at the pool level is due to some of the considerations about our internal stacking of devicemapper targets combined with our discussions with the cryptsetup maintainers so that is unlikely to change. However, we have discussed the option of enabling encryption after the fact. I will let @drckeefe and @mulkieran know that this is something that is of interest.

mberndt123:
Hey @jbaublitz,

awesome, thanks! Good to know that users' requirements are taken seriously around here. I don't regret trusting stratis with my data.

PR #2163, opened 2020-08-10
Move terminal settings handling to stratisd

jbaublitz:
Related to stratis-storage/project#209

### DIFF
--- a/src/dbus_api/api/manager_2_2/methods.rs
+++ b/src/dbus_api/api/manager_2_2/methods.rs
@@ -1,0 +1,11 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use dbus::tree::{MTFn, MethodInfo, MethodResult};
+
+use crate::dbus_api::{api::shared::set_key_shared, types::TData};
+
+pub fn set_key(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
+    set_key_shared(m, true)
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -411,6 +411,7 @@ dependencies = [
  "serde_derive 1.0.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "tempfile 3.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "termios 0.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "timerfd 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "uuid 0.7.4 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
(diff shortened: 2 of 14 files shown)


jbaublitz:
@mulkieran I'm holding off on this until I can get a working prototype for stratis-min in the initramfs. I may need to do some more moving stuff around depending on how that goes.

jbaublitz:
We may want to reconsider this. See [here](https://github.com/stratis-storage/stratisd/pull/2160#issuecomment-677948138) for details.

jbaublitz:
@mulkieran This tentatively should resolve the testing problem.

jbaublitz:
@mulkieran This is looking like it resolves the two test failures that we saw. I'm going to do one last pass and put it up for review.

jbaublitz:
@mulkieran I think this is ready. We should talk in stand up about one issue, but other than that, I think it's ready for your initial thoughts.

src/dbus_api/api/manager_2_2/methods.rs
```diff
@@ -0,0 +1,11 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use dbus::tree::{MTFn, MethodInfo, MethodResult};
+
+use crate::dbus_api::{api::shared::set_key_shared, types::TData};
+
+pub fn set_key(m: &MethodInfo<MTFn<TData>, TData>) -> MethodResult {
+    set_key_shared(m, true)
```
mulkieran, line 10: I don't understand here why the second argument is hard-coded to ```true``` here, because I had thought that there were some situations where it would still be undesirable to change the settings on the daemon side.

jbaublitz: This is part of what I want to talk about in stand up. There are two points to discuss here:
1. The bug that I introduced that is resolved here.
2. Whether we should require a terminal to be present when interactively providing a passphrase over the D-Bus. My initial suggestion of allowing the user to choose through a parameter over the D-Bus was influenced by the errors I was seeing in point #1 which are now resolved.

mulkieran: Ok!

mulkieran: The answer is that we believe that when the D-Bus is up, altering the settings in stratisd always makes sense. It is only for stratis-min that we wan't to be able to decide whether or not to do this.

mulkieran:
@jbaublitz: I had one last request on this one...

jbaublitz:
@mulkieran I'm waiting to do any further work on this for now until I get a prototype of the systemd service files in place in dracut so that I can get a sense for the overall state of things. I may be able to make the code simpler once I have an environment in which I can test it.

mulkieran:
Ok!

jbaublitz:
I'm going to address your change request and then put this up for review @mulkieran. Now that #2160 is a working prototype, I have more confidence that we don't need to make anymore changes here.

jbaublitz:
Ready for review!

mulkieran (changes requested):
This looks ready! Please close and open a new PR for final review.

PR #2166, opened 2020-08-11
Fix udev rule to probe with blkid

### LINKED ISSUE #2162: ID_FS_USAGE etc are not set in udev immediately after "stratis fs create"
After creating a new filesystem with "stratis fs create" or via the D-Bus API, the ID_FS_USAGE etc properties are not set in udev.

```
$ stratis fs create pool0 fsys
$ udevadm info $(readlink /stratis/pool0/fsys)
P: /devices/virtual/block/dm-7
N: dm-7
L: 0
S: disk/by-id/dm-uuid-stratis-1-a1654aed04c14ec69e595aa151b32b76-thin-fs-f4529e5796b24c2ba8c69f64bab89ae4
S: mapper/stratis-1-a1654aed04c14ec69e595aa151b32b76-thin-fs-f4529e5796b24c2ba8c69f64bab89ae4
S: disk/by-id/dm-name-stratis-1-a1654aed04c14ec69e595aa151b32b76-thin-fs-f4529e5796b24c2ba8c69f64bab89ae4
E: DEVPATH=/devices/virtual/block/dm-7
E: DEVNAME=/dev/dm-7
E: DEVTYPE=disk
E: MAJOR=253
E: MINOR=7
E: SUBSYSTEM=block
E: USEC_INITIALIZED=198498677
E: DM_UDEV_PRIMARY_SOURCE_FLAG=1
E: DM_ACTIVATION=1
E: DM_NAME=stratis-1-a1654aed04c14ec69e595aa151b32b76-thin-fs-f4529e5796b24c2ba8c69f64bab89ae4
E: DM_UUID=stratis-1-a1654aed04c14ec69e595aa151b32b76-thin-fs-f4529e5796b24c2ba8c69f64bab89ae4
E: DM_SUSPENDED=0
E: DM_UDEV_RULES_VSN=2
E: E2_SIZE=0
E: DEVLINKS=/dev/disk/by-id/dm-uuid-stratis-1-a1654aed04c14ec69e595aa151b32b76-thin-fs-f4529e5796b24c2ba8c69f64bab89ae4 /dev/mapper/stratis-1-a1654aed04c14ec69e595aa151b32b76-thin-fs-f4529e5796b24c2ba8c69f64bab89ae4 /dev/disk/by-id/dm-name-stratis-1-a1654aed04c14ec69e595aa151b32b76-thin-fs-f4529e5796b24c2ba8c69f64bab89ae4
E: TAGS=:systemd:
```

The properties how up after `udevadm trigger`, or a reboot.

```
$ udvadm trigger
$ udevadm info $(readlink /stratis/pool0/fsys)
P: /devices/virtual/block/dm-7
N: dm-7
L: 0
S: disk/by-uuid/f4529e57-96b2-4c2b-a8c6-9f64bab89ae4
S: mapper/stratis-1-a1654aed04c14ec69e595aa151b32b76-thin-fs-f4529e5796b24c2ba8c69f64bab89ae4
S: disk/by-id/dm-name-stratis-1-a1654aed04c14ec69e595aa151b32b76-thin-fs-f4529e5796b24c2ba8c69f64bab89ae4
S: disk/by-id/dm-uuid-stratis-1-a1654aed04c14ec69e595aa151b32b76-thin-fs-f4529e5796b24c2ba8c69f64bab89ae4
E: DEVPATH=/devices/virtual/block/dm-7
E: DEVNAME=/dev/dm-7
E: DEVTYPE=disk
E: MAJOR=253
E: MINOR=7
E: SUBSYSTEM=block
E: USEC_INITIALIZED=198498677
E: DM_UDEV_PRIMARY_SOURCE_FLAG=1
E: DM_UDEV_RULES_VSN=2
E: DM_NAME=stratis-1-a1654aed04c14ec69e595aa151b32b76-thin-fs-f4529e5796b24c2ba8c69f64bab89ae4
E: DM_UUID=stratis-1-a1654aed04c14ec69e595aa151b32b76-thin-fs-f4529e5796b24c2ba8c69f64bab89ae4
E: DM_SUSPENDED=0
E: ID_FS_UUID=f4529e57-96b2-4c2b-a8c6-9f64bab89ae4
E: ID_FS_UUID_ENC=f4529e57-96b2-4c2b-a8c6-9f64bab89ae4
E: ID_FS_TYPE=xfs
E: ID_FS_USAGE=filesystem
E: E2_SIZE=0
E: DEVLINKS=/dev/disk/by-uuid/f4529e57-96b2-4c2b-a8c6-9f64bab89ae4 /dev/mapper/stratis-1-a1654aed04c14ec69e595aa151b32b76-thin-fs-f4529e5796b24c2ba8c69f64bab89ae4 /dev/disk/by-id/dm-name-stratis-1-a1654aed04c14ec69e595aa151b32b76-thin-fs-f4529e5796b24c2ba8c69f64bab89ae4 /dev/disk/by-id/dm-uuid-stratis-1-a1654aed04c14ec69e595aa151b32b76-thin-fs-f4529e5796b24c2ba8c69f64bab89ae4
E: TAGS=:systemd:
```

Cockpit wont show a block device as a filesystem unless it has `ID_FS_USAGE==filesystem`.

jbaublitz:
Closes #2162 

This PR makes some changes due to the added import statement not being triggered on `add` events without an explicit condition for that. This also removes the need for some code in `stratis_uuids_to_names`.

@bgurney-rh I'm requesting your approval because the name of the udev rule has changed to make the priority numeric prefix consistent with other storage rules. This will require some changes to packaging.

@drckeefe I would also ask that you take a look at the name change of the udev rule for packaging purposes. Let me know if you have any concerns.

Before we merge, I'd be interested to see if this fixes your issue, @mvollmer. I was able to reproduce your issue before the change, and the issue seems resolved after the change, but I'd like to make sure this is resolved on your system as well.

### DIFF
--- a/src/bin/stratis_uuids_to_names.rs
+++ b/src/bin/stratis_uuids_to_names.rs
@@ -145,16 +145,6 @@ fn main_report_error() -> Result<Option<(String, String)>, StratisUdevError> {
             ));
         }
     };
-    match args.next().as_deref() {
-        Some(action) => {
-            if action != "change" && action != "add" {
-                return Ok(None);
-            }
-        }
-        None => {
-            return Err(StratisUdevError::new("udev action required as argument."));
-        }
-    };
 
     let managed_objects = get_managed_objects()?;
 
--- a/udev/60-stratisd.rules
+++ b/udev/60-stratisd.rules
@@ -1,0 +1,1 @@
+ACTION="add|change", ENV{DM_NAME}=="stratis-1-[0-9a-f]*-thin-fs-[0-9a-f]*", PROGRAM="stratis_uuids_to_names $env{DM_NAME}", SYMLINK+="stratis/%c{1}/%c{2}", IMPORT{builtin}="blkid"
--- a/udev/99-stratisd.rules
+++ b/udev/99-stratisd.rules
@@ -1,1 +1,0 @@
-ENV{DM_NAME}=="stratis-1-[0-9a-f]*-thin-fs-[0-9a-f]*", PROGRAM="stratis_uuids_to_names $env{DM_NAME} $env{ACTION}", SYMLINK+="stratis/%c{1}/%c{2}"


mulkieran:
We'll wait for a response from @mvollmer.

mulkieran:
@jbaublitz Please rebase at your convenience.

mvollmer:
This fix doesn't immediately make sense to me. `blkid` is supposed to be called by the regular udev rules, and we should make sure that those rules run correctly instead of duplicating them, I'd say.

(I haven't tried your fix, so maybe I have to eat my words later, but... :-) )

For example, the usual `/dev/by-uuid/...` symlink is also missing for me when `ID_FS_USAGE` is missing, and while your fix here does set `ID_FS_USAGE`, I don't think it adds that symlink. (Nor should it.)

Normally, udev rules run when mkfs is done. The rules are triggered when a file descriptor is closed that was open for writing for the block device in question. You can simulate it like this:

```
# echo -n /dev/stratis/pool0/fsys
```

This is enough to trigger udev for me after `stratis fs create`.

So right now I am thinking that something prevents udev from running rules as necessary.  This might happen when the block device is locked with `flock`. If that's the case, the right thing would be to emit a artificial "change" event when the flock lock is removed.  See https://systemd.io/BLOCK_DEVICE_LOCKING/

That's all I have on this point, and we need to start talking to the udev developers/maintainers for more,

mulkieran:
@mvollmer Ok, thanks, we'll take another pass at this.

mvollmer:
I did a quick experiment by replacing `/usr/bin/mkfs.xfs` with a wrapper.

When adding a long `sleep` in front of it like so
```
#! /bin/sh
sleep 30
exec /usr/sbin/mkfs.xfs.orig "$@"
```
I get all the expected udev properties.

A shorter `sleep 1` is not enough.  It needs to be something like 4 or 5 seconds in my case.

When comparing the two scenarios, there is one more udev event in the case when I get all expected properties. This udev evetn seems to be blocked by something in the broken case.


I don't want to say what that all means. :-) I find the device mapper udev rules quite complicated, and they might well interfere.

jbaublitz:
@mvollmer See [this comment](https://github.com/stratis-storage/stratisd/issues/2175#issuecomment-674853804) for my theory on what is happening. I'm not sure if we're acquiring an exclusive `flock` which according to the documentation posted, I think we should be.

jbaublitz:
@mvollmer I tested out using `flock` and this did not resolve it. I'll dig into this more. It seems to be related to the rule that we added.

PR #2172, opened 2020-08-12
Add a script to generate outdatedness information about cargo dependencies

waltdisgrace:
Preliminary PR


waltdisgrace:
Rebased

.travis.yml
```diff
@@ -35,6 +35,9 @@ before_install:
   # remove due to system dependencies
   - sudo dpkg --purge --force-all libcryptsetup12
 
+install:
```
mulkieran, line 38: You only need requests to be available for the Python linting tasks. So you should add requests to the list of things that are installed for those tasks (rather than installing for every task). You should also specify the version, which should be whatever is the default on Fedora 31.

developer_tools/cargo/cargo.py
```diff
@@ -154,7 +154,10 @@ def main():
     print("\n\nNOW PRINTING KEY RESULTS\n")
 
     for key in cargo_outdated_output:
-        if key == "Name" or key == "----":
+        if key == "Name":
```
mulkieran, line 157: Try to follow the hint that was in the original lint:
```
cargo/cargo.py:157: [R1714(consider-using-in), main] Consider merging these comparisons with "in" to "key in ('Name', '----')"
```
The lint message suggests the preferred alternative.

waltdisgrace:
Rebased

mulkieran:
@waltdisgrace: It looks like, for recommended Python interpreter the requests version should be 2.22.0, for lowest supported Python interpreter it should be 2.20.0.

.travis.yml
```diff
@@ -98,6 +98,7 @@ jobs:
           dbus-python-client-gen==0.7
           psutil==5.6.7
           pyudev==0.22.0
+          requests==2.7.0
```
mulkieran, line 101: Might as well fix this up so it's the correct version number, see comment.

.travis.yml
```diff
@@ -124,6 +125,7 @@ jobs:
           dbus-python-client-gen==0.7
           psutil==5.4.3
           pyudev==0.22.0
+          requests==2.7.0
```
mulkieran, line 128: Fix version number here as well.

developer_tools/Makefile
```diff
@@ -2,14 +2,17 @@ PYTEST_OPTS = --verbose
...
 .PHONY: lint
 lint:
-	./check.py stratis_id 
+	./check.py stratis_id
+	./check.py cargo
 
 .PHONY: fmt
 fmt:
 	isort --recursive check.py stratis_id 
+	isort --recursive check.py cargo
```
mulkieran, line 11: For isorting, you should specify all the files and directories to be sorted on just one line.

developer_tools/Makefile
```diff
@@ -2,14 +2,17 @@ PYTEST_OPTS = --verbose
...
 	black .
 
 .PHONY: fmt-travis
 fmt-travis:
 	isort --recursive --diff --check-only check.py stratis_id 
+	isort --recursive --diff --check-only check.py cargo
```
mulkieran, line 17: Just use one line here, also.

developer_tools/cargo/cargo.py
```diff
@@ -0,0 +1,504 @@
...
+            if include:
+                not_found.append(key)
+            else:
+                not_included.append(key)
+
+    print("\n\nRESULTS")
+
+    print(
+        "\nThe following crates that were outputted by 'cargo outdated' are outdated"
+        + " with respect to the koji repo:"
```
mulkieran, line 466: There should be a way to get rid of this '+'. Try just removing it and leaving the rest the same and see if the program still runs.

developer_tools/cargo/cargo.py
```diff
@@ -0,0 +1,501 @@
...
+    command = ["cargo", "outdated"]
+    proc = subprocess.Popen(command, stdout=subprocess.PIPE)
+
+    pattern = r"([^\s]*)\s*([^\s]*)\s*([^\s]*)\s*([^\s]*)\s*([^\s]*)\s*(.*)"
+    my_reg_ex = re.compile(pattern)
+
+    while True:
+        line_bo = proc.stdout.readline()
+
+        if not line_bo:
```
mulkieran, line 307: Please avoid boolean overloading here and use an equality check instead. Does 'bo' stand for "boolean overloading"? Probably "line" is just as good as ```line_bo```.

developer_tools/cargo/cargo.py
```diff
@@ -0,0 +1,501 @@
...
+    command = ["rustc", "--print", "cfg"]
+    proc = subprocess.Popen(command, stdout=subprocess.PIPE)
+
+    rustc_cfg_dict = {}
+
+    pattern = r'target_([^\s]*)="([^\s]*)"'
+    my_reg_ex = re.compile(pattern)
+
+    while True:
+        line_bo_2 = proc.stdout.readline()
```
mulkieran, line 47: Same comments apply here as for ```line_bo``` above.

developer_tools/cargo/cargo.py
waltdisgrace: 'bo' stands for byte object. `readline()` returns an object of the class 'bytes', and I later call `decode("utf-8")` in order to convert it into a string which I can perform regular expression on.

I'm not sure what you mean by "boolean overloading" or how I would use an equality check here; do you mind explaining further?

Renaming line_bo to line is reasonable.

developer_tools/cargo/cargo.py
```diff
@@ -0,0 +1,504 @@
...
+        if cargo_outdated_match.group(1) in ("Name", "----"):
+            continue
+
+        platform = cargo_outdated_match.group(6)
+        include = parse_platform(platform)
+
+        dependencies = cargo_outdated_match.group(1)
+        version = cargo_outdated_match.group(2)
+
+        if "->" not in dependencies:
```
mulkieran, line 325: You might as well take advantage of the semantics of ```split()``` here.

```
>>> "bindgen->proc-macro2".split("->")
['bindgen', 'proc-macro2']
>>> "bindgen".split("->")
['bindgen']
```

Knowing this makes the check for whether "->" is contained in the string unnecessary and you can be a lot more succinct in the code block. Something like:

```
dependencies_split = dependencies.split("->")
dependency = dependencies.pop(-1)
pulled_in_by = None if dependencies_split == [] else dependencies_split.pop(-1)

# add  new item to cargo_outated_dict
```

developer_tools/cargo/cargo.py
```diff
@@ -0,0 +1,562 @@
...
+        dependencies_split = dependencies.split("->")
+
+        dependency = dependencies_split.pop(-1)
+
+        version = cargo_outdated_match.group(2)
+        pulled_in_by = None if dependencies_split == [] else dependencies_split[0]
+        platform = cargo_outdated_match.group(6)
+        include = parse_platform(platform)
+
+        cargo_outdated_dict[dependency] = (
```
mulkieran, line 376: There is a difficulty here, because the ```dependency``` key is not necessarily unique. A dependency may be pulled in by multiple different packages and that dependency _may_ have different versions each time.

developer_tools/cargo/cargo.py
```diff
@@ -0,0 +1,562 @@
...
+            version,
+            pulled_in_by,
```
mulkieran, line 378: I'm not sure that the value of ```pulled_in_by``` is used anywhere in the source. If it is not, then it is better not to store it.

waltdisgrace: I just added a commit that utilizes pulled_in_by.

waltdisgrace: Upon further pondering, my newly added commit needs additional work since I intended for it to categorize dependencies which are exclusively pulled in by dependencies with irrelevant platforms, but I did not implement the "exclusive" part of that.

developer_tools/cargo/cargo.py
waltdisgrace: This is true. The keys could be string-manipulated to include unique identifiers, perhaps. For example, "dependency-$-_i_", where _i_ is a natural number. @jbaublitz, what do you think of this idea?

mulkieran:
@waltdisgrace: Please start getting ready to put up a final PR. It will be most convenient for us to merge the work as is so far, as the script is for our internal use, and we won't be packaging it downstream.

PR #2178, opened 2020-08-18
Fix for udev rule priority and switch to IMPORT

### LINKED ISSUE #2175: Follow-up work on udev rule for filesystems
See comment: https://github.com/stratis-storage/stratisd/pull/2166#issuecomment-674833563

jbaublitz:
Closes #2175 

@mvollmer If you could give this a test, that would be great. My initial tests show that this works with both our symlink and all of the filesystem attributes without reimplementing anything.

@mulkieran I have switched to using import because in the udev documentation, it mentions that `PROGRAM` is usually used for matching devices so we would be doing something non-standard. `IMPORT`, however, gets used the way it is being used here quite frequently from what I've seen of other udev rules and this seems to be a more standard way to populate the udev entry with additional information. Beyond that, this solution seems just generally less fragile. Take a look and let me know what you think.

@bgurney-rh I've requested your review because I had to change the udev rule priority again. Sorry for the confusion!

### DIFF
--- a/udev/11-stratisd.rules
+++ b/udev/11-stratisd.rules
@@ -1,0 +1,1 @@
+ACTION=="add|change", ENV{DM_NAME}=="stratis-1-[0-9a-f]*-thin-fs-[0-9a-f]*", IMPORT{program}="stratis_uuids_to_names $env{DM_NAME}", SYMLINK+="$env{STRATIS_SYMLINK}"
--- a/src/bin/stratis_uuids_to_names.rs
+++ b/src/bin/stratis_uuids_to_names.rs
@@ -173,7 +173,7 @@ fn main() -> Result<(), StratisUdevError> {
             )
             .map_err(StratisUdevError::new)?;
 
-            println!("{} {}", pool_name, fs_name);
+            println!("STRATIS_SYMLINK=stratis/{}/{}", pool_name, fs_name);
             Ok(())
         }
         Ok(None) => Ok(()),
--- a/udev/60-stratisd.rules
+++ b/udev/60-stratisd.rules
@@ -1,1 +1,0 @@
-ACTION=="add|change", ENV{DM_NAME}=="stratis-1-[0-9a-f]*-thin-fs-[0-9a-f]*", PROGRAM="stratis_uuids_to_names $env{DM_NAME}", SYMLINK+="stratis/%c{1}/%c{2}", IMPORT{builtin}="blkid"


mvollmer:
With this rule, I do get all expected `ID_FS_*` udev properties immediately after `stratis fs create`.

However, the `/dev/stratis/` symlinks are missing after a reboot, presumably because `stratis_uuid_to_names` can't contact `stratisd` via D-Bus early enough... :-(

mulkieran:
@mvollmer Hmm, that seems likely to me. We'll take another look.

mulkieran:
Perhaps there is a possibility for a fallback in stratis-min which does not depend on the D-Bus at all. Or possibly make that a primary mechanism, even.

mvollmer:
Ok, I think I have figured this all out.  I just can't leave udev race conditions alone... :-) Let me drop my findings here; feel free to copy them to better places, of course.

Whether or not the ID_FS_* properties have their expected values depends on subtle timing variations, and moving the rules around and changing slightly what they do is enough to change the outcome.

It so happens that mkfs.xfs and udev always run in parallel and uncoordinated during "stratis fs create" and it all comes down to whether blkid runs before mkfs.xfs has written the superblock or after.

There is supposed to be coordination, but it fails here for the following reasons:

- mkfs.xfs relies on udevd to synthesize a "change" event when the block device changes (this works via inotify). However, udevd doesn't watch a block device while processing rules. It doesn't queue these events for later, it just ignores them completely. 

- One should be using flock to avoid writing to block devices while rules are running for them, but this doesn't work here because udevd doesn't take any locks for device mapper devices, for whatever reason.

Thus, the thing we should do is trigger our own "change" event right after mkfs.xfs is done, by writing "change" to /sys/block/dm-X/uevent`.

jbaublitz:
@mvollmer I reproduced the issue and it appears from the logs that D-Bus is not up yet. You were correct with your first guess. Please see the edits I've made to the systemd service file and the udev rule. On my machine, these changes result in the symlinks being created reliably on reboot.

udev/11-stratisd.rules
```diff
@@ -0,0 +1 @@
+ACTION=="add|change", ENV{DM_NAME}=="stratis-1-[0-9a-f]*-thin-fs-[0-9a-f]*", IMPORT{program}="stratis_uuids_to_names $env{DM_NAME}", SYMLINK+="$env{STRATIS_SYMLINK}", ENV{SYSTEMD_WANTS}+="stratisd.service"
```
mvollmer, line 1: `SYSTEMD_WANTS` doesn't do what I think you think it does. :-) It does not influence at all when the udev rule runs.

But I don't think you need to order the rule: The `ENV{DM_NAME}=="stratis-1-[0-9a-f]*-thin-fs-[0-9a-f]*"` will only match for devices that have been created by stratisd, so the reminder of the rule will only execute when stratisd is running (or has already been running at least once if it ever exits after a idle period, which I think it doesn't, right?).
The remaining thing then is to make sure that `stratis_uuids_to_names` can actually communicate with stratisd, maybe via D-Bus, maybe via something else.

stratisd.service
```diff
@@ -1,6 +1,7 @@
 [Unit]
 Description=Stratis daemon
 Documentation=man:stratisd(8)
+After=dbus.socket
```
mvollmer, line 4: This unfortunately leads to a dependency cycle:

```
Aug 20 12:13:38 dev systemd[1]: local-fs.target: Found ordering cycle on local-fs-pre.target/start
Aug 20 12:13:38 dev systemd[1]: local-fs.target: Found dependency on stratisd.service/start
Aug 20 12:13:38 dev systemd[1]: local-fs.target: Found dependency on dbus.socket/start
Aug 20 12:13:38 dev systemd[1]: local-fs.target: Found dependency on sysinit.target/start
Aug 20 12:13:38 dev systemd[1]: local-fs.target: Found dependency on ldconfig.service/start
Aug 20 12:13:38 dev systemd[1]: local-fs.target: Found dependency on local-fs.target/start
Aug 20 12:13:38 dev systemd[1]: local-fs.target: Job local-fs-pre.target/start deleted to break ordering cycle starting with local-fs.target/start
```

jbaublitz: I was warned by @mulkieran that this might not be the way to go. I'm reevaluating this piece.

udev/11-stratisd.rules
jbaublitz: I'm reading the documentation in more depth, and thinking over some possible solutions. My initial reaction is that providing a systemd unit file for our D-Bus service name of type `dbus` may help here. I'm still doing research about what action that unit file should trigger, but it may work to just trigger a synthetic change event for all of the Stratis filesystem devices once the D-Bus service comes up fully. Do you have any initial reactions to this?

jbaublitz: This may present some problems as the `dbus` type imposes a requirement on `dbus.socket` which you mentioned causes a dependency cycle below.

jbaublitz: I'm going to test if I can create a unit file simply for the D-Bus service separate from the stratisd service file that depends on `stratisd.service` and see how systemd reacts to that.

mvollmer: >  I'm still doing research about what action that unit file should trigger, but it may work to just trigger a synthetic change event for all of the Stratis filesystem devices once the D-Bus service comes up fully. Do you have any initial reactions to this?

The goal is to allow entries in /etc/fstab for Stratis filesystems with the "auto" option, right?  That's why stratisd.service is `Before=local-fs-pre.target´. If the symlinks haven't been created by that time, people can't use them in /etc/fstab. That would be surprising, I think.

jbaublitz: @mvollmer We have addressed that problem without stratisd. I've created a minimal Rust CLI using the daemon code to be used in contexts that don't have D-Bus available like initramfs.

PR #2180, opened 2020-08-18
Clean up FetchProperties signal handling and add LockedPools

### LINKED ISSUE #2179: Expose key descriptions for encrypted, locked pools
Currently if a pool is locked, there is no way to programmatically determine what the key description is that needs to be set to unlock the pool. Based on @mvollmer's feedback for Cockpit integration and related issues that have come up in #2134, it would be helpful to determine what key description needs to be set simply by pool UUID which we know prior to the pool being unlocked. This would allow scripts to determine what key description to set so that the user would only need to remember the passphrase, not the key description too.

The solution to this seems to be modifying the `LockedPoolUuids` key to be `LockedPools` and change the value to be a dictionary instead of a list and include a mapping of UUIDs to key descriptions in the next minor version D-Bus interface bump.

jbaublitz:
Closes #2179


src/engine/strat_engine/liminal.rs
```diff
@@ -651,6 +651,25 @@ impl LiminalDevices {
             .collect()
     }
 
+    /// Get a mapping of pool UUIDs from all of the LUKS2 devices that are currently
+    /// locked to their key descriptions in the set of pools that are not yet set up.
+    // Precondition: All devices for a given errored pool have been determined to have
+    // the same key description.
+    pub fn locked_pools(&self) -> HashMap<PoolUuid, KeyDescription> {
```
mulkieran, line 658: Right here, I think this deviates from the logic of ```locked_pool_uuids()``` and I think it shouldn't, because if any devices in the set are still unlocked we need to consider the thing a "locked pool". Also, if you could make ```locked_pool_uuids``` depend on this method perhaps, since I think it should just represent the keys from the result of this method OR get rid of it altogether and rewrite StratEngine's ```locked_pool_uuids()``` method just to get the keys that would tidy things up a bit.

jbaublitz: Okay this should be resolved!

src/dbus_api/filesystem/mod.rs
```diff
@@ -84,8 +89,6 @@ pub fn get_initial_properties(
             consts::FILESYSTEM_DEVNODE_PROP => shared::fs_devnode_prop(fs, pool_name, fs_name),
             consts::FILESYSTEM_POOL_PROP => parent,
             consts::FILESYSTEM_CREATED_PROP => shared::fs_created_prop(fs)
-        },
-        consts::PROPERTY_FETCH_INTERFACE_NAME => {},
-        consts::PROPERTY_FETCH_INTERFACE_NAME_2_1 => {}
+        }
```
mulkieran, line 92: Are you sure about these removals? I know they have no properties, so it's pointless to include these, but I think we follow the D-Bus spec more closely by leaving them in. They show up in the result of GetManagedObjects(), so I think it's the right thing to do.

mulkieran: A related question. Are we also supposed to send and InterfacesAdded signal when we create the root object?

jbaublitz: > Are you sure about these removals? I know they have no properties, so it's pointless to include these, but I think we follow the D-Bus spec more closely by leaving them in. They show up in the result of GetManagedObjects(), so I think it's the right thing to do.

I moved these to be autogenerated in the macro because they will never have any D-Bus properties associated with them. They are still sent, but I moved them so that we don't have to change it in multiple places every time we add a new FetchProperties interface version.

jbaublitz: > A related question. Are we also supposed to send and InterfacesAdded signal when we create the root object?

That's a very good question. I'm not entirely sure because technically it is a D-Bus object which, to me, would indicate yes. However, we will never be able to add or remove the top object during execution, only at start up and tear down, so I'm not sure how useful it is to send a signal when the D-Bus connection is just starting or shutting down. @mvollmer Are you familiar with how common/necessary this sort of thing is?

src/engine/strat_engine/liminal.rs
```diff
@@ -629,24 +629,25 @@ impl LiminalDevices {
...
         self.errored_pool_devices
             .iter()
             .filter_map(|(pool_uuid, map)| {
-                let has_locked = map.iter().any(|(_, info)| {
-                    if let LInfo::Luks(_) = info {
-                        true
+                map.iter().next().and_then(|(_, info)| {
```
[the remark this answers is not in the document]
jbaublitz: Based on our discussion yesterday where we talked about the precondition, I'm not sure what you mean by the devices can have different key descriptions if they're open. We never allow different key descriptions in the LUKS2 metadata whether the devices are open or not if the encrypted devices are created through stratisd so I'll need some clarification around why the precondition doesn't hold. If the encrypted device is created outside of stratisd, it will not be registered as any kind of Stratis device until it has been unlocked and once it is unlocked, it would be registered as an unencrypted device simply because we are not managing the LUKS setup in that case. Based on this, I'm under the impression that the precondition does hold and that we can simply check the first device.

src/dbus_api/filesystem/mod.rs
mulkieran: > > Are you sure about these removals? I know they have no properties, so it's pointless to include these, but I think we follow the D-Bus spec more closely by leaving them in. They show up in the result of GetManagedObjects(), so I think it's the right thing to do.
> 
> I moved these to be autogenerated in the macro because they will never have any D-Bus properties associated with them. They are still sent, but I moved them so that we don't have to change it in multiple places every time we add a new FetchProperties interface version.

Aaah, I see it now. But it was a bit confusing, so I think this change should be moved out to a separate PR. I'll do that myself, so you can focus on your current udev experiments.

jbaublitz: Resolved.

jbaublitz:
Rebased.

src/engine/strat_engine/liminal.rs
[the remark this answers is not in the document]
mulkieran: At present, the multiple key descriptions for the same pool UUUID could not be accomplished w/out manually editing the LUKS metadata. But @bgurney-rh edits metadata all the time, so it is doable.

My concern is really consistency w/in the code base. Our code should not implicitly assert that this is possible and implicitly assert that it is impossible at the same time. But I'm willing to remove the existing check in setup_pool for this unlikely event, and that will make me happier w/ not checking for it here.
See: https://github.com/stratis-storage/stratisd/pull/2182, commit https://github.com/stratis-storage/stratisd/pull/2182/commits/92c3d9ef4bce0be6adc3781833e7c8c4fdcb6b41.

Please go ahead and resolve the first concern, about the possibility that _some_ but not all of the devices belonging to the pool may be locked, which is easy to achieve by unlocking and then bringing another device online, but also feel free to assume that there is only one key description among all the devices. I'll put https://github.com/stratis-storage/stratisd/pull/2182 up for review.

src/dbus_api/filesystem/mod.rs
mvollmer: > > A related question. Are we also supposed to send and InterfacesAdded signal when we create the root object?
> 
> That's a very good question. I'm not entirely sure because technically it is a D-Bus object which, to me, would indicate yes. However, we will never be able to add or remove the top object during execution, only at start up and tear down, so I'm not sure how useful it is to send a signal when the D-Bus connection is just starting or shutting down. @mvollmer Are you familiar with how common/necessary this sort of thing is?

Is `/org/storage/stratis2` the root object?

You should not include the object that implements o.fd.Dbus.ObjectManager in the output of GetManagedObjects, nor should you send InterfacesAdded or InterfacesRemoved for it.

mulkieran: Ok! We'll make sure it's not in the output of GetManagedObjects, and if it is, we'll deal w/ that.

mulkieran: @mvollmer It is in the output of GetManagedObjects, because that's how dbus-rs implements it: https://github.com/stratis-storage/project/issues/216. Can you point us at a part of the D-Bus specification that indicates that this is incorrect?

mvollmer: > @mvollmer It is in the output of GetManagedObjects, because that's how dbus-rs implements it: [stratis-storage/project#216](https://github.com/stratis-storage/project/issues/216). Can you point us at a part of the D-Bus specification that indicates that this is incorrect?

Oh! Here is the spec: https://dbus.freedesktop.org/doc/dbus-specification.html#standard-interfaces-objectmanager

```
All returned object paths are children of the object path implementing this interface, i.e. their object paths start with the ObjectManager's object path plus '/'. 
```

mvollmer: I am not aware of any actual harm that comes from including the root in the output of GetManagedObjects, but I can imagine some endless loops if a client tries to handle nested ObjectManagers. But I don't think anybody nests ObjectManagers.

mulkieran: > I am not aware of any actual harm that comes from including the root in the output of GetManagedObjects, but I can imagine some endless loops if a client tries to handle nested ObjectManagers. But I don't think anybody nests ObjectManagers.

Ok. We won't make it a high priority, but we will talk to the dbus-rs developer about it.

PR #2186, opened 2020-08-21
Fix for udev rule priority and switch to IMPORT

### LINKED ISSUE #2175: Follow-up work on udev rule for filesystems

jbaublitz:
Supersedes #2178 
Closes #2175

### DIFF
--- a/src/bin/stratis_uuids_to_names.rs
+++ b/src/bin/stratis_uuids_to_names.rs
@@ -173,7 +173,7 @@ fn main() -> Result<(), StratisUdevError> {
             )
             .map_err(StratisUdevError::new)?;
 
-            println!("{} {}", pool_name, fs_name);
+            println!("STRATIS_SYMLINK=stratis/{}/{}", pool_name, fs_name);
             Ok(())
         }
         Ok(None) => Ok(()),
--- a/stratisd.service
+++ b/stratisd.service
@@ -1,7 +1,7 @@
 [Unit]
 Description=Stratis daemon
 Documentation=man:stratisd(8)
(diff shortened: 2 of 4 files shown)


mvollmer:
So now stratisd.service is `After=dbus.socket`.  This means Stratis filesystems can not be in /etc/fstab without extra magic, right?  (Extra magic: something like the x-systemd.requires=stratisd.service.  I experiment a bit with that.)

mvollmer:
> So now stratisd.service is `After=dbus.socket`. This means Stratis filesystems can not be in /etc/fstab without extra magic, right? (Extra magic: something like the x-systemd.requires=stratisd.service. I experiment a bit with that.)

Yes, I could confirm it: If you put a Stratis filesystem into /etc/fstab, the next boot will stop for a minute or so while waiting for the Stratis block device to appear and then continue without ever mounting it.

This can be worked around by putting `_netdev,x-systemd.requires=stratisd.service` into the mount options, but that's ugly, no?

What about giving up on using D-Bus in stratis_uuids_to_names? Maybe stratisd can maintain a trivial mapping in /run/stratisd?

jbaublitz:
@mvollmer We've addressed this by making stratis-min compatible with the initramfs. stratis-min does not require D-Bus and allows users to unlock pools which wouldn't be possible with D-Bus for encrypted devices in /etc/fstab which would be a very bad solution. Encryption requires some level of interaction so we need some way to perform operations without D-Bus.

PR #2200, opened 2020-09-01
Add PhysicalPath property

### LINKED ISSUE #2191: Consider exposing the physical path for a block device on the D-Bus
~~We can do this via the FetchProperties interface.~~ Actually, it seems like it makes more sense to expose it as a D-Bus property: it is not exactly volatile and it is a useful identifier of a blockdev, i.e., you might want to find the blockdev with the specified physical path, and so it is better to export it via GetManagedObjects, and so forth.

mulkieran:
Closes #2191


mulkieran:
I've decided to designate the property Const, because that makes it consistent with the Devnode property.

jbaublitz:
@mulkieran The way that I wrote up the documentation specifies that interface should be `2_3` and `r3` when we release this. The version number should be the minor version of the release in which it was introduced. We should either make this consistent with the documentation we've already written or update the documentation to specify the versioning scheme you've introduced here.

jbaublitz:
Whoops! Typed the wrong version. I'll take a look at this.

src/dbus_api/blockdev/mod.rs
```diff
@@ -50,13 +51,25 @@ pub fn create_dbus_blockdev<'a>(
...
+                .add_p(blockdev_2_2::physical_path_property(&f)),
+        )
         .add(
             f.interface(consts::PROPERTY_FETCH_INTERFACE_NAME, ())
                 .add_m(fetch_properties_2_0::get_all_properties_method(&f))
                 .add_m(fetch_properties_2_0::get_properties_method(&f)),
         )
         .add(
-            f.interface(consts::PROPERTY_FETCH_INTERFACE_NAME_2_1, ())
+            f.interface(consts::PROPERTY_FETCH_INTERFACE_NAME_2_2, ())
```
jbaublitz, line 72: I don't think this is supposed to be bumped because usually we do this for this interface across all types at once when something changes in the `FetchProperties` implementation for one type. I don't think you've changed anything here so this should stay at the same version.

src/dbus_api/blockdev/mod.rs
```diff
@@ -89,9 +102,19 @@ pub fn get_initial_properties(
             consts::BLOCKDEV_HARDWARE_INFO_PROP => shared::blockdev_hardware_info_prop(dev),
             consts::BLOCKDEV_USER_INFO_PROP => shared::blockdev_user_info_prop(dev),
             consts::BLOCKDEV_INIT_TIME_PROP => shared::blockdev_init_time_prop(dev),
-            consts::BLOCKDEV_POOL_PROP => parent,
+            consts::BLOCKDEV_POOL_PROP => parent.to_owned(),
```
jbaublitz, line 105: Can you give me a little more information about this change?

src/dbus_api/blockdev/mod.rs
mulkieran: Nope. Why did it compile?

src/dbus_api/blockdev/mod.rs
mulkieran: Didn't compile w/out it. I didn't look closely into the macro to see why.

src/dbus_api/blockdev/mod.rs
jbaublitz: I don't think this would break anything at compile time. It would break our CLI though.

src/dbus_api/blockdev/mod.rs
jbaublitz: Ok!

src/dbus_api/blockdev/mod.rs
mulkieran: Aaah.

src/dbus_api/blockdev/mod.rs
mulkieran: Yeah, there's two vecs, one for each interface, each evaluated in the order that it is defined. And so ```parent``` can't be shared. I don't know that there's a prettier way to deal with this.

mulkieran:
rebased.

mvollmer (approved):
This works nicely, thanks!

mulkieran:
Superceded by #2209

PR #2206, opened 2020-09-02
Improve write support

mulkieran:
Put write support into a separate module.

Compile only for test implementations that are only for test.

Flush the bufwriter before syncing the underlying implementation. If the data has not been flushed, syncing the underlying implementation ought not to have the intended effect. This fixes a bug, where we could not be confident that the start of a device was appropriately wiped, which could cause a failure to create a thinpool if the meta device had non-zero bytes in its first eight sectors.

Do not allocate more space in the buffer than can possibly be used. For the places where it is used in our actual source the vector was 256 times as big as it needed to be, which is excessive.


mulkieran:
rebased.

jbaublitz (approved):

PR #2209, opened 2020-09-04
Add PhysicalPath D-Bus property

### LINKED ISSUE #2191: Consider exposing the physical path for a block device on the D-Bus

mulkieran:
Closes #2191

Supercedes #2200


mulkieran:
Identical to previous, merging.

PR #2211, opened 2020-09-04
Prototype of Clevis integration

jbaublitz:
Closes stratis-storage/project#192

The initial commit is a prototype of handling keys from the kernel keyring securely in memory for operations that do not integrate with the kernel keyring and require a keyfile.

@mulkieran This is not ready for review, but if you'd like to take a look, I'm putting this up so that you can.


mulkieran (commented):
This seems to me to make sense so far.

jbaublitz:
TODO: Fix the unshare path to be the mounted tmpfs inside `/run`, not `/run` itself.

PR #2216, opened 2020-09-08
Change clippy task to stable as allowed failure

bgurney-rh:
Related: stratis-storage/project#224

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,8 @@ jobs:
   allow_failures:
     # Allow audit task to fail
     - env: TASK=audit
+    # Allow clippy task to fail
+    - env: TASK=clippy
   include:
 
     # MANDATORY CHECKS USING CURRENT DEVELOPMENT COMPILER
@@ -48,11 +50,6 @@ jobs:
       before_script:
         - rustup component add rustfmt
       env: TASK=fmt-travis
-    - name: "lint Rust source using current development toolchain"
-      rust: 1.46.0
-      before_script:
-        - rustup component add clippy
-      env: TASK=clippy
 
     # MANDATORY TESTING ON STABLE
     - name: "build using stable toolchain"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -48,11 +50,6 @@ jobs:
       before_script:
         - rustup component add rustfmt
       env: TASK=fmt-travis
-    - name: "lint Rust source using current development toolchain"
```
mulkieran, line 51: We want to keep this...this and stable will still diverge w/ every update.

.travis.yml
```diff
@@ -141,8 +138,8 @@ jobs:
     # INTERMITTENTLY ALLOWED FAILURES
```
mulkieran, line 138: You'll want to restructure the file a bit, so that it doesn't even mention intermittently allowed failures, since after this change there won't be any.

mulkieran (changes requested):

.travis.yml
bgurney-rh: I don't know if I follow what you mean.  Do we still want two lint tasks, but have them be the same until the update occurs, where one follows to the new stable version?

(Also, the task wasn't removed, it was just moved down to the "intermittently allowed" section, which I suppose will fold into the "allowed failures" section.)

bgurney-rh: Also, there's no `env: TASK=clippy` task for stable.  Should there be?

bgurney-rh: I did a little bit of restructuring; how does it look now?

mulkieran (changes requested):
@bgurney-rh Now both clippy tasks are allowed failures; we only want that to be true for the stable clippy task.

.travis.yml
```diff
@@ -155,6 +144,18 @@ jobs:
       rust: stable
       env: TASK=audit
 
+    - name: "lint Rust source using stable toolchain"
+      rust: stable
+      before_script:
+        - rustup component add clippy
+      env: TASK=clippy
+
+    - name: "lint Rust source using current development toolchain"
```
mulkieran, line 153: Please move this one back up to where it came from.

mulkieran (changes requested):

.travis.yml
bgurney-rh: Resolved.

bgurney-rh:
Rebased.

mulkieran (changes requested):
This looks good to me. Please close and open a final PR, also follow up w/ PRs for devicemapper and libcryptsetup.

PR #2217, opened 2020-09-08
Send PropertiesChanged signals for multiple interfaces

### LINKED ISSUE #2215: PropertiesChanged signal is not emitted for org.storage.stratis2.pool.r1
When calling SetName for a pool, there is a corresponding PropertiesChanged signal on D-Bus, but only for org.storage.stratis2.pool, not for org.storage.stratis2.pool.r1.

```
‣ Type=method_call  Endian=l  Flags=4  Version=1  Priority=0 Cookie=148
  Sender=:1.237  Destination=org.storage.stratis2  Path=/org/storage/stratis2/1  Interface=org.storage.stratis2.pool.r1  Member=SetName
  UniqueName=:1.237
  MESSAGE "s" {
          STRING "pool3";
  };

‣ Type=signal  Endian=l  Flags=1  Version=1  Priority=0 Cookie=649
  Sender=:1.3  Path=/org/storage/stratis2/1  Interface=org.freedesktop.DBus.Properties  Member=PropertiesChanged
  UniqueName=:1.3
  MESSAGE "sa{sv}as" {
          STRING "org.storage.stratis2.pool";
          ARRAY "{sv}" {
                  DICT_ENTRY "sv" {
                          STRING "Name";
                          VARIANT "s" {
                                  STRING "pool3";
                          };
                  };
          };
          ARRAY "s" {
          };
  };

‣ Type=method_return  Endian=l  Flags=1  Version=1  Priority=0 Cookie=650  ReplyCookie=148
  Sender=:1.3  Destination=:1.237
  UniqueName=:1.3
  MESSAGE "(bs)qs" {
          STRUCT "bs" {
                  BOOLEAN true;
                  STRING "64b4b34442ee45939e29a372ce5ae370";
          };
          UINT16 0;
          STRING "Ok";
  };
```

mulkieran:
Closes #2215 

The first commit is simply a refactoring and is not essential for the correctness of the second commit. The two commits are so closely related that I'm keeping them in the same PR.

### DIFF
--- a/src/dbus_api/consts.rs
+++ b/src/dbus_api/consts.rs
@@ -64,32 +64,50 @@ pub fn fetch_properties_interfaces() -> Vec<String> {
     .collect()
 }
 
-/// Get a list of all interfaces supported by a pool object.
-pub fn pool_interface_list() -> InterfacesRemoved {
-    let mut interfaces = [POOL_INTERFACE_NAME, POOL_INTERFACE_NAME_2_1]
+// Get a list of all the standard pool interfaces; i.e., all the revisions of
+// org.storage.stratis2.pool.
+fn standard_pool_interfaces() -> Vec<String> {
+    [POOL_INTERFACE_NAME, POOL_INTERFACE_NAME_2_1]
         .iter()
         .map(|s| (*s).to_string())
-        .collect::<InterfacesRemoved>();
+        .collect()
+}
+
+// Get a list of all the standard filesystem interfaces; i.e., all the
+// revisions of org.storage.stratis2.filesystem.
+fn standard_filesystem_interfaces() -> Vec<String> {
+    [FILESYSTEM_INTERFACE_NAME]
+        .iter()
+        .map(|s| (*s).to_string())
+        .collect()
+}
+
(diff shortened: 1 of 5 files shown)


mulkieran:
Tested. Now shows signals on both interfaces.

jbaublitz (changes requested):
Wouldn't it make more sense to use the refactoring you did here and call into `standard_pool_interfaces` from `signal_on_name_change_pool_interfaces`, etc.? I don't see a downside here.

mulkieran:
> Wouldn't it make more sense to use the refactoring you did here and call into `standard_pool_interfaces` from `signal_on_name_change_pool_interfaces`, etc.? I don't see a downside here.

If we turn on sending signals for a property later, which is actually possible, that approach would be a little bit broken (that was my first idea).

mulkieran:
@jbaublitz Back in review.

jbaublitz (approved):

mvollmer (approved):
Works, thanks!

mulkieran:
Superceded by #2221

PR #2204, opened 2020-09-02
Establish new raw_devices module

mulkieran:
This PR does the following:
 * move wipe_blockdevs() into blockdevmgr module
* splits backtore/devices module into devices and initialize module
* exports fewer names from the backstore module and makes some test only
* moves the identify module into the liminal module
* makes the crypt and udev modules submodules of strat_engine

PR #2218, opened 2020-09-08
Refactor initialization and discovery code

mulkieran:
Supercedes #2204


mulkieran:
It is gratifying that the stratisd non-rust test failed; this indicates that we encounter Change events on unowned devices.

mulkieran:
We have so many change events on devices that have not yet been added that we should add a different data structure just for caching them, somehow.

mulkieran:
Obsolete.

PR #2222, opened 2020-09-09
Remove beta clippy and add stable clippy as allowed failure

bgurney-rh:
Supersedes: #2216 
Related: stratis-storage/project#224

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,9 @@ jobs:
   allow_failures:
     # Allow audit task to fail
     - env: TASK=audit
+    # Allow stable clippy task to fail
+    - rust: stable
+      env: TASK=clippy
   include:
 
     # MANDATORY CHECKS USING CURRENT DEVELOPMENT COMPILER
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -155,6 +149,12 @@ jobs:
       rust: stable
       env: TASK=audit
 
+    - name: "lint Rust source using stable toolchain"
```
mulkieran, line 152: Whoops! I missed this previously.

Could you keep the comments from the former beta task? These are still true, and explain why this _is_ an allowed failure.

mulkieran (changes requested):

.travis.yml
bgurney-rh: I can, though I wonder if we may need to edit the comment to better fit the context:

```
    # Allowed if a failure occurs after a new Rust release until the
    # failure is addressed.
```

mulkieran: Maybe.

```
# Allowed because a failure may occur after any new Rust version is released.
```

bgurney-rh: Okay; I'm going to try a variant of that:
```
    # Allowed because a failure may occur after a new Rust stable
    # version is released.
```

...along with changing the `allow_failures` annotation to the following:
```
# Allow stable clippy task to fail
```

jbaublitz (approved):

mulkieran (approved):

mulkieran:
@bgurney-rh Please rebase, thx.

bgurney-rh:
Rebased.

PR #2223, opened 2020-09-09
Add a new filesystem interface

mulkieran:


mulkieran:
What about doing this via a DeferredAction? A new constructor called Change.

mulkieran:
@jbaublitz: I'm gonna give the idea in https://github.com/stratis-storage/stratisd/pull/2223#issuecomment-689908554 a try. Let me know what you think of it, thanks!

jbaublitz:
@mulkieran This seems like a reasonable way to handle this. I'll keep an eye on how this progresses, but no immediate objections. This looks very similar to what I had to do, so I don't think this will be a concern.

mulkieran:
Test failures most likely not spurious. Most likely a failing expect() in pool rename.

mulkieran:
@jbaublitz It's still a little rough, but all the parts are there. The most interesting bit is that all the properties changed signals are constructed in the pool rename method (because some information needs to be looked up in the engine). But I still think this is an improvement.

src/dbus_api/pool/shared.rs
```diff
@@ -197,3 +199,27 @@ pub fn pool_name_prop(name: &Name) -> String {
...
+    );
+    r0_properties.interface_name = consts::POOL_INTERFACE_NAME.into();
+    properties_changed.push(r0_properties);
+
+    let mut r1_properties = PropertiesPropertiesChanged::default();
+    r1_properties.changed_properties.insert(
+        consts::POOL_NAME_PROP.into(),
+        Variant(Box::new(pool_name_prop(pool_name))),
+    );
+    r1_properties.interface_name = consts::POOL_INTERFACE_NAME.into();
```
jbaublitz, line 221: Should this be `POOL_INTERFACE_NAME_2_1`?

jbaublitz (changes requested):

src/dbus_api/pool/shared.rs
mulkieran: Fixed. But there's also a further thing I'ld like to try, to move more processing into the ActionQueue implementation.

mulkieran:
Obsoleted by multi-threading work.

PR #2228, opened 2020-09-14
Returned LockedPoolUuids property in FetchProperties.r2

mulkieran:


jbaublitz:
@mulkieran I had removed that intentionally because users can easily get the locked pool UUIDs by iterating through only the keys of `LockedPools`. I don't think we made any kind of commitment that we'd carry forward all methods to the new interfaces, simply that the old interfaces would continue to work. What are your thoughts on this?

mulkieran:
> @mulkieran I had removed that intentionally because users can easily get the locked pool UUIDs by iterating through only the keys of `LockedPools`. I don't think we made any kind of commitment that we'd carry forward all methods to the new interfaces, simply that the old interfaces would continue to work. What are your thoughts on this?

They can use the new key to get the information that they got from the previous key, but I'ld still consider it API breakage. Right now, if the CLI starts using r2 of the FetchProperties interface, it will raise an unexpected exception. I'ld rather it be able to move up to the new revision w/out requiring code changes to fix the breakage.

jbaublitz (approved):
I'm happy to keep it if this is the process you want to standardize on.

mulkieran:
ok!

PR #2229, opened 2020-09-15
Set D-Bus timeout to maximum possible

### LINKED ISSUE #2220: stratis_uuids_to_names times out sometimes
When creating a snapshot of a filesystem, the symlink for the original filesystem disappears, but `udevadm trigger` brings it back:

```
# ls /dev/stratis/pool3
myfs6
# stratis fs snapshot pool3 myfs6 myfs6@now
# ls /dev/stratis/pool3
myfs6@now
# udevadm trigger
# ls /dev/stratis/pool3
myfs6 myfs6@now
```

I can find this in the journal:

```
Sep 09 13:19:34 dev stratis_uuids_t[13797]: Could not convert Stratis UUIDs to names. "Did not receive a reply. Possible causes include: the remote application did not send a reply, the message bus security policy blocked the reply, the reply timeout expired, or the network connection was broken."
```

stratisd seems to be slow to reply to D-Bus messages when it is busy with something else, such as creating a filesystem or taking a snapshot.

It is possible to disable the timeout in `stratis_uuids_to_names` (and I would recommend that as a rule), but there is another timeout for udev rules.

I think this is another reason to stop relying on D-Bus in udev rules.

mulkieran:
Closes #2220

### DIFF
--- a/src/bin/stratis_uuids_to_names.rs
+++ b/src/bin/stratis_uuids_to_names.rs
@@ -42,8 +42,10 @@ const STRATIS_MANAGER_OBJECT: &str = "/org/storage/stratis2";
 const STRATIS_POOL_IFACE: &str = "org.storage.stratis2.pool.r1";
 const STRATIS_FS_IFACE: &str = "org.storage.stratis2.filesystem";
 const DBUS_OM_IFACE: &str = "org.freedesktop.DBus.ObjectManager";
+
 lazy_static! {
-    static ref TIMEOUT: Duration = Duration::new(5, 0);
+    // The value chosen is the maximum timeout permitted for the Stratis CLI
+    static ref TIMEOUT: Duration = Duration::new(1_073_741, 823);
 }
 
 struct StratisUdevError(Option<String>);
--- a/src/bin/stratis_uuids_to_names.rs
+++ b/src/bin/stratis_uuids_to_names.rs
@@ -44,8 +44,11 @@ const STRATIS_FS_IFACE: &str = "org.storage.stratis2.filesystem";
 const DBUS_OM_IFACE: &str = "org.freedesktop.DBus.ObjectManager";
 
 lazy_static! {
-    // The value chosen is the maximum timeout permitted for the Stratis CLI
-    static ref TIMEOUT: Duration = Duration::new(1_073_741, 823);
+    // The value is 2**31 - 1 millisecond, the largest representable value
+    // in a 32-bit signed int. -1, which is often used to indicate something
+    // unbounded is interpreted by libdbus as a request for its default timeout,
+    // which we consider unacceptably small.
+    static ref TIMEOUT: Duration = Duration::from_millis(0x7fff_ffff);
 }
 
 struct StratisUdevError(Option<String>);
--- a/src/bin/stratis_uuids_to_names.rs
+++ b/src/bin/stratis_uuids_to_names.rs
@@ -14,6 +14,7 @@
 
 use std::{
     collections::HashMap,
+    convert::TryFrom,
     env::args,
     error::Error,
     fmt::{self, Debug, Display},
@@ -48,7 +49,8 @@ lazy_static! {
     // in a 32-bit signed int. -1, which is often used to indicate something
     // unbounded is interpreted by libdbus as a request for its default timeout,
     // which we consider unacceptably small.
-    static ref TIMEOUT: Duration = Duration::from_millis(0x7fff_ffff);
+    static ref TIMEOUT: Duration =
+        Duration::from_millis(TryFrom::try_from(std::i32::MAX).expect("statically verified"));
 }
 
 struct StratisUdevError(Option<String>);
--- a/src/bin/stratis_uuids_to_names.rs
+++ b/src/bin/stratis_uuids_to_names.rs
@@ -14,7 +14,7 @@
 
 use std::{
     collections::HashMap,
-    convert::TryFrom,
+    convert::TryInto,
     env::args,
     error::Error,
     fmt::{self, Debug, Display},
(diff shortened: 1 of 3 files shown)


src/bin/stratis_uuids_to_names.rs
```diff
@@ -42,8 +42,10 @@ const STRATIS_MANAGER_OBJECT: &str = "/org/storage/stratis2";
 const STRATIS_POOL_IFACE: &str = "org.storage.stratis2.pool.r1";
 const STRATIS_FS_IFACE: &str = "org.storage.stratis2.filesystem";
 const DBUS_OM_IFACE: &str = "org.freedesktop.DBus.ObjectManager";
+
 lazy_static! {
-    static ref TIMEOUT: Duration = Duration::new(5, 0);
+    // The value chosen is the maximum timeout permitted for the Stratis CLI
+    static ref TIMEOUT: Duration = Duration::new(1_073_741, 823);
```
mvollmer, line 48: This is a very magic number, and stratis-cli also doesn't explain it. In my experience, there is no maximum timeout with regard to D-Bus method calls and it is totally fine to just wait forever (or u64::MAX seconds).

Moreover, stratis-cli has ` maximum_dbus_timeout_ms = 1073741823`, but the second argument to `Duration::new` is nanoseconds. Thus, this should be `Duration::new(1_073_741, 823_000_000)`, no?

mvollmer:
I have tested this manually and the timeouts reported in #2220 don't occur anymore.

jbaublitz:
`Duration` does appear to take nanoseconds as the second parameter.

src/bin/stratis_uuids_to_names.rs
mulkieran: Probably hex makes it more obvious: 0x3fffffff = 1073741823, 2^30 - 1. The _reason_ the exponent was chosen to be 30 has much to do with the mechanics of floating point conversion; it was the largest integer value that could be guaranteed to be transmitted without its value being changed via numeric conversions through the Python bindings that we use in the CLI to the underlying C library w/out our doing sophisticated analysis of floating point conversions (which seemed a waste of time). But there is no floating point conversion in the Rust source! So we can do something simpler.

mulkieran: Looks like we can be direct and use the maximum value that can be stored in a 32 bit signed int.

mulkieran: We can't play it safe and use the maximum value that can be stored in a 16-bit unsigned int, that is way too small for practical purposes.

jbaublitz: @mulkieran Just to make sure I understand, the value `1073741823` is in milliseconds because that's what D-Bus ultimately takes if I remember correctly, right? If so, `823` as the 2nd parameter will not yield that value because that's 823 nanoseconds. This would result in 1,073,741.000000823 seconds or 1,073,741,000.000823 milliseconds. Am I missing something here?

mulkieran:
Ok, this is ready to go, again.

jbaublitz (approved):

src/bin/stratis_uuids_to_names.rs
mvollmer: Right, as we can see [here](https://docs.rs/dbus/0.8.4/src/dbus/channel.rs.html#269-279), the dbus crate uses `Duration::as_millis` to convert the timeout Duration object to the integer value used with libdbus.

libdbus treats i32::MAX as infinite as documented [here](https://dbus.freedesktop.org/doc/api/html/group__DBusPendingCall.html#ga21384c9c5b0da54f7d0a92012522f213). So I think this deserves a named constant in the dbus crate, but the second best thing would be a named constant in stratisd.

src/bin/stratis_uuids_to_names.rs
```diff
@@ -42,8 +43,14 @@ const STRATIS_MANAGER_OBJECT: &str = "/org/storage/stratis2";
...
 const DBUS_OM_IFACE: &str = "org.freedesktop.DBus.ObjectManager";
+
 lazy_static! {
-    static ref TIMEOUT: Duration = Duration::new(5, 0);
+    // The value is 2**31 - 1 millisecond, the largest representable value
+    // in a 32-bit signed int. -1, which is often used to indicate something
+    // unbounded is interpreted by libdbus as a request for its default timeout,
+    // which we consider unacceptably small.
+    static ref TIMEOUT: Duration =
+        Duration::from_millis(TryFrom::try_from(std::i32::MAX).expect("statically verified"));
```
mvollmer, line 53: `i32::MAX` does indeed mean infinite to libdbus, so I vote for naming this constant `DBUS_TIMEOUT_INFINITE`, to match libdbus. See https://dbus.freedesktop.org/doc/api/html/group__DBusPendingCall.html#ga21384c9c5b0da54f7d0a92012522f213.

(For my own education: I am surprised at all the ceremony; why not just `Duration::from_millis(std::i32::MAX)`?)

mulkieran: ```Duration::from_millis()```'s parameter type is u64, Rust's type system requires a conversion.

src/bin/stratis_uuids_to_names.rs
mulkieran: We should probably prefer to be good citizens and submit to the upstream crate...

mulkieran:
Not considering any upstream contributions...a Duration "constant" would requires use of lazy_static, which the dbus developer isn't using.

mulkieran:
Ready again.

src/bin/stratis_uuids_to_names.rs
```diff
@@ -27,6 +28,7 @@ use dbus::{
     Path,
 };
 use lazy_static::lazy_static;
+use libdbus_sys::DBUS_TIMEOUT_INFINITE;
```
mvollmer, line 31: Nice!

mvollmer (approved):

jbaublitz (approved):

mulkieran:
Superceded by #2235

PR #2231, opened 2020-09-15
Allow non-privileged communication all revisions of FetchProperties

mulkieran:


mulkieran:
Trivial, merging as soon as tests are passed.

PR #2233, opened 2020-09-16
Add mechanism for easily flock'ing devices

jbaublitz:
This is an exploratory PR for adding the equivalent of mutability locks on devices that are outside of Rust's mutability memory safety guarantees.


mulkieran:
Took a brief look...makes sense. I'm not sure how best to handle a failure to unlock; warning is fine for now.

jbaublitz:
@mulkieran Based on possible errors listed [here](https://linux.die.net/man/2/flock), I don't think any of those errors would be thrown for our case. We can guarantee we have a valid file descriptor, we won't need an allocation when we're unlocking (it's probably just a memory deallocation which usually has a `void` return type) and I don't think we'll have to worry about any signal handling. Overall this is more caution than anything, but I don't see it being a problem for unlocking. Locking is another story.

PR #2235, opened 2020-09-17
Set D-Bus timeout to infinite

### LINKED ISSUE #2220: stratis_uuids_to_names times out sometimes

mulkieran:
Closes #2220
Superces #2229

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -396,6 +396,7 @@ dependencies = [
  "lazy_static 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.68 (registry+https://github.com/rust-lang/crates.io-index)",
  "libcryptsetup-rs 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


mulkieran:
rebased.

PR #2236, opened 2020-09-18
Use Ubuntu focal distribution

mulkieran:


mulkieran:
Doesn't look like focal distribution can help us simplify anything.

PR #2238, opened 2020-09-21
No longer iteratively try to set up the D-Bus connection

### LINKED ISSUE #2198: Remove cycling in event-loop order to set up the D-Bus connection
If we stick w/ making stratisd service of type dbus this will be unnecessary. See https://github.com/stratis-storage/stratisd/pull/2197.

mulkieran:
Closes #2198

### DIFF
--- a/src/stratis/dbus_support.rs
+++ b/src/stratis/dbus_support.rs
@@ -11,7 +11,10 @@
 
 use std::{cell::RefCell, rc::Rc};
 
-use crate::engine::{Engine, Name, Pool, PoolUuid};
+use crate::{
+    engine::{Engine, Name, Pool, PoolUuid},
+    stratis::StratisResult,
+};
 
 #[cfg(feature = "dbus_enabled")]
 use crate::{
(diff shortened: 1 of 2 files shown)


mulkieran:
@bgurney-rh Would you be able to test this? What we want to ensure is that, when systemd starts stratisd, stratisd will always be started after enough other conditions have been satisfied that it will always be able to make a D-Bus connection on the first try. We believe that this is truebecause the systemd configuration file implicitly require that the dbus.socket service is available.

mulkieran:
Whoops, sorry. Missed a spot, fixing.

mulkieran:
One Travis test failed; the one that builds and tests with no defaults enabled. clippy didn't detect the problem because it runs with all features enabled, and so had compiled out the offending code.

mulkieran:
This is a follow up to #2197. If we feel confident that we've taken the correct path there, then we should feel able to go ahead with this PR.

jbaublitz (approved):

mulkieran:
We already know that the udev tests are passing, and that these stop and start the daemon multiple times and verify that the previously created pools are available on the D-Bus. We do not have test for stopping and starting in the blackbox tests, but we do have an issue for it: https://github.com/stratis-storage/stratis-cli/issues/557.

bgurney-rh (approved):
I built a test package, and it looks like it works.  I can see the following in the "systemd-analyze dump" output for stratisd.service:
```
After: dbus.socket (origin-file)
```

PR #2252, opened 2020-09-29
Flock

jbaublitz:
Supersedes #2233


jbaublitz:
I'm resurrecting this PR. @mulkieran and I had discussed where flock should be used. I'm going to flock:
* Flock physical devices when they're being read or altered.
* Flock devicemapper devices recursively when they're being read or altered. This will lock the devicemapper device and all devicemapper devices below it. This should provide us with plenty of safety as:
  * A write on a devicemapper device cannot be done if a devicemapper anywhere in the stack is being read.
  * A read on a devicemapper device cannot be done if a devicemapper anywhere in the stack is being written.
  * Reads can happen concurrently anywhere in the stack.

We may be able to fine tune the locking eventually given that mutating a device higher up on the stack may not affect reading a devicemapper device below, but I'd need to look into it to be sure. If this turns out to be the case, the write flock could be altered to not be recursive so that my flock data structure would only block if a read is attempted above a devicemapper device being written, not below.

@mulkieran Do you have any additional feedback on this? It seems like the two types of devices we'd need to flock are either physical devices or devicemapper devices so I see this as adequate as flock will only provide filesystem-based safety and we'll rely on Rust's mutability guarantees and locking mechanisms for the internal data structures.

jbaublitz:
Also here is a question I asked the tokio team about making flock asynchronous: https://github.com/tokio-rs/tokio/issues/3148.

PR #2262, opened 2020-10-05
Make handling of idempotent return values consistent

jbaublitz:
Related to stratis-storage/project#186.

We may have to change the CLI to handle this fix.


jbaublitz:
The test failure seems to be valid. Fixing now.

mulkieran (approved):

Issue #385, opened 2017-06-02
Replace &[&Path] parameters with more organized structure


mulkieran:
~~Why can't we just do ```&[Path]```. They are kind of cumbersome to deal with, and I'm not sure that the internal reference is really helpful.~~

We would be better off with a purpose built data structure that enforced some invariants, for example, that all the paths are absolute, not duplicates, etc.

agrover:
See https://doc.rust-lang.org/std/path/struct.Path.html

`Path` is an unsized type, like &str, so the alternative isn't `Path` it's `PathBuf`.

mulkieran:
Right. But maybe ```&[PathBuf]``` is really the right thing.

agrover:
`&[P] where P: AsRef<Path>`?

Kintaro:
That would be way more in line with idiomatic Rust.

mulkieran:
I think it is tricky.

tasleson:
I took a stab at this and didn't get very far, I tried taking

```rust
fn add_blockdevs(&mut self, paths: &[&Path], force: bool) -> EngineResult<Vec<DevUuid>>;
```
changing it to:
```rust
fn add_blockdevs<P>(&mut self, paths: &[P], force: bool) -> EngineResult<Vec<DevUuid>>
        where P: AsRef<Path>;
```
Which results in a compile error:
```
error[E0038]: the trait `engine::engine::Pool` cannot be made into an object
   --> src/engine/strat_engine/engine.rs:122:5
    |
122 |     fn pools(&self) -> Vec<&Pool> {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `engine::engine::Pool` cannot be made into an object
    |
    = note: method `add_blockdevs` has generic type parameters
```

I'm pretty new to rust, so I maybe I'm not taking the correct approach as suggested in the above comment by @agrover

mulkieran:
Yeah, I removed the trivial label for a reason. I'm not sure how to do it.

mulkieran:
We are gradually introducing these purpose built data structures, to address the invariants that we need to enfoce, so this one is obsolete.

PR #2264, opened 2020-10-07
Verify that all device paths passed to trait methods are absolute

mulkieran:
Closes: https://github.com/stratis-storage/project/issues/229

### DIFF
--- a/src/engine/shared.rs
+++ b/src/engine/shared.rs
@@ -213,6 +213,30 @@ pub fn validate_name(name: &str) -> StratisResult<()> {
     Ok(())
 }
 
+/// Verify that all paths are absolute.
+pub fn validate_paths(paths: &[&Path]) -> StratisResult<()> {
+    let non_absolute_paths: Vec<&Path> = paths
+        .iter()
+        .filter(|path| !path.is_absolute())
+        .cloned()
+        .collect();
+    if non_absolute_paths.is_empty() {
+        Ok(())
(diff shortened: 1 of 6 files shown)


mulkieran:
I think that this will do. But I updated an old issue: https://github.com/stratis-storage/stratisd/issues/385. It would be better to rely on our source to enforce what we need, but the proposed change would be quite invasive, so best to defer it.

src/engine/shared.rs
```diff
@@ -162,3 +162,116 @@ pub fn set_key_shared(key_fd: RawFd, interactive: Option<bool>) -> StratisResult
...
+
+/// Validate a str for use as a Pool or Filesystem name.
+pub fn validate_name(name: &str) -> StratisResult<()> {
+    let name_path = Path::new(name);
+    if name.contains('\u{0}') {
+        return Err(StratisError::Engine(
+            ErrorEnum::Invalid,
+            format!("Name contains NULL characters : {}", name),
+        ));
+    }
```
jbaublitz, line 174: Is there a reason that this doesn't come before `let name_path ...`?

mulkieran: I doubt it! I didn't look inside this method when I moved it. I'll swap the order.

jbaublitz: Perfect, thanks!

jbaublitz (approved):

PR #2266, opened 2020-10-09
Change if let statements to map_or

bgurney-rh:
Related: stratis-storage/project#230

### DIFF
--- a/src/engine/strat_engine/backstore/backstore.rs
+++ b/src/engine/strat_engine/backstore/backstore.rs
@@ -556,13 +556,10 @@ impl Backstore {
     pub fn teardown(&mut self) -> StratisResult<()> {
         match self.cache {
             Some(ref mut cache) => cache.teardown(get_dm()),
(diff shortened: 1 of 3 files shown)


drckeefe (approved):

mulkieran (changes requested):
This lint has a bunch of problems. Its suggestions are not good in some places; for example in keys.rs it adds extra parens for no obvious reason. It also has a habit of mixing up if-else chains and calls to map_or(), as here: https://github.com/rust-lang/rust-clippy/issues/6137#issuecomment-705605688 and in our code in liminal.rs. And it's a pedantic lint anyway, and we explicitly allow ourselves to allow pedantic lints.

So instead of making the code changes, please make a change in the Makefile. Add this lint as an allowed lint, and add a comment explaining why we choose to allow it.

bgurney-rh:
OK; I created a new PR with the Makefile change, in #2267.

Issue #2265, opened 2020-10-07
Need to restrict the names of filesystems and pools so that no characters are escaped in forming symlinks


mulkieran:
The method we use to check the format of symlinks is here: https://github.com/stratis-storage/stratisd/blob/develop-2.2.0/src/engine/shared.rs#L167.

Need to add additional checks to exclude any characters that might be escaped to form symlinks.

Need to add additional tests as well, in test module.

Will also have to modify blackbox tests, so that they do not generate file and pool names with forbidden characters.

May also have to modify udev tests.

PR #2269, opened 2020-10-12
Add regex to exclude characters filtered by udev

bgurney-rh:
Related: #2265

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ itertools = "0.8.0"
 libcryptsetup-rs = "0.4.0"
 semver = "0.9.0"
 termios = "0.3"
+regex = "1.3"
 
 [dependencies.dbus]
 version = "0.8"
@@ -52,9 +53,6 @@ optional = true
 [dependencies.libdbus-sys]
 version = "0.2"
 optional=true
-
-[dependencies.regex]
-version = "1.3"
 
 [dependencies.uuid]
 version = "0.7"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,7 +55,6 @@ optional=true
 
 [dependencies.regex]
 version = "1.3"
-optional = true
 
 [dependencies.uuid]
 version = "0.7"
@@ -73,5 +72,5 @@ matches = "0.1.8"
 
 [features]
 default = ["dbus_enabled"]
-dbus_enabled = ["dbus", "regex", "libdbus-sys"]
+dbus_enabled = ["dbus", "libdbus-sys"]
 extras = []
--- a/src/engine/shared.rs
+++ b/src/engine/shared.rs
@@ -198,7 +198,6 @@ pub fn validate_name(name: &str) -> StratisResult<()> {
             format!("Name contains control characters : {}", name),
         ));
     }
-    //let name_udevregex = Regex::new(r"[\$!]").expect("regex is invalid");
     let name_udevregex = Regex::new(r"^[0-9A-Za-z#+-.:=@_]*$")
         .expect("regex is invalid");
     if !name_udevregex.is_match(name) {
--- a/src/engine/shared.rs
+++ b/src/engine/shared.rs
@@ -284,5 +284,7 @@ mod tests {
         assert_matches!(validate_name("\u{211D}"), Ok(_));
         assert_matches!(validate_name("☺"), Ok(_));
         assert_matches!(validate_name("ok_name"), Ok(_));
+        assert_matches!(validate_name("ユニコード"), Ok(_));
+        assert_matches!(validate_name("ユニコード?"), Err(_));
     }
 }
--- a/src/engine/shared.rs
+++ b/src/engine/shared.rs
@@ -197,9 +197,11 @@ pub fn validate_name(name: &str) -> StratisResult<()> {
             format!("Name contains control characters : {}", name),
         ));
     }
-    let name_udevregex =
-        Regex::new(r"[[:ascii:]&&[^0-9A-Za-z#+-.:=@_/]]+").expect("regex is valid");
-    if name_udevregex.is_match(name) {
+    lazy_static! {
+        static ref NAME_UDEVREGEX: Regex =
+            Regex::new(r"[[:ascii:]&&[^0-9A-Za-z#+-.:=@_/]]+").expect("regex is valid");
+    }
+    if NAME_UDEVREGEX.is_match(name) {
         return Err(StratisError::Engine(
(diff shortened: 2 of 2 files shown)


src/engine/shared.rs
```diff
@@ -196,6 +198,13 @@ pub fn validate_name(name: &str) -> StratisResult<()> {
             format!("Name contains control characters : {}", name),
         ));
     }
+    let name_udevregex = Regex::new(r"[\$!]").unwrap();
```
mulkieran, line 201: You can change the unwrap() into an expect(). If it is a valid regular expression, then the function must succeed, and you can guarantee that ti's a valid regular expression by looking at it.

mulkieran: You can also follow the advice of the regex maintainers about using lazy_static(), I should think. The instructions are here: https://docs.rs/regex/1.4.0/regex/index.html#example-avoid-compiling-the-same-regex-in-a-loop. There was no point in using lazy_static in stratis_uuids_to_names, because the regular expression is used only once, but since stratisd is a long running program it's definitely worth it here.

mulkieran (changes requested):

mulkieran:
You'll also want to edit Cargo.toml to make the regex dependency non-optional.

src/engine/shared.rs
bgurney-rh: OK; I just changed it to an expect() call.

drckeefe (approved):

Cargo.toml
```diff
@@ -55,7 +55,6 @@ optional=true
 
 [dependencies.regex]
 version = "1.3"
```
jbaublitz, line 57: For consistency's sake, it might be best to move this up with all of the other key value pairs under `[dependencies]`.

jbaublitz (changes requested):

Cargo.toml
bgurney-rh: Resolved.

jbaublitz (approved):

src/engine/shared.rs
```diff
@@ -198,8 +198,10 @@ pub fn validate_name(name: &str) -> StratisResult<()> {
             format!("Name contains control characters : {}", name),
         ));
     }
-    let name_udevregex = Regex::new(r"[\$!]").expect("regex is invalid");
-    if name_udevregex.is_match(name) {
+    //let name_udevregex = Regex::new(r"[\$!]").expect("regex is invalid");
+    let name_udevregex = Regex::new(r"^[0-9A-Za-z#+-.:=@_]*$")
```
mulkieran, line 202: Please leave the '/' in here, for completeness/ sake. It's fine if its redundant w/ the path checks below.

mulkieran (changes requested):
This looks nearly done. Please add some tests in the test method at the bottom of the file.

bgurney-rh:
Interesting: one of the tests failed while I was trying to run `build-and-package.sh` from the ci repository:

```
failures:

---- engine::shared::tests::test_validate_name stdout ----
thread 'engine::shared::tests::test_validate_name' panicked at 'assertion failed: `Err(Engine(Invalid, "Name contains characters not allowed by udev : 𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏𐌏"))` does not match `Ok(_)`', src/engine/shared.rs:277:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

bgurney-rh:
That last test is failing on line 277 of shared.rs:
```
assert_matches!(validate_name(&"𐌏".repeat(63)), Ok(_));
```

src/engine/shared.rs
```diff
@@ -196,6 +198,14 @@ pub fn validate_name(name: &str) -> StratisResult<()> {
             format!("Name contains control characters : {}", name),
         ));
     }
+    let name_udevregex =
```
mulkieran, line 201: Use lazy_static trick as described here: https://docs.rs/regex/1.4.1/regex/index.html#example-avoid-compiling-the-same-regex-in-a-loop. We want to make sure that even if this method is called many times, the regular expression is compiled only once.

src/engine/shared.rs
```diff
@@ -196,6 +198,14 @@ pub fn validate_name(name: &str) -> StratisResult<()> {
...
+        Regex::new(r"[[:ascii:]&&[^0-9A-Za-z#+-.:=@_/]]+").expect("regex is valid");
+    if name_udevregex.is_match(name) {
+        return Err(StratisError::Engine(
+            ErrorEnum::Invalid,
+            format!("Name contains characters not allowed by udev : {}", name),
```
mulkieran, line 206: "not allowed in udev symlinks" might be better.

mulkieran (changes requested):
A few requests, but looks good generally.

src/engine/shared.rs
```diff
@@ -198,7 +198,7 @@ pub fn validate_name(name: &str) -> StratisResult<()> {
             format!("Name contains control characters : {}", name),
         ));
     }
-    let name_udevregex = Regex::new(r"[\$!]").unwrap();
+    let name_udevregex = Regex::new(r"[\$!]").expect("regex is invalid");
```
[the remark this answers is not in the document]
bgurney-rh: Resolved.

src/engine/shared.rs
```diff
@@ -16,6 +16,8 @@ use termios::Termios;
 use devicemapper::Bytes;
 use libcryptsetup_rs::SafeMemHandle;
 
+use regex::Regex;
```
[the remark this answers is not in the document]
bgurney-rh: Resolved.

src/engine/shared.rs
bgurney-rh: Resolved.

src/engine/shared.rs
bgurney-rh: Resolved.

mulkieran (changes requested):
Looks good! Please close and open new PR for final review.

Issue #2156, opened 2020-08-05
Move use of KeyDescription type all the way up to the D-Bus layer


jbaublitz:
After the `KeyDescription` type was moved up to a shared scope between the sim_engine and strat_engine, some return types for key descriptions were still left as `&str` or `String`. The conversion should be done in the D-Bus layer only.

mulkieran:
@jbaublitz You would like to change the types of trait KeyActions methods so that they return or take an argument of a KeyDescription rather than an &str or String as the case might be?

jbaublitz:
That's correct. I think the D-Bus layer should be responsible for these conversions (from strongly typed to D-Bus compatible data types).

mulkieran:
Ok. Tentatively re-assigning.

PR #2273, opened 2020-10-14
Revise KeyAction method arguments to use KeyDescription

lleshchi:
Related: #2156 

KeyActions methods set() and unset() and StratKeyActions::set_no_fd() have been adjusted to use KeyDescription rather than &str. 

This revision required a test adjustment in which a keydescription is cloned. Is this the correct approach? Should the methods mentioned above take a &KeyDescription as an argument instead?

### DIFF
--- a/src/bin/stratis-min/key.rs
+++ b/src/bin/stratis-min/key.rs
@@ -5,7 +5,7 @@
 use std::{fs::File, io, os::unix::io::AsRawFd};
 
 use libstratis::{
-    engine::{DeleteAction, KeyActions, MappingCreateAction, StratKeyActions},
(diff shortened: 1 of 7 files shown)


mulkieran:
This looks fine to me, but we will need to make a decision about the failures to compile w/ stratis-min.

mulkieran (changes requested):
@lleshchi We decided it makes the most sense for you to go ahead and fix the stratis-min compiler errors. We expect that that code will all be removed pretty soon, due to other changes, but it seems best to move this PR along, rather than wait for that.

mulkieran (changes requested):
Looks good! Please close this PR and open a new one for final review.

PR #2280, opened 2020-10-21
Rework update_cargo_crates script and do all updates that seem desirable at this time

mulkieran:
Related: https://github.com/stratis-storage/project/issues/232


mulkieran:
For preliminary review.

mulkieran:
Actually, I think I want to add some more command-line options and make the output a bit nicer first, just in case anybody wants to try it.

jbaublitz:
@mulkieran In the title, it says "do some updates". Is that why `Cargo.lock` has changes?

Other than that, I've taken a preliminary look and think that using an existing graph solution seems like a very good idea, and this solves the problem of removing children of eliminated nodes in the graph, so I have no concerns so far. Let me know when you're ready for me to try it out, and I'll pull it down and give feedback on the interface.

mulkieran:
> @mulkieran In the title, it says "do some updates". Is that why `Cargo.lock` has changes?
> 
> Other than that, I've taken a preliminary look and think that using an existing graph solution seems like a very good idea, and this solves the problem of removing children of eliminated nodes in the graph, so I have no concerns so far. Let me know when you're ready for me to try it out, and I'll pull it down and give feedback on the interface.

Ok! I tried out its recommended updates to see what the results would be, so the updates are just a demo of what we'ld get if we used it and a proof that it's working up to a point.

mulkieran:
Ok! I'm putting this up for final review. The script is still a work-in-progress, and it probably should be ported to Rust if we can possibly take advantage of cargo-tree or cargo-outdated internal implementation. But that'll be a while yet. I'll want to discuss the update decisions a bit more in standup.

jbaublitz (approved):

PR #2281, opened 2020-10-21
Revise KeyAction method arguments to use KeyDescription

lleshchi:
Closes: #2156
Supercedes: #2273

### DIFF
--- a/src/bin/stratis-min/main.rs
+++ b/src/bin/stratis-min/main.rs
@@ -2,7 +2,10 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+use std::convert::TryFrom;
+
 use clap::{App, Arg, ArgGroup, SubCommand};
+use libstratis::engine::KeyDescription;
 
 mod key;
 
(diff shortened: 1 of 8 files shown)


mulkieran (approved):

drckeefe (approved):

src/dbus_api/api/shared.rs
```diff
@@ -130,7 +130,7 @@ pub fn set_key_shared(
     let message: &Message = m.msg;
     let mut iter = message.iter_init();
 
-    let key_desc: &str = get_next_arg(&mut iter, 0)?;
+    let key_desc_str: &str = get_next_arg(&mut iter, 0)?;
```
jbaublitz, line 133: For future reference, if you change this line to `key_desc_string: String`, `get_next_arg` will parse a `String` for you and you can drop the `.to_owned()` below.

jbaublitz (approved):

src/dbus_api/api/shared.rs
lleshchi: I see, because `get_next_arg` returns `Result<T, MethodErr>`. @jbaublitz I can change this if you'd like.

jbaublitz: Unless @mulkieran has objections, I'm happy to keep it the way it is!

mulkieran: @lleshchi: Actually, you can get the benefit of that property in two places, so you might as well make the change.

mulkieran (approved):

jbaublitz (approved):

mulkieran:
There's a failure in the stratis_min code changes. @lleshchi You should be able to see it directly by running the "make" command that builds it. The target is ```stratis-min```; that executable is not built by default because it is optional.

src/bin/stratis-min/main.rs
```diff
@@ -46,13 +49,18 @@ fn main() -> Result<(), String> {
     if let Some(subcommand) = args.subcommand_matches("key") {
         if let Some(args) = subcommand.subcommand_matches("set") {
             key::key_set(
-                args.value_of("key_desc").expect("required"),
+                &KeyDescription::try_from(args.value_of("key_desc").expect("required").to_owned())
+                    .expect("try_from"),
```
lleshchi, line 53: @mulkieran What message should I print here? Is there a better way to handle errors that may arise in try_from?

mulkieran: @lleshchi Yes. You only want to use expect() if you can prove it must succeed. The argument for the expect() on value_of() is that the parser has been set up so that it ensures that a value for "key_desc" is always available, and so that can never fail. But the user could pass an invalid string for the key description, our code can't prevent that.

The thing to look at is the return type of the method, ```Result<(), String>```. When you return an error, it must be in the form of a String. There's examples in the method itself, using ```map_err```, to convert another kind of error to a String. Try following that exampel.

mulkieran (changes requested):
Looks good. Please close and open for final review.

PR #2282, opened 2020-10-23

### LINKED ISSUE #1906: Potential use case for mio
The comment [here](https://github.com/stratis-storage/stratisd/blob/develop-2.0.1/src/bin/stratisd.rs#L369-L379) makes a good case for avoiding nix. However, mio does not have the same limitations and provides a good cross-platform implementation of the poll interface for all supported operating systems. Moreover, because file descriptors should be unique if they are not duplicated which we do not do, there seems to be a clever way in which we could store the file descriptor as the `Token` value for use with D-Bus. While we should investigate this option as it doesn't seem to be the exact case for which `Token` was intended, there is also another possibility of implementing the `Source` trait for our own data structure wrapping the file descriptor which could do resource clean up using the `Drop` trait as well. Both of these options would take some investigation but seem like viable options over libc.

### LINKED ISSUE #2277: Move D-Bus request processing into a separate thread
This is work that will make the addition of JSON RPC as another IPC mechanism easier and cleaner.

jbaublitz:
Closes #2277 
Closes #1906 

@mulkieran This is not ready for review, but feel free to take a look at my approach so far.


jbaublitz:
@bgurney-rh Rerunning CI with a rebase. The CI failure I saw here may be valid, but I just want to see if it's reproducible.

jbaublitz:
@bgurney-rh I think it was spurious failure. I can't reproduce it locally or in the CI pipeline.

PR #2283, opened 2020-10-28
Fix up audit command

mulkieran:


mulkieran:
Trivial...a fix to a command-line argument w/ a task that is an allowed failure anyway.

Issue #2133, opened 2020-07-16
Reading the static header may automatically repair the header when repair is undesired (by diagnostic tools)


bgurney-rh:
The new stratis_dumpmetadata program is a diagnostic tool to help analyze the static header and pool metadata.  In an ideal case, the tool would only read the portions of metadata necessary to print the headers, and report any issues.  However, the `StaticHeader::setup` method currently may attempt repair a signature block if the other block is intact.

Normally, this behavior is desired, but now that we have a diagnostic tool, it would be ideal for this call to be able to take no corrective action, and to report the presence of a faulty signature block.

PR #2285, opened 2020-10-29
Decompose StaticHeader::setup() into two functions

lleshchi:
Related: #2133 

Step 1 of 3

Next steps:
2. Completely replace setup() with calls to these two new functions.
3. Adapt BDA::load() and also probably expose the new method read_buffer()
so that it can be used in the stratis_dump_metadata script. Make changes
elsewhere as necessary.

### DIFF
--- a/src/engine/strat_engine/metadata/static_header.rs
+++ b/src/engine/strat_engine/metadata/static_header.rs
@@ -34,6 +34,8 @@ const STRAT_MAGIC: &[u8] = b"!Stra0tis\x86\xff\x02^\x41rh";
 
 const STRAT_SIGBLOCK_VERSION: u8 = 1;
 
+pub type ReadResult = StratisResult<StratisResult<Option<StaticHeader>>>;
+
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub enum MetadataLocation {
     Both,
@@ -83,7 +85,8 @@ pub fn device_identifiers<F>(f: &mut F) -> StratisResult<Option<StratisIdentifie
 where
     F: Read + Seek + SyncAll,
 {
-    StaticHeader::setup(f).map(|sh| sh.map(|sh| sh.identifiers))
+    let read_results = StaticHeader::read_sigblocks(f);
+    StaticHeader::repair_sigblocks(f, read_results).map(|sh| sh.map(|sh| sh.identifiers))
 }
 
 /// Remove Stratis identifying information from device.
@@ -209,6 +212,22 @@ impl StaticHeader {
         BDAExtendedSize::new(self.mda_size.bda_size().sectors() + self.reserved_size.sectors())
     }
 
+    pub fn read_sigblocks<F>(f: &mut F) -> (ReadResult, ReadResult)
+    where
+        F: Read + Seek,
+    {
+        let (maybe_buf_1, maybe_buf_2) = StaticHeader::read(f);
+
+        (
+            maybe_buf_1
+                .map(|buf| StaticHeader::sigblock_from_buf(&buf))
+                .map_err(StratisError::from),
+            maybe_buf_2
+                .map(|buf| StaticHeader::sigblock_from_buf(&buf))
+                .map_err(StratisError::from),
+        )
+    }
+
     /// Try to find a valid StaticHeader on a device.
     /// Return the latest copy that validates as a Stratis BDA, however verify both
     /// copies and if one validates but one does not, re-write the one that is incorrect.  If both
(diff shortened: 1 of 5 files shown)


jbaublitz (approved):

jbaublitz (changes requested):
Whoops! Just saw there's a linting error.

mulkieran (commented):
Looks good! Feel free to proceed w/ step (2).

mulkieran:
@lleshchi You'll probably want to define an alias for the complex type, to make clippy happy. Here's a basic reference: https://doc.rust-lang.org/reference/items/type-aliases.html.

lleshchi:
@mulkieran @jbaublitz Should I use type definitions?

lleshchi:
Travis failures are on account of the unused StaticHeader import in stratis_dumpmetadata (as expected). @mulkieran Would you like for me to edit stratis_dumpmetadata as well to incorporate the use of this method? 

Additionally, what are your thoughts on maintaining a StaticHeader::setup() method, which calls StaticHeader::read_buffer() and Static Header::repair(), for cases where repairs are necessary?

mulkieran:
> Travis failures are on account of the unused StaticHeader import in stratis_dumpmetadata (as expected). @mulkieran Would you like for me to edit stratis_dumpmetadata as well to incorporate the use of this method?
> 
> Additionally, what are your thoughts on maintaining a StaticHeader::setup() method, which calls StaticHeader::read_buffer() and Static Header::repair(), for cases where repairs are necessary?

I'm in favor of dropping ```StaticHeader:;setup()``` entirely. On reflection, I think ```read_buffer``` should probably be called ```read_sigblocks```, to be more precise. Don't bother with ```stratis_dumpmetadata``` yet, until we've taken a good luck at the rest of the methods that are within the engine module.

src/engine/strat_engine/metadata/static_header.rs
```diff
@@ -483,17 +501,20 @@ pub mod tests {
...
             let buf_size = bytes!(static_header_size::STATIC_HEADER_SECTORS);
             let mut buf = Cursor::new(vec![0; buf_size]);
-            prop_assert!(StaticHeader::setup(&mut buf).unwrap().is_none());
+            let read_results = StaticHeader::read_sigblocks(&mut buf);
+            prop_assert!(StaticHeader::repair_sigblocks(&mut buf, read_results).unwrap().is_none());
 
             sh.write(&mut buf, MetadataLocation::Both).unwrap();
 
-            prop_assert!(StaticHeader::setup(&mut buf)
+            let read_results_2 = StaticHeader::read_sigblocks(&mut buf);
```
lleshchi, line 509: @mulkieran Could you recommend a better way of doing this while still indicating a clear difference between these reads? Would you prefer simply shadowing these variables in the same scope?

mulkieran: In ```test_ownership``` you have a good argument for not invoking ```repair_sigblocks``` at all. This test was never designed to test the repair behavior; there are other tests that do. Instead, it makes sense to assert a useful fact about the result of ```read_sigblocks```. That approach will resolve your question about this identifier, also.

src/engine/strat_engine/metadata/static_header.rs
```diff
@@ -483,17 +501,20 @@ pub mod tests {
         fn test_ownership(ref sh in static_header_strategy()) {
             let buf_size = bytes!(static_header_size::STATIC_HEADER_SECTORS);
             let mut buf = Cursor::new(vec![0; buf_size]);
-            prop_assert!(StaticHeader::setup(&mut buf).unwrap().is_none());
+            let (s1, s2) =  StaticHeader::read_sigblocks(&mut buf);
+            prop_assert!(s1.unwrap().unwrap().is_none() && s2.unwrap().unwrap().is_none());
```
mulkieran, line 505: You might as well use ```prop_assert_eq!``` here for better error messages.

mulkieran (changes requested):
I have one small request, otherwise looks like you're headed in the right direction. I think that the Jenkins errors are spurious, probably some old LUKS metadata got left on a device.

mulkieran:
@bgurney-rh Could you see if there are any stray headers sitting around on some devices? I think they're the cause of the Jenkins failures most likely.

mulkieran (changes requested):
Newest changes look good. Go ahead and make the fixes that will eliminate the lint, just in case there are some other lints that that lint is blocking. Now is also a good time to port your entire PR over to the new development branch. You'll want to close this PR and open a completely new one. Feel free to squash your changes when you do.

PR #2286, opened 2020-10-30
Prototype of potential workaround for D-Bus connection hang up

jbaublitz:
Related to stratis-storage/project#237

### DIFF
--- a/src/bin/stratis-min/key.rs
+++ b/src/bin/stratis-min/key.rs
@@ -15,20 +15,16 @@ use libstratis::{
 /// settings and settings such as `NOECHO` are not set. This option should be
 /// used carefully as it will cause the password to be echoed on the screen if
 /// invoked interactively.
-pub fn key_set(
-    key_desc: &KeyDescription,
-    keyfile_path: Option<&str>,
-    no_tty: bool,
-) -> StratisResult<()> {
+pub fn key_set(key_desc: &KeyDescription, keyfile_path: Option<&str>) -> StratisResult<()> {
     let ret = match keyfile_path {
         Some(kp) => {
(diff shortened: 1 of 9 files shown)


jbaublitz:
@mulkieran I've switched to a non-blocking way of checking if there are more bytes available after `STRATIS_MAX_PASS_SIZE` is reached. The previous way in the PR would have had the same hanging problem if the user provided _exactly_ `STRATIS_MAX_PASS_SIZE` bytes in a terminal as it would read exactly that many bytes, trigger the if statement, and then if the file descriptor were blocking, the `.next()` call on the byte iterator would block as there would be no more data left to read. This change will ensure that we get the appropriate information without ever blocking.

mulkieran:
> @mulkieran I've switched to a non-blocking way of checking if there are more bytes available after `STRATIS_MAX_PASS_SIZE` is reached. The previous way in the PR would have had the same hanging problem if the user provided _exactly_ `STRATIS_MAX_PASS_SIZE` bytes in a terminal as it would read exactly that many bytes, trigger the if statement, and then if the file descriptor were blocking, the `.next()` call on the byte iterator would block as there would be no more data left to read. This change will ensure that we get the appropriate information without ever blocking.

Great! Thanks for catching that. AFAICT, nothing that a caller can do could now cause this chunk of code to hang.

mulkieran (commented):
Looks good.

mvollmer (approved):
This works fine for the Cockpit Stratis PoC. I have tested it together with https://github.com/stratis-storage/stratis-cli/pull/670.

mulkieran (changes requested):
jbaublitz: Please close this and prepare a final PR.

PR #2291, opened 2020-11-04
Jbaublitz issue project 192

mulkieran:


mulkieran:
Should be passing stratis-nonrust tests now.

mulkieran:
stalled on test_unique_instance, redoing

mulkieran:
ok! stratisd_nonrust got a lot further. I'll continue investigating.

mulkieran (commented):
Checked over code in dbus_api, looks good.

jbaublitz:
@mulkieran A few responses:
* With regards to the regex version pin, my typical approach when specifying dependencies is that if there is a question about whether a lower versions will compile, I think it is good to explicitly state which minor version is our minimum for the sake of anyone looking through our `Cargo.toml` file for research purposes. I think that is a clearer indication of what our minimums are than Cargo.lock which some people do ignore.
* `sha-1` 0.9 did not compile on the toolchain we were using last time I tried to use it. I'm glad that this has been resolved.
* I'm not sure that returning the identity idempotent return type makes sense for unencrypted pools. We very clearly return an error when trying to bind unencrypted pools. Why would we do something different for unbinding? This seems inconsistent and confusing.

mulkieran:
>     * I'm not sure that returning the identity idempotent return type makes sense for unencrypted pools. We very clearly return an error when trying to bind unencrypted pools. Why would we do something different for unbinding? This seems inconsistent and confusing.

I'm also fine w/ returning an error. I've put up a few more patches.

mulkieran:
> @mulkieran A few responses:
> 
>     * With regards to the regex version pin, my typical approach when specifying dependencies is that if there is a question about whether a lower versions will compile, I think it is good to explicitly state which minor version is our minimum for the sake of anyone looking through our `Cargo.toml` file for research purposes. I think that is a clearer indication of what our minimums are than Cargo.lock which some people do ignore.

I don't think our versioning in Cargo.toml has followed a consistent rule, until now. But I can see that making Cargo.lock the sole authority might cause some trouble. We can see about in Cargo.toml specifying the lowest version that is compatible with the version in Cargo.lock and that still allows our code to compile, for all.

mulkieran:
> > @mulkieran A few responses:
> > ```
> > * With regards to the regex version pin, my typical approach when specifying dependencies is that if there is a question about whether a lower versions will compile, I think it is good to explicitly state which minor version is our minimum for the sake of anyone looking through our `Cargo.toml` file for research purposes. I think that is a clearer indication of what our minimums are than Cargo.lock which some people do ignore.
> > ```
> 
> I don't think our versioning in Cargo.toml has followed a consistent rule, until now. But I can see that making Cargo.lock the sole authority might cause some trouble. We can see about in Cargo.toml specifying the lowest version that is compatible with the version in Cargo.lock and that still allows our code to compile, for all.

Actually, I think I'ld like to stick w/ this decision. I looked through our other dependencies, and it seems like where we specify only the major version, any failure of compilation that might occur is caused only be requirements of our dependencies. That is, there is no instance in which _stratisd_ requires a higher version because it uses some enhancement provided by the new minor version. The requirement of a higher version is always due solely to some dependency having pinned the version. And that is one of the several things that the Cargo.lock file is for.

I think it's likely that what we should be doing in some cases is generalizing our dependencies in the libraries that we control. In a lot of cases, they seem to be too specific.

jbaublitz:
> > > @mulkieran A few responses:
> > > ```
> > > * With regards to the regex version pin, my typical approach when specifying dependencies is that if there is a question about whether a lower versions will compile, I think it is good to explicitly state which minor version is our minimum for the sake of anyone looking through our `Cargo.toml` file for research purposes. I think that is a clearer indication of what our minimums are than Cargo.lock which some people do ignore.
> > > ```
> > 
> > 
> > I don't think our versioning in Cargo.toml has followed a consistent rule, until now. But I can see that making Cargo.lock the sole authority might cause some trouble. We can see about in Cargo.toml specifying the lowest version that is compatible with the version in Cargo.lock and that still allows our code to compile, for all.
> 
> Actually, I think I'ld like to stick w/ this decision. I looked through our other dependencies, and it seems like where we specify only the major version, any failure of compilation that might occur is caused only be requirements of our dependencies. That is, there is no instance in which _stratisd_ requires a higher version because it uses some enhancement provided by the new minor version. The requirement of a higher version is always due solely to some dependency having pinned the version. And that is one of the several things that the Cargo.lock file is for.
> 
> I think it's likely that what we should be doing in some cases is generalizing our dependencies in the libraries that we control. In a lot of cases, they seem to be too specific.

I'm a little confused by this decision mainly because, even though we may not be in that situation currently, semantic versioning could allow us to eventually bump into that problem quite easily.

mulkieran:
> > > > @mulkieran A few responses:
> > > > ```
> > > > * With regards to the regex version pin, my typical approach when specifying dependencies is that if there is a question about whether a lower versions will compile, I think it is good to explicitly state which minor version is our minimum for the sake of anyone looking through our `Cargo.toml` file for research purposes. I think that is a clearer indication of what our minimums are than Cargo.lock which some people do ignore.
> > > > ```
> > > 
> > > 
> > > I don't think our versioning in Cargo.toml has followed a consistent rule, until now. But I can see that making Cargo.lock the sole authority might cause some trouble. We can see about in Cargo.toml specifying the lowest version that is compatible with the version in Cargo.lock and that still allows our code to compile, for all.
> > 
> > 
> > Actually, I think I'ld like to stick w/ this decision. I looked through our other dependencies, and it seems like where we specify only the major version, any failure of compilation that might occur is caused only be requirements of our dependencies. That is, there is no instance in which _stratisd_ requires a higher version because it uses some enhancement provided by the new minor version. The requirement of a higher version is always due solely to some dependency having pinned the version. And that is one of the several things that the Cargo.lock file is for.
> > I think it's likely that what we should be doing in some cases is generalizing our dependencies in the libraries that we control. In a lot of cases, they seem to be too specific.
> 
> I'm a little confused by this decision mainly because, even though we may not be in that situation currently, semantic versioning could allow us to eventually bump into that problem quite easily.

What's the specific scenario that you're thinking of, assuming that we make our versions in Cargo.toml general and continue to release the Cargo.lock file?

jbaublitz:
Even if we release Cargo.lock, we've already bumped into situations where people have ignored Cargo.lock. We can definitely take a hard line on that, but I don't think that will stop people from doing it. If we eventually take advantage of a feature in a minor (or patch version prior to the v1.0.0 release for a crate) which we've already done with dbus-rs v0.8.4 in one of my PRs, Cargo.toml will have no indication of the minimum version of the crate that stratisd will compile with. Going lower than that will fail to compile even though it's a feature addition, not a breaking change. I think it's better to be explicit rather than implicit in Cargo.toml.

mulkieran:
> Even if we release Cargo.lock, we've already bumped into situations where people have ignored Cargo.lock. We can definitely take a hard line on that, but I don't think that will stop people from doing it. If we eventually take advantage of a feature in a minor (or patch version prior to the v1.0.0 release for a crate) which we've already done with dbus-rs v0.8.4 in one of my PRs, Cargo.toml will have no indication of the minimum version of the crate that stratisd will compile with. Going lower than that will fail to compile even though it's a feature addition, not a breaking change. I think it's better to be explicit rather than implicit in Cargo.toml.

If, in our stratisd source, there is code that will not compile with any version of dbus-rs less than 0.8.4, then we'll specify 0.8.4 in Cargo.toml.

jbaublitz:
I am okay with that. Making it as specific as it needs to be in Cargo.toml sounds like a good compromise.

mulkieran:
Superceded.

PR #2293, opened 2020-11-05
Bump a few more version in lock file

mulkieran:


mulkieran:
just following policy, so we can call it trivial.

mulkieran:
rebased.

PR #2300, opened 2020-11-10
Jbaublitz issue project 192 new

mulkieran:
Supercedes: #2291


src/engine/strat_engine/backstore/crypt.rs
```diff
@@ -461,14 +479,96 @@ impl CryptHandle {
...
+            let map = if let Some(m) = json.as_object_mut() {
+                m
+            } else {
+                return Err(LibcryptErr::Other(format!(
+                    "configuration for Clevis is is not in JSON object format: {}",
+                    json
+                )));
+            };
+
+            map.remove(CLEVIS_TANG_TRUST_URL).is_some()
```
jbaublitz, line 545: Given that the value for this is a boolean, wouldn't it make more sense to do `map.remove(CLEVIS_TANG_TRUST_URL).and_then(|b| b.as_bool()).unwrap_or(false)` or something similar just in case someone passes a `false` value? In that case, this code would still pass `-y` which could be problematic for API users.

jbaublitz (changes requested):
One high level question: if we're making `bind_clevis` take a `&mut self` reference, do you see any value in storing the clevis configuration in our data structures so that we don't have to recalculate it each time?

mulkieran:
> One high level question: if we're making `bind_clevis` take a `&mut self` reference, do you see any value in storing the clevis configuration in our data structures so that we don't have to recalculate it each time?

I think there is..but it's a bit more to do:

* At initial bind, reread the configuration from a bound device and store it. More work.
* On addition of new devices, get the stored configuration, rather than reading it off the devices. Less work.
Probably evens out.

* In liminal processing, read the Clevis configuration as well as the key description.  More work.
But there's some value in doing that, because we ensure that the pool is well-formed w/ respect to the clevis configuration on its devices. We already do lots of checks for the key description, we would just add more. At present, there is a tiny loophole where it would be possible, I think, to unlock and setup a pool where different devices had different Clevis configurations.

Ultimately, we'll want to publish something on the D-Bus wrt. Clevis and a pool or device. One thing we could publish is the Clevis configuration subset that the ```clevis_info()``` method yields.

jbaublitz:
@mulkieran Do you want to do that part of it or do you want me to?

mulkieran:
> @mulkieran Do you want to do that part of it or do you want me to?

Let's check in about this in standup.

mulkieran:
Ok! So the breakdown is:

@jbaublitz:
* Take key_desc out of BlockDevMgr struct
* Stow Clevis config information in every StratBlockDev
* Handle setting of the Clevis config information in the StratBlockDev after a bind operation
* Export the Clevis config for a pool via the FetchProperties interface over the D-Bus
* Handle getting the Clevis config during an add operation

@mulkieran:
* Handle all tracking of Clevis config info during setup, mostly analogously to
how key description is tracked.

mulkieran:
Superceded.

PR #2301, opened 2020-11-10

jbaublitz:
Supersedes #2282


jbaublitz:
@mulkieran I think I know what's going on with the test failure. I'm going to try to reproduce locally as it's definitely not a 100% occurrence rate.

mulkieran:
> @mulkieran I think I know what's going on with the test failure. I'm going to try to reproduce locally as it's definitely not a 100% occurrence rate.

Ok! It does look non-deterministic. The two Create4TestCases are nearly identical, but one passed, because it detected the already existing filesystem, while the other failed because it didn't.

```
test_2_create (integration.logical.test_create.Create4TestCase)
Creation of 3 volumes, of which 1 already exists, must fail. ... FAIL
test_create (integration.logical.test_create.Create4TestCase)
Creation of 2 volumes, of which 1 already exists, must fail. ... Execution failed:
It appears that you issued an unintended command: The 'create' action has no effect for resource oubliette but does for resource mnemosyne 
```

jbaublitz:
I've reviewed the code and tried to reproduce. I was unable to reproduce even using a shell script, but after reviewing the code, I think this comes down to `RwLock` scheduling in Rust. The D-Bus tree is wrapped in a read-write lock. D-Bus method calls require read access. D-Bus object addition requires write access. These happen in different threads. My guess is that in the automated tests, because the filesystem create calls are moving so quickly, the D-Bus object adder thread does not have time to acquire a write lock and add the object to the D-Bus before the next D-Bus filesystem create method call acquires a read lock on the tree, preventing the write lock from occurring. My guess is that this results in the filesystems not being visible on the D-Bus when the final filesystem create operation is performed, resulting in the call to create the filesystem not detecting the filesystems in the CLI before passing it to the engine.

Two notes:
1. If we keep it as a `RwLock`, we may see more `IncoherenceError`s on the stratis-cli side.
2. If we change it to a `Mutex`, my guess is that we will no longer see these errors as `Mutex` scheduling seems to be a queue, not a priority queue that depends on whether a read lock is already acquired. However, we will lose the benefit of being able to process multiple method calls simultaneously on the D-Bus. Some of this may be resolved if we eventually move to the new async-minded dbus 0.9 however when it's ready.

mulkieran:
Some notes about the test failure:

*. We can think of most mutating commands for stratis-cli to be a sequence of calls. First, a call to ```GetManagedObjects()```, then a call to some mutating method, like, eg. ```CreateFilesystems()```. So, CLI command ```fs create ...``` = ```R; M``` where ```R``` is a read-only D-Bus call, and ```M``` is a mutating method call. Call ```fs_create ...``` twice in a row, and you get a sequence like ```R_1;M_1;R_2;M_2```, i.e., a read D-Bus call, followed by a mutate D-Bus call, followed by a second read D-Bus call, followed by a second mutating D-Bus call.
*. It is possible to imagine two users who use the CLI at the same time. So user 1 calls ```fs create...``` and user 2 calls ```fs create...``` more or less simultaneously. This could cause the D-Bus calls from the two methods to be interleaved, as ```R_1;R_2;M_!;M_2```. We explicitly allow this to happen. There could be ways to prevent this, but this work here doesn't introduce that problem and does not attempt to address it. To prevent this unlikely problem would require additional infrastructure in the D-Bus layer and on the CLI side.
*. What we are instead interested in is the way that the sequence of calls: ```R_1;M_1;R_2;M_2``` is handled on the stratisd side. For example, for ```CreateFilesystems()```, a mutating D-Bus method:
  1. _Rec_: The message is received.
  2. _Eng_: An engine method is invoked.
  3. Following:
  *. _Rep_: A message reply is sent.
  *. _Upd_: Additional objects are added to the D-Bus layer to represent the filesystems created
It seems that it is currently possible that the _Upd_ action for ```M_1``` can occur after the read action corresponding to ```R_2```. What we want is to impose an ordering in stratisd to ensure that the _Upd_ action corresponding to ```M_1``` is completed before the read action associated with ```R_2```.

jbaublitz:
@mulkieran For now, I'm going to switch the `RwLock` to a `Mutex` because the engine is wrapped in a `Mutex` and add a fixme.

jbaublitz:
The `Mutex` does not solve the problem. There's still not enough ordering. I'll see if I can queue requests for the mutex somehow.

jbaublitz:
@mulkieran The last commit may fix the synchronization issues. Instead of acquiring the lock on each addition of the object paths to D-Bus, I hold the lock until the entire buffer is drained then release it. It's a little hard to tell if this solved it just because the tests pass this time, but let's look out for the issue as I continue to rebase. We have at least one success so far!

mulkieran:

PR #2310, opened 2020-11-18
Decompose StaticHeader::setup() into two functions

lleshchi:
Closes: #2133
Supersedes: #2285

### DIFF
--- a/src/engine/strat_engine/metadata/bda.rs
+++ b/src/engine/strat_engine/metadata/bda.rs
@@ -60,7 +60,8 @@ impl BDA {
     where
         F: Read + Seek + SyncAll,
     {
-        let header = match StaticHeader::setup(f)? {
+        let read_results = StaticHeader::read_sigblocks(f);
+        let header = match StaticHeader::repair_sigblocks(f, read_results)? {
             Some(header) => header,
             None => return Ok(None),
         };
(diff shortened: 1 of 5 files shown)


drckeefe (approved):

bgurney-rh (approved):

jbaublitz (approved):

src/engine/strat_engine/metadata/bda.rs
```diff
@@ -60,7 +60,8 @@ impl BDA {
     where
         F: Read + Seek + SyncAll,
     {
-        let header = match StaticHeader::setup(f)? {
+        let read_results = StaticHeader::read_sigblocks(f);
```
mulkieran, line 63: The code all looks good. The best thing to do is to consider this method, ```BDA::load```. If it's being called by stratis_dumpmetdata what should it do, instead of what it is doing right now? It might be worthwhile to write that out in words, and then we can think about how we should implement our decision in code.

mulkieran (changes requested):
The changes all look good. See inline comment.

mulkieran:
stratisd-nonrust-2 can be ignored.

mulkieran:
@lleshchi Please take a look at https://github.com/stratis-storage/stratisd/pull/2323. I think that the rest of this project requires these things:
1. The changes in #2323 which enrich the return type of ```read_sigbuf()```.
2. The proposed change to ```BDA::load()``` so that it takes a StaticHeader argument.
3. A modification to the repair function so that it can be pass through, which will amount to choosing to skip the write() call in ```write_header()```. There's still some technical details to discuss in that.
4. The modifications to stratis_dumpmetadata to make use of the newly exposed StaticHeader functionality so that the header information can be dumped out better and so that it does not write to any device.

mulkieran:
Superceded by #2357 and #2348.

PR #2315, opened 2020-11-19
Include revision r3 of FetchProperties interface in D-Bus conf

mulkieran:


mulkieran:
This fixes blackbox test errors.

PR #2322, opened 2020-11-20
Flock

jbaublitz:
Supersedes #2252

### DIFF
--- a/src/engine/strat_engine/flock.rs
+++ b/src/engine/strat_engine/flock.rs
@@ -1,0 +1,52 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use std::{
+    fs::File,
+    os::unix::io::AsRawFd,
+    path::{Path, PathBuf},
+};
+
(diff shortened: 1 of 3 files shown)


jbaublitz:
Copied from #2252

jbaublitz:
There's actually an interesting consideration with our metadata API. We expose that to some of our `extras` utilities. Do we want our metadata module to implement the flock code too? Currently, because we're simply passing in a type parameter to a lot of the metadata functions, we can't acquire an flock on the data structure that's passed in. We have a few options here:
* Rename the existing initialize method for use with testing our `Cursor` implementation of the metadata and make it private and expose a public method wrapping it that has the additional restriction of `AsRawFd`.
* Remove the type parameter and require that a file is passed in. This would make our tests invalid.

I prefer the first option personally and think it would allow us to maintain our tests while also providing a metadata interface that implements flock internally so that everything that we expose from the engine can be used in a multithreaded context.

jbaublitz:
This may be harder than anticipated. I've made my way up the stack from the metadata operations and given that we're doing operations on device node paths alone, there's not really a good way to determine what we need to lock below that layer. For example, for crypt devices, we need to lock the crypt device and the physical device below it, but there's no way to differentiate this from a physical device simply from a device path.

PR #2337, opened 2020-12-03
Issue stratisd 2277

jbaublitz:
Supersedes #2301


mulkieran:
I think the cli-with-stratisd error might need investigating.

jbaublitz:
I think we should have a conversation about whether `IncoherenceError`s are acceptable once we add multithreading. I changed the lock on the `Tree` data structure back to a `RwLock` which probably caused this error. If we change it back to a `Mutex`, the D-Bus layer will effectively no longer be multithreaded and if we do that, it really doesn't matter if we multithread the engine at that point as only one request can be processed at a time. As I'd alluded to before, we can try queuing the read/write requests in the order that they're received if that's preferable as that way we theoretically could handle multiple method calls at once if they are received sequentially. What is your preference? Do you think an `IncoherenceError` is unacceptable? It will certainly cause issues with testing and I can see that being a major drawback.

src/stratis/run.rs
```diff
@@ -4,61 +4,80 @@
...
-                Ok(true)
+fn udev_thread(sender: Sender<UdevEngineEvent>, should_exit: Arc<AtomicBool>) -> StratisResult<()> {
+    let context = libudev::Context::new()?;
+    let mut udev = UdevMonitor::create(&context)?;
+
+    let mut pollers = [PollFd::new(udev.as_raw_fd(), PollFlags::POLLIN)];
+    loop {
+        match poll(&mut pollers, 100)? {
+            0 => {
+                if should_exit.load(Ordering::Relaxed) {
```
mulkieran, line 44: Does it make more sense to check ```should_exit``` unconditionally, regardless of return value of ```poll()```?

mulkieran (changes requested):
Just one comment so far...

src/stratis/run.rs
jbaublitz: I'll explain my original thought process here and we can discuss what the best way forward is. In the case of a poll value of 0, this means no event was detected and the poll timed out. My design is meant to process all data available to stratisd prior to exiting. Now in the case of the engine, I think this is absolutely necessary so that we don't kill an operation part way through its execution. However, it may be less useful in the case of udev given that the udev database is outside of the engine and we are not responsible for udev state consistency like we are for our own engine. What are your thoughts? Do you think this is unnecessarily cautious? It may honestly be sufficient to exit as long as we're not in the middle of an operation which checking `should_exit` unconditionally would still satisfy.

mulkieran (changes requested):
Please take a look at https://github.com/stratis-storage/stratisd/pull/2346. The only significant change is removal of the Eventable trait. Let me know if you have any questions. I'm continuing review more or less from the top down.

jbaublitz:
This all looks reasonable so far. I also saw that 1.48.0 will add `AsRawFd` for `RawFd` which will hopefully make things a lot easier.

mulkieran:

jbaublitz:
@mulkieran Rebased.

mulkieran:
Please rebase when convenient.

jbaublitz:
Rebased!

mulkieran:
@jbaublitz: Please fix up conflict when able. Looks like git made a big conflict out of as_raw_fd() change.

jbaublitz:
Rebased

jbaublitz:
Whoops! Hadn't updated the system Rust toolchain.

mulkieran:

jbaublitz:
Rebased.

jbaublitz:
The CI failure appears to be a timing issue. I'm going to see what I can do about resolving it.

jbaublitz:
@mulkieran  This is still a timing error in the udev tests. Because all of the udev events do not appear to be coming in at once, the pool being set up is checked before all devices have been detected by the udev thread. I have a proposal for these timing errors and I'll discuss it with you on Monday.

mulkieran:
@jbaublitz After our talk on Tuesday I see that we're agreed that a way to simplify DmFd is to pass the devicemapper context to the ```dm_event_thread()``` method in the ```run()``` method. Please do that whenever is convenient, and I'll pull in your changes to my review PR.

mulkieran:
@jbaublitz Also, I remember that tokio v1 is now available. I thought I posted an issue, but I can't find it anywhere. In any case, if you can move up to tokio v1 that would be ideal. I know that it will require some adjustments; I tried it myself and apparently the "stream" feature has been discontinued.

jbaublitz:
Blocked by #2388 and CI failures.

mulkieran:
We believe that the test failures are due to the fact that they were previously succeeding due to an artifact of the single-threaded implementation: all udev events that became available were always processed before any D-Bus method was attended to. With multi-threading, that is not guaranteed. So, we may have to rethink the tests.

mulkieran:
@jbaublitz I have an alternate explanation of the udev test failures which may be more interesting. Search for the string "ERROR: test_generic (tests.udev.test_udev.UdevTest1)" and you'll find a Python stack trace for the test and a ugly looking printout of the stderr from stratisd. This represents the second time that stratisd is started within the ```test_generic``` test. If you look at the end of stderr, you can see that the pool, which is unencrypted, was set up, meaning all the block devices were discovered. Right after that there is a panic: '[2021-01-25T14:59:56Z INFO  libstratis::engine::strat_engine::keys] Mount found at /run/stratisd/keyfiles; unmounting\n', "thread 'tokio-runtime-worker' panicked at 'Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.0.2/src/runtime/enter.rs:39:9\n". I'm not able to interpret what is causing this, but I think it points to an actual bug in the multi-threading code.

jbaublitz:
Okay I will take a look!

jbaublitz:
@mulkieran This is a different error from before. I don't think this invalidates my previous concern, but I'll resolve this and see if we still bump into the same problem.

jbaublitz:
@mulkieran I reviewed your comment on my use of `pending!()`. In one case, it was unnecessary (the devicemapper thread) and I've removed that one but in the IPC thread, it is necessary or stratisd will immediately shut down if no IPC mechanism is present (`make build-no-default`). I'm happy to explain in more detail if you're interested.

mulkieran:
@jbaublitz Please rebase when you get the chance. Also, note that you can remove ```async_fd.rs``` entirely now that lowest supported toolchain is 1.49.

PR #2338, opened 2020-12-03
Remove randomizer code in sim_engine

### LINKED ISSUE #2336: Drop randomization in sim engine
We used to use it in our tests, but those were the old Python-based tests in stratisd which we eliminated. We no longer use randomization of the sim engine for any testing anywhere, and I expect we'll never use it again. It's a bit of a pain to maintain, we should drop it.

Related: #2301

jbaublitz:
Closes #2336


mulkieran (approved):
Great!

PR #2341, opened 2020-12-05
Issue project 52 u128

jbaublitz:
Supersedes #2320 

@mulkieran I think this is ready for review. Based on [this](https://github.com/rust-lang/rfcs/pull/1504), it appears that `u128` is supported on all architectures that Rust supports, even 32 bit ones. Given that our concern was stability, I think that should no longer be an issue.

### DIFF
--- a/src/dbus_api/blockdev/fetch_properties_2_0/methods.rs
+++ b/src/dbus_api/blockdev/fetch_properties_2_0/methods.rs
@@ -34,7 +34,7 @@ fn get_properties_shared(
                 result_to_tuple(blockdev_operation(
                     m.tree,
                     object_path.get_name(),
-                    |_, bd| Ok((*bd.size()).to_string()),
+                    |_, bd| Ok((*bd.size().bytes()).to_string()),
                 )),
             )),
             _ => None,
--- a/src/dbus_api/blockdev/fetch_properties_2_0/methods.rs
+++ b/src/dbus_api/blockdev/fetch_properties_2_0/methods.rs
@@ -26,24 +26,20 @@ fn get_properties_shared(
 
     let return_message = message.method_return();
 
-    let return_value: HashMap<String, (bool, Variant<Box<dyn RefArg>>)> =
-        properties
-            .unique()
-            .filter_map(|prop| match prop.as_str() {
-                consts::BLOCKDEV_TOTAL_SIZE_PROP => Some((
-                    prop,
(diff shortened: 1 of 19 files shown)


jbaublitz:
I'm doing on last quick pass over this to clean up rebase cruft and then rebasing.

jbaublitz:
Okay, this is ready.

src/engine/strat_engine/backstore/blockdevmgr.rs
```diff
@@ -345,7 +345,7 @@ impl BlockDevMgr {
             current_time
         };
 
-        let data_size = Bytes(metadata.len() as u64);
+        let data_size = Bytes(metadata.len() as u128);
```
mulkieran, line 348: We might as well not waste all those ```From``` impls in the devicemapper PR. This could just be ```Bytes::from(metadata.len())``` or some variation thereof. There's a couple other places where I think one of those would make more sense.

mulkieran (changes requested):
Just two points, otherwise looks good.

src/dbus_api/blockdev/fetch_properties_2_0/methods.rs
```diff
@@ -26,24 +26,20 @@ fn get_properties_shared(
...
-                    )),
+    let return_value: HashMap<String, (bool, Variant<Box<dyn RefArg>>)> = properties
+        .unique()
+        .filter_map(|prop| match prop.as_str() {
+            consts::BLOCKDEV_TOTAL_SIZE_PROP => Some((
+                prop,
+                result_to_tuple(blockdev_operation(
+                    m.tree,
+                    object_path.get_name(),
+                    |_, bd| Ok((*bd.size()).to_string()),
```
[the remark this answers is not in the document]
jbaublitz: Resolved.

src/engine/strat_engine/backstore/blockdevmgr.rs
jbaublitz: Okay, yes, this makes sense. I've gone through the entire code base and I think I found everywhere that that could be changed over in the most recent commit.

mulkieran (commented):
Yes, this looks good. Moving to pending.

mulkieran (changes requested):
Ok! Please close and make new PR for final review. Please add in as a separate commit use of AsRawFd as well.

PR #2343, opened 2020-12-10
Clear out a little more configure_simulator related cruft

mulkieran:


jbaublitz (commented):
It looks like CI is passing and I'm a little bit confused as to why. You've removed `configure_simulator` from the `Engine` trait but haven't removed that engine call in the D-Bus layer, and I feel like the trait type safety should have caused that to fail. Do you know what's going on here? I'm specifically confused as to why [these lines](https://github.com/mulkieran/stratisd/blob/develop-2.3.0-configure-simulator/src/dbus_api/api/manager_2_0/methods.rs#L82-L85) don't cause it to fail.

jbaublitz (approved):

PR #2344, opened 2020-12-10
Add an is_sim method to the Engine trait

mulkieran:


mulkieran:
There are some invariants related to this fact that I'ld like to be able to check programmatically.

jbaublitz (approved):

Issue #2219, opened 2020-09-09
filesystem.Devnode does not emit PropertiesChanged when calling SetName


mvollmer:
When calling SetName on a filesystem, there is a PropertiesChanged signal for the Name property, but not for the Devnode property, which changed as well.  In fact, Devnode is marked as "const", although it can be changed by a SetName call.

```
‣ Type=method_call  Endian=l  Flags=4  Version=1  Priority=0 Cookie=142
  Sender=:1.1729  Destination=org.storage.stratis2  Path=/org/storage/stratis2/2  Interface=org.storage.stratis2.filesystem  Member=SetName
  UniqueName=:1.1729
  MESSAGE "s" {
          STRING "myfs5";
  };

‣ Type=signal  Endian=l  Flags=1  Version=1  Priority=0 Cookie=349
  Sender=:1.1622  Path=/org/storage/stratis2/2  Interface=org.freedesktop.DBus.Properties  Member=PropertiesChanged
  UniqueName=:1.1622
  MESSAGE "sa{sv}as" {
          STRING "org.storage.stratis2.filesystem";
          ARRAY "{sv}" {
                  DICT_ENTRY "sv" {
                          STRING "Name";
                          VARIANT "s" {
                                  STRING "myfs5";
                          };
                  };
          };
          ARRAY "s" {
          };
  };

‣ Type=method_return  Endian=l  Flags=1  Version=1  Priority=0 Cookie=350  ReplyCookie=142
  Sender=:1.1622  Destination=:1.1729
  UniqueName=:1.1622
  MESSAGE "(bs)qs" {
          STRUCT "bs" {
                  BOOLEAN true;
                  STRING "f84ed73749744e408e5a67a5b09e451a";
          };
          UINT16 0;
          STRING "Ok";
  };
```

```
org.storage.stratis2.filesystem         interface -         -                                  -
.SetName                                method    s         (bs)qs                             -
.Created                                property  s         "2020-09-08T11:47:02Z"             const
.Devnode                                property  s         "/dev/stratis/pool3/myfs5"         const
.Name                                   property  s         "myfs5"                            emits-change
.Pool                                   property  o         "/org/storage/stratis2/1"          const
.Uuid                                   property  s         "f84ed73749744e408e5a67a5b09e451a" const
```

mulkieran:
Whoops! Expect  a fix in about a day.

mulkieran:
[...]

PR #2346, opened 2020-12-11
Jbaublitz issue stratisd 2277

mulkieran:
Also closes https://github.com/stratis-storage/stratisd/issues/2219.


mulkieran:
rebased.

mulkieran:
@jbaublitz Please see commit https://github.com/stratis-storage/stratisd/pull/2346/commits/f0554020190c6757d22012bae2c3f91cbaaaf1c6. It initializes DM context outside strat engine and unconditionally checks for DM events in DmFd implementation. Let me know if that looks good, thx. Other changes are trivial.

jbaublitz:
I took a look at that and had some questions about it. I read the commit message for this work, and it seems like requiring devicemapper for the sim_engine imposes an arbitrary restriction that is not required for the sim_engine to operate so I'm wondering if there's a way to simply this so that we can still take references to `DM` out of the sim_engine by removing it from the `Engine` trait while still only initializing it for the strat_engine. For example, we have `sim: bool` available in `run` and we could simply make [this `dm` variable](https://github.com/stratis-storage/stratisd/pull/2346/commits/f0554020190c6757d22012bae2c3f91cbaaaf1c6#diff-3604c519d3539d03673ac6805e179a57af023dbee55fee25322beb1be846798dR94) an optional value so if `sim == true`, it would be `None` and if `sim == false`, it would be `Some(DM)`.

mulkieran:
> I took a look at that and had some questions about it. I read the commit message for this work, and it seems like requiring devicemapper for the sim_engine imposes an arbitrary restriction that is not required for the sim_engine to operate so I'm wondering if there's a way to simply this so that we can still take references to `DM` out of the sim_engine by removing it from the `Engine` trait while still only initializing it for the strat_engine. For example, we have `sim: bool` available in `run` and we could simply make [this `dm` variable](https://github.com/stratis-storage/stratisd/pull/2346/commits/f0554020190c6757d22012bae2c3f91cbaaaf1c6#diff-3604c519d3539d03673ac6805e179a57af023dbee55fee25322beb1be846798dR94) an optional value so if `sim == true`, it would be `None` and if `sim == false`, it would be `Some(DM)`.

That won't really help, because dm itself is just a local variable. Basically, anything that only checks for DM events if strat engine is requested will just disperse the conditionals to different, somewhat confusing, places. I decided to go w/ something really simple instead: https://github.com/stratis-storage/stratisd/pull/2346/commits/34292590e695b00babe294d68611aed5dfc52df0.

src/stratis/run.rs
```diff
@@ -83,10 +82,7 @@ async fn dm_event_thread(engine: Option<Arc<Mutex<dyn Engine>>>) -> StratisResul
                 dm_fd.next().await;
             }
         }
-        None => {
-            pending!();
```
jbaublitz: Unless I'm misunderstand the documentation of `select!()`, this will cause stratisd to always exit when the sim engine is in use. If one future resolves, `select!()` cancels all of the other futures. The only way to avoid this is to return a future that never resolves.

jbaublitz (changes requested):
I'll double check the comment below and test it out, but I think this needs to change.

src/stratis/run.rs
mulkieran: I'll reread and take a look...the interesting thing is that the stratis-cli tests actually succeed, except for two that look like incoherence errors.

mulkieran:
simply waiting for the number of pools expected may do the trick.

mulkieran:
I _think_ the prior failure to pass was due to our eventual consistency problem. The engine is fully set up by the time the D-Bus method is invoked, but the D-Bus interface is not yet populated with all the objects, so ```GetManagedObjects``` does not return the full compliment of pools in all cases. I do not believe this could ever happen in the single-threaded implementation.

mulkieran:
GitHub Actions didn't run for some reason.

mulkieran:
@jbaublitz Please take a look at https://github.com/stratis-storage/stratisd/pull/2346/commits/f615c96029fa7a19e8c20db526afad7dfb8be5a0 and https://github.com/stratis-storage/stratisd/pull/2346/commits/2510cf387b21c6710c1f706b7a291ad2c865d900, principally. The intervening commits all just remove the old event handling framework.

mulkieran:
@mvollmer Could you take a look at this PR and see if you like the signals? The newer one #2407 , called "fresh" is under a bit more active development and this one has all the relevant changes.

mvollmer:
> @mvollmer Could you take a look at this PR and see if you like the signals?

Could you point me to the relevant changes?  This PR is quite big... :)

mulkieran:
@mvollmer I apologize; on Friday I thought our work had settled down to the point that it was ready for you to review, but it has moved on a bit since. I'm going to close this PR and tag you on the newer PR instead.

PR #2353, opened 2020-12-16
Flock

jbaublitz:
This PR adds flock support in preparation for multithreading.

Currently, I'm leaving liminal devices as is (using `EncryptionInfo` for encryption detection as opposed to storing a `CryptHandle` directly) but this could be changed if `CryptHandle` is preferable as using `EncryptionInfo` requires acquiring the crypt context twice.


jbaublitz:
Blocked by #2388

jbaublitz:
@mulkieran The wrapper type that I mentioned is pushed.

PR #2348, opened 2020-12-11
Prototype 1 - Use flag to indicate repair_sigblocks behavior

lleshchi:
Related: #2285 
Related: #2133

PR #2357, opened 2020-12-17
Prototype 2 - Pass closure to repair_sigblocks

lleshchi:
Related: #2285
Related: #2133

Compare with alternative prototype: #2348

### DIFF
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -12,7 +12,7 @@ use std::{
 
 use serde_json::Value;
 
-use libstratis::engine::BDA;
+use libstratis::engine::{StaticHeader, BDA};
 
 fn run(devpath: &str) -> Result<(), String> {
     let mut devfile = OpenOptions::new()
@@ -20,7 +20,11 @@ fn run(devpath: &str) -> Result<(), String> {
         .open(&devpath)
         .map_err(|the_io_error| format!("Error opening device: {}", the_io_error))?;
 
-    let bda = BDA::load(&mut devfile)
+    let read_results = StaticHeader::read_sigblocks(&mut devfile);
+    let header =
+        StaticHeader::repair_sigblocks(&mut devfile, read_results, StaticHeader::do_nothing);
+    // Handle printing unrepaired header here
+    let bda = BDA::load(header, &mut devfile)
         .map_err(|bda_load_error| format!("BDA detected but error found: {}", bda_load_error))?
         .ok_or_else(|| "No Stratis BDA metadata found".to_string())?;
     println!("{:#?}", bda);
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -23,7 +23,9 @@ fn run(devpath: &str) -> Result<(), String> {
     let read_results = StaticHeader::read_sigblocks(&mut devfile);
     println!("{:#?}", read_results);
     let header =
-        StaticHeader::repair_sigblocks(&mut devfile, read_results, StaticHeader::do_nothing);
+        StaticHeader::repair_sigblocks(&mut devfile, read_results, StaticHeader::do_nothing)
+            .map_err(|repair_error| format!("No valid StaticHeader found: {}", repair_error))?
+            .ok_or_else(|| "No valid Stratis signature found".to_string())?;
     let bda = BDA::load(header, &mut devfile)
         .map_err(|bda_load_error| format!("BDA detected but error found: {}", bda_load_error))?
         .ok_or_else(|| "No Stratis BDA metadata found".to_string())?;
--- a/src/bin/stratis_dumpmetadata.rs
+++ b/src/bin/stratis_dumpmetadata.rs
@@ -21,9 +21,9 @@ fn run(devpath: &str) -> Result<(), String> {
         .map_err(|the_io_error| format!("Error opening device: {}", the_io_error))?;
 
     let read_results = StaticHeader::read_sigblocks(&mut devfile);
+    println!("{:#?}", read_results);
     let header =
         StaticHeader::repair_sigblocks(&mut devfile, read_results, StaticHeader::do_nothing);
-    // Handle printing unrepaired header here
     let bda = BDA::load(header, &mut devfile)
         .map_err(|bda_load_error| format!("BDA detected but error found: {}", bda_load_error))?
         .ok_or_else(|| "No Stratis BDA metadata found".to_string())?;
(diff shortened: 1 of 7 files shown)


lleshchi:
@mulkieran I'm simply printing the read results with the generic pretty-print format. Is there a more specific output we were looking for in dumpmetadata?

mulkieran:
> @mulkieran I'm simply printing the read results with the generic pretty-print format. Is there a more specific output we were looking for in dumpmetadata?

Give whatever you're doing a try and see what it looks like. The Debug format is a fairly good choice, I think. You can post how it prints when you're done, and we can take a look at that. I think what's most important is that you make sure to show both signature buffers, clearly identified.

lleshchi:
@mulkieran Here is the current print format:
```
(
    StaticHeaderResult {
        bytes: Err(
            Io(
                Custom {
                    kind: UnexpectedEof,
                    error: "failed to fill whole buffer",
                },
            ),
        ),
        header: None,
    },
    StaticHeaderResult {
        bytes: Err(
            Io(
                Custom {
                    kind: UnexpectedEof,
                    error: "failed to fill whole buffer",
                },
            ),
        ),
        header: None,
    },
)
Error encountered: BDA detected but error found: Engine error: Unable to read data at sigblock locations.
```

src/bin/stratis_dumpmetadata.rs
```diff
@@ -12,15 +12,19 @@ use std::{
...
 fn run(devpath: &str) -> Result<(), String> {
     let mut devfile = OpenOptions::new()
         .read(true)
         .open(&devpath)
         .map_err(|the_io_error| format!("Error opening device: {}", the_io_error))?;
 
-    let bda = BDA::load(&mut devfile)
+    let read_results = StaticHeader::read_sigblocks(&mut devfile);
+    println!("{:#?}", read_results);
+    let header =
```
mulkieran, line 25: You'll want to scrutinize the error messages here a bit. If there was an error just reading the bytes, then you don't know that a BDA was detected.

This might be a reason to change the signature of BDA::load(). Right now, one parameter is ```StratisResult<Option<StaticHeader>>```. Maybe it should just be ```Option<StaticHeader>``` or ```StaticHeader```. Try having the argument be just ```StaticHeader```, and see how that goes.

mulkieran (changes requested):
Looks good! See comment about parameter type.

src/engine/strat_engine/liminal/setup.rs
```diff
@@ -38,7 +38,19 @@ use crate::{
 /// is returned.
 pub fn get_bdas(infos: &HashMap<DevUuid, &LStratisInfo>) -> StratisResult<HashMap<DevUuid, BDA>> {
     fn read_bda(info: &LStratisInfo) -> StratisResult<BDA> {
-        BDA::load(&mut OpenOptions::new().read(true).open(&info.ids.devnode)?)?.ok_or_else(|| {
+        let f = &mut OpenOptions::new().read(true).open(&info.ids.devnode)?;
```
mulkieran, line 41: The ```collect``` method is doing something pretty sophisticated in the code below: if read_bda() returns an error on reading any BDA, then that is the error that will be returned from ```get_bdas()```. So, whatever error you do return from ```read_bda()``` it should be at least as identifying as the BDA one, which says what the expected ids of the device are.

Your choices of what to interpret as an error an what not to are exactly correct.

mulkieran (changes requested):
This looks essentially correct. Please take a look at the collect() comment first, and then we can talk about the lifetimes question.

src/engine/strat_engine/metadata/static_header.rs
```diff
@@ -653,9 +769,12 @@ pub mod tests {
             let mut buf = Cursor::new(vec![0; buf_size]);
             sh_older.write(&mut buf, older_location).unwrap();
             sh_newer.write(&mut buf, newer_location).unwrap();
+            let read_results = StaticHeader::read_sigblocks(&mut buf);
```
mulkieran, line 772: Please tidy up the header comment on this closure.

src/engine/strat_engine/metadata/static_header.rs
```diff
@@ -209,20 +222,86 @@ impl StaticHeader {
...
+                },
+                Err(err) => StaticHeaderResult {
+                    bytes: Err(err.into()),
+                    header: None,
+                },
+            },
+        )
+    }
+
+    pub fn write_header<F>(
```
mulkieran, line 255: Put a comment on this method.

src/engine/strat_engine/metadata/static_header.rs
```diff
@@ -209,20 +222,86 @@ impl StaticHeader {
...
+        repair_location: MetadataLocation,
+    ) -> StratisResult<Option<StaticHeader>>
+    where
+        F: Seek + SyncAll,
+    {
+        sh.write(f, repair_location)?;
+        Ok(Some(sh))
+    }
+
+    // Replacement function for write_header
```
mulkieran, line 267: Might as well use ```///``` here, instead of ```//```.

src/engine/strat_engine/metadata/static_header.rs
```diff
@@ -209,20 +222,86 @@ impl StaticHeader {
         BDAExtendedSize::new(self.mda_size.bda_size().sectors() + self.reserved_size.sectors())
     }
 
+    pub fn read_sigblocks<F>(f: &mut F) -> (StaticHeaderResult, StaticHeaderResult)
```
mulkieran, line 225: Please add header doc here.

src/engine/strat_engine/metadata/static_header.rs
```diff
@@ -34,6 +34,17 @@ const STRAT_MAGIC: &[u8] = b"!Stra0tis\x86\xff\x02^\x41rh";
 
 const STRAT_SIGBLOCK_VERSION: u8 = 1;
 
+/// Data structure to hold results of reading and parsing a signature buffer.
+/// Invariant: bytes is Err <-> header == None, because if there was an error
+/// reading the data then there is no point in parsing.
+#[derive(Debug)]
+pub struct StaticHeaderResult {
```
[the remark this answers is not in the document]
mulkieran: Done.

src/bin/stratis_dumpmetadata.rs
mulkieran: Done!

mulkieran (changes requested):
A bunch of documentation requests, otherwise this looks ready.

lleshchi:
@mulkieran Could we discuss whether there is a way to remove f as an argument as well, and if not, why that is the case?

mulkieran (changes requested):
Looks good! Please close this and open a PR for final review.

PR #2362, opened 2020-12-24
Formalize range allocation with a new data structure

mulkieran:
Supercedes #2361

PR #2364, opened 2021-01-04
Formalize range allocation with a new data structure

mulkieran:
Supercedes #2362


mulkieran:
stable clippy failure only, merging.

PR #2339, opened 2020-12-04
Issue stratisd 2134

jbaublitz:
Supersedes #2302

PR #2372, opened 2021-01-08
Issue stratisd 2134

jbaublitz:
Supersedes #2339


cgwalters:
Is this the main stratis-in-initramfs PR?

jbaublitz:
Yes, it is.

jbaublitz:
Blocked by #2337

jbaublitz:
First install Fedora 33 into a virtual machine with two partitions that can be used for `/boot`, one for the host installation and one for the Stratis-based installation later. Do not add a mount point for the second `/boot` partition yet. From there, install the first Fedora 33 installation however you'd like.

One you've booted into your Fedora 33 VM after the install, you may install stratisd and stratis-cli through dnf. Create a pool and filesystem through Stratis and mount it wherever you'd like. I had the most success attaching persistent virtual drives to the VM and using those for the Stratis pool. Mount the unused boot partition under `/path/to/mount/boot`. Run

`anaconda --text --dirinstall=/path/to/mount/point/of/stratis/filesystem`

and install in text mode. Chroot into the Fedora installation on the Stratis filesystem.

From here, [clone my fork](https://github.com/jbaublitz/stratisd) and checkout the branch `issue-stratisd-2134`.

The dependencies that you will need are:
* clang
* cryptsetup-devel
* systemd-devel
* dbus-devel
* libblkid-devel
* rust
* cargo
* make

For the Rust builds on this branch run:
```
make build
make build-min
make build-extras
```

This will generate all of the executables that you need. You will need the following from the ./target/debug/ directory and copy them to these absolute paths:
* `stratisd-min` -> `/usr/libexec/stratisd-min`: This is the executable daemon that will run in the initramfs and shortly after the initramfs before D-Bus comes up.
* `stratis-min` -> `/usr/bin/stratis-min`: This is the Rust CLI client for use in the initramfs (where there's no Python) and shortly after the initramfs before D-Bus comes up.
* `stratisd` -> `/usr/libexec/stratisd`: This is the D-Bus enabled stratisd that will come up after D-Bus is available.
* `stratis_uuids_to_names` -> `/usr/lib/udev/stratis_uuids_to_names`: This is the helper executable that we use in our udev rule to generate our easy-to-use symlinks of /dev/stratis/pool_name/filesystem_name.
* `stratis-setup-generator` -> `/usr/lib/systemd/system-generators/stratis-setup-generator`: This generates a systemd service file in the initramfs based on kernel command line parameters to ensure that encrypted devices are unlocked and the root filesystem is set up prior to switching root and exiting the initramfs. Please note that the following kernel command line parameters are required for the root filesystem to come up properly:

* `root=[STRATIS_SYMLINK]`: This command line parameter is built into dracut and you simply need to specify `/dev/stratis/[POOLNAME]/[FILESYSTEMNAME]` to indicate that this is the root filesystem.
* `stratis.rootfs.uuids_paths`: This kernel command line parameter can be specified multiple times and should be set for every disk in the pool housing the root filesystem. The value should be `/dev/disk/by-uuid/[UUID_OF_DISK_IN_STRATIS_POOL]`.
* `stratis.rootfs.pool_uuid`: This should be the UUID of the pool that contains the root filesystem.

This will provide the necessary executables for stratisd to run both during and after boot. For configuration files, copy the following files:

* `systemd/stratisd-min-postinitrd.service` -> `/usr/lib/systemd/system/stratisd-min-postinitrd.service`: This is the service file for stratisd-min enabling non-root filesystem /etc/fstab Stratis mounts.
* `systemd/stratisd-min.service` -> `/usr/lib/systemd/system/stratisd-min.service`: This is the service file for stratisd-min in the initramfs.
* `systemd/stratisd.service` -> `/usr/lib/systemd/system/stratisd.service`: This is the service file for stratisd once the D-Bus is available.
* `systemd/stratis-fstab-setup` -> `/usr/lib/systemd/stratis-fstab-setup`: This is the executable that performs handling for Stratis filesystems listed in /etc/fstab.
* `systemd/stratis-fstab-setup@.service` -> `/usr/lib/systemd/system/stratis-fstab-setup@.service`: This should be referenced in all Stratis filesystem fstab entries for non-root filesystems like so: add the options:

`x-systemd.requires=stratis-fstab-setup@[POOL_UUID].service,x-systemd.after=stratis-fstab-setup@[POOL_UUID].service`

* `udev/11-stratisd.rules` -> `/usr/lib/udev/rules.d/11-stratisd.rules`: This is the udev rule for our symlinks.
* `dracut/90-stratis.conf` -> `/usr/lib/dracut/dracut.conf.d/90-stratis.conf`: This enables the dracut module when building the initramfs.
* `dracut/90stratis` -> `/usr/lib/dracut/modules.d/90stratis`: This is a directory and needs to be copied recursively. Note that while we have removed the old, out of date Stratis dracut module upstream, it may not have been removed in current Fedora releases, so if you find /usr/lib/dracut/modules.d/90stratis already exists, you can safely remove the contents of that directory. Hopefully this problem will be resolved in the next release or two.

Regenerate the initramfs using dracut, set up the grub configuration to specify the kernel command line parameters above and the second boot partition pointing to your rebuilt dracut initramfs. Put your mount entries in either /etc/fstab or mount files. From here, you should be able to boot into the installation.

jbaublitz:
cc @cgwalters

I've attached the set up instructions.

cgwalters:
> You will need the following from the ./target/debug/ directory and copy them to these absolute paths:

Let's put all that knowledge under `make install`?  Doing it by hand is just way too fragile; xref https://github.com/stratis-storage/stratisd/pull/2400

jbaublitz:
@cgwalters How does c8c0a6aa4f5472c259539349eedff048ad1ef6f0 look?

jbaublitz:
@cgwalters Okay this should be resolved now.

jbaublitz:
Okay! This is ready for review now.

mulkieran (changes requested):
Took a quick pass...one very minor comment.

Makefile
```diff
@@ -235,6 +274,7 @@ clippy:
 	license
 	outdated
 	release
+	build-min
```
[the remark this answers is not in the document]
jbaublitz: Ah yes! This should be resolved now.

.github/workflows/main.yml
```diff
@@ -62,10 +66,18 @@ jobs:
...
             image: ubuntu:groovy
+          - task: make -f Makefile release-min
+            toolchain: stable
+            components: cargo
+            image: ubuntu:groovy
           - task: make -f Makefile build
             toolchain: stable
             components: cargo
             image: fedora:32
+          - task: make -f Makefile build-min
```
mulkieran, line 77: I think we should consider dropping the ```build-no-default``` Makefile task and the GitHub Actions directed at it. It was put in from a time where stratisd was never released without the defaults and wasn't really interesting without them. It was just useful for catching the occasional error when something didn't get the ```[cfg(dbus)]``` flag put on that should or vice-versa. But at this point, now that there are actual targets which build w/out dbus enabled, I think its usefulness is gone.

src/engine/structures.rs
```diff
@@ -43,6 +43,26 @@ where
     }
 }
 
+impl<U, T> Extend<(Name, U, T)> for Table<U, T>
```
mulkieran, line 46: The code is good, but I'm convinced it's dead; should remove or add tests.

mulkieran (changes requested):
Still working, two minor requests, continues to look good.

src/engine/structures.rs
jbaublitz: Whoops! It looks like the code I was using that for got removed.

mulkieran:
Please rebase and make the additions you were planning for once Clevis PR went in.

dracut/90stratis-clevis/module-setup.sh
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/bash
+# This dracut module requires the kernel command line parameter rd.neednet=1 if
```
mulkieran, line 2: Has this comment become obsolete or are we sticking with the neednet=1 requirement?

mulkieran (changes requested):
Everything continues to look fine. I'm assuming that under the new hard-link regime we'll want to combine the two systemd generators into a single executable? As far as I can tell, though, all other executable should stay distinct.

There are major conflicts...please feel free to drop this PR, start a new one, squash all the commits, etc. Whatever makes it easier.

dracut/90stratis-clevis/module-setup.sh
jbaublitz: This is still necessary but not in all cases so it is up to the user to determine if they need it (only when using Tang).

PR #2374, opened 2021-01-11
Update to devicemapper 0.29.0

jbaublitz:
Supersedes #2341

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ required-features = ["extras"]
 [dependencies]
 clap = "2"
 nix = "0.18"
-devicemapper = "0.28"
+devicemapper = "0.29"
 crc = "1"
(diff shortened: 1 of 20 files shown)


Cargo.toml
```diff
@@ -24,7 +24,7 @@ required-features = ["extras"]
 [dependencies]
 clap = "2"
 nix = "0.18"
-devicemapper = "0.28"
+devicemapper = "0.29.0"
```
mulkieran, line 27: Please drop the patch version here.

mulkieran (changes requested):
Please feel free to force push the requested change, and then this should be good to go.

Cargo.toml
jbaublitz: Resolved.

mulkieran (approved):

PR #2379, opened 2021-01-17
Makefile: Add an install rule

cgwalters:
Drain the install logic from the spec file:
https://src.fedoraproject.org/rpms/stratisd/blob/master/f/stratisd.spec#_74
to upstream.

This is better because it keeps the logic for where files go
aligned with the code, not with spec files which live in
a separate git repository on a separate lifecycle.  It allows sharing
installation logic across distributions (and even just Red Hat
has 3 distributions).

It will allow us to avoid the hacky way we `mv` the binaries
after the RPM `%cargo_install` macro.

It allows more easily building from source upstream without
going through RPM at all.

It makes it easier to move towards auto-generated spec
files.

In short: it's better ;)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,14 @@ else
   TARGET_ARGS = --target=${TARGET}
 endif
 
+DESTDIR ?=
+PREFIX ?= /usr
+LIBEXECDIR ?= $(PREFIX)/libexec
+DATADIR ?= $(PREFIX)/share
+UDEVDIR ?= $(PREFIX)/lib/udev
+MANDIR ?= $(DATADIR)/man
+UNITDIR ?= $(PREFIX)/lib/systemd/system
+
 RUST_2018_IDIOMS = -D bare-trait-objects \
                    -D ellipsis-inclusive-range-patterns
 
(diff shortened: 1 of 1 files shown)


cgwalters:
More at https://github.com/cgwalters/build-api

mulkieran:
ok to test

mulkieran (approved):
I'm willing to give this a go, on the basis that getting this upstream is better than having it in spec files, even unified ones. We can work on calling if correctly from spec files later. I hope this is a good model: https://src.fedoraproject.org/rpms/rust-bootupd/blob/master/f/rust-bootupd.spec#_61.

cgwalters:
> I hope this is a good model: https://src.fedoraproject.org/rpms/rust-bootupd/blob/master/f/rust-bootupd.spec#_61.

Yep, that's the idea - the spec file just invokes the upstream logic.

jbaublitz (approved):

drckeefe (approved):

bgurney-rh (approved):
Looks good.  (Please ignore the failure on stratisd-nonrust-2; it's a spurious test failure.)

jbaublitz:
@cgwalters I had a question about this PR. From the looks of this, if `make release` isn't run before `make install`, `make install` will install the debug version because it depends on `make build` not `make release`. Was this intended? If not, I'm going to switch this over to `install: release`.

PR #2381, opened 2021-01-18
mdv: Use openat-ext to write state JSON

cgwalters:
This uses `O_TMPFILE`, which means we avoid having a `.temp`
file until the very end (when the file is fully written).

We also avoid serializing the state into memory; instead
serde directly writes to the file descriptor.

At some point hopefully the Linux kernel will pick up
these patches
https://lwn.net/Articles/707504/
which means there will never be any temporary file at
all.  And when that happens I'll update `openat-ext`
to use it and projects like this one will transparently
benefit.

(Perhaps at some point the Rust stdlib will gain this too)


src/engine/strat_engine/thinpool/mdv.rs
```diff
@@ -128,31 +130,13 @@ impl MetadataVol {
...
-        let data = serde_json::to_string(&fs.record(name, uuid))?;
-        let path = self
-            .mount_pt
-            .join(FILESYSTEM_DIR)
-            .join(uuid.to_simple_ref().to_string())
-            .with_extension("json");
-
-        let temp_path = path.with_extension("temp");
-
         let _mount = MountedMDV::mount(self)?;
```
cgwalters, line 133: Actually also related to this - the `MountedMDV` could hold an `openat::Dir` directory file descriptor for the mount point, which means it would more properly "own" the data and we wouldn't have to use the underscore `_` prefix to say "keep this alive even though it's not used".

jbaublitz:
ok to test

jbaublitz:
ok to test

mulkieran (commented):
I'm happy with this, modulo conflicts to be resolved.

jbaublitz:
This looks good to me too.

cgwalters:
Rebased :surfer:

cgwalters:
One caveat I'd add is while I've invested a not-small amount of time in the `openat-ext` crate, I recently came across https://github.com/bytecodealliance/cap-std which is much better than `openat` in various ways and am considering a port to it.  But we have enough usage of `openat-ext` in some of the coreos/ Rust code that I'm not going to stop maintaining it overnight - and a benefit of "cap-std" is (like async-std) the API is intended to closely mirror "std" instead of openat's more custom interface, so porting to it should feel familiar.

jbaublitz (approved):

jbaublitz:
@cgwalters @mulkieran Are you okay with me pulling this branch, making the revisions, putting it up for review, and then squashing the commits under @cgwalters's info for the commit? I'm sorry, reviewing this completely slipped my mind.

Cargo.toml
```diff
@@ -48,6 +48,8 @@ regex = "1"
 base64 = "0.13"
 sha-1 = "0.9"
 either = "1"
+openat-ext = "0.1.10"
```
mulkieran, line 51: For openat-ext please drop the patch release number, here, but make it 0.1.6 in the Cargo.lock (so as not to get ahead of Fedora, or @cgwalters can bump the Fedora release if necessary).

Cargo.toml
```diff
@@ -48,6 +48,8 @@ regex = "1"
...
+openat = "0.1.19"
```
mulkieran, line 52: Patch release is good, but just pin it in the Cargo.lock, not here.

mulkieran (changes requested):
@jbaublitz I'm fine if you rebase it and  push it in. Just requesting the two changes I noted.

mulkieran:
@jbaublitz Please rebase this one and see how it reacts to our new CI.

PR #2382, opened 2021-01-20
Assert that libcryptsetup is always being called from the same thread

jbaublitz:
Related to #2277

### DIFF
--- a/src/engine/strat_engine/backstore/crypt.rs
+++ b/src/engine/strat_engine/backstore/crypt.rs
@@ -8,6 +8,7 @@ use std::{
     io,
     path::{Path, PathBuf},
     sync::Arc,
+    thread::{current, ThreadId},
 };
 
 use base64::{decode, encode_config, CharacterSet, Config};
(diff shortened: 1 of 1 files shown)


jbaublitz:
I'm using this PR as a way to explore putting in a canary so that if we make future changes to how we're using the crypt module, we will be notified immediately if we attempt to accidentally access it from multiple threads as Rust will not protect us from that in this case.

jbaublitz:
@bgurney-rh The failure appears to be a udev problem.

bgurney-rh:
OK; I'm trying a re-run on a different test machine.

jbaublitz:
@mulkieran Given that we've determined that the way we're currently doing things is safe even in the multithreading PR, do we want to put this into review to make sure that we never violate this contact?

mulkieran:
> @mulkieran Given that we've determined that the way we're currently doing things is safe even in the multithreading PR, do we want to put this into review to make sure that we never violate this contact?

Yes, I think this is fine. Pls go ahead.

mulkieran (approved):

jbaublitz:
This may be obsoleted by #2388. I'm leaving this open until we merge that one.

PR #2385, opened 2021-01-21
Replace '~' with '-' in version strings

mulkieran:
Miscellaneous improvements to update_cargo_crates.py.


mulkieran:
This is more for an FYI than anything else. The only interesting thing is that Fedora was using SemVer standard version strings in https://kojipkgs.fedoraproject.org/repos/rawhide/latest/x86_64/pkglist for all the crates that we were interested in until recently. I don't know if running into invalid version strings now is just a coincidence or if they changed some rule.

Look for this string "toplink/packages/rust-clap/3.0.0~beta.2/1.fc34/noarch/rust-clap+lazy_static-devel-3.0.0~beta.2-1.fc34.noarch.rpm". It's the '~' before the "beta" that is non-standard. The corresponding cargo crate just has a '-'. So, in Cargo.toml, we would require the crate version with the '-', if we wanted to be that specific.

mulkieran:
In any case, it looks like cargo-outdated ignores beta releases, so our script will never recommend to update to them. So all that seems pretty safe.

drckeefe (approved):

PR #2391, opened 2021-01-29
Make target for upload-vendor-asset and accompanying script

### LINKED ISSUE #2387: Script process for making a new release from a tag on GitHub
It can be uploaded as part of the notes, but there is no manual way to upload it as an _Asset_. The problem w/ uploading it as part of the notes is that its URL in unpredictable, currently it's https://github.com/stratis-storage/stratisd/files/5859142/stratisd-2.3.0-vendor.tar.gz, note the randomly generated number.

GitHub has some information on Releases: https://docs.github.com/en/github/administering-a-repository/about-releases#about-releases, and this appears to indicate that even though there is no manual way, there is some programmatic way to upload an Asset: https://docs.github.com/en/rest/reference/repos#releases. 

The two zip files that are under the "Assets" heading are actually the ```tarball_url``` and ```zipball_url``` respectively.

Magic method found: https://docs.github.com/en/rest/reference/repos#upload-a-release-asset. So the trick is to already have a release, so that the release id can be obtained. Then use the release id to specify the release to attach the asset to.

Perhaps we can find a script or GitHub Action setup to do this for us. Here are some GitHub Actions:
- https://github.com/marketplace/actions/upload-a-release-asset
- https://github.com/alexellis/upload-assets

And a blog post about it: https://trstringer.com/github-actions-create-release-upload-artifacts/

I think this is doable by making a GitHub workflow, that we manually trigger that creates a release and that adds the additional Asset. We can create the release as a draft.

- [x] Create Makefile target for creating vendored zip file with extension ```tar.gz``` and associated compression app. Otherwise, format of the name can be anything desired, like ```stratisd-vendor.tar.gz```. Do not require version string in name at this time. I don't think the target should require any sub-targets; it should do the whole task.
- [x] Use Makefile target to create a GitHub Action to both make a release and add the vendored packages Asset. Things to do:
   * Give all Assets the proper version names that we like best. This is optional, so long as the names are known the files can always be downloaded and renamed.
   * Make the release a Draft release. That way, the releaser, probably me, can review the release before hitting the button.
   * Have the release follow the format of the previous releases, which is intentionally quite simple. Just a name and a link to the changelog.
   * Make the GitHub Action, manually, rather than automatically, triggered.

jbaublitz:
Related #2387

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -160,6 +160,7 @@ jobs:
               networkx==2.5
               requests==2.22.0
               semantic_version==2.8.4
+              PyGithub==1.54.1
             task: >
               (cd code_maintenance; make -f Makefile lint) &&
               (cd developer_tools; make -f Makefile lint) &&
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -160,7 +160,7 @@ jobs:
               networkx==2.5
               requests==2.22.0
               semantic_version==2.8.4
-              PyGithub==1.54.1
+              PyGithub==1.45
             task: >
               (cd code_maintenance; make -f Makefile lint) &&
               (cd developer_tools; make -f Makefile lint) &&
--- a/code_maintenance/create_release.py
+++ b/code_maintenance/create_release.py
@@ -17,10 +17,10 @@
 Uploads the vendored dependency tarball as a release asset.
 """
 
-from getpass import getpass
 import os
-import sys
 import re
+import sys
+from getpass import getpass
 
 from github import Github
 
--- a/code_maintenance/create_release.py
+++ b/code_maintenance/create_release.py
@@ -48,13 +48,13 @@ def main():
     repo = git.get_repo("stratis-storage/stratisd")
     branch_name = repo.default_branch
 
-    repo.create_git_release(
+    release = repo.create_git_release(
         "v%s" % release_version,
         "Version %s" % release_version,
         "Version %s of stratisd" % release_version,
+        True,
     )
 
-    release = repo.get_latest_release()
     tag_name = release.tag_name
     release.upload_asset(
         "stratisd-vendor.tar.gz", "stratisd-vendor-%s.tar.gz" % tag_name
--- a/code_maintenance/create_release.py
+++ b/code_maintenance/create_release.py
@@ -51,8 +51,8 @@ def main():
     release = repo.create_git_release(
         "v%s" % release_version,
         "Version %s" % release_version,
-        "Version %s of stratisd" % release_version,
-        True,
+        "See changelog here: https://github.com/stratis-storage/stratisd/blob/master/CHANGES.txt",
+        draft=True,
     )
 
     tag_name = release.tag_name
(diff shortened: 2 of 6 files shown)


code_maintenance/upload_vendor_asset.py
```diff
@@ -0,0 +1,48 @@
...
+from github import Github
+
+
+def main():
+    """
+    Main function
+    """
+    api_key = os.environ.get("GITHUB_API_KEY")
+    if api_key is None:
+        raise RuntimeError("GITHUB_API_KEY environment variable is required")
```
mulkieran, line 32: What kind of key is this? The reason I'm unsure is that when I was trying out the API I authenticated as me. Would this be a GitHub Secret instead?

mulkieran (commented):
I just have one question...

the title of the issue is obsolete, though, so I've updated it. What I'ld really like is to get the whole release process scripted from when the tag is pushed, through making the release, to the final step for now which is when the vendored tarball is uploaded. That way is the most benefit.

This is a fine start, and will help us and also downstream packaging. We can go ahead and push this, which would be fine, or we can hold back and try to address the larger goal in this PR. Either one is fine w/ me.

code_maintenance/upload_vendor_asset.py
jbaublitz: This is a personal access token.

jbaublitz:
If you want to script the whole release process, I'm happy to wait on this!

mulkieran:
Cool, are you able to carry on then?

jbaublitz:
Yes, I'll look into how I can create a release through the Github API without having to clone the repo.

code_maintenance/create_release.py
```diff
@@ -0,0 +1,69 @@
...
+
+    git = Github(api_key)
+
+    repo = git.get_repo("stratis-storage/stratisd")
+    branch_name = repo.default_branch
+
+    release = repo.create_git_release(
+        "v%s" % release_version,
+        "Version %s" % release_version,
+        "Version %s of stratisd" % release_version,
```
mulkieran, line 54: For the message, I'ld like to stick w/ the link to the changelog, which is always the same:

```See https://github.com/stratis-storage/stratisd/blob/master/CHANGES.txt.```

**Review by mulkieran — CHANGES_REQUESTED** [review]
A few small requests + please set up linting for the new file in ```check.py``` and ```Makfile```. After that it should be ready-to-merge. We can work to build on it later, perhaps next to have it explicitly write the correct tag for the blackbox tests into the message.

**Comment by jbaublitz** [comment]
Ready for review again.

**Review by mulkieran — CHANGES_REQUESTED** [review]
Aside from CI failure, this is ready.

File: .github/workflows/main.yml [hunk 3]
```diff
@@ -160,6 +160,7 @@ jobs:
               networkx==2.5
               requests==2.22.0
               semantic_version==2.8.4
+              PyGithub==1.54.1
```
mulkieran, line 163: Please change version to 1.45, it seems to be standard on current reference version of Fedora.

mulkieran (changes requested):
That's all...

.github/workflows/main.yml
jbaublitz: I made a similar comment in the `ci` pull request. 1.54 is the latest available with pip.

PR #2397, opened 2021-02-10
Dynamic Clevis executable check

jbaublitz:
Related to stratis-storage/project#247

### DIFF
--- a/src/engine/strat_engine/cmd.rs
+++ b/src/engine/strat_engine/cmd.rs
@@ -102,7 +102,7 @@ lazy_static! {
     .iter()
     .cloned()
     .collect();
-    pub static ref USE_CLEVIS: bool = CLEVIS_EXECS.iter().all(|exec| find_binary(exec).is_some());
+    static ref USE_CLEVIS: bool = CLEVIS_EXECS.iter().all(|exec| find_binary(exec).is_some());
 }
 
 /// Verify that all binaries that the engine might invoke are available at some
--- a/src/engine/strat_engine/cmd.rs
+++ b/src/engine/strat_engine/cmd.rs
@@ -56,7 +56,6 @@ const CLEVIS_UNLOCK: &str = "clevis-luks-unlock";
 const CLEVIS_DECRYPT: &str = "clevis-decrypt";
 const CLEVIS_DECRYPT_TANG: &str = "clevis-decrypt-tang";
 const CLEVIS_DECRYPT_TPM2: &str = "clevis-decrypt-tpm2";
-const CLEVIS_ENCRYPT: &str = "clevis-encrypt";
 const CLEVIS_ENCRYPT_TANG: &str = "clevis-encrypt-tang";
 const CLEVIS_ENCRYPT_TPM2: &str = "clevis-encrypt-tpm2";
 const JOSE: &str = "jose";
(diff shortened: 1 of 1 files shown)


src/engine/strat_engine/cmd.rs
```diff
@@ -49,9 +49,46 @@ const UDEVADM: &str = "udevadm";
...
+const JOSE: &str = "jose";
+const JQ: &str = "jq";
+const CRYPTSETUP: &str = "cryptsetup";
+const CURL: &str = "curl";
+const TPM2_CREATEPRIMARY: &str = "tpm2_createprimary";
+const TPM2_UNSEAL: &str = "tpm2_unseal";
+const TPM2_LOAD: &str = "tpm2_load";
+const MKTEMP: &str = "mktemp";
+
+const CLEVIS_EXECS: &[&str] = &[
```
mulkieran, line 71: Could you put a comment here, indicating how you came up with this list?

src/engine/strat_engine/cmd.rs
```diff
@@ -61,14 +98,11 @@ lazy_static! {
...
         (XFS_GROWFS.to_string(), find_binary(XFS_GROWFS)),
-        (CLEVIS.to_string(), find_binary(CLEVIS)),
-        (CLEVIS_BIND.to_string(), find_binary(CLEVIS_BIND)),
-        (CLEVIS_UNBIND.to_string(), find_binary(CLEVIS_UNBIND)),
-        (CLEVIS_UNLOCK.to_string(), find_binary(CLEVIS_UNLOCK)),
     ]
     .iter()
     .cloned()
     .collect();
+    static ref USE_CLEVIS: bool = CLEVIS_EXECS.iter().all(|exec| find_binary(exec).is_some());
```
mulkieran, line 105: There's something we overlooked when we first added the clevis commands: for all the other commands we use the absolute path that is found by ```find_binary``` to invoke the binary. I believe that the reason was that the ```$PATH``` environment variable had been found to be missing in certain situations. We should have this solution do that also.

mulkieran (changes requested):
Just two requests...

src/engine/strat_engine/cmd.rs
jbaublitz: The way that I came up with this was from the dracut modules that Clevis provides and my development of the dracut modules for booting from Stratis filesystems as the root filesystem. Do you want me to say something to that effect? For example, `mktemp` was not included in the distributed Clevis dracut modules, but that was required for our altered, full install of Clevis whereas it does not appear to be required for the Clevis dracut module architecture because they only add a subset of Clevis functionality.

mulkieran: Yes, please do. Even if it needs to be rather long comment, it will be good to have that record right in the source.

jbaublitz:
This is ready again.

jbaublitz:
Actually I'm going to clean up one part.

jbaublitz:
Okay, this is ready.

mulkieran (changes requested):
Looks good! Please close this and open a new PR for final review.

jbaublitz:
@mulkieran Before I put up a final PR, do you want to take a look at the failed udev test? It looks like a timing issue to me (not all of the encrypted devices were processed as encrypted devices).

mulkieran:
@jbaublitz I saw the error and I couldn't see how your change could have been related.

PR #2402, opened 2021-02-12
stratisd: clippy fixes for Rust 1.50 (in progress)

bgurney-rh:


mulkieran:
Closing...we're trying a new system that avoids eagerly pulling in new pedantic lints that are likely to prove mostly frivolous and require too many code changes with every new rust release: #2403.

PR #2404, opened 2021-02-15

jbaublitz:
Supersedes #2337 
Closes https://github.com/stratis-storage/project/issues/216.


mulkieran:

jbaublitz:
Done

mulkieran:
@jbaublitz Please rebase so I can rebase. My changes in #2415 are inducing a conflict.

jbaublitz:
Rebased.

src/engine/strat_engine/thinpool/filesystem.rs
```diff
@@ -291,6 +291,22 @@ impl StratFilesystem {
 }
 
 impl Filesystem for StratFilesystem {
+    fn send_udev_change(&self) -> StratisResult<()> {
```
mulkieran, line 294: We already do these actions, slightly differently, in devlinks.rs. Can you factor that out, since it becomes a shared action now, with this PR?

Also, did you decide against only triggering the rule on a change event? That still seems sensible to me, to avoid an initial failure of the script.

mulkieran (changes requested):

src/engine/strat_engine/thinpool/filesystem.rs
jbaublitz: I decided against only doing it on a change event because `mkfs.xfs` will cause a change event even on filesystem creation so we should potentially ask the question of whether we should be logging at any level higher than debug so that people can filter it out if they're not actively debugging a problem.

src/dbus_api/connection.rs
```diff
@@ -89,6 +93,13 @@ impl DbusConnectionHandler {
                             warn!("Failed to send reply to D-Bus client");
                         }
                     }
+                } else {
+                    let reply = default_reply(&msg);
+                    if let Some(r) = reply {
+                        if connection.send(r).is_err() {
+                            warn!("Failed to send reply to D-Bus client");
+                        }
+                    }
```
mulkieran, line 102: Can you make this obviously impossible to hang? Right now it still would hang if ```default_reply``` changed its definition, or we for some reason started matching on something other than just a method call.

mulkieran (changes requested):

src/dbus_api/connection.rs
jbaublitz: I'm not really sure how we can accomplish that. Are you suggesting simply sending an empty message?

src/engine/strat_engine/thinpool/filesystem.rs
mulkieran: Ok! We decided in standup that the best thing would be to reduce the severity of the messages from the udev script to ```debug```.

src/dbus_api/connection.rs
mulkieran: We'll let this go for now, since we believe that it solves the problem of hanging completely at this time.

mulkieran (changes requested):
And also, in standup, we decided we'ld move the calculation of the object paths to remove from the D-Bus layer into the tree-handler rather than leaving it where it is, in the message handler.

src/engine/strat_engine/thinpool/filesystem.rs
jbaublitz: @mulkieran I'm not sure that this is actually a shared action just because the only shared piece is really writing to the udev file. The way the path to the udev file is generated is entirely different. Do you want me to just factor out the `.write_all("change".as_bytes())` call?

mulkieran: > @mulkieran I'm not sure that this is actually a shared action just because the only shared piece is really writing to the udev file. The way the path to the udev file is generated is entirely different. Do you want me to just factor out the `.write_all("change".as_bytes())` call?

I see your point. And it's a toss-up whether just the writing part is worth abstracting yet. Let's wait until the next time we find a need to generate a synthetic event to worry about it.

mulkieran (changes requested):
Please pick up the last two commits found here: https://github.com/stratis-storage/stratisd/pull/2422.

PR #2407, opened 2021-02-18
Jbaublitz issue stratisd 2277 fresh

### LINKED ISSUE #2219: filesystem.Devnode does not emit PropertiesChanged when calling SetName

mulkieran:
Also closes #2219.
Supercedes #2346.


mulkieran:
Looks like a real failure to me, i.e., not a problem with the containerized version.

jbaublitz:
My only comment is can't the invalidation and update signals be sent at the same time? They're both fields of the same signal. Perhaps we should ask Marius about this?

mulkieran:
> My only comment is can't the invalidation and update signals be sent at the same time? They're both fields of the same signal. Perhaps we should ask Marius about this?

We should definitely ask. @mvollmer: This PR implements a solution to https://github.com/stratis-storage/stratisd/issues/2219. But when for example, a filesystem name is changed, two signals are put out: one a property changed signal for the Name property and the other a property invalidated signal for the Devnode property. Since the signal can carry both changed and invalidated parts, is it better for both to be part of the same signal?

For a pool name change, things are different. The pool name gets a property changed signal, and its filesystems get a property invalidated signal, so there's no way of combining those.

mulkieran:
@mvollmer The relevant commits are: https://github.com/stratis-storage/stratisd/pull/2407/commits/f615c96029fa7a19e8c20db526afad7dfb8be5a0, https://github.com/stratis-storage/stratisd/pull/2407/commits/2510cf387b21c6710c1f706b7a291ad2c865d900, and https://github.com/stratis-storage/stratisd/pull/2407/commits/feb73a3b54380c35240b4992a15a03fe16cf41cd.

mulkieran:
@mvollmer Also, on a filesystem name change, there is just one signal for the name change and the invalidated devnode.

mulkieran:
@jbaublitz Please take a look at the last change also. I was concerned that the number of messages sent on the channel would increase in proportion to the number of filesystems in the pool.

mvollmer:
> Since the signal can carry both changed and invalidated parts, is it better for both to be part of the same signal?

I'd say it would be better, but not very important.  It would be better just to avoid there being a time when the name and devnode are inconsistent, not because of D-Bus overhead or anything like that.

What I would consider important is that all change notifications go out before the return message of the SetName method call that triggered the changes. That way clients can rely on having seen the expected changes already when the method call finishes.

mvollmer:
This works as expected with https://github.com/cockpit-project/cockpit/pull/14109.

The PropertiesChanged signal is emitted after the method return message, but I don't think it will be a problem in practice.

mvollmer:
Random observations, probably unrelated:

- after playing around with this, I had stratisd sitting at around 95% CPU indefinitely.
- stratisd does not reply to introspection requests for "/", "/org", or "/org/storage", only for "/org/storage/stratis2" and below.  This renders "busctl tree org.storage.stratis2" useless, for example.

mvollmer:
>     * after playing around with this, I had stratisd sitting at around 95% CPU indefinitely.

This only happens with this branch, the `develop-2.3.0` branch does not show excessive CPU usage.

mvollmer:
> > * after playing around with this, I had stratisd sitting at around 95% CPU indefinitely.
> 
> This only happens with this branch, the `develop-2.3.0` branch does not show excessive CPU usage.

and it goes to 95% right after startup, there is no need to perform any operations (such as renaming a fsys).

jbaublitz:
@mvollmer @mulkieran My PR #2404 does not have the CPU usage problem so I'm wondering if this was introduced in some of the code edits in this PR or whether the most recent edit that I made to the devicemapper tokio polling fixed that and it hasn't yet been applied to this PR yet. I'll take more of a look into this.

jbaublitz:
After rolling back the devicemapper changes, this is still not a problem for my PR. @mulkieran Would you be willing to do an strace on stratisd and see where the rogue polling is coming from? That's my guess for what's causing this because I fixed a similar problem early on in my PR (I think it was in the udev polling).

jbaublitz:
> Random observations, probably unrelated:
> 
>     * after playing around with this, I had stratisd sitting at around 95% CPU indefinitely.
> 
>     * stratisd does not reply to introspection requests for "/", "/org", or "/org/storage", only for "/org/storage/stratis2" and below.  This renders "busctl tree org.storage.stratis2" useless, for example.

@mvollmer Do most D-Bus services make the root object `/` as opposed to what we have? I feel like I've traditionally seen the service name translated from dots to slashes to the top object but I might be wrong. If we are doing what most projects do, I'm wondering if this is related to the D-Bus library that we're using.

mvollmer:
> Would you be willing to do an strace on stratisd and see where the rogue polling is coming from?

Here is a log:

[stratisd-loop.log](https://github.com/stratis-storage/stratisd/files/6034403/stratisd-loop.log)

Commit e3ee692fed30578464695aa73fa13c94b8533b10 seems to be the culprit.

mvollmer:
> Do most D-Bus services make the root object / as opposed to what we have? I feel like I've traditionally seen the service name translated from dots to slashes to the top object but I might be wrong. 

Your object tree is just fine, as far as I can tell, but introspection is supposed to work for all objects in the tree, including "/", "/org", and "/org/storage".  This is so that tools like "busctl tree" can work.

```
# busctl introspect org.storage.stratis2 /
Failed to introspect object / of service org.storage.stratis2: Connection timed out
# busctl introspect org.storage.stratis2 /org
Failed to introspect object /org of service org.storage.stratis2: Connection timed out
# busctl introspect org.storage.stratis2 /org/storage
Failed to introspect object /org/storage of service org.storage.stratis2: Connection timed out
# busctl introspect org.storage.stratis2 /org/storage/stratis2
NAME                                    TYPE      SIGNATURE   RESULT/VALUE  FLAGS
org.freedesktop.DBus.Introspectable     interface -           -             -
.Introspect                             method    -           s             -
org.freedesktop.DBus.ObjectManager      interface -           -             -
[...]
```

Compare with any other D-Bus service:

```
# busctl introspect org.freedesktop.systemd1 /
NAME                                TYPE      SIGNATURE RESULT/VALUE FLAGS
org.freedesktop.DBus.Introspectable interface -         -            -
.Introspect                         method    -         s            -
org.freedesktop.DBus.Peer           interface -         -            -
.GetMachineId                       method    -         s            -
[...]
# busctl introspect org.freedesktop.systemd1 /blah
Failed to introspect object /blah of service org.freedesktop.systemd1: Unknown object '/blah'.
```

Note how systemd replies for "/" and that it gives a proper error for objects that don't exist in its tree (instead of timing out as stratisd does).

The introspection reply for "/" also includes a list of children.  They are not shown by "busctl introspect", but are used by "busctl tree" to discover all objects.  This is the `<node name=\"org\"/>\n</node>` part in the output below.

```
# busctl call org.freedesktop.systemd1 / org.freedesktop.DBus.Introspectable Introspect
s "<!DOCTYPE node PUBLIC \"-//freedesktop//DTD D-BUS Object Introspection 1.0//EN\"\n\"http://www.freedesktop.org/standards/dbus/1.0/introspect.dtd\">\n<node>\n <interface name=\"org.freedesktop.DBus.Peer\">\n  <method name=\"Ping\"/>\n  <method name=\"GetMachineId\">\n   <arg type=\"s\" name=\"machine_uuid\" direction=\"out\"/>\n  </method>\n </interface>\n <interface name=\"org.freedesktop.DBus.Introspectable\">\n  <method name=\"Introspect\">\n   <arg name=\"data\" type=\"s\" direction=\"out\"/>\n  </method>\n </interface>\n <interface name=\"org.freedesktop.DBus.Properties\">\n  <method name=\"Get\">\n   <arg name=\"interface\" direction=\"in\" type=\"s\"/>\n   <arg name=\"property\" direction=\"in\" type=\"s\"/>\n   <arg name=\"value\" direction=\"out\" type=\"v\"/>\n  </method>\n  <method name=\"GetAll\">\n   <arg name=\"interface\" direction=\"in\" type=\"s\"/>\n   <arg name=\"properties\" direction=\"out\" type=\"a{sv}\"/>\n  </method>\n  <method name=\"Set\">\n   <arg name=\"interface\" direction=\"in\" type=\"s\"/>\n   <arg name=\"property\" direction=\"in\" type=\"s\"/>\n   <arg name=\"value\" direction=\"in\" type=\"v\"/>\n  </method>\n  <signal name=\"PropertiesChanged\">\n   <arg type=\"s\" name=\"interface\"/>\n   <arg type=\"a{sv}\" name=\"changed_properties\"/>\n   <arg type=\"as\" name=\"invalidated_properties\"/>\n  </signal>\n </interface>\n <node name=\"org\"/>\n</node>\n"
```

And, yeah, I would expect all these details to be handled by the Rust D-Bus bindings.

mulkieran:
@mvollmer @jbaublitz See new issue #2412 for introspectable concern. This is not related to our multi-threading work.

mulkieran:
> @mvollmer @jbaublitz See new issue #2412 for introspectable concern. This is not related to our multi-threading work.

It's not entirely unrelated either. The timeout part seems to have been introduced by the multi-threading. Still working on it in #2412, though.

mvollmer (approved):
The signals work for us, thanks!

mulkieran:
PR obsoleted. We believe we are addressing all concerns.

Issue #2408, opened 2021-02-19
Start denying cargo clippy lints


mulkieran:
It seems to be mostly missing package metadata; it's probably not wrong to fill that in. Hopefully the cargo manual will give some hint as to what the legitimate values are.

PR #2411, opened 2021-02-23
Deny cargo clippy lints

bgurney-rh:
Closes: #2408

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ build = "build.rs"
 description = "Stratis daemon"
 license = "MPL-2.0"
 repository = "https://github.com/stratis-storage/stratisd/"
-keywords = ["storage", "filesystem"]
+keywords = ["Linux", "filesystem", "storage"]
 categories = ["filesystem"]
 
 [[bin]]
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -4,6 +4,11 @@ version = "2.4.0"
 authors = ["Stratis Developers <stratis-devel@lists.fedorahosted.com>"]
 edition = "2018"
 build = "build.rs"
+description = "Stratis daemon"
+license = "MPL-2.0"
+repository = "https://github.com/stratis-storage/stratisd/"
+keywords = ["storage", "filesystem"]
```
mulkieran, line 10: We get a maximum of 5 keywords. We should definitely use "linux", also. You might want to look for likely words on crates.io.

mulkieran (changes requested):
One request, otherwise looks good.

Cargo.toml
bgurney-rh: The Cargo Book states that "you may choose any words that would help someone find this crate.  (Ah, the wonderful world of Search Engine Optimization; thankfully, it's limited to five words.)

Compare to the "categories" field, which must have strings exactly matching the category strings in https://crates.io/category_slugs (which, unfortunately, doesn't seem to have any storage-related categories, beyond "filesystem".)

mulkieran (changes requested):
Looks good! Please close and open a final PR.

PR #2425, opened 2021-03-03
Add LockedPoolsWithDevs key to FetchProperties.r4

### LINKED ISSUE #2423: RFE: Add LUKS device identification to LockedPools property
I'm currently working on encrypted pools support for the [Stratis installer support](https://github.com/storaged-project/blivet/pull/915) and I think the API around encrypted pools a little too complicated to use, especially when dealing with multiple pools. The `UnlockPool` method requires pool UUID which can be obtained from the `FetchProperties.r2` interface, but there is no information about the underlying LUKS devices. This is a problem for Blivet (and other storage tools like UDisks too). The usual use case is to detect a LUKS device and ask user for a passphrase to unlock it. This means we need to scan through the LUKS tokens on the device and parse the pool UUID from the stratis token JSON metadata. This is certainly possible but Stratis already does this (I assume) so adding UUIDs of LUKS devices to the `LockedPools` would make this way more user friendly.

jbaublitz:
Closes #2423

### DIFF
--- a/src/dbus_api/api/fetch_properties_2_4/methods.rs
+++ b/src/dbus_api/api/fetch_properties_2_4/methods.rs
@@ -1,0 +1,99 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use std::collections::HashMap;
+
+use dbus::{
+    arg::{RefArg, Variant},
+    tree::{MTFn, MethodInfo, MethodResult},
+    Message,
+};
+use itertools::Itertools;
+
+use crate::dbus_api::{
+    api::shared::{list_keys, locked_pool_uuids, locked_pools},
+    consts,
+    types::TData,
+    util::result_to_tuple,
+};
+
+const ALL_PROPERTIES: [&str; 4] = [
+    consts::KEY_LIST_PROP,
+    consts::LOCKED_POOLS,
+    consts::LOCKED_POOL_UUIDS,
+    consts::LOCKED_POOL_DEVS,
+];
+
+pub fn locked_pools_with_devs(
+    info: &MethodInfo<MTFn<TData>, TData>,
+) -> Result<HashMap<String, Variant<Box<dyn RefArg>>>, String> {
+    let dbus_context = info.tree.get_data();
+
+    let engine = dbus_context.engine.borrow();
+    Ok(engine
+        .locked_pools()
+        .into_iter()
+        .map(|(u, locked)| {
+            (
+                uuid_to_string!(u),
+                Variant(Box::new(
+                    vec![
+                        (
+                            "key_description".to_string(),
+                            Variant(Box::new(
+                                locked.info.key_description.as_application_str().to_string(),
+                            ) as Box<dyn RefArg>),
+                        ),
+                        (
+                            "devs".to_string(),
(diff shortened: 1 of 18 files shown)


jbaublitz:
@vojtechtrefny Feel free to test this as well. I've done some preliminary testing using `busctl` and the output looks reasonable to me so far.

src/engine/strat_engine/liminal/device_info.rs
```diff
@@ -410,6 +410,46 @@ impl DeviceSet {
...
+                            strat_info.luks.as_ref().map(|l| l.ids.devnode.clone())
+                        }
+                        LInfo::Luks(luks_info) => Some(luks_info.ids.devnode.clone()),
+                    };
+                    devnode.map(|devnode| LockedPoolDevice {
+                        devnode,
+                        uuid: *uuid,
+                    })
+                })
+                .fold(Some(Vec::new()), |vec, dev_info| {
```
mulkieran, line 438: I think this fold returns None if the LUKS device node was not known for any of the devices belonging to the set. Could you add that information in the header comment?

src/dbus_api/api/fetch_properties_2_4/methods.rs
```diff
@@ -0,0 +1,99 @@
...
+const ALL_PROPERTIES: [&str; 4] = [
+    consts::KEY_LIST_PROP,
+    consts::LOCKED_POOLS,
+    consts::LOCKED_POOL_UUIDS,
+    consts::LOCKED_POOL_DEVS,
+];
+
+pub fn locked_pools_with_devs(
+    info: &MethodInfo<MTFn<TData>, TData>,
+) -> Result<HashMap<String, Variant<Box<dyn RefArg>>>, String> {
```
mulkieran, line 30: I don't understand exactly why the value type of the HashMap is a Variant, instead of being a HashMap of String * Variant.

mulkieran (changes requested):
Two remarks, otherwise it looks ready.

src/dbus_api/api/fetch_properties_2_4/methods.rs
jbaublitz: Resolved.

src/engine/strat_engine/liminal/device_info.rs
jbaublitz: I'm not sure I entirely agree. The devnode will always be known if it is an encrypted Stratis device. What it's checking here is whether or not it's encrypted, not whether or not it knows the device node. The device node is not optional, the encryption status is.

mulkieran:
@jbaublitz Please rebase to pull in introspection data changes. Once that's done, please update introspection data used by tests for new FetchProperties interface. You should be able to copy ```_introspect.py``` exactly from https://github.com/stratis-storage/stratis-cli/pull/718.

src/engine/strat_engine/liminal/device_info.rs
mulkieran: I think we are saying the same thing. One of the devices belonging to the set could be represented as a Stratis device w/out an underlying LUKS device. Unlikely but possible if the LUKS device hasn't been picked up via udev. In that case, you would not return this set as part of the result. This seems to me like a reasonable decision, but it should be documented, because it is not obvious without mulling over the ```fold``` fairly carefully.

jbaublitz: Okay, I understand now! I'll add a comment.

PR #2435, opened 2021-03-09
Fix error handling for crypt module

jbaublitz:


jbaublitz:
@mulkieran This resolves some of the ugliness around using methods like the Clevis bind shell out method (which returns `StratisError`) inside of a method that returns `LibcryptErr` as an error.

mulkieran (approved):

PR #2443, opened 2021-03-12
Fix logging levels in crypt module

jbaublitz:


jbaublitz:
@mulkieran The logging level particularly for determining whether it's a Stratis device or not should absolutely not be `warn!()`. That's popped up in my boot from root testing just because there are other encrypted non-Stratis devices in the VM and it seems silly to warn on that.

mulkieran (approved):

PR #2461, opened 2021-03-18
Remove stratis_uuids_to_names

### LINKED ISSUE #2307: Make stratis_uuids_to_names go away entirely
Building stratis_uuids_to_names in debug creates a file of 21M, without debug the file is about 2MB.  Other files in /usr/lib/udev
are typically not over 40K.  So, would it be possible to optimize this file?

jbaublitz:
Closes #2307


jbaublitz:
@mvollmer Any thoughts on the shell invocation in the udev rule? Will this be a problem given that we're doing a simple string comparison?

mulkieran (changes requested):
Code looks great...but reviewing made me wonder about two things:
1. What do we do for snapshots, anyway?
2. Is there any way to make it possible for all of the events to be synthesized in the engine? I feel like synthesizing any in the D-Bus layer may be a bit risky.

jbaublitz:
> Code looks great...but reviewing made me wonder about two things:
> 
>     1. What do we do for snapshots, anyway?

I'll look into this a bit more.

> 
>     2. Is there any way to make it possible for all of the events to be synthesized in the engine? I feel like synthesizing any in the D-Bus layer may be a bit risky.

This originally happened because we needed to synthesize an event _after_ the D-Bus had the necessary information to respond to `stratis_names_to_uuids`. Given that we're removing this executable in this PR, I see what you mean about moving this into the engine. Let me see what I can do.

jbaublitz:
@prajnoha Is there a way to preserve udev variables across runs? Currently what I'm bumping into is that a change event that is not synthetic and not initiated by Stratis will completely remove the symlink. The rule only matches synthetic udev events so udev cleans up the symlink when a non-synthetic udev event is fired as `/dev/stratis` symlinks do not get added to the `SYMLINK` variable. Attempting to store the pool and filesystem names across udev runs does not appear to work either as the environment seems to be rebuilt by each udev run. For example, setting something like `ENV{STRATIS_POOL_NAME}=...` does not work because it will not be available in the next run and therefore the symlink will get removed.

prajnoha:
> @prajnoha Is there a way to preserve udev variables across runs? 

Yup, there's IMPORT{db}="KEY_NAME" to import the KEY_NAME from preceding udev db state.

jbaublitz:
Oh perfect! That resolved it! @mulkieran We should be good to go now!

mulkieran (changes requested):
Looks great! Please close and open a final PR.

PR #2439, opened 2021-03-09
Multithreading capabilities for stratisd

jbaublitz:
Supersedes #2404
Closes stratis-storage/project#216.

PR #2465, opened 2021-03-19
Fix for in-memory keyfiles with multithreading

jbaublitz:
Related to #2439


mulkieran (approved):
Code all looks good to me.

PR #2466, opened 2021-03-19
Try running CI in containers with systemd

jbaublitz:
Related to #2421


jbaublitz:
We can't start systemd to manage udev on the primary container running the tests, but I've tested it and we can make it work for our service containers in the Clevis/Tang tests.

PR #2471, opened 2021-03-21
Rewrite code and add some CI for dependency version management tasks

mulkieran:
* Choose dependencies in Cargo.toml that are the lowest that allow this to be both compiled and basic tests run
* Alphabetize the dependencies, since touching them anyway, and I'm tired of chaos
* Change update_cargo_crates.py entirely. Now reports three things:
  * If there is a newer package in Fedora for some direct dependency
  * If there are no corresponding Fedora packages for some direct dependency
  * If there is a corresponding Fedora package for some direct dependency, but it's too old


mulkieran:
We have a working script! See the Makefile for two proposed targets. The question is, when and how to run the script.

The part that identifies that the version should be bumped doesn't need to be run any more frequently than nightly. It also doesn't require running any cargo commands that have any effect.

The part that identifies missing Fedora packages or Fedora packages that are too old might ideally be run in the CI. But it's a bit more complicated than the other.

mulkieran:
rebased.

jbaublitz:
Does this supersede #2458?

mulkieran:
> Does this supersede #2458?

Yes. Please go ahead and close that one.

mulkieran:
rebased.

PR #2479, opened 2021-03-24
Trigger udev symlink creation on ThinPool setup

### LINKED ISSUE #2477: Symlinks may not be created on boot
When looking at #2475, I realized that we are now fully dependent on synthetic udev events for symlink creation. We have covered the case of filesystem creation, filesystem snapshots, filesystem renaming, and pool renaming, but we have not covered the case of filesystem setup. This could affect #2372 as we need the symlinks when stratisd starts.

jbaublitz:
Supersedes #2478 
Closes #2477


mulkieran (approved):

jbaublitz:
Making one last change.

PR #2482, opened 2021-03-25
Fix for udev rule

### LINKED ISSUE #2477: Symlinks may not be created on boot

jbaublitz:
Closes #2477 
Supersedes #2479


src/engine/strat_engine/backstore/crypt.rs
```diff
@@ -720,7 +720,8 @@ fn tang_dispatch(json: &Value) -> StratisResult<Value> {
     let mut hasher = Sha1::new();
     hasher.update(thp.as_bytes());
     let array = hasher.finalize();
-    let thp = encode_config(array, Config::new(CharacterSet::UrlSafe, false));
+    let thp = BASE64URL_NOPAD.encode(array.as_slice());
+    //let thp = encode_config(array, Config::new(CharacterSet::UrlSafe, false));
```
mulkieran, line 724: Stray commented bit.

mulkieran (changes requested):
Looks great! Just one request, then please close this and open final PR.

PR #2493, opened 2021-03-29
mdv: Use openat-ext to write state JSON

jbaublitz:
Supersedes #2381


Cargo.toml
```diff
@@ -39,6 +39,8 @@ libmount = "0.1.9"
 libudev = "0.2.0"
 log = "0.4.8"
 nix = "0.20.0"
+openat-ext = "0.1.10"
```
mulkieran, line 42: Please move this down to 0.1.6, it's the highest available in Fedora.

mulkieran (changes requested):
Just one request. If you're fine with that, please just force push and I'll merge more-or-less immediately.

mulkieran:
Well, the good news is that my CI is working. The bad news is that we require at least openat-ext 1.7 which isn't in rawhide yet. So, pushing back to pending.

mulkieran:
Closing. We're still not vendoring in Fedora, and it may be quite a while before a Fedora package that we can use is available in rawhide.

PR #2509, opened 2021-04-07
Fix bug in devicemapper event handling

jbaublitz:
Blocked by #2508
Closes stratis-storage/project#280

### DIFF
--- a/src/engine/strat_engine/engine.rs
+++ b/src/engine/strat_engine/engine.rs
@@ -6,8 +6,6 @@ use std::{clone::Clone, collections::HashMap, path::Path};
 
 use serde_json::Value;
 
-use devicemapper::DmNameBuf;
-
 use crate::{
     engine::{
         engine::KeyActions,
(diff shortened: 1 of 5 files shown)


src/engine/types/mod.rs
```diff
@@ -353,3 +353,9 @@ impl<'a> From<&'a libudev::Device<'a>> for UdevEngineDevice {
         }
     }
 }
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct DevicemapperInfo {
```
mulkieran, line 358: I'm having trouble justifying why this summary information is a sufficient proxy, for two reasons:

1. What if the highest event_nr is one that hasn't changed? For example, for some reason the thinpool has the highest event number, but there's a filesystem that has experienced a change, and has had it's event number bumped. I think the pool should still ```event_on``` in that case.
2. ```device_count``` might be the same if a filesystem had been removed and another one added. But in that case, I think an ```event_on``` would still be indicated.

Maybe ```DevicemapperInfo``` can only be accurate if it's just a map of names to event numbers?

mulkieran (changes requested):
Code looks great, but I do have one general question.

src/engine/types/mod.rs
jbaublitz: Okay, this should be resolved!

src/engine/strat_engine/engine.rs
```diff
@@ -303,16 +301,35 @@ impl Engine for StratEngine {
...
+                .iter()
+                .filter_map(|(dm_name, event_nr)| {
+                    if dev_names.contains(dm_name) {
+                        Some((dm_name.clone(), *event_nr))
+                    } else {
+                        None
+                    }
+                })
+                .collect::<HashMap<_, _>>();
+            let device_count = event_nrs.len();
```
mulkieran, line 316: I think ```device_count``` is now redundant.

mulkieran (changes requested):

src/engine/strat_engine/engine.rs
jbaublitz: Resolved!

mulkieran (changes requested):
Looks great! Please close this and open a final PR.

PR #2507, opened 2021-04-06
Issue project 270

mulkieran:
Related https://github.com/stratis-storage/project/issues/270

PR #2513, opened 2021-04-08
Make script for predicting space usage in a pool

mulkieran:
Resolves https://github.com/stratis-storage/project/issues/270
Supercedes #2507


bgurney-rh (approved):

Makefile
```diff
@@ -202,8 +204,10 @@ install-cfg: docs/stratisd.8
 install: $(PROFILETARGET) install-cfg
 	install -Dpm0755 -t $(DESTDIR)$(LIBEXECDIR) target/$(PROFILEDIR)/stratisd
 	install -Dpm0755 -t $(DESTDIR)$(UDEVDIR) target/$(PROFILEDIR)/stratis-utils
-	mv $(DESTDIR)$(UDEVDIR)/stratis-utils $(DESTDIR)$(UDEVDIR)/stratis-str-cmp
-	ln $(DESTDIR)$(UDEVDIR)/stratis-str-cmp $(DESTDIR)$(UDEVDIR)/stratis-base32-decode
+	mv -fv $(DESTDIR)$(UDEVDIR)/stratis-utils $(DESTDIR)$(UDEVDIR)/stratis-str-cmp
+	ln -fv $(DESTDIR)$(UDEVDIR)/stratis-str-cmp $(DESTDIR)$(UDEVDIR)/stratis-base32-decode
+	mkdir -p $(DESTDIR)$(BINDIR)
```
jbaublitz, line 209: Can you explain this line? I don't think we should be creating standard directories for users and if they are installing into a non-standard directory, my thought is that it should already be there.

jbaublitz (changes requested):
One initial question. Should we use clap for all executables if we're already including it? I'm going to review this in more detail soon.

jbaublitz:
Fully reviewed. Other than those two comments, this looks good to me!

Makefile
mulkieran: You're right about that. I was a little too focused on getting it working in our Jenkins tests, but for real use we'ld better drop it.

mulkieran:
> One initial question. Should we use clap for all executables if we're already including it? I'm going to review this in more detail soon.

I like having one locally defined parser per executable name, it keeps things simpler than, for example, using subcommand. For uniformity, I can set up parsers for the other two executable names. Is that what you were thinking of?

mulkieran:
@jbaublitz If https://github.com/stratis-storage/stratisd/pull/2513/commits/603f39e98fec5896c28bdefca17d9fd84f4caa23 looks good to you, I'll go ahead and do stratis-base32-decode.

jbaublitz:
> @jbaublitz If [603f39e](https://github.com/stratis-storage/stratisd/commit/603f39e98fec5896c28bdefca17d9fd84f4caa23) looks good to you, I'll go ahead and do stratis-base32-decode.

Other than clap arguments being optional by default, this looks good to me! I think we have to explicitly declare that they're required.

mulkieran:
jbaublitz: Ok! I believe this is ready again. Ignore the clippy error, it's bogus.

jbaublitz (approved):

mulkieran:
Superceded

PR #2514, opened 2021-04-09
Issue stratisd 2134

### LINKED ISSUE #2134: Add a script to set the Stratis keys in the initramfs

jbaublitz:
Supersedes #2372 
Closes #2134


mulkieran (changes requested):
A few requests. Should you be unifying the two systemd generators into a single executable? If not possibly something else newly added?

Makefile
```diff
@@ -294,6 +339,8 @@ verify-dependency-bounds: set-lower-bounds
 	outdated
 	release
 	set-lower-bounds
+	build-min
```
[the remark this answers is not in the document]
jbaublitz: > A few requests. Should you be unifying the two systemd generators into a single executable? If not possibly something else newly added?

Whoops! This slipped my mind. I'll do that now.

jbaublitz:
@mulkieran Okay, this should all be resolved! I'll keep an eye on CI just to make sure it passes.

mulkieran (changes requested):
Please pick up final commit https://github.com/stratis-storage/stratisd/pull/2448/commits/8c05b4c20d2574b026e55a33a7d2215d84ec5060 from https://github.com/stratis-storage/stratisd/pull/2448.

Makefile
```diff
@@ -292,8 +337,9 @@ verify-dependency-bounds: set-lower-bounds
 	install-cfg
 	license
 	outdated
-	release
 	set-lower-bounds
+	release
```
mulkieran, line 341: Pls swap w/ set-lower-bounds.

mulkieran (changes requested):

jbaublitz:
Ready for review.

src/bin/stratis-utils.rs
```diff
@@ -166,6 +171,64 @@ fn parse_args() -> Result<(), Box<dyn Error>> {
...
+                        .required(true)
+                        .help("Directory in which to write a unit file for early priority"),
+                )
+                .arg(
+                    Arg::with_name("late_priority_dir")
+                        .required(true)
+                        .help("Directory in which to write a unit file for late priority"),
+                );
+            let matches = parser.get_matches_from(&args);
+            stratis_clevis_setup_generator::generator(
```
mulkieran, line 196: I can't make sense of three required arguments, only one of which is used.

src/jsonrpc/client/utils.rs
```diff
@@ -0,0 +1,197 @@
...
+    (30, "GiB"),
+    (20, "MiB"),
+    (10, "KiB"),
+    (1, "B"),
+];
+
+#[allow(clippy::cast_precision_loss)]
+pub fn to_suffix_repr(size: u128) -> String {
+    SUFFIXES.iter().fold(String::new(), |acc, (div, suffix)| {
+        let div_shifted = 1 << div;
```
mulkieran, line 186: ```acc``` ends up as empty string if size is 0.

mulkieran (changes requested):
Two observations...also please pick up https://github.com/stratis-storage/stratisd/pull/2448/commits/8743c564319adb3e07cb77cfccb8dc9412357a42; it is trivial but makes for an easier read.

src/bin/stratis-utils.rs
jbaublitz: This is the systemd generator interface. It is mandatory to accept three positional arguments.

src/jsonrpc/client/utils.rs
jbaublitz: There may also be a problem with the (1, "B") value too. I think it should be (0, "B"). I'll clean this up.

jbaublitz:
@mulkieran Take a look at 27b2c5b9bb27852ce0eb40c56c44300aa847a7c4. I've added tests, reworked it slightly, and added a special case for `size == 0`.

jbaublitz:
I've tested booting from Stratis as a root filesystem in my test environment and so it's back in review.

mulkieran (changes requested):
I think it's time for this to go in. If you're satisfied, please close this and open a new PR for final review.

PR #2525, opened 2021-04-14
stratis_dumpmetadata: Add log initializer

bgurney-rh:


mulkieran (changes requested):
Please drop the test warn message, and this PR should be ready to go.

PR #2534, opened 2021-04-16
Issue stratisd 2530

### LINKED ISSUE #2530: Support encrypted Stratis pools in /etc/fstab
Currently, encrypted Stratis pools in /etc/fstab can be unlocked with custom configuration, but we should add a `.service` file to enable this and establish some best practices.

jbaublitz:
Closes #2530

### DIFF
--- a/src/bin/stratis-min/stratisd-min.rs
+++ b/src/bin/stratis-min/stratisd-min.rs
@@ -64,7 +64,7 @@ fn trylock_pid_file() -> StratisResult<File> {
     let stratisd_min_file = match flock(f.as_raw_fd(), FlockArg::LockExclusiveNonblock) {
         Ok(_) => {
             f.write_all(getpid().to_string().as_bytes())?;
-            Ok(f)
+            f
         }
         Err(_) => {
             let mut buf = String::new();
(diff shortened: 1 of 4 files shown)


src/bin/stratis-min/stratisd-min.rs
```diff
@@ -27,13 +41,82 @@ fn parse_args() -> App<'static, 'static> {
...
+            if f.read_to_string(&mut buf).is_err() {
+                buf = "<unreadable>".to_string();
+            }
+
+            Err(StratisError::Error(format!(
+                "Daemon already running with supposed pid: {}",
+                buf
+            )))
+        }
+    };
```
mulkieran, line 81: Should it return immediately if it can't get the lock on the "stratisd-min" file instead of continuing and attempting the lock on the stratisd file?

mulkieran (changes requested):
One question, otherwise code looks great.

src/bin/stratis-min/stratisd-min.rs
jbaublitz: Yes, that actually makes more sense. I'll push an edit.

PR #2539, opened 2021-04-20
Run cargo clippy on some cfg flagged subsets of the code

mulkieran:


jbaublitz:
I'm not sure, but you may be able to add back in `--all-features` now that I've changed the `ipc_support` module.

mulkieran:
> I'm not sure, but you may be able to add back in `--all-features` now that I've changed the `ipc_support` module.

I think I'ld like to stick with the first clippy target command kind of mimicking the build command. I added some Makefile variables, though, which makes the clippy target clearer and make the convenience targets stratisd-min and stratis-min build without dbus enabled.

jbaublitz (approved):

mulkieran:
rebased.

PR #2544, opened 2021-04-22
Remove an unnecessary get_mut call

mulkieran:


mulkieran:
This change should have been put in as part of the original multi-threading work; it became possible when the pool object no longer stored its object path.

mulkieran:
stratisd-nonrust failure is unique-instance test only, can be ignored.

jbaublitz (approved):

PR #2551, opened 2021-04-24
Unique instance

mulkieran:


mulkieran:
Would like to do an intregated test of both lock files. Also, still need to clean up logging.

mulkieran:
Obsolete. We should really do a complete rewrite of this step, including stratisd-min in the story.

PR #2554, opened 2021-04-26
github actions: move allowed_failures to nightly

bgurney-rh:
Related: stratis-storage/project#244

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -260,41 +260,6 @@ jobs:
       - name: Test ${{ matrix.task }} on ${{ matrix.toolchain }} toolchain
         run: ${{ matrix.task }}
 
-  # ALLOWED FAILURES
-  allowed_failures:
-    continue-on-error: true
-    strategy:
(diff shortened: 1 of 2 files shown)


mulkieran (changes requested):
Please work on inserting the clippy stable check in w/ the other allowed failures.

bgurney-rh:
> Please work on inserting the clippy stable check in w/ the other allowed failures in nightly.

When I look at the `.github/workflows/main.yml' file, in the `checks:` section, the only other clippy tasks I see are for the current development compiler (currently 1.51.0).

mulkieran:
The task is parameterized on the toolchain, (see https://github.com/stratis-storage/stratisd/pull/2554/files#diff-0d5658b415099a82c11c03a06ca4ec765b4003a1f4b2f3f1943980a882cf8aa6R52), so it should work correctly to just slot it into the matrix.

bgurney-rh:
Ok, now I understand: the two jobs are so similar, we can run them as one.

PR #2557, opened 2021-04-26
Use udevadm settle to synchronize in /etc/fstab setup service

jbaublitz:


mulkieran:
stratisd-nonrust was hung up on unique instance test again.

mulkieran (approved):

PR #2565, opened 2021-04-29
Split crypt.rs into several submodules

### LINKED ISSUE #2505: Should try to break up crypt.rs at some time
It's 2000 lines, almost. We should try for some useful division. One possibility that occurred to me was to break out the support code that massages Clevis configuration info, but there's not really much of that.

jbaublitz:
Closes #2505


jbaublitz:
@mulkieran I've left the tests in `mod.rs` because they're not really specific to any of the submodules.

mulkieran (approved):

PR #2578, opened 2021-05-06
Expose newer interfaces in signals when objects are added and removed

jbaublitz:


mulkieran (changes requested):
Looks good. Please close and submit new PR for final review.

PR #2581, opened 2021-05-07
Update Rust lints for 1.52

bgurney-rh:
Related: stratis-storage/project#301


mulkieran (approved):
I'm fine with this as is.

PR #2583, opened 2021-05-08
Use containers for Python checks in CI

mulkieran:
Related https://github.com/stratis-storage/project/issues/302


mulkieran:
By running in containers, we use the Fedora package versions of Python dependencies and don't have to update the individual dependencies and install via pip. So I think these are better checks all around.

bgurney-rh (approved):

jbaublitz (approved):

PR #2591, opened 2021-05-13
Fix for hanging synchronization lock in D-Bus layer

jbaublitz:
Supersedes #2585


mulkieran:
GitHub Actions failure, but it was because of Fedora 29 container, so it may be our problem. Let's see.

mulkieran:
Nope. looks like it should be good.

mulkieran (approved):

PR #2593, opened 2021-05-14
Add step in Python tasks to display Python version

mulkieran:


mulkieran:
This both works and is useful, let us put it in.

bgurney-rh (approved):

PR #2604, opened 2021-05-19
Add crypt module support for changing passphrases

### LINKED ISSUE #2358: Support changing passphrase via key description and clevis-luks-regen
While @mulkieran brought up the question of Tang key rotation for our Clevis integration and this process does appear to be manual, I do want to look into the details of handling key rotation. There is some information that indicates that newer Clevis clients automatically pick up new key advertisements, but we should verify this with the Clevis maintainers and determine whether the same keyslot will be used and overwritten as automatically binding to a new keyslot and leaving the old keyslot tied to the old keys could be a problem.

jbaublitz:
Closes #2358

### DIFF
--- a/src/dbus_api/api/mod.rs
+++ b/src/dbus_api/api/mod.rs
@@ -92,6 +92,17 @@ pub fn get_base_tree<'a>(
                 .add_p(manager_2_0::version_property(&f)),
         )
         .add(
+            f.interface(consts::MANAGER_INTERFACE_NAME_2_5, ())
+                .add_m(manager_2_4::create_pool_method(&f))
+                .add_m(manager_2_2::set_key_method(&f))
+                .add_m(manager_2_1::unset_key_method(&f))
+                .add_m(manager_2_3::unlock_pool_method(&f))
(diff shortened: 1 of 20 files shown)


jbaublitz:
So far, I've tested the keyring passphrase change when there are no Clevis bindings. The old keyslot is wiped, the token is appropriately changed, attempting to open the device with the old keyslot and passphrase fails, and the new passphrase successfully unlocks the device. So far it looks like it's behaving as we want it to.

jbaublitz:
I've now tested with Clevis bindings, the keyslot allocation happens as expected, the tokens are in the correct places, and both bindings can be used to unlock the pool. So far I think this call is ready for anyone to test who wants to!

jbaublitz:
@mulkieran This PR addresses part of https://github.com/stratis-storage/project/issues/300. Do you want me to put up a PR first fully addressing that issue?

mulkieran:
> @mulkieran This PR addresses part of [stratis-storage/project#300](https://github.com/stratis-storage/project/issues/300). Do you want me to put up a PR first fully addressing that issue?

Yes. Please bump all the revisions in a separate PR.

PR #2608, opened 2021-05-20
Remove stray tar file and ignore vendor tar files in future

mulkieran:


mulkieran:
It was hanging on test_unique_instance, otherwise fine. That test should be dropped until we fix it, I think.

PR #2613, opened 2021-05-21
Bump Fedora versions in CI

mulkieran:
Related https://github.com/stratis-storage/project/issues/263


mulkieran:
Actually, we should be able to run the tests on f:30 as we do in stratis-cli; that's probably best.

mulkieran:
Ok! I believe this is ready.

bgurney-rh (approved):

drckeefe (approved):

PR #2625, opened 2021-05-27
Use header_contents function in build script

mulkieran:


mulkieran:
> Did you change your mind about leaving this as configuration?

Originally my proposal was that build.rs write out the file, dynamically, and that bindgen then use the file. The argument against that was that it was more complexity in build.rs and more dynamic. Now that I have been told that the bindgen Builder is actually set up to do the work for us (https://github.com/stratis-storage/devicemapper-rs/pull/624#pullrequestreview-669804786) , rather than our having to write the code, I'm in favor of using its facility. It looks like bindgen doesn't make a file at all, but simply add the contents to an internal data structure. I'm particularly in favor of the change in this case, since the header file is only needed when the systemd_compat flag is set, but I think it makes sense in most cases.

jbaublitz:
Okay that makes sense! Approved.

jbaublitz (approved):

PR #2627, opened 2021-05-27
add xfs_db in dracut module

mberndt123:
Hey there,

I'm running Fedora 34 with the root fs on a stratis volume. When I upgraded to the 5.12 kernel, my initramfs was suddenly no longer able to mount the root fs. I investigated and found that apparently the xfs_db binary is needed in the initramfs. I modified the 90stratis dracut module to include it, and now it's working again, so I thought I'd share the fix.

Unfortunately I can't explain why it was working with the 5.11 kernel and then suddenly stopped when I upgraded to 5.12. Presumably some other dracut module (probably 99fs-lib) was including xfs_db before and for whatever reason stopped doing so 🤷🏻. Either way, if stratisd needs that tool, it needs to be included in the initramfs, so I think my fix is the correct thing to do regardless of the exact cause.

### DIFF
--- a/dracut/90stratis/module-setup.sh
+++ b/dracut/90stratis/module-setup.sh
@@ -11,6 +11,7 @@ check() {
 	    xfs_admin \
 	    xfs_growfs \
 	    xfs_db \
+	    udevadm \
 	    plymouth \
 	    /usr/sbin/plymouthd \
 	    /usr/lib/udev/stratis-str-cmp \
@@ -41,6 +42,7 @@ install() {
 	    xfs_admin \
 	    xfs_growfs \
 	    xfs_db \
+	    udevadm \
 	    plymouth \
 	    /usr/sbin/plymouthd \
 	    /usr/lib/udev/stratis-str-cmp
--- a/dracut/90stratis/module-setup.sh
+++ b/dracut/90stratis/module-setup.sh
@@ -10,6 +10,7 @@ check() {
 	    mkfs.xfs \
 	    xfs_admin \
 	    xfs_growfs \
+	    xfs_db \
 	    plymouth \
 	    /usr/sbin/plymouthd \
 	    /usr/lib/udev/stratis-str-cmp \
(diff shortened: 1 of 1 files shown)


drckeefe:
mberndt123, thank you for reporting this issue.  It might be mid next week before it can be reviewed this in detail.

mulkieran:
linked bz https://bugzilla.redhat.com/show_bug.cgi?id=1965534

jbaublitz:
Hi @mberndt123! Thanks so much for the report. Your theory about `xfs_db` no longer being included sounds accurate. Would you be willing to modify your PR to include `udevadm` just in case that gets removed from module dependencies at some point? I'd like to avoid this problem in the future explicitly.

mberndt123:
Hey @jbaublitz,

Well, udevadm doesn't do much good unless udevd is also present. So I think it's better to add a dependency to the systemd-udevd module instead. That will ensure that both udevadm and udevd are present.

jbaublitz:
Hi @mberndt123! I'd prefer explicitly including `udevadm`. For one, the systemd-udevd dracut module does not seem to be installed by default on Fedora on the filesystem. From what I can tell, it appears it may be internal to dracut as no package in Fedora provides it. Given this and the fact that the dracut udev hook must always be available according to the documentation, I don't think we need to worry about udev being started. The hook infrastructure is relatively crucial to dracut operating the way it does and while this may be a bad assumption, I find that far less likely to go away than `udevadm` as a command line dependency.

My rationale behind asking you to include `udevadm` explicitly is that we do a pre-check when starting stratisd to ensure that all of the executables that we need are there. If `udevadm` is not present, stratisd will fail to start. Even if udev is not started, we will at the very least be able to start up stratisd. While udev is built into the hook infrastructure of dracut, we have no guarantee that `udevadm` will be included moving forward. There is no way to depend on a specific command other than specifying it explicitly.

mberndt123:
ok

jbaublitz (approved):
Thanks! Approved.

bgurney-rh:
ok to test

bgurney-rh (approved):
All tests passed (I ran the stratisd test manually, and it passed; there may be a failure indicated in one of the ci stratisd tests, but it looks like an infrastructure issue.)

I also ran the blackbox tests on this PR, and they passed.

mulkieran (approved):

PR #2629, opened 2021-06-01
Unify roll back code for BlockDevMgr

### LINKED ISSUE #2610: Should we make a better effort to roll back encryption operations?
After reviewing some of the encryption operations like unbinding, I noticed that I did not attempt to roll back operation as they may not be able to be rolled back successfully. I have some ideas for how to attempt to roll some of them back if they fail by doing some prechecks before attempting to roll them back. This may give us better consistency if a requested operation fails.

jbaublitz:
Closes #2610


jbaublitz:
@mulkieran I'm just requesting preliminary review.

PR #2634, opened 2021-06-02
Test fixups

mulkieran:


.github/workflows/main.yml
```diff
@@ -265,27 +265,35 @@ jobs:
       matrix:
         include:
           # PYTHON CHECKS ON CURRENT FEDORA AND PYTHON TOOLS VERSION
+          - dependencies: >
+              pylint
+              python3-requests
+              python3-pygithub
```
bgurney-rh, line 271: This test is probably failing because `python3-semantic_version` needs to be added as a dependency.

mulkieran: Whoops! I thought all the tests had passed...

bgurney-rh: The first series of tests passed, prior to when the second commit added.  But it passes now.

bgurney-rh (approved):

PR #2646, opened 2021-06-07
Test nightly actions for cargo vendor inventory

bgurney-rh:


.github/workflows/release.yml
```diff
@@ -0,0 +1,21 @@
...
+on:
+  release:
+    types:
+      - created
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  allowed_failures:
```
mulkieran, line 14: We don't want this one to be an allowed failure, so we should give the job a different title. Might as well drop continue-on-error setting as well. With those changes, should be ready to go in.

mulkieran (changes requested):

mulkieran (changes requested):
I think this is good enough to get started with. Please just squash the two commits, and then I'll merge it and we can begin serious development.

PR #2653, opened 2021-06-14
Generalize the Makefile over the manifest path

mulkieran:
Related https://github.com/stratis-storage/project/issues/307


mulkieran:
@bgurney-rh The approach I have here, if followed in libblkid-rs, would require two separate Makefile targets, one for the ```-sys``` subdirectory and one for the top-level directory. It is the most general approach, though; get the paths correct and it works from anywhere.

mulkieran:
Whoops!

bgurney-rh (approved):

PR #2657, opened 2021-06-17
Prepare for upgrade to Rust 1.53

mulkieran:
This PR does two separate things:

* removed some old allow unknown lints directives that are no longer necessary
* changes all tests on the stable toolchain in the CI to tests on the current development toolchain. The reasoning here is that since we compile with warnings on, sometimes the upgrade to stable will cause our tests to fail to compile which is inconvenient. The errors do get caught by the clippy task in nightly, which runs on stable.


mulkieran:
This should go in before we start on https://github.com/stratis-storage/project/issues/315

jbaublitz (approved):

bgurney-rh (approved):

PR #2687, opened 2021-06-30
Run fedora check against f34 also

mulkieran:
Supercedes #2685


mulkieran:
Ignore the PYthon-based udev tests failure once more, they are irrelevant.

drckeefe (approved):

mulkieran:
I had to add one more fix and an additional check. But it sure is tested now.

drckeefe (approved):

PR #2689, opened 2021-07-01
Make devicemapper event processing and D-Bus request handling errors non-fatal

### LINKED ISSUE #2686: Rework thread error handling
Currently, there are cases in our threading model where failed operations cause stratisd to shut down that should not. Rework the threading model so that setup actions that fail cause stratisd to shut down, but processing actions after stratisd has been set up cause an `warn` level log instead of the daemon exiting.

Related to #2638

jbaublitz:
Closes #2686

### DIFF
--- a/src/dbus_api/connection.rs
+++ b/src/dbus_api/connection.rs
@@ -351,7 +351,9 @@ impl DbusConnectionHandler {
             }),
         );
         loop {
-            self.connection.process(Duration::from_millis(100))?;
+            if let Err(e) = self.connection.process(Duration::from_millis(100)) {
+                warn!("Failed to process D-Bus request: {}", e);
+            }
             match self.should_exit.try_recv() {
                 Ok(()) => {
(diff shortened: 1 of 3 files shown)


jbaublitz:
@mulkieran There were really only two places that this needed to be fixed:
* D-Bus request processing
* devicemapper event handling

One question that I do have, however, is whether channels being closed should be treated as a fatal error. My thought is yes because I see this as similar to setup errors. Once the channel closes, stratisd loses the ability to do things like processing udev events, updating the D-Bus tree, quitting on `ctrl-C` etc. This seems like enough of a degraded state to allow stratisd to shut down.

mulkieran (approved):
Looks good! Regarding exit on channel close, that seems like a reasonable decision. If we find we need to reevaluate in future; we can do that then.

jbaublitz:
@mulkieran Rebased

PR #2693, opened 2021-07-06
Remove unnecessary methods from Pool trait

jbaublitz:


mulkieran (approved):

mulkieran:
Several became unused in the course of doing the multi-threading work; when the event-handling no longer required that we stow the object path in the struct used by the engine.

mulkieran:
Not waiting for stratisd-2.

PR #2696, opened 2021-07-07
Add support for changing passphrases and regenerating Clevis bindings

### LINKED ISSUE #2358: Support changing passphrase via key description and clevis-luks-regen

jbaublitz:
Closes #2358

### DIFF
--- a/src/dbus_api/pool/mod.rs
+++ b/src/dbus_api/pool/mod.rs
@@ -21,6 +21,7 @@ mod pool_2_0;
 mod pool_2_1;
 mod pool_2_3;
 mod pool_2_4;
+mod pool_2_5;
 mod shared;
 
 pub fn create_dbus_pool<'a>(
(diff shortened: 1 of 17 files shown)


mulkieran (approved):
Will merge once CLI is ready.

mulkieran:
Jenkins failure is a self-cancellation, I think I'll ignore it.

mulkieran:
jenksin re-run succeeded.

PR #2709, opened 2021-07-14
Pass optional size parameter for new filesystem

mulkieran:
Related https://github.com/stratis-storage/project/issues/299

Changes in this PR:
1. Change the CreateFileystem D-Bus method to accept size parameter.
2. Include more filesystem information, specifically the size, in the JSON report.
3. Set a minimum size of 32 MiB and a maximum of 8 EiB.
4. Log the size of the filesystem on creation.
5. Use an idempotency check that requires that the size and the name of the filesystem be the same. If the filesystem already exists, but the size is different, return an error.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -344,6 +344,12 @@ jobs:
         run: PROFILEDIR=debug make -f Makefile install
       - name: Check out stratis-cli
         run: git clone https://github.com/stratis-storage/stratis-cli.git
+      - name: Switch to PR branch
+        # yamllint disable rule:line-length
+        run: |
+          git checkout -b mulkieran-issue-project-299 develop-2.4.1
+          git pull https://github.com/mulkieran/stratis-cli.git issue-project-299
+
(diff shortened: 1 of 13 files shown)


mulkieran:
Expect test failures

mulkieran:
ignore jenkins failures...

mulkieran:
Want to check for sectors in engine, not D-Bus.

mulkieran:
Note that the filesystem can be made as small as one sector, which is obviously far too small. We can set a lower bound later.

mulkieran:
ignore annoyingly volatile strats-cli check failure, it is an annoying GitHub Actions related problem.

mulkieran:
rebased.

mulkieran:
Proposal...cap the filesystem size at 1 TiB for this PR.

mulkieran:
Proposal...store original size of filesystem; increase the size by maximum of 8 times.

drckeefe (approved):

jbaublitz (approved):

PR #2710, opened 2021-07-15
Update _introspect.py

mulkieran:
Related https://github.com/stratis-storage/project/issues/316


mulkieran:
Ignoring lower bounds check failure, which is clearly unrelated. Unexpected flakiness there.

PR #2711, opened 2021-07-16
Add test for Clevis rebind

### LINKED ISSUE #2708: Test rollback code for keyring and Clevis binding
We can add a devicemapper device to the pool that starts as a linear target and gets switched to the error target after the device is initialized.

jbaublitz:
Closes #2708


jbaublitz:
It looks like there's residual udev issues on stratis-tester-8.

mulkieran (changes requested):
It looks like you may have left some unintended ```unwrap```s in "tests/crypt.rs". Otherwise, this looks all set.

PR #2717, opened 2021-07-20
Cargo update

mulkieran:


mulkieran:
Might as well start running against newest version of devicemapper-rs.

mulkieran:
Given that one of these is a tokio update and I'm not sure that I've seen the udev test error before, I'll want to rerun the tests, at least.

mulkieran:
Ok! I'm convinced I'ld like to start using the new version of devicemapper-rs and that the bump of tokio probably doesn't matter.

jbaublitz (approved):

PR #2725, opened 2021-07-30
Optimized lock for per-pool locking

### LINKED ISSUE #2724: Optimize lock for per-pool locking
We've introduced multithreading into stratisd but have no yet managed to optimize a lot for per-pool locking.

jbaublitz:
Closes #2724


jbaublitz:
@mulkieran As I've tried to start using the data structure defined here, I've determined that we have two options. Because of the unsized types in our codebase (`dyn Pool`, `dyn Engine`, etc.) we will not be able to generalize the data structure here very easily.

We can:
1. Switch to generics and associated types for our `Engine`, `Pool`, etc. types (I've already done an evaluation and while the changes are a little bit tedious, it is possible to do).
2. Make this data structure specifically designed for pools.

Which would you prefer here? I'd rather get rid of our trait objects personally just because this has caused a number of problems for me while designing new locking data structures.

jbaublitz:
Blocked by #2729

jbaublitz:
TODO:
* Prevent two `CreatePool` invocations on the same devices from disrupting each other. Currently, I've confirmed that this will causes errors when done in parallel.
* Figure out cause of trace-level logs causing a deadlock on `Stderr`.
* Investigate lost wake ups in my `AllOrSomeLock` data structure.

jbaublitz:
@mulkieran Outside of the `EBUSY` errors, the good news about the test failures here is that they appear to be related to the fact that I made the D-Bus layer `CreatePool` method synchronous but not the engine API. The udev tests are now passing which is great. This does pose a challenge, however, because if we create a pool in unit tests, we don't really have a good way to set it up through liminal devices. My recommendation here is that we call `StratPool::setup()` to resolve this. What are your thoughts?

PR #2727, opened 2021-07-31
Ignore that nix version in rawhide is higher than our version

mulkieran:


mulkieran:
This works but will require two updates, one for the rawhide Fedora bump and one for the f34 Fedora bump. It would be nice to have something that will make the commit history less cluttered.

mulkieran:
Obsolete.

PR #2730, opened 2021-08-02
Jbaublitz issue project 310

mulkieran:
Related https://github.com/stratis-storage/project/issues/310


drckeefe (approved):

mulkieran:
rebased.

mulkieran:
rebased.

mulkieran:
Why not use the three argument ```handle_action``` macro for all the calls. Is it more expensive?

src/dbus_api/consts.rs
```diff
@@ -21,6 +21,7 @@ pub const POOL_NAME_PROP: &str = "Name";
 pub const POOL_UUID_PROP: &str = "Uuid";
 pub const POOL_HAS_CACHE_PROP: &str = "HasCache";
 pub const POOL_ENCRYPTED_PROP: &str = "Encrypted";
+pub const POOL_STATE_PROP: &str = "State";
```
mulkieran, line 24: Would like to call this something less generic than "State".

jbaublitz: Sounds good! I'll go with `AvailableActions` as we decided in stand up.

jbaublitz:
> Why not use the three argument `handle_action` macro for all the calls. Is it more expensive? Most of the time the D-Bus layer won't have to analyze the error, so I feel it ought not to be. Minor code bloat.

This seems reasonable to me as a future proofing measure!

mulkieran:
stratis-cli test should fail

PR #2735, opened 2021-08-12
Pass optional size parameter for new filesystem

mulkieran:
Related https://github.com/stratis-storage/project/issues/299
Supercedes #2709


mulkieran:
I think that if the udev tests fail, they always fail on the duplicate pool name test, and always on the Fedora 30 machine.

PR #2748, opened 2021-08-17
Fix dracut module after addition of rebind command

### LINKED ISSUE #2745: Fix dracut module now that we depend on clevis-luks-regen
After #2633, the dracut module was not updated. The result of this is that our `stratisd` dracut module still works but our `stratisd-clevis` dracut module is broken as `clevis-luks-regen` is not included in the initramfs and so any Clevis-based operation will fail.

jbaublitz:
Closes #2745


jbaublitz:
@mulkieran I've confirmed that this resolves the issue.

mulkieran (approved):

PR #2753, opened 2021-08-19
Switch from trait objects to type parameters and associated types

### LINKED ISSUE #2729: Investigation into Engine, Pool, Filesystem, and BlockDev trait definitions
Currently, we use trait objects to unify the return types between the sim and strat engine. Trait objects are causing problems for per-pool locking due to them being unsized, they incur a performance penalty, and I'd like to propose switching to type parameters and associated types.

@mulkieran brought up the concern of monomorphization and bloat of our binary size. Due to this, we'd like to explore the following steps:
* Change from trait objects to type parameters and associated types
* Compile out the sim engine in the release build
* Check the binary size using trait objects and compare it against the release build using type parameters.

jbaublitz:
Closes #2729


jbaublitz:
@mulkieran Okay, I looked over this and it's ready for review.

mulkieran:
@jbaublitz Please rebase, fix conflicts, etc.

jbaublitz:
@mulkieran Okay, this is finally rebased and back in review!

mulkieran (changes requested):
Looks good! Please close and open new PR for final review.

Wierd note: For cast like ```<<E as Engine>::Pool as Pool>::Filesystem``` it turns out that ```<E::Pool as Pool>::Filesystem``` works just as well. It's puzzling because I would have thought that the rule requiring disambiguation of the associated type by specifying the trait should be required for the type variable as well as for its associated type. Not requesting any change, just noting it's a bit peculiar.

mulkieran:
@jbaublitz On second thoughts, could you subsitute ```E::Pool``` for ```<E as Engine>::Pool```? Theres no mystery why this works, the ```as Engine``` is truly redundant because of the ```where``` clause on every method where this expression is used.

jbaublitz:
@mulkieran The more explicit `<<E as Engine>::Pool as Pool>::Filesystem` was what was suggested by the compiler. If `<E::Pool as Pool>::Filesystem` works, I would much rather use that and stay away from `<E as Engine>::Pool` and use `E::Pool` instead. We've already gotten into maintenance issues by specifying more than is required by the compiler to succeed. I think of the avoidance of `Self` as being one case where this happened once we added in our use of tokio. What are your thoughts? Are you okay with me using the least verbose representation of the associated types?

mulkieran:
> @mulkieran The more explicit `<<E as Engine>::Pool as Pool>::Filesystem` was what was suggested by the compiler. If `<E::Pool as Pool>::Filesystem` works, I would much rather use that and stay away from `<E as Engine>::Pool` and use `E::Pool` instead. We've already gotten into maintenance issues by specifying more than is required by the compiler to succeed. I think of the avoidance of `Self` as being one case where this happened once we added in our use of tokio. What are your thoughts? Are you okay with me using the least verbose representation of the associated types?

Yes, do please use shorter form. My theory now is that it doesn't start that type-checking phase until ```E```` has already been bound to concrete types ```SimEngine``` or ```StratEngine``` as the case may be; hence there is nothing to disambiguate.

mulkieran (changes requested):
Looks good. Please close and new PR for final review.

PR #2761, opened 2021-08-24
Clean up threading to handle both join and thread runtime errors

jbaublitz:


jbaublitz:
@mulkieran The difference here is subtle but it will now catch both join and runtime errors in the thread. I think this is an improvement.

mulkieran (approved):

mulkieran:
@jbaublitz Please feel free to merge this today if that will be convenient. The straits-cli test failure should be fixed now.

PR #2773, opened 2021-09-08
Clean up test_config.json documentation

bgurney-rh:
Closes: stratis-storage/project#345


mulkieran (changes requested):
This looks good. Two additional requests:
* add README_tests.md to the list of ignored files in the relevant GitHub workflows. 
* Edit README.md so that it refers to README_tests.md not tests/README.md

.github/workflows/support.yml
```diff
@@ -12,6 +12,7 @@ on:
       - "**/*.yml"
       - "tests/**"
       - "!tests/**/README.*"
+      - "!README_tests.md"
```
mulkieran, line 15: Might as well leave these two changes out actually. These ```!``` s are negation symbols, so this entry has no effect. The entry means ```(all *.yml files AND everything under the tests directory) EXCEPT README's```; excluding ```README_tests.md``` has no effect, because its not part of the included set.

mulkieran (changes requested):
One last request then this should be ready to go in...

.github/workflows/support.yml
bgurney-rh: OK; removed.

mulkieran (changes requested):
Looks good. Please squash the last two commits into one, then this should be ready.

mulkieran (approved):

PR #2772, opened 2021-09-07
Refactor for devicemapper event handling

jbaublitz:
Closes #2712
Closes #1675 
Supersedes #2716

PR #2786, opened 2021-09-16
Refactor for devicemapper event handling

### LINKED ISSUE #2712: Work on allowing devicemapper events to trigger D-Bus signals
Currently we do not expose filesystem size over the D-Bus. We are considering adding this in stratis-storage/project#299. Because the filesystem size can change, we will need to support sending a signal over the D-Bus to indicate that the size has been updated. Until now, we did not need to communicate between the devicemapper event handling thread and the D-Bus API threads. However, devicemapper events trigger filesystem resizing so we need to be able to signal to the D-Bus layer after the filesystem has been resized that the property has changed. This requires communication between the devicemapper thread after the event has been processed and the D-Bus API.

My current thoughts for designing this are adding a new D-Bus thread to communicate with the devicemapper thread and look up the object path of the filesystem whose size has changed to pass to the D-Bus tree updating thread. The channel between these two threads will be optional as our JSON RPC layer does not need to send out signals and this thread would therefore be unnecessary.

Related to https://github.com/stratis-storage/project/issues/323

jbaublitz:
Supersedes #2772
Closes #2712


mulkieran (approved):

mulkieran:
@jbaublitz Plz rebase, thx.

jbaublitz:
@mulkieran Ok, ready!

PR #2797, opened 2021-09-28
Clean up after failed rollback tests to avoid lingering devicemapper device

jbaublitz:
Related to #2724


jbaublitz:
@mulkieran This problem was discovered while working on #2724. A failure in the test would cause `stratis_fail_device` to not be cleaned up because of the Stratis devicemapper devices on top of it were never torn down so the `Drop` method would warn that `stratis_fail_device` was busy.

mulkieran (approved):

mulkieran (approved):

PR #2807, opened 2021-10-13
Remove explicit libdbus-sys dependency

mulkieran:


mulkieran:
The udev error is a bit wierd. I'm going to look into it.

PR #2808, opened 2021-10-13
Verify udev information using libblkid

### LINKED ISSUE #1506: Stratis should double check that devices are not used with blkid
In the event there is an issue with udev, Stratis should double check whether a block device is in use.
One method would be to use blkid <device>.

Here is an instance where udev didn't have a device listed in the db.
https://bugzilla.redhat.com/show_bug.cgi?id=1591832

jbaublitz:
Closes #1506


mulkieran:
@jbaublitz I believe that alll you have to do is rebase and all tests will pass.

jbaublitz:
@mulkieran There's an interesting case that we may want to handle. The most recent test failure in the udev tests is due to a lag between udev and initializing the Stratis metadata on the mapped device for encryption. I wonder if it makes more sense to make it a little bit more specific in the case of udev and the superblock not matching where if a Stratis superblock is found, we proceed with the normal Stratis metadata handling.

mulkieran:
@jbaublitz Change looks good to me.

PR #2831, opened 2021-11-05
Make chrono depend on branch with call to localtime_r removed

mulkieran:


mulkieran:
Closing. We last had a CVE from chrono dependency in stratisd 3.2.0.

PR #2852, opened 2021-11-22
Change liminal device code to respect udev remove events

### LINKED ISSUE #2456: Unnecessary warnings on encrypted pool destruction
Related stratis-storage/project#242

```
[2021-03-17T14:18:27Z WARN  libstratis::engine::strat_engine::liminal::identify] udev identified device /dev/dm-0 as a Stratis device but device /dev/dm-0 could not be opened for reading: No such file or directory (os error 2), disregarding the device
```

This appears to occur because when Stratis rescans the udev database, the device is still present in the database but the device node has already been removed.

This may be a use case for udev synchronization, but we need to be aware that libdevmapper is not thread safe so we should use caution before introducing that.

jbaublitz:
Closes #2456

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -252,6 +252,8 @@ install: $(PROFILETARGET) $(PROFILEMINTARGET) install-cfg
 	install -Dpm0755 -t $(DESTDIR)$(UDEVDIR) target/$(PROFILEDIR)/stratis-utils
 	mv -fv $(DESTDIR)$(UDEVDIR)/stratis-utils $(DESTDIR)$(UDEVDIR)/stratis-str-cmp
 	ln -fv $(DESTDIR)$(UDEVDIR)/stratis-str-cmp $(DESTDIR)$(UDEVDIR)/stratis-base32-decode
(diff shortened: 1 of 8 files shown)


jbaublitz:
@mulkieran We may need another addition to stratis-utils to fully do what I'm looking to do to fix the problem. If we had a udev utility to report the Stratis device and pool UUIDs for Stratis LUKS devices, we would hardly have to make any changes to your code in liminal devices.

mulkieran (changes requested):
Looks good! Please grab commit https://github.com/stratis-storage/stratisd/commit/522834eaeb43b0846cb8a88a988c85efc21b1a16 and apply it. It simply moves the ```update_on_remove``` method back to its previous position.

jbaublitz:
Ready for review!

mulkieran (approved):

jbaublitz:
@mulkieran Feel free to cherry pick this commit onto another branch after per-pool locking is merged and merge this as well if you would like. Otherwise I will rebase it in January.

mulkieran:
Thanks, it is in.

PR #2870, opened 2021-12-08
Fix Rust 1.57 clippy errors

bgurney-rh:
Related: stratis-storage/project#368

### DIFF
--- a/src/engine/strat_engine/liminal/identify.rs
+++ b/src/engine/strat_engine/liminal/identify.rs
@@ -591,7 +591,6 @@ mod tests {
         )
         .unwrap();
 
-        #[allow(clippy::redundant_closure)]
         for path in paths {
             let device_path = DevicePath::new(path).expect("our test path");
             let info = block_device_apply(&device_path, process_stratis_device)
@@ -637,7 +636,6 @@ mod tests {
     fn test_process_device_uninitialized(paths: &[&Path]) {
         assert!(!paths.is_empty());
 
-        #[allow(clippy::redundant_closure)]
         for path in paths {
             let device_path = DevicePath::new(path).expect("our test path");
             assert_eq!(
@@ -654,7 +652,6 @@ mod tests {
             );
         }
 
-        #[allow(clippy::redundant_closure)]
         for path in paths {
             create_fs(path, None, false).unwrap();
             let device_path = DevicePath::new(path).expect("our test path");
--- a/src/engine/strat_engine/liminal/identify.rs
+++ b/src/engine/strat_engine/liminal/identify.rs
@@ -591,9 +591,10 @@ mod tests {
         )
         .unwrap();
 
+        #[allow(clippy::redundant_closure)]
         for path in paths {
             let device_path = DevicePath::new(path).expect("our test path");
-            let info = block_device_apply(&device_path, process_stratis_device)
+            let info = block_device_apply(&device_path, |dev| process_stratis_device(dev))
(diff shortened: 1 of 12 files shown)


bgurney-rh:
Not quite done yet; still 5 redundant closures to fix, and the fix in the last commit might not be correct.

bgurney-rh:
I have only one clippy error left:
```
error: called `expect` on `max_size` after checking its variant with `is_none`
   --> src/engine/strat_engine/tests/real.rs:118:28
    |
115 |         if max_size.is_none() {
    |         --------------------- help: try: `if let Some(..) = max_size`
...
118 |             let max_size = max_size.expect("!max_size.is_none()");
    |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
```

...but I can never understand the rust-lang tutorial on "if let"; I think it might need more insight into what variable to put in there.

mulkieran:
> I have only one clippy error left:
> 
> ```
> error: called `expect` on `max_size` after checking its variant with `is_none`
>    --> src/engine/strat_engine/tests/real.rs:118:28
>     |
> 115 |         if max_size.is_none() {
>     |         --------------------- help: try: `if let Some(..) = max_size`
> ...
> 118 |             let max_size = max_size.expect("!max_size.is_none()");
>     |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>     |
> ```
> 
> ...but I can never understand the rust-lang tutorial on "if let"; I think it might need more insight into what variable to put in there.

They want you to switch around a bunch of things to get this statement:
```
    // Retain only devices that are less than max_size
    let (mut matches, mut too_large): (Vec<(&Path, Sectors)>, Vec<(&Path, Sectors)>) =
        if let Some(max_size) = max_size {
            matches.iter().partition(|&&(_, s)| s <= max_size)
        } else {
            (matches, vec![])
        };
```

bgurney-rh:
Okay; it was changing from `if ...is_none()` to `if let Some(...)`.

This latest set of commits has no clippy errors, and builds cleanly in 1.57.0 and 1.56.0.

src/engine/strat_engine/liminal/identify.rs
```diff
@@ -591,6 +591,7 @@ mod tests {
         )
         .unwrap();
 
+        #[allow(clippy::redundant_closure)]
```
mulkieran, line 594: Plz remove any of the redundant closures that you can remove in this file.

mulkieran (changes requested):
One request...

src/engine/strat_engine/liminal/identify.rs
bgurney-rh: I just cherry-picked the commit from when I tried that the last time... and it worked; hooray.

mulkieran: Great, keep going and remove all allows that you can remove, I _think_ it should be all.

bgurney-rh: Ah yes, I have to remove the allow comments now.

bgurney-rh: I think this is as far as I can go.  The only ones left are the dbus_api ones, that show "expected signature" errors if I try to remove the allow.

mulkieran (approved):
Looks good. Plz squash into something reasonable, and I'll merge.

PR #2883, opened 2021-12-21
Strip out redundancy field in engines

mulkieran:


mulkieran:
This is just a consequence of clippy alertly noting that the redundancy fields were entirely dead. Also, the bit rot had begun, so it is more than time for this to go.

PR #2884, opened 2021-12-22
Use version 2 of crc crate

mulkieran:


mulkieran:
We already have an existing PR for this.

PR #2892, opened 2022-01-05
Run cargo-update

mulkieran:


mulkieran:
Pushing to drop dependency on yanked sha2 crate.

mulkieran:
Filed a bug just in case we find we need to do anything about vendored tarfiles: `https://github.com/stratis-storage/project/issues/387.`

PR #2896, opened 2022-01-10
Include stratis dracut module by default if it is installed

### LINKED ISSUE #2893: Dracut module inclusion problem
@refi64 I'm pulling out our discussion from #2828 here.

The cause for this change was that we noticed a failure on some of our machines during a Fedora upgrade. The behavior that we were seeing seemed to indicate that the configuration file was the cause of this failure and we were trying to avoid unbootable systems for our users. Unfortunately, this seems to have caused other issues.

I want to propose an option that I think should remove the failure case we were seeing with the config file as well as the failure case you noticed.

The failure case we noticed is that if there is a config file enabling a module and the module is not present, the automatic initramfs regeneration on kernel updates fails silently and the system will be unbootable on the next reboot. This failure case was triggered during Fedora upgrades and we wanted to avoid that. However the only reason we need the configuration file is because of this line of code: https://github.com/stratis-storage/stratisd/blob/master/dracut/90stratis/module-setup.sh#L20.

If we were to change the return code to 0, the dracut module would always be included if it is present on the system without the need to ever have a configuration file. This both avoids the problem you bumped into and the problem we bumped into. Would you be satisfied with that as a fix in the next patch release?

jbaublitz:
Closes #2893

### DIFF
--- a/dracut/90stratis/module-setup.sh
+++ b/dracut/90stratis/module-setup.sh
@@ -17,7 +17,7 @@ check() {
 		/usr/lib/udev/stratis-str-cmp \
 		/usr/lib/udev/stratis-base32-decode ||
 		return 1
-	return 255
+	return 0
 }
 
 # called by dracut


jbaublitz:
@refi64 @mulkieran I'm leaving the stratis-clevis module as opt-in. Any objections?

mulkieran (approved):

mulkieran:
attributing tang test failure to GitHub flakiness.

PR #2899, opened 2022-01-14
No longer ignore a high devicemapper version in rawhide

mulkieran:


mulkieran:
trivial if COMPARE_FEDORA_VERSIONS test passes

PR #2925, opened 2022-02-10
Require libcryptsetup-rs 0.5.0

mulkieran:


jbaublitz (changes requested):
We should be also able to remove the `unsafe impl Send {}` on the safe key memory data structure.

mulkieran:
> We should be also able to remove the `unsafe impl Send {}` on the safe key memory data structure.

Done!

jbaublitz (approved):

PR #2934, opened 2022-02-22
Stop allowing high versions of libcryptsetup

mulkieran:


mulkieran:
I canceled the irrelevant ones and now merging.

Issue #2814, opened 2021-10-19
Thin provisioning rework proposal


jbaublitz:
Recently we've bumped into some reliability and usability problems with our thin provisioning solution. Some of the notable problems, some of which have been resolved, are:
* Filesystem extension not being triggered
* Imprecise metadata size calculation
* No guard rails for out of space conditions on the thin pool

After a number of discussions with the LVM team and other storage engineers, I'm suggesting the following changes:
* Decoupling filesystem extension from the thin pool low water mark (this change has already been merged)
  * We have added a timer to periodically check for the need to extend the filesystem.
  * Thin devices have sizes, but the size is not constrained by physical storage so there is no event to listen for when the thin device (and filesystem on top of it) begins to run out of space.
  * Previously, relying on the low water mark alone could occasionally cause the filesystem extension to fail and the filesystem would fill up before resizing was retried.
* Switch to lazy allocation for the data and metadata thin pool device
  * Currently, our metadata calculation is simply 1/1000th of the total available space.
  * The proposed solution involves the following:
    * Allocate data and metadata device space as needed.
    * Disable flush on device extension. This involves disabling flush on suspend, extending the table and allocating the storage from the cap device, and disabling flush on resume. This will prevent unnecessary IO storms on device extension.
    * Data device extension will be triggered by the thin data device low water mark and metadata extension will be calculated accordingly.
    * We do not need to listen for the metadata low water mark because of the restrictions below on metadata growing operations.
    * The benefit of this is increased flexibility around usage allocated to data and metadata space depending on whether data device size and number of filesystems on the pool changes over time.
* For metadata extension, this can be triggered by either growing the data device (background) or increasing the limit on thin devices (IPC request). We will disable filesystem creation and data device extension if the metadata device cannot be extended accordingly.
  * Calculate the necessary metadata space using the `thin_metadata_size` tool, and resize the metadata accordingly.
  * This allows us to disregard the metadata device low water mark because we will always be guaranteed to have enough space for metadata operations.
  * Add a limit on thin devices that can be created on a specific pool to ensure that the user does not create more thin devices than there is room to record metadata for.
* D-Bus API:
  * Writable property for limit on filesystems for a pool. Changing the value may cause a metadata device resize.
    * This limit will be applied to the pool on filesystem creation and any attempt to create more filesystems than the limit allows will be prevented.
  * (Optional) Writable property for data device resize allocation size (how much the data device should grow when it begins to run out of space)
[...]

PR #2942, opened 2022-03-02
Add macro for retry operations and use for MDV unmount

jbaublitz:
Pulled out from #2814


mulkieran (changes requested):
I think that top rule of ```retry_operation``` must always be matched, so the desired expansion will not occur.

I'm not absolutely certain that a macro is the best choice. Would the retry crate be useful here[1]? It is already packaged in Fedora.

[1] https://crates.io/crates/retry

mulkieran:
One more thing, the expand_macros test actually does expand, I think correctly. https://github.com/stratis-storage/stratisd/runs/5407644334?check_suite_focus=true.

jbaublitz:
Okay this is ready for review!

mulkieran (approved):

mulkieran:
@jbaublitz Plz sqash and I'll merge.

PR #2948, opened 2022-03-07
Add UnownedDevices

mulkieran:
Closes: https://github.com/stratis-storage/stratisd/issues/1968

* Does check for an empty list of devices earlier in ```init_cache``` and ```create_pool``` methods. The idea is to treat an empty list as much as possible as simply an invalid argument, even though it is typed correctly.
* In StratEngine uses ```ProcessedPathInfos::try_from``` to process the paths and discover their information in the ```add_blockdevs```, ```init_cache```, and ```create_pool``` methods. Checks for potential errors in the result _before_ invoking the code that will actually initialize the device.
* Retains the use of the ```*idempotent*``` methods for the strat engine, but only invokes them if all the passed paths could conceivably belong to the given pool.

Advantages:
* Raises the inspection of devices up to a higher level, to the pool, or in the case of creating a pool, the engine level. This is the principal goal. The reasons for doing this are to allow stratisd to have a full understanding of the properties of the devices it has been given as an argument at the same time as it can inspect information about all the existing devices in _both_ tiers, in order to decide whether they should be considered compatible or not. Currently, when creating a pool, everything about the devices to be used for the pool are known at the time of creation. But when initializing a cache, only the info about the proposed cache devices is known. And when adding devices to either tier, only the information about that tier is known. It is not known, at all, whether the devices are cache or data devices, at the point where they are inspected. This last could be remedied by passing a Tier argument down to the blockdev manager methods, but the other issues could not be.
* When creating a pool, the lock on the possibly existing pool doesn't need to be taken until after the information on the devices has been read. This is a bit of an improvement, because the lock, which would conflict w/, say a global write lock, is held for a shorter time.
* When initializing or extending the cache, it is necessary to suspend the thinpool. By reading the data from the devices before initializing or extending the cache, the suspend/resume interval is made a bit shorter. Not that much, probably, compared to the time that it takes to initialize the devices, but still an improvement.

NOTES:
* There is a lot of code, now, that is transitively invoked by the initialization methods, that inspects the devices a second time, that I expect to remove in future, if I continue with this. I left it there for now, because it will require a lot of tinkering w/ tests, etc., to clean it up.
* We have a lot of code in the CLI to try to guard against and give good messages in case the user specifies devices that belong to another pool or the same pool but a different tier, when adding, creating, or initializing. We can't really remove this, and handle it in stratisd, because although the inspection of devices is hoisted, it would all have to be hoisted up into the engine to make this doable. We could, however, send out good error messages for when a user specifies devices in one tier, but they belong to the opposite tier. I haven't taken the trouble to implement really complete error messages in this case, as of now, but it is possible.
* I think the ```can_unlock()``` check taken before initializing devices Ishould be retained in the blockdev manager's add method because it is checking whether it is possible to go ahead and actually initialize the devices.

Questions:
* The result of these changes is that in ```StratEngine::CreatePool``` the devices are read and processed, although not initialized, outside the blocking task where this would have previously happened. How should this change be viewed? I don't see any reason why reading the data from the devices need to be in a blocking thread.
* etc.

### DIFF
--- a/src/engine/sim_engine/pool.rs
+++ b/src/engine/sim_engine/pool.rs
@@ -191,17 +191,18 @@ impl Pool for SimPool {
     ) -> StratisResult<SetCreateAction<DevUuid>> {
         validate_paths(blockdevs)?;
 
+        if blockdevs.is_empty() {
+            return Err(StratisError::Msg(
+                "At least one blockdev path is required to initialize a cache.".to_string(),
+            ));
+        }
+
         if self.is_encrypted() {
             return Err(StratisError::Msg(
                 "Use of a cache is not supported with an encrypted pool".to_string(),
             ));
         }
         if !self.has_cache() {
-            if blockdevs.is_empty() {
-                return Err(StratisError::Msg(
-                    "At least one blockdev path is required to initialize a cache.".to_string(),
-                ));
-            }
             let blockdev_pairs: Vec<_> = blockdevs.iter().map(|p| SimDev::new(p, None)).collect();
             let blockdev_uuids: Vec<_> = blockdev_pairs.iter().map(|(uuid, _)| *uuid).collect();
             self.cache_devs.extend(blockdev_pairs);
--- a/src/engine/sim_engine/pool.rs
+++ b/src/engine/sim_engine/pool.rs
@@ -189,13 +189,13 @@ impl Pool for SimPool {
         _pool_name: &str,
         blockdevs: &[&Path],
     ) -> StratisResult<SetCreateAction<DevUuid>> {
-        validate_paths(blockdevs)?;
-
         if blockdevs.is_empty() {
             return Err(StratisError::Msg(
                 "At least one blockdev path is required to initialize a cache.".to_string(),
             ));
         }
+
+        validate_paths(blockdevs)?;
 
         if self.is_encrypted() {
             return Err(StratisError::Msg(
@@ -268,8 +268,6 @@ impl Pool for SimPool {
         paths: &[&Path],
         tier: BlockDevTier,
     ) -> StratisResult<(SetCreateAction<DevUuid>, Option<PoolDiff>)> {
-        validate_paths(paths)?;
-
         if tier == BlockDevTier::Cache && !self.has_cache() {
             return Err(StratisError::Msg(
                     "The cache has not been initialized; you must use init_cache first to initialize the cache.".to_string(),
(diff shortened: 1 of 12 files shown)


jbaublitz:
> Questions:
> 
> The result of these changes is that in StratEngine::CreatePool the devices are read and processed, although not initialized, outside the blocking task where this would have previously happened. How should this change be viewed? I don't see any reason why reading the data from the devices need to be in a blocking thread.
> etc.
> 

Per the tokio docs, any IO should likely be in a blocking thread because if that device hangs for some reason, for example, it would lock up all tasks spawned on that thread. Blocking simply means that a single task is allocated to that thread.

src/engine/sim_engine/pool.rs
```diff
@@ -191,17 +191,18 @@ impl Pool for SimPool {
     ) -> StratisResult<SetCreateAction<DevUuid>> {
         validate_paths(blockdevs)?;
 
+        if blockdevs.is_empty() {
```
jbaublitz, line 194: Should this go before `validate_paths`?

src/engine/strat_engine/engine.rs
```diff
@@ -351,19 +353,51 @@ impl Engine for StratEngine {
 
         validate_paths(blockdev_paths)?;
 
+        if blockdev_paths.is_empty() {
```
jbaublitz, line 356: Should this go before `validate_paths`?

src/engine/strat_engine/pool.rs
```diff
@@ -422,22 +422,48 @@ impl Pool for StratPool {
     ) -> StratisResult<SetCreateAction<DevUuid>> {
         validate_paths(blockdevs)?;
 
+        if blockdevs.is_empty() {
```
jbaublitz, line 425: Same comment about `validate_paths`.

jbaublitz (commented):
Just one note in a few places.

src/engine/sim_engine/pool.rs
mulkieran: Done.

src/engine/strat_engine/engine.rs
mulkieran: Done.

src/engine/strat_engine/pool.rs
mulkieran: Done.

src/engine/strat_engine/pool.rs
```diff
@@ -584,31 +625,96 @@ impl Pool for StratPool {
...
                     pool_name
                 )
             ));
-        } else if paths.is_empty() {
-            //TODO: Substitute is_empty check with process_and_verify_devices
+        }
+
+        validate_paths(paths)?;
+
+        if paths.is_empty() {
```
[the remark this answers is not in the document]
mulkieran: Done.

mulkieran:
rebased.

PR #2956, opened 2022-03-21
Use checkout action to select tag

mulkieran:


mulkieran:
This is nightly only; it shouldn't get in anybody's way.

PR #2970, opened 2022-04-09
Initial work for removing redundancy option

### LINKED ISSUE #2969: Remove redundancy option from PoolCreate method
it does nothing.

mulkieran:
Closes #2969

### DIFF
--- a/src/dbus_api/api/manager_3_1/api.rs
+++ b/src/dbus_api/api/manager_3_1/api.rs
@@ -1,0 +1,46 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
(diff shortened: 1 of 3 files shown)


mulkieran:
This will compete w/ other work currently happening w/ CreatePool, so I'm putting it in Pending and not completing it now.

faangbait:
> This will compete w/ other work currently happening w/ CreatePool, so I'm putting it in Pending and not completing it now.

Can you specify/link to what functionality this will compete with? Is redundancy coming in from another direction in the near future?

jbaublitz:
Hi @faangbait! I just want to chime in here to say that this would simply remove the redundancy operation which is effectively dead at the moment. The only option that it currently accepts is "none". While we eventually plan to add redundancy, our current recommendation is to use a RAID device underneath Stratis and provide that to the create pool method if redundancy is desired. This is not removing redundancy as a feature, just a placeholder parameter.

mulkieran:
This might be completely out-of-date now.

PR #2975, opened 2022-04-21
Increase libblkid-rs dependency to 0.2.0

mulkieran:


mulkieran:
Defer until just before next upstream release.

mulkieran:
I think we can blame the test failure on udev.

mulkieran:
@bgurney-rh @jbaublitz @bmr-cymru This is a good time to bump the dependency, although the changes in libblkid-rs 0.2.0 will not be needed until the next release. In Fedora, that means will have to include the new version of libblkid-rs in the side-tag, along w/ stratis-cli and stratisd, but we might as well get that over with in this Fedora release.

bgurney-rh (approved):

jbaublitz (approved):

bmr-cymru (approved):

PR #2980, opened 2022-04-23
Create MDV filesystem with bigtime option

mulkieran:


mulkieran:
Looks like creating the MDV w/ bigtime set is causing problems.

mulkieran:
This may be a problem w/ the tests, there is a failure to teardown the fail device, used for testing, consistently.

mulkieran:
I want to investigate why the CI failed; but the motivation for doing this has gone away as we're using the private namespace for the MDV instead.

mulkieran:
This must be obsolete, since bigtime=1 is the default in xfs these days. Closing.

PR #2949, opened 2022-03-07
Add ability to start and stop pools

jbaublitz:
Closes stratis-storage/project#296

PR #2988, opened 2022-05-03
Move namespace code to separate module

jbaublitz:
Related to #2949 
Related https://github.com/stratis-storage/project/issues/454

### DIFF
--- a/src/engine/strat_engine/engine.rs
+++ b/src/engine/strat_engine/engine.rs
@@ -24,8 +24,9 @@ use crate::{
         strat_engine::{
             cmd::verify_binaries,
             dm::get_dm,
-            keys::{MemoryFilesystem, StratKeyActions},
+            keys::StratKeyActions,
             liminal::{find_all, LiminalDevices},
+            ns::MemoryFilesystem,
             pool::StratPool,
         },
         structures::{
@@ -60,12 +61,9 @@ pub struct StratEngine {
     // Handler for key operations
     key_handler: Lockable<Arc<RwLock<StratKeyActions>>>,
 
-    // TODO: Remove this code when Clevis supports reading keys from the
-    // kernel keyring.
-    // In memory filesystem for passing keys to Clevis.
-    // See GitHub issue: https://github.com/stratis-storage/project/issues/212.
+    // In memory filesystem for private namespace mounts.
     #[allow(dead_code)]
-    key_fs: MemoryFilesystem,
+    fs: MemoryFilesystem,
 }
 
 impl StratEngine {
@@ -79,6 +77,7 @@ impl StratEngine {
     /// Returns an error if there was an error reading device nodes.
     /// Returns an error if the binaries on which it depends can not be found.
     pub fn initialize() -> StratisResult<StratEngine> {
+        let fs = MemoryFilesystem::new()?;
         verify_binaries()?;
 
         let mut liminal_devices = LiminalDevices::default();
@@ -92,7 +91,7 @@ impl StratEngine {
             liminal_devices: Lockable::new_shared(liminal_devices),
             watched_dev_last_event_nrs: Lockable::new_shared(HashMap::new()),
             key_handler: Lockable::new_shared(StratKeyActions),
-            key_fs: MemoryFilesystem::new()?,
+            fs,
         })
     }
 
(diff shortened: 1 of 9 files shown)


src/engine/strat_engine/ns.rs
```diff
@@ -0,0 +1,90 @@
...
+/// Path to the root mount namespace
+const INIT_MNT_NS_PATH: &str = "/proc/1/ns/mnt";
+/// Path to where private namespace mounts are mounted
+pub const NS_TMPFS_LOCATION: &str = "/run/stratisd/ns_mounts";
+
+/// Check if the stratisd mount namespace for this thread is in the root namespace.
+pub fn is_in_root_namespace() -> StratisResult<bool> {
+    let pid_one_stat = stat(INIT_MNT_NS_PATH)?;
+    let self_stat = stat(format!("/proc/self/task/{}/ns/mnt", gettid()).as_str())?;
+    Ok(pid_one_stat.st_ino == self_stat.st_ino)
```
bmr-cymru, line 38: Should this also verify that `pid_one_stat.st_dev == self_stat.st_dev`? Man 7 namespaces says this:

```
In  Linux  3.7  and earlier, these files were visible as hard links.  Since Linux 3.8,
they appear as symbolic links.  If two processes are in the same namespace,
then the device IDs and inode numbers of their /proc/[pid]/ns/xxx symbolic links
will be the same; an application can check this using the stat.st_dev and
stat.st_ino fields returned by stat(2). The content of this symbolic link is a string
containing the namespace type and inode number as in the following example:

           $ readlink /proc/$$/ns/uts
           uts:[4026531838]
```

I can't think of a "normal" situation where the `stat.st_dev` of two paths in procfs would differ (all files in procfs report the same device - 0x16/22d) but historically it was possible to bind mount files or directories over parts of `/proc`, which could make this assumption false.

Trying that on a recent kernel appears to fail in all cases I can think of so far, but equally I can't think of a case where checking for an `st_dev` match would cause a problem?

src/engine/strat_engine/ns.rs
```diff
@@ -0,0 +1,90 @@
...
+        )?;
+        Ok(MemoryFilesystem)
+    }
+}
+
+impl Drop for MemoryFilesystem {
+    fn drop(&mut self) {
+        if let Err(e) = umount(NS_TMPFS_LOCATION) {
+            warn!(
+                "Could not unmount temporary in memory storage for Clevis keyfiles: {}",
```
bmr-cymru, line 95: Should this error string be made more general, since `MemoryFilesystem` will also be used as the mount location for the MDV? Although it's an unlikely error in practice the reference to Clevis may be consusing for users who have not configured encryption? (but who do need the private MDV mount).

bmr-cymru: "Could not unmount temporary in memory storage for private file system mounts"?

bmr-cymru (changes requested):
I think the error string for `MemoryFilesystem`'s drop warning could be improved, and I wonder if the check for the root namespace couldn't be made a bit more robust? (although this may be a purely theoretical concern with modern kernels).

src/engine/strat_engine/ns.rs
jbaublitz: Yes, thanks for catching this, this was an oversight. That's along the lines of text I had before I checked out a file and forgot to rework it again. I'll add that to the PR!

src/engine/strat_engine/ns.rs
jbaublitz: I'm fine with adding additional protections! I'll make it check both.

src/engine/strat_engine/engine.rs
```diff
@@ -65,7 +66,7 @@ pub struct StratEngine {
     // In memory filesystem for passing keys to Clevis.
     // See GitHub issue: https://github.com/stratis-storage/project/issues/212.
     #[allow(dead_code)]
-    key_fs: MemoryFilesystem,
+    fs: MemoryFilesystem,
```
mulkieran, line 69: Are you able to drop comment and also dead code annotation above?

mulkieran (changes requested):
Looks good, just the one request...

src/engine/strat_engine/engine.rs
jbaublitz: The dead code annotation is still required but I've modified the comment!

mulkieran (approved):
Looks good.

PR #2854, opened 2021-11-24
Lazy allocation (thin provisioning rework)

jbaublitz:
Follow up to #2830 
Related to #2814

PR #2830, opened 2021-11-03
Thin pool rework (lazy allocation and more precise metadata size calculation)

jbaublitz:
Related to #2814

PR #2991, opened 2022-05-04
Fix for thin provisioning rework in boot work

jbaublitz:
Related to #2854

I've tested this and verified it resolves the issue.

### DIFF
--- a/dracut/90stratis/module-setup.sh
+++ b/dracut/90stratis/module-setup.sh
@@ -14,6 +14,7 @@ check() {
 		udevadm \
 		plymouth \
 		/usr/sbin/plymouthd \
+		/usr/sbin/thin_metadata_size \
 		/usr/lib/udev/stratis-str-cmp \
 		/usr/lib/udev/stratis-base32-decode ||
 		return 1
@@ -45,6 +46,7 @@ install() {
 		udevadm \
 		plymouth \
 		/usr/sbin/plymouthd \
+		/usr/sbin/thin_metadata_size \
 		/usr/lib/udev/stratis-str-cmp
 	ln ${initdir}/usr/lib/udev/stratis-str-cmp ${initdir}/usr/lib/udev/stratis-base32-decode
 


mulkieran (approved):

mulkieran:
fedora repos are funky; might as well ignore those failures.

bmr-cymru:
The patch looks completely correct, but I wanted to pull this branch onto my Stratis rootfs test system and try out some tests there. I've had to bump up my VM's memory allocation to get the current branches building reliably - that's finished now but I'm still seeing a boot failure with the current start-stop branch including this change. I'm just waiting for the dracut timeouts to expire so that I can try to debug it. I'll update the PR again when I have a better idea what's going on.

This did prompt me to take another look at #2830 and in particular the commit that introduces `thin_metadata_size`:

```
commit 0d1d50972cc8e5869e7154294455640c1877e358
Author: John Baublitz <jbaublitz@redhat.com>
Date:   Wed Nov 3 16:36:42 2021 -0400

    Transition from greedy to lazy allocation
    
    This commit adds a few notable changes. One is a transition to lazy
    allocation with migration code to ensure that older versions of stratisd
    still function properly after migration. Another is a filesystem limit
    to allow precise calculation of the necessary metadata space for the
    thin pool. This ensures that a user can never run out of metadata space
    as the amount required is calculated and allocated upfront.
```

My only concern here is the naming - thin-provisioning-tools has long used `thin_` as the prefix for the thinp tools that it provides, and I wonder if this might have the potential to cause some confusion? It could also lead to conflicts if thin-provisioning-tools ever introduced a similarly named command (there is already `thin_metadata_pack` and `thin_metadata_unpack`.

bmr-cymru:
Ah never mind - I had mis-read the commit that I mentioned, and I thought that `thin_metadata_size` was being provided by stratisd. I can see now that it's just invoking the tool supplied by thin-provisioning-tools, which is absolutely fine!

bmr-cymru (approved):
Tested on an F35 Stratis root fs system and I can successfully boot with this patch included in the start-stop branch.

PR #2994, opened 2022-05-09
Make fs_limit on setup no less than number of filesystems present

### LINKED ISSUE #2993: Must ensure that fs_limit is never less than the number of filesystems in a pool
Currently, if it is not set in the metadata, it is set to the default limit, it should be set at least to the maximum of the number of fileystems in the pool and the default limit.

mulkieran:
Closes #2993


bmr-cymru (approved):

jbaublitz (approved):

mulkieran:
rebased.

PR #2941, opened 2022-03-02
Update lowest supported Fedora to 34

bgurney-rh:
Related: stratis-storage/project#400

PR #2998, opened 2022-05-14
Drop support for skipping duplicate names test

### LINKED ISSUE #2984: Might be able to remove old infrastructure associated with SKIP_DUPLICATE_NAMES_TEST
It seemed to fail specifically on Fedora 33 containers and we're no longer running tests on Fedora 33 anywhere. See previous issue: https://github.com/stratis-storage/stratisd/issues/2719

mulkieran:
Closes #2984


mulkieran:
I think this could just as well have been included in #2941, so marking as trivial. Relevant test passed, so merging.

PR #3023, opened 2022-06-16
Fix space calculation algorithm

### LINKED ISSUE #3022: Convergence problem in metadata size calculation
@bmr-cymru found an issue where the data/metadata calculation algorithm fails to converge on disks that have 18872320 sectors of size.

My best guess for what is happening here is that the convergence condition is incorrect. It requires both the aligned halfway point between the upper and lower limit to be equal to the lower limit AND the metadata size calculated based on the upper limit to be equal to the metadata size calculated based on the lower limit.

This second condition does not make any sense. The upper limit by definition is larger than the total space so it is entirely possible that the upper limit would have a metadata size that is different from the lower limit even in the case where the value has correctly converged.

jbaublitz:
Closes #3022

### DIFF
--- a/src/engine/strat_engine/thinpool/thinpool.rs
+++ b/src/engine/strat_engine/thinpool/thinpool.rs
@@ -204,80 +204,106 @@ impl fmt::Display for ThinPoolStatusDigest {
 
 /// Returns the determined size of the data and metadata devices.
 ///
-/// This method implements something similar to a binary search. The upper and lower
-/// limits are data device sizes.  The upper limit should be the total available
-/// space (total_space) as this leaves no room for metadata. We can assume the data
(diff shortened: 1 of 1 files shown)


jbaublitz:
@mulkieran I'm fine setting the lower bound to 0 if you are.

mulkieran:
> @mulkieran I'm fine setting the lower bound to 0 if you are.

@jbaublitz That one change would make me more confident in the algorithm's continued correctness if changes might happen elsewhere in the code in the future. Plz go ahead and do that. Otherwise, I think it looks good.

jbaublitz:
Okay ready for merge!

mulkieran (approved):

Issue #1556, opened 2019-06-11
[feature request] Guix install target


FruityWelsh:
I would think building a Guix package would be useful for both end users and developers of this project. 

For both:
1. Easy setup and dependency setup, 
2. distro-agnostic
For developers:
1. Easy ability to change dependency versions (while creating an easy way to check if these have been changed.
2. Build reproducibility to help reduce variation among devs
3. guix environments allow you to quickly change the dev environment 

For advanced users:
1. Easy way to build closer to the metal, and optimize for a particular machine.

PR #3064, opened 2022-08-02
Reflect changes to LIBEXECDIR and UNITEXECDIR in systemd units

NickCao:


mulkieran (approved):

cgwalters:
Looks reasonable to me.  Templating systemd units like this is quite common.

It'd be good to have some motivation in the commit message; I'm assuming this is for e.g. Debian where there isn't a libexecdir as is?

NickCao:
I'm utilizing these changes for packaging in nixpkgs: https://github.com/NixOS/nixpkgs/pull/184675, in general all the hardcoded paths should be parametrizable to accommodate non-fhs distributions (NixOS or Guix https://github.com/stratis-storage/stratisd/issues/1556, for example) or distributions with their own policies for where these files should go.

jbaublitz (approved):

PR #3068, opened 2022-08-03
Replace invocatins to thin-provisioning-tools with native rust binding

NickCao:
Related https://github.com/stratis-storage/project/issues/383

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 3
 
 [[package]]
+name = "adler"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+
+[[package]]
 name = "aho-corasick"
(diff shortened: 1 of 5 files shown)


NickCao:
Just noticed that stratisd it published on crates.io, so maybe git deps is a no go?

mulkieran:
@NickCao Thanks for the PR! Unfortunately, the Cargo.toml change will make Fedora packaging impossible. It's a change we have wanted to make though, so we'll check into some of the ideas we were offered, and may be able to find a solution that will push it to completion. But this will take a bit longer.

mulkieran:
> Just noticed that stratisd is published on crates.io, so maybe git deps is a no go?

I guess that, too!

mulkieran:
Some expected but some interesting errors. We got a stack overflow in one place, which is unusual.

NickCao:
The current implementation is not 100% percent identical to the rust based cli, lacking some sanity checks, let alone the difference from the c version.

mulkieran:
@NickCao Do you mind if I close this? It's still a goal for us, but our code has moved on and the PR has gained some significant conflicts.

PR #3069, opened 2022-08-04
Build stratisd with the udev sync devicemapper-rs branch

bmr-cymru:
Update the build dependencies for devicemapper-rs to reference the bmr-udev-sync branch and include `ilbdevmapper.h` in the dependencies for CI environments.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -52,6 +52,7 @@ jobs:
           curl
           cryptsetup-devel
           dbus-devel
+          device-mapper-devel
           libblkid-devel
           make
           ncurses
@@ -100,8 +101,9 @@ jobs:
           clevis-luks
           cracklib-dicts
           curl
-          device-mapper-persistent-data
           dbus-devel
+          device-mapper-devel
+          device-mapper-persistent-data
           libblkid-devel
           make
           ncurses
@@ -132,6 +134,7 @@ jobs:
           dbus-daemon
           dbus-tools
           dbus-devel
+          device-mapper-devel
           libblkid-devel
           git
           glibc-static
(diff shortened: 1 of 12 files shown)


.github/workflows/main.yml
```diff
@@ -153,7 +156,7 @@ jobs:
           --print-address
       - name: Run stratis-cli tests
         run: >
-          RUST_LOG=stratisd=debug
+          RUST_LOG=trace
```
mulkieran, line 159: For stratis-cli tests, might as well stick with DEBUG level. They are run against the sim engine, and so don't tend to cause udev events of interest.

bmr-cymru: Thanks - I will make that change on my branch, and disable the other tests that I'm not concerned with to keep the runs to a minimum.

bmr-cymru:
The CI failures from Friday are due to missing udev rules / support programs in the test runner. This is due to the stratis-ci-container-action using the old "build-and-install" target that was removed in January. I've pushed an update to the GitHub action and modified this branch to point to it (stratis-ci-container-action@v2).

bmr-cymru:
Rebased & tidied up

mulkieran (approved):

jbaublitz (commented):
Could you explain why we need the remaining udev_settle calls? I was under the impression that this would would make those unnecessary.

mulkieran:
@bmr-cymru Ok! New devicemapper-rs release is up; so you can fix up Cargo.toml and put it up for final review.

jbaublitz (approved):

mulkieran:
@bmr-cymru I think we're all good with this. Plz squash any commits if you want to and rebase, and I'll merge.

mulkieran:
Oh, I see that this was actually due to https://bugzilla.redhat.com/show_bug.cgi?id=2176263 . Thanks all for identifying that.

mulkieran:
We have to hold this a bit to do some spec file testing.

PR #3079, opened 2022-08-10
Allow starting pools by name

### LINKED ISSUE #3078: Allow starting pools by name
We should leave the ability to start by UUID, but if we have a complete pool, we should also allow starting by name. This can be accomplished on the crypt side by adding the pool name into the LUKS2 Stratis token.

jbaublitz:
Closes #3078


jbaublitz:
Blocked by #3100

jbaublitz:
@mulkieran @bmr-cymru Okay, this is nearing completion. I think I've fixed all of the bugs caught by tests, and done a little bit of clean up in certain areas. I want to take one last pass at review of the diffs, and then this will be ready for review.

I will probably do that last pass later because I want to focus on #3097 and #3100 first.

src/engine/strat_engine/liminal/setup.rs
```diff
@@ -281,7 +297,7 @@ pub fn get_blockdevs(
     fn check_and_sort_devs(
         mut devs: Vec<StratBlockDev>,
         dev_map: &HashMap<DevUuid, (usize, &BaseBlockDevSave)>,
-    ) -> StratisResult<Vec<StratBlockDev>> {
+    ) -> BDARecordResult<Vec<StratBlockDev>> {
```
mulkieran, line 300: Suggest that you not change ```check_and_sort_devs``` at all for this PR. Instead, if ```check_and_sort_devs``` returns an error, then you know that you need to pull the BDAs out of the blockdevs + cachedevs combined and you can do that easily at line 345 and line 359.

I think that way is clearer and just a little more robust, too.

mulkieran (changes requested):

src/engine/strat_engine/liminal/setup.rs
jbaublitz: That doesn't work. `check_and_sort_devs` consumes the vector so we either need to return `Vec<StratBlockdev>` or `HashMap<DevUuid, BDA>` in the `Err(_)` variant. The best we can do is pass a mutable reference to `check_and_sort_devs` which still requires a change.

mulkieran: Whoops! Yes, that's unavoidable. I no longer have a concern about this code so you should feel free to carry on.

jbaublitz: Ok!

mulkieran:
@jbaublitz This really is minor, but I noticed that the introspect data shows:

```
    <method name="StartPool">
      <arg name="id" type="s" direction="in" />
      <arg name="unlock_method" type="(bs)" direction="in" />
      <arg name="id_type" type="s" direction="in" />
      <arg name="result" type="(b(oaoao))" direction="out" />
      <arg name="return_code" type="q" direction="out" />
      <arg name="return_string" type="s" direction="out" />
    </method>
```

Note that the "id" key and the "id_type" key are separated by the "unlock_method" key. Should we re-arrange? I wouldn't mind having "id_type" as the first parameter, then "id", then "unlock_method".

src/dbus_api/api/manager_3_4/methods.rs
```diff
@@ -0,0 +1,128 @@
...
+    let mut iter = message.iter_init();
+    let dbus_context = m.tree.get_data();
+    let default_return: (
+        bool,
+        (Path<'static>, Vec<Path<'static>>, Vec<Path<'static>>),
+    ) = (false, (Path::default(), Vec::new(), Vec::new()));
+    let return_message = message.method_return();
+
+    let id_str: &str = get_next_arg(&mut iter, 0)?;
+    let unlock_method = {
```
mulkieran, line 36: Need to reverse order in which values are unmarshaled...id_type before unlock_method.

mulkieran (changes requested):
One more request about D-Bus API...

mulkieran:
Ignore stratis-cli errors, they are expected.

mulkieran:
In standup, we agreed that we could include the name along with the UUID of the stopped pool in the StoppedPools property on the Manager object.

jbaublitz:
@drckeefe @mulkieran I've added the requested functionality. Please note that I discovered #3153 while I was looking into it so we should eventually take care of that.

src/engine/sim_engine/engine.rs
```diff
@@ -221,41 +221,44 @@ impl Engine for SimEngine {
...
+                    StoppedPoolInfo {
+                        info: pool.encryption_info(),
+                        devices: pool
+                            .blockdevs()
+                            .into_iter()
+                            .map(|(dev_uuid, _, bd)| PoolDevice {
+                                devnode: bd.devnode().to_path_buf(),
+                                uuid: dev_uuid,
+                            })
+                            .collect::<Vec<_>>(),
```
mulkieran, line 256: Please add updates for uuid_to_name in sim engine also.

mulkieran (changes requested):
One minor request so far...

src/engine/sim_engine/engine.rs
jbaublitz: Okay, this is resolved!

jbaublitz:
Blocked by #3161

src/engine/strat_engine/mod.rs
```diff
@@ -15,12 +15,14 @@ mod names;
...
 mod serde_structs;
+mod shared;
 mod thinpool;
+mod types;
 mod udev;
 mod writing;
 
 pub use self::{
-    backstore::{crypt_metadata_size, set_up_crypt_logging, CLEVIS_TANG_TRUST_URL},
+    backstore::{crypt_metadata_size, set_up_crypt_logging, StratBlockDev, CLEVIS_TANG_TRUST_URL},
```
mulkieran, line 25: I do  not believe you need to export StratBlockDev.

mulkieran (changes requested):
Two very minor things...still working.

mulkieran (changes requested):
Last two comments...

src/engine/strat_engine/liminal/liminal.rs
```diff
@@ -190,47 +212,49 @@ impl LiminalDevices {
             (Err(e), _) => return Err(e),
         };
 
+        let uuids = unlocked_devices
```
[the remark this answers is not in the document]
jbaublitz: I think `unzip()` may actually be less performant because we would actually have to collect the discarded elements into a `Vec` and then immediately drop it. At the very least, this is an unnecessary allocation.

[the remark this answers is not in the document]
mulkieran: ok!

mulkieran (approved):

PR #3088, opened 2022-08-17
Fix bug in calculating initial upper limit for data and metadata device size

### LINKED ISSUE #3086: assertion failed: datablocks_to_sectors(upper_data_aligned) + upper_limit_meta_size * 2u64 > total_space
I just upgraded to Stratis 3.2.1 yesterday, and while working with some files today (basic operations, just updating the cover art on some music files), all I/O completely froze & I got this error in the journal:

```
Aug 17 09:45:54 hp-fedora stratisd[1732]: thread 'stratis-wt-657' panicked at 'assertion failed: datablocks_to_sectors(upper_data_aligned) + upper_limit_meta_size * 2u64 >\n    total_space', src/engine/strat_engine/thinpool/thinpool.rs:303:5
Aug 17 09:45:54 hp-fedora stratisd[1732]: stack backtrace:
Aug 17 09:45:54 hp-fedora stratisd[1732]: note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
Aug 17 09:45:54 hp-fedora stratisd[1732]: [2022-08-17T14:45:54Z WARN  stratisd::engine::strat_engine::engine] Failed to get status for thread handling filesystem checks: task 1301 panicked
```

I'm...admittedly not sure how to debug this, I'm guessing there's some FS metadata or similar that would help?

jbaublitz:
Closes #3086 

@refi64 I reproduced the failure in our unit tests and I *think* this fix resolves your issue too. If it doesn't, we can discuss some more questions I have in the linked issue. Hope this helps!

### DIFF
--- a/.github/workflows/fedora.yml
+++ b/.github/workflows/fedora.yml
@@ -69,6 +69,7 @@ jobs:
           curl
           cryptsetup-devel
           dbus-devel
+          device-mapper-persistent-data
           libblkid-devel
           make
           systemd-devel
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -65,6 +65,7 @@ jobs:
           apt-get install -y
           clang
           curl
+          device-mapper-persistent-data
           libblkid-dev
           libcryptsetup-dev
           libdbus-1-dev
--- a/src/engine/strat_engine/thinpool/thinpool.rs
+++ b/src/engine/strat_engine/thinpool/thinpool.rs
@@ -275,25 +275,12 @@ fn divide_space(
     current_meta_size: Sectors,
     fs_limit: u64,
 ) -> StratisResult<(Sectors, Sectors)> {
-    // We add the expression at the end because sectors_to_datablocks is a floor
-    // function. If the available space is not aligned to DATA_BLOCK_SIZE, the
-    // remainder will be discarded.
-    //
-    // The below expression is added because if the remainder is zero, the
-    // total_space is guaranteed to be less than the total size calculated by
-    // the upper limit. The value of upper_data_aligned will be exact, not
-    // rounded down at all and so adding the additional metadata space guarantees
-    // that it will be larger than total_space. However, if the remainder is
-    // non-zero, we round up to the nearest data block which will always be
-    // a valid upper limit.
+    // The below expression of DataBlocks(1) is added because we must round up to the
+    // next largest data block which will always be a valid upper limit.
     //
     // This is required to satisfy the preconditions of search.
-    let upper_data_aligned = sectors_to_datablocks(current_data_size + available_space)
-        + if available_space % DATA_BLOCK_SIZE == Sectors(0) {
-            DataBlocks(0)
-        } else {
-            DataBlocks(1)
-        };
+    let upper_data_aligned =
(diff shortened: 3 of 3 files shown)


mulkieran (approved):

refi64:
Unfortunately, I don't think this resolved the issue. I rebuilt the Fedora package but applied this patch (and confirmed in the logs that the patch actually applied), then swapped out stratisd & restarted, but the same panics are appearing.

mulkieran:
> Unfortunately, I don't think this resolved the issue. I rebuilt the Fedora package but applied this patch (and confirmed in the logs that the patch actually applied), then swapped out stratisd & restarted, but the same panics are appearing.

Thanks again! @jbaublitz I think we need to allow for the possiblity that the predicted ```thin_metadata_size()``` requirement is less than the actual allocated thin metadata size, and I think we need to allow for that in the ```search()``` method as well.

jbaublitz:
@mulkieran Okay, that makes sense!

@refi64 So it appears that the bug that I found was a valid bug but a different one from the one you bumped into. Let me know if this process sounds similar to what you did, but I created a pool in v2.4.3 and then upgraded stratisd to 3.2.1. The panic was immediately visible. This branch appears to resolve that issue if that is indeed the same issue that you bumped into and not just an identical but unrelated panic message. @bmr-cymru and @mulkieran are going to review this now from a code correctness standpoint, but please do let us know if this fixes the problem on your end!

mulkieran:
@jbaublitz Could the test failure be due to ```available_size``` being exactly 0?

mulkieran:
These tests I wrote are failing:

```
    fn test_divide_space3() {
        divide_space(Sectors(0), Sectors(20 * IEC::Ki), Sectors(8), 100).unwrap();
    }

    #[test]
    fn test_divide_space4() {
       # increase metadata to 1024 and it succeeds
        divide_space(Sectors(0), Sectors(20 * IEC::Ki), Sectors(512), 100).unwrap();
    }

    #[test]
    fn test_divide_space5() {
        // increase available space to 1024 and it succeeds
        divide_space(Sectors(256), Sectors(20 * IEC::Ki), Sectors(512), 100).unwrap();
    }

    #[test]
    fn test_search1() {
        search(
            Sectors(0),
            sectors_to_datablocks(Sectors(20 * IEC::Ki)) + DataBlocks(1),
            sectors_to_datablocks(Sectors(0)),
            sectors_to_datablocks(Sectors(20 * IEC::Ki)),
            Sectors(512),
            100,
        )
        .unwrap();
    }
```

Proposed explanation: If the predicted required metadata size is large and the current metadata device size is small AND the available space is small, this condition will be true until the upper limit for the data device is driven to be so small that the half value is just a single data block and then the thin_metadata_size error occurs.

```
else if datablocks_to_sectors(half_data_ext) + half_meta_ext * 2u64 > available_space
```

jbaublitz:
That was the conclusion I came to as well. How do you want to resolve that? One idea I had was now that the MDV is much larger so the maximum number of filesystems is too, should we perhaps use the current filesystem limit as the input to the metadata size calculation? My thought is that we can check the total sizes of all filesystems to see if additional metadata can be allocated in no overprovisioning mode as opposed to allocating room for it upfront.

jbaublitz:
Replaced by #3096

PR #3096, opened 2022-08-22
Issue stratisd 3086 v2

### LINKED ISSUE #3086: assertion failed: datablocks_to_sectors(upper_data_aligned) + upper_limit_meta_size * 2u64 > total_space

jbaublitz:
Closes #3086 
Alternative to #3088

### DIFF
--- a/src/engine/strat_engine/thinpool/thinpool.rs
+++ b/src/engine/strat_engine/thinpool/thinpool.rs
@@ -234,17 +234,27 @@ fn search(
     total_space: Sectors,
     upper_limit: DataBlocks,
     lower_limit: DataBlocks,
+    current_meta_size: Sectors,
     fs_limit: u64,
 ) -> StratisResult<(Sectors, Sectors)> {
     let diff = upper_limit - lower_limit;
     let half = lower_limit + diff / 2u64 + diff % 2u64;
-    let half_meta = thin_metadata_size(DATA_BLOCK_SIZE, datablocks_to_sectors(half), fs_limit)?;
+    let half_meta = max(
+        thin_metadata_size(DATA_BLOCK_SIZE, datablocks_to_sectors(half), fs_limit)?,
+        current_meta_size,
+    );
 
     if diff == DataBlocks(1) {
         let upper_limit_sectors = datablocks_to_sectors(upper_limit);
         let lower_limit_sectors = datablocks_to_sectors(lower_limit);
-        let upper_meta = thin_metadata_size(DATA_BLOCK_SIZE, upper_limit_sectors, fs_limit)?;
-        let lower_meta = thin_metadata_size(DATA_BLOCK_SIZE, lower_limit_sectors, fs_limit)?;
+        let upper_meta = max(
+            thin_metadata_size(DATA_BLOCK_SIZE, upper_limit_sectors, fs_limit)?,
+            current_meta_size,
+        );
+        let lower_meta = max(
+            thin_metadata_size(DATA_BLOCK_SIZE, lower_limit_sectors, fs_limit)?,
+            current_meta_size,
+        );
         trace!(
             "Upper data: {}, upper meta: {}, lower data: {}, lower meta: {}, total space: {}",
             upper_limit_sectors,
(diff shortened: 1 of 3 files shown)


src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -257,9 +267,9 @@ fn search(
         assert!(lower_limit_sectors + lower_meta * 2u64 <= total_space);
         Ok((lower_limit_sectors, lower_meta))
     } else if datablocks_to_sectors(half) + half_meta * 2u64 > total_space {
```
mulkieran, line 269: suggest extra condition here to allow for the unlikely possibility that ```current_data_size``` > than optimum data size.

```
if datablocks_to_sectors(half) + half_meta * 2u64 > total_space && half > current_data_size { ...
```

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -234,17 +234,27 @@ fn search(
...
 ) -> StratisResult<(Sectors, Sectors)> {
     let diff = upper_limit - lower_limit;
     let half = lower_limit + diff / 2u64 + diff % 2u64;
-    let half_meta = thin_metadata_size(DATA_BLOCK_SIZE, datablocks_to_sectors(half), fs_limit)?;
+    let half_meta = max(
+        thin_metadata_size(DATA_BLOCK_SIZE, datablocks_to_sectors(half), fs_limit)?,
+        current_meta_size,
+    );
 
     if diff == DataBlocks(1) {
```
mulkieran, line 247: suggest additional distinct termination condition in case ```current_data_size``` > than optimum data size.

```
if half == current_data_size {
    Ok((current_data_size, (total - current_data_size) / 2))
}
```

mulkieran (changes requested):
I think the algorithm looks quite robust for the previously identified thin metadata problem. I propose a way of hardening it against the not yet observed thin data size too large for optimum problem as well.

src/engine/strat_engine/thinpool/thinpool.rs
mulkieran: > suggest additional distinct termination condition in case `current_data_size` > than optimum data size.
> 
> ```
> if half == current_data_size {
>     Ok((current_data_size, (total - current_data_size) / 2))
> }
> ```
Thinking you should return an error here, because metadata size can not be enough.

mulkieran: But also thinking that an error is wrong, because in the usual overprovisioned case, even if the metadata is not enough, then you still want to expand. Otoh, in the switch to no-overprovisioned calculation, I think you want to reject the result, because there isn't room enough for the metadata. So, maybe it would be best to enrich the return type somehow.

jbaublitz:
Okay I think this should resolve the issue!

mulkieran (changes requested):
Please look at use of ```min_datadev_size()``` in ```check_fs()```. I believe that the change as it stands will mean that the ```current_data_size``` will never be chosen in the max operation. Maybe ```search()``` should return ```Result<Option<(Sectors, Sectors)>>```.

mulkieran:
> Please look at use of `min_datadev_size()` in `check_fs()`. I believe that the change as it stands will mean that the `current_data_size` will never be chosen in the max operation. Maybe `search()` should return `Result<Option<(Sectors, Sectors)>>`.

And then maybe ```min_datadev_size()``` could use the info to simply return the max of the search result or the current data size.

mulkieran:
> > Please look at use of `min_datadev_size()` in `check_fs()`. I believe that the change as it stands will mean that the `current_data_size` will never be chosen in the max operation. Maybe `search()` should return `Result<Option<(Sectors, Sectors)>>`.
> 
> And then maybe `min_datadev_size()` could use the info to simply return the max of the search result or the current data size.

@jbaublitz: Also need to fix up dependencies for test failures.

mulkieran (approved):

refi64:
Just tested and confirmed this fixes my issue!

PR #3106, opened 2022-09-08
Always allow multiple crate versions clippy lint

mulkieran:


mulkieran:
Since this is purely a clippy change, merging once clippy tests have passed.

PR #3116, opened 2022-09-22
Refactor processing of devices intended for initialization by stratisd

mulkieran:
Parts extracted from https://github.com/stratis-storage/stratisd/pull/2948

### DIFF
--- a/src/engine/strat_engine/backstore/devices.rs
+++ b/src/engine/strat_engine/backstore/devices.rs
@@ -366,6 +366,17 @@ impl TryFrom<&[&Path]> for ProcessedPathInfos {
     }
 }
 
+/// A list of device paths that have been determined to be unowned, and thus
+/// can be initialized by stratisd.
+/// Invariants:
+/// * DeviceInfo devnode values are unique.
+/// * DeviceInfo devno values are unique.
+/// * DeviceInfo.size value meets the required Stratis minimum.
+#[derive(Debug)]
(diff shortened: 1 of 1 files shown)


src/engine/strat_engine/backstore/devices.rs
```diff
@@ -264,6 +264,54 @@ pub struct DeviceInfo {
...
+                    "devices ({}) appear to belong to Stratis pool with UUID {}",
+                    devs.iter()
+                        .map(|(_, info)| info.devnode.display().to_string())
+                        .collect::<Vec<_>>()
+                        .join(", "),
+                    pool_uuid
+                )
+            })
+            .collect::<Vec<_>>()
+            .join("; ");
```
jbaublitz, line 301: I think it may make sense to remove this here and invoke it on line 308.

For example:

```
let error_strings = ... .collect::<Vec<_>>();

...

let error_message = format!(..., error_strings.join("; "));
```

I don't think testing for an empty string is as clear as testing for an empty `Vec`.

jbaublitz (changes requested):
Other than one minor stylistic thing, this looks good!

src/engine/strat_engine/backstore/devices.rs
mulkieran: Done!

jbaublitz (approved):

Issue #2880, opened 2021-12-16
Prohibit certain changes in block size combinations that might occur in the devices that make up our thinpool


mulkieran:
Sub of https://github.com/stratis-storage/project/issues/372

PR #3119, opened 2022-09-26
Process device paths at the top level

### LINKED ISSUE #1968: Hoist use of devices::process_and_verify_devices out of blockdevmgr to backstore and then to StratPool

mulkieran:
Supercedes #2948
Closes #1968 
Related #2880 

* Defines or extends some structs and methods for obtaining and processing sets of device information in the backstore/devices.rs file.
* Uses these methods to check for an empty set of to-be-initialized devices in ```init_cache``` and ```create_pool``` at the top level. This ensure a correct evaluation of emptiness, requiring little reasoning about the intervening reading steps. Also checks for emptiness when adding devices, to return early and omit expensive processing if no devices are to be added.
* Retains the use of the ```*idempotent*``` methods for the strat engine. Unpacks some device sets to obtain paths to pass to the methods, but only those device sets that wouldn't already have caused an error. This decision is motivated by the fact that the ```idempotent*``` methods are shared by the sim engine and the strat engine.
* Removes the process_and_verify_devices() and the check_device_ids() method entirely. Substitutes use of structs and their related impls in tests throughout the code.

Advantages:
* Raises the inspection of devices up to a higher level, to the pool, or in the case of creating a pool, the engine level. This is the principal goal. The reasons for doing this are to allow stratisd to have a full understanding of the properties of the devices it has been given as an argument at the same time as it can inspect information about all the existing devices in _both_ tiers, in order to decide whether they should be considered compatible or not. Currently, when creating a pool, everything about the devices to be used for the pool are known at the time of creation. But when initializing a cache, only the info about the proposed cache devices is known. And when adding devices to either tier, only the information about that tier is known. It is not known, at all, whether the devices are cache or data devices, at the point where they are inspected. This last could be remedied by passing a Tier argument down to the blockdev manager methods, but the other issues could not be.
* When initializing or extending the cache, it is necessary to suspend the thinpool. By reading the data from the devices before initializing or extending the cache, the suspend/resume interval is made a bit shorter. Not that much, probably, compared to the time that it takes to initialize the devices, but still an improvement.
* We have a lot of code in the CLI to try to guard against and give detailed messages in case the user specifies devices that belong to another pool or the same pool but a different tier, when adding, creating, or initializing. We can't really remove this, and achieve all that precision in stratisd, because although the inspection of devices is hoisted higher, it would all have to be hoisted up into the engine to make the level of detail in error messages comparable to that in the CLI. We do, however, send out precise error messages for when a user specifies devices in one tier, but they belong to the opposite tier, which is an improvement over the previous situation.

NOTES:
* This is an improvement over the previous iteration, which had convoluted and not generally useful abstractions in some of the impls in devices.rs. It drops the NonEmptyUnownedDevices, which turned out to be unhelpful in enforcing non-emptiness requirements.
* It keeps calls to ```validate_paths()``` before other processing, because ```validate_paths()``` verifies that the paths are absolute and emptiness checks now occur after the paths have been processed and information has been read from their corresponding devices.

### DIFF
--- a/src/engine/strat_engine/backstore/backstore.rs
+++ b/src/engine/strat_engine/backstore/backstore.rs
@@ -669,12 +669,12 @@ impl Recordable<BackstoreSave> for Backstore {
 
 #[cfg(test)]
 mod tests {
-    use std::{collections::HashSet, fs::OpenOptions, path::Path};
+    use std::{fs::OpenOptions, path::Path};
 
     use devicemapper::{CacheDevStatus, DataBlocks, DmOptions, IEC};
 
     use crate::engine::strat_engine::{
-        backstore::devices::process_and_verify_devices,
+        backstore::devices::{ProcessedPathInfos, UnownedDevices},
         metadata::device_identifiers,
         tests::{loopbacked, real},
     };
@@ -708,6 +708,15 @@ mod tests {
         assert!(backstore.next <= backstore.size())
     }
 
+    fn get_devices(paths: &[&Path]) -> StratisResult<UnownedDevices> {
+        ProcessedPathInfos::try_from(paths)
+            .map(|ps| ps.unpack())
+            .map(|(sds, uds)| {
+                sds.error_on_not_empty().unwrap();
+                uds
+            })
+    }
+
     /// Test adding cachedevs to the backstore.
     /// When cachedevs are added, cache tier, etc. must exist.
     /// Nonetheless, because nothing is written or read, cache usage ought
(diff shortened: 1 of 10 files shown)


jbaublitz (approved):

mulkieran:
rebased

mulkieran:
rebased.

bmr-cymru (approved):

mulkieran:
rebased.

PR #3132, opened 2022-10-11
Remove clippy allows for some false positives

mulkieran:


mulkieran:
Ignore cancelled test.

PR #3139, opened 2022-10-17
Fix underestimation of crypt metadata size

jbaublitz:


jbaublitz:
@mulkieran Okay, I found the issue causing the CI failure. It was a unit problem when converting from `u64` from libcryptsetup to `Bytes`. Because `Bytes` is in devicemapper, we have to do the conversions manually in stratisd unless we factor the units component out into a Rust `justbytes`-type equivalent.

src/engine/strat_engine/backstore/crypt/consts.rs
```diff
@@ -22,12 +24,12 @@ pub const STRATIS_TOKEN_TYPE: &str = "stratis";
 /// each block device.
 pub const STRATIS_MEK_SIZE: usize = 512 / 8;
 
-/// Sector size as determined in `cryptsetup/lib/internal.h`
-pub const SECTOR_SIZE: u64 = 512;
+/// Sector size as defined in the LUKS2 specification documentation. 4k bytes.
+pub const SECTOR_SIZE: Sectors = Sectors(8);
```
bmr-cymru, line 28: What's the advantage of typing this as `Sectors`? It seems a little confusing to have a variable named `SECTOR_SIZE` with the value `Sectors(8)`?

When the value is used it appears to always be converted to `.bytes()` so would it perhaps be more readable to just make it a u64 value in units of bytes from the outset?

I also wonder if it might be better named something like `LUKS_SECTOR_SIZE` to make it clear that this is specific to LUKS and nothing to do with "device sectors"?

bmr-cymru (changes requested):

src/engine/strat_engine/backstore/crypt/consts.rs
bmr-cymru: @jbaublitz everything else makes complete sense, I was just a bit thrown by the name/type here.

bmr-cymru (approved):

mulkieran (approved):

PR #3145, opened 2022-10-25
Add workflow to generate module tree with cargo-modules

bmr-cymru:
Related https://github.com/stratis-storage/project/issues/537

Add a new workflow to generate visualization artefacts with cargo-modules.


mulkieran:
cargo-modules is nice and I should use it. But it seems so easy to use from the command-line, that we might as well stick to that, like ```cargo-tree```.

PR #3155, opened 2022-11-02
Prohibit certain block size combinations when creating a new pool or adding devices to an existing pool (prior)

mulkieran:
Related #2880

* Add in_use detection to individual block devices. (1)A block devices is considered in use if the space that is allocated from it exceeds that allocated for Stratis metadata.
* Add an AllocatedAbove struct to tidy up management of segments allocated to the data tier or cache tier from their block devices.
* Assert an invariant that blockdev is considered to be in use by criterion (1) exactly when it is considered in use because it is included in the AllocatedAbove data structure.
* Read BLKSSZGET and BLKPBSZGET ioctls along with other device information when examining devices.
* Calculate whether an existing pool should have reduced ActionAvailability. If the data cap device could be extended so that its logical block size could change, then ActionAvailability is set to no actions. There is no check on the cache tier; if the pool has a cache tier and it is working, then the situation has to be assumed to be correct.
* Require identical block sizes for devices being newly added to either the cache or the data tier.
* Do not allow initializing a cache where the logical block size of the resulting cache would differ from that of the data tier.
* Require the logical block size of the devices to be added to the cache or data tier to match the current logical block size.
* When creating a new pool, require block sizes to match.

### DIFF
--- a/src/engine/strat_engine/backstore/backstore.rs
+++ b/src/engine/strat_engine/backstore/backstore.rs
@@ -16,8 +16,7 @@ use crate::{
         strat_engine::{
             backstore::{
                 blockdev::StratBlockDev, blockdevmgr::BlockDevMgr, cache_tier::CacheTier,
-                data_tier::DataTier, devices::UnownedDevices, shared::map_to_dm,
-                transaction::RequestTransaction,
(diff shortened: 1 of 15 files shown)


mulkieran:
rebased.

PR #3157, opened 2022-11-04
Fix clippy 1.65 lint errors

bgurney-rh:


bgurney-rh:
Force-pushed to fix a formatting error in the commit `Remove binary comparison to literal 'Option::None'`

PR #3166, opened 2022-11-10
Clean up for AllOrSomeLock

jbaublitz:


src/engine/structures/lock.rs
```diff
@@ -767,7 +797,14 @@ where
 }
 
 /// Future returned by AllOrSomeLock::real_all().
-struct AllRead<U, T>(AllOrSomeLock<U, T>, AtomicBool, u64);
+struct AllRead<U: AsUuid, T>(AllOrSomeLock<U, T>, AtomicBool, u64);
+
+impl<U, T> Unpin for AllRead<U, T>
```
mulkieran, line 802: I _think_ this might be unnecessary.

mulkieran (changes requested):
I think it might be useful to add a comment explaining why the UnsafeCell is necessary. Also, please see the one review comment.

src/engine/structures/lock.rs
jbaublitz: It is! Thanks for catching that!

mulkieran (approved):

PR #3167, opened 2022-11-15
Use timestamp_opt instead of soon to be deprecated timestamp method

mulkieran:
Related https://github.com/stratis-storage/project/issues/538

Use timestamp_opt() instead of timestamp() to construct new timestamps from component parts. timestamp() is deprecated in chrono 0.4.23 which we expect to begin depending on when we make our upcoming release. timestamp() just panicked when it was unable to construct a timestamp from the arguments. timestamp_opt() returns a non-standard error result in a principled way; that error result is now handled by the calling code. Also, store the initialization time as a DateTime in the StaticHeader struct.

### DIFF
--- a/src/engine/shared.rs
+++ b/src/engine/shared.rs
@@ -10,6 +10,7 @@ use std::{
     path::{Path, PathBuf},
 };
 
+use chrono::{DateTime, LocalResult, TimeZone, Utc};
 use nix::poll::{poll, PollFd, PollFlags};
 use regex::Regex;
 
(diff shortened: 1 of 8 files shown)


mulkieran:
checks_with_udev test failures might be real.

src/engine/shared.rs
```diff
@@ -292,6 +293,32 @@ pub fn total_allocated(allocated: &Diff<Bytes>, metadata_size: &Diff<Bytes>) ->
     }
 }
 
+/// Convert a u64 value representing seconds, and an optional u32 value
+/// representing nanoseconds to a timestamp.
+pub fn unsigned_to_timestamp(secs: u64, nanos: Option<u32>) -> StratisResult<DateTime<Utc>> {
```
jbaublitz, line 298: I'm usually not a big supported of multiple values evaluating to the same input when you can help it. For example, `0` and `None` are identical from an input perspective, so I'm not seeing what the value of the `Option` input is here.

jbaublitz (changes requested):
Other than one stylistic comment, this looks good!

src/engine/shared.rs
mulkieran: Simplified!

jbaublitz (approved):

mulkieran:
rebased.

Issue #2989, opened 2022-05-03
Drop udevadm settle calls in source


mulkieran:
There are two calls, both to be executed before cleanup after some failure.

1. ```StratFilesystem::initialize()``` if ```create_fs``` (for creating xfs filesystem fails).
2. ```ThinPool::create_filesystem``` if saving the metadata to the mdv fails.

Can we just use the retry crate for both?

mulkieran:
@jbaublitz, @bmr-cymru Do you see a compelling reason to use udevadm settle on these error paths instead of just retrying the cleanup using the facilities in the retry crate?

jbaublitz:
@mulkieran Based on my experience with udev, I'm wondering if retries are really necessary for the MDV save case. Just writing to a filesystem shouldn't trigger udev processing but @bmr-cymru correct me if I'm wrong. As for failing to create the filesystem, that would trigger a udev event for sure but I think waiting until the whole udev queue is cleared is a bit overboard. We have similar cases where we use the retry crate elsewhere so this seems like it would work here too.

bmr-cymru:
@mulkieran I don't see a strong reason for waiting on udev here and I think there are other (rare/unlikely) cases where we might legitimately need to retry the cleanup.

@jbaublitz agreed - a close of a writeable file descriptor on the block device that contains the file system will trigger udev events, but ordinary IO to files in a mounted file system will not.

I need to look at the retry crate and understand how it works better (I just had a quick look at the umount retry in `StratFilesystem::snapshot()`), but I don't see any reason to require the full udev settle, and a retry has some minor benefit.

Creating the file system and mounting the MDV both require the thin device node but cleaning up the DM device doesn't - it just requires ioctl access to `/dev/mapper/control` (which is already open in any case I think, looking at devicemapper-rs?), and for the device to be in a state where removal can succeed (i.e. not mounted or open). Changing these to retries seems like it would cover cases that the udev settle doesn't - for example another process having the device open and blocking the remove.

PR #3173, opened 2022-11-18
do not rely on udevadm cli

leseb:
Related #2989

### DIFF
--- a/src/engine/strat_engine/cmd.rs
+++ b/src/engine/strat_engine/cmd.rs
@@ -63,8 +63,6 @@ fn find_executable(name: &str) -> Option<PathBuf> {
 const MKFS_XFS: &str = "mkfs.xfs";
 const THIN_CHECK: &str = "thin_check";
 const THIN_REPAIR: &str = "thin_repair";
-#[cfg(test)]
-const UDEVADM: &str = "udevadm";
 const THIN_METADATA_SIZE: &str = "thin_metadata_size";
 const XFS_DB: &str = "xfs_db";
 const XFS_GROWFS: &str = "xfs_growfs";
@@ -120,8 +118,6 @@ lazy_static! {
         (MKFS_XFS.to_string(), find_executable(MKFS_XFS)),
         (THIN_CHECK.to_string(), find_executable(THIN_CHECK)),
         (THIN_REPAIR.to_string(), find_executable(THIN_REPAIR)),
-        #[cfg(test)]
-        (UDEVADM.to_string(), find_executable(UDEVADM)),
         (XFS_DB.to_string(), find_executable(XFS_DB)),
         (XFS_GROWFS.to_string(), find_executable(XFS_GROWFS)),
         (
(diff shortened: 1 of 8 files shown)


mulkieran:
@leseb Thanks for the PR! We have two targets in our Makefile, "fmt" and "clippy" and we enforce their rules in our CI. Running those in your development environment will save you the trouble of unnecessary round trips w/ our CI.

src/engine/strat_engine/udev.rs
```diff
@@ -157,10 +157,60 @@ where
...
+
+    // If the file exists then the queue is not empty and we return an error
+    if udev_queue_file_path.exists() {
+        Err(StratisError::Msg(format!(
+            "Udev queue file {} still exists",
+            UDEV_QUEUE_FILE_PATH
+        )))
+    } else {
+        Ok(true)
+    }
```
jbaublitz, line 184: Mainly stylistic, but I'd think this return type should either be `StratisResult<()>` or `bool`. I don't see any case where `Ok(false)` is returned.

jbaublitz (commented):
I'm unsure whether this conflicts with the udev sync work, but if that PR is still using udevadm and CI doesn't fail for this PR, this may be okay to merge prior to that.

src/engine/strat_engine/cmd.rs
```diff
@@ -14,6 +14,9 @@
 // the existence of the file is checked before the command is invoked, and
 // an explicit error is returned if the executable can not be found.
 
+use libc::c_uint;
+use libcryptsetup_rs::SafeMemHandle;
+use serde_json::Value;
```
jbaublitz, line 19: We generally want the imports to be separated into the following categories:

`std`

external crates (like libc, serde_json)

crates that are external that we develop (like devicemapper)

`crate`

all separated by a newline so they're ordered appropriately within those categories.

src/engine/strat_engine/udev.rs
```diff
@@ -157,10 +157,60 @@ where
...
         .scan_devices()?
         .filter(|dev| dev.is_initialized())
         .find(|x| x.devnode().map_or(false, |d| **device_path == *d))
         .map(|ref d| f(&UdevEngineDevice::from(d))))
 }
+
+/// Return true if udev queue is empty and no events are being processed.
+#[cfg(test)]
+pub fn is_udev_queue_empty() -> StratisResult<()> {
+    // Path of udev queue's file, typically this file is empty, it just asses that udev events are still
```
jbaublitz, line 170: This should be "it just assesses"

jbaublitz (changes requested):
After these are handled, I think we can accept this PR.

jbaublitz:
Oops, missed that CI is failing. We'll have to investigate that first too @leseb.

jbaublitz:
@leseb Looks like we've fixed all of the failures by doing that except for `check_with_tang` which appears to not have been modified. If you apply the same changes there, I have a feeling that we should be good to go with CI.

mulkieran:
I think we're trying to fix up some CI problems, see #3211

mulkieran:
The CI problems seem to be caused by the non-cli implementation of settle starting and finishing too fast. We aren't sure, but it might be that the udev synchronization work would resolve this problem.

leseb:
> The CI problems seem to be caused by the non-cli implementation of settle starting and finishing too fast. We aren't sure, but it might be that the udev synchronization work would resolve this problem.

Agreed, hence the retry in this PR. This might be environment specific since the runners are small VMs, something we might not encounter on larger machines. Also, AFAIC this is only executed during the unit tests so retry is not uncommon in CI :).

leseb:
Closing this PR since it will soon be superseded by another implementation (udev synchronization improvements). We may re-open in the future if necessary. Thanks @mulkieran @jbaublitz for your support on this.

PR #3174, opened 2022-11-18
Prohibit certain block size combinations when creating a new pool or adding devices to an existing pool 

mulkieran:
Related #2880

* Add in_use detection to individual block devices. (1)A block devices is considered in use if the space that is allocated from it exceeds that allocated for Stratis metadata.
* Add an AllocatedAbove struct to tidy up management of segments allocated to the data tier or cache tier from their block devices.
* Assert an invariant that blockdev is considered to be in use by criterion (1) exactly when it is considered in use because it is included in the AllocatedAbove data structure.
* Read BLKSSZGET and BLKPBSZGET ioctls along with other device information when examining devices.
* Calculate whether an existing pool should have reduced ActionAvailability. If the data cap device could be extended so that its logical or physical block size could change, then ActionAvailability is set to no actions. There is no check on the cache tier; if the pool has a cache tier and it is working, then the situation has to be assumed to be correct.
* Require identical block sizes for devices being newly added to either the cache or the data tier.
* Do not allow initializing a cache where the logical block size of the resulting cache would differ from that of the data tier.
* Require the block sizes of the devices to be added to the cache or data tier to match the current block sizes.
* When creating a new pool, require block sizes to match.

### DIFF
--- a/src/engine/strat_engine/backstore/backstore.rs
+++ b/src/engine/strat_engine/backstore/backstore.rs
@@ -16,7 +16,8 @@ use crate::{
         strat_engine::{
             backstore::{
                 blockdev::StratBlockDev, blockdevmgr::BlockDevMgr, cache_tier::CacheTier,
-                data_tier::DataTier, devices::UnownedDevices, transaction::RequestTransaction,
+                data_tier::DataTier, devices::UnownedDevices, shared::BlockSizeSummary,
+                transaction::RequestTransaction,
             },
             dm::get_dm,
             metadata::{MDADataSize, BDA},
@@ -27,8 +28,8 @@ use crate::{
             writing::wipe_sectors,
         },
         types::{
-            BlockDevTier, DevUuid, EncryptionInfo, KeyDescription, Name, PoolEncryptionInfo,
-            PoolUuid,
+            ActionAvailability, BlockDevTier, DevUuid, EncryptionInfo, KeyDescription, Name,
+            PoolEncryptionInfo, PoolUuid,
         },
     },
     stratis::{StratisError, StratisResult},
@@ -711,6 +712,35 @@ impl Backstore {
 
     pub fn rename_pool(&mut self, new_name: &Name) -> StratisResult<()> {
         self.data_tier.block_mgr.rename_pool(new_name)
+    }
+
+    /// A summary of block sizes
+    pub fn block_size_summary(&self, tier: BlockDevTier) -> Option<BlockSizeSummary> {
+        match tier {
+            BlockDevTier::Data => Some(self.data_tier.partition_by_use().into()),
+            BlockDevTier::Cache => self
+                .cache_tier
+                .as_ref()
+                .map(|ct| ct.partition_cache_by_use().into()),
+        }
+    }
+
+    /// What the pool's action availability should be
+    pub fn action_availability(&self) -> ActionAvailability {
+        let data_tier_bs_summary = self
+            .block_size_summary(BlockDevTier::Data)
+            .expect("always exists");
+        let cache_tier_bs_summary: Option<BlockSizeSummary> =
+            self.block_size_summary(BlockDevTier::Cache);
+        if let Err(err) = data_tier_bs_summary.validate() {
+            warn!("Disabling pool changes for this pool: {}", err);
+            ActionAvailability::NoPoolChanges
+        } else if let Some(Err(err)) = cache_tier_bs_summary.map(|ct| ct.validate()) {
+            warn!("Disabling pool changes for this pool: {}", err);
+            ActionAvailability::NoPoolChanges
+        } else {
+            ActionAvailability::Full
+        }
     }
 }
 
--- a/src/engine/strat_engine/backstore/backstore.rs
+++ b/src/engine/strat_engine/backstore/backstore.rs
@@ -15,11 +15,8 @@ use crate::{
     engine::{
         strat_engine::{
             backstore::{
-                blockdev::StratBlockDev,
-                blockdevmgr::{map_to_dm, BlockDevMgr},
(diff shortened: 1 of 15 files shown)


src/engine/strat_engine/device.rs
```diff
@@ -22,3 +34,29 @@ pub fn blkdev_size(file: &File) -> StratisResult<Bytes> {
...
+    let mut val = 0i32 as c_int; // util-linux uses int* as out-arg for ioctl
+    unsafe { blksszget(file.as_raw_fd(), &mut val) }.map_err(|e| {
+        StratisError::Msg(format!(
+            "Error reading logical sector size (BLKSSZGET): {}",
+            e
+        ))
+    })?;
+    // Allowed because the size should be less than u16::MAX
+    #[allow(clippy::cast_possible_truncation)]
+    Ok(Bytes::from(val as u16))
```
jbaublitz, line 48: Why are you using `as` when we have the macros for integer conversion and this method returns a `StratisResult` anyway?

src/engine/strat_engine/device.rs
```diff
@@ -22,3 +34,29 @@ pub fn blkdev_size(file: &File) -> StratisResult<Bytes> {
...
+    let mut val = 0i32 as c_int; // util-linux uses int* as out-arg for ioctl
+    unsafe { blkpbszget(file.as_raw_fd(), &mut val) }.map_err(|e| {
+        StratisError::Msg(format!(
+            "Error reading physical sector size (BLKPBSZGET): {}",
+            e
+        ))
+    })?;
+    // Allowed because the size should be less than u16::MAX
+    #[allow(clippy::cast_possible_truncation)]
+    Ok(Bytes::from(val as u16))
```
jbaublitz, line 61: I think the integer conversion macro could be used here too.

src/engine/strat_engine/engine.rs
```diff
@@ -418,6 +418,12 @@ impl Engine for StratEngine {
                 ));
             }
 
+            let block_size_summary = unowned_devices.blocksizes();
+            if block_size_summary.len() > 1 {
+                let err_str = "The devices specified for initializing the pool do not have uniform physical and logical sector sizes.".into();
+                return Err(StratisError::Msg(err_str));
+            }
```
jbaublitz, line 425: Maybe this is better to discuss in stand up, but what is the danger of having a different logical and physical block sizes? Is this the torn write problem?

src/engine/strat_engine/pool.rs
```diff
@@ -219,16 +234,28 @@ impl StratPool {
...
         if let Err(e) = check_metadata(metadata) {
             return Err((e, tiers_to_bdas(datadevs, cachedevs, None)));
         }
 
         let mut backstore =
             Backstore::setup(uuid, &metadata.backstore, datadevs, cachedevs, timestamp)?;
+        let action_avail = max(
+            get_pool_state(encryption_info),
+            backstore.action_availability(),
+        );
```
jbaublitz, line 248: I think our original intention with `get_pool_state()` was to have a single place where the action availability was determined. We may want to pass a reference to the backstore into `get_pool_state()` to unify the logic of determining the action availability.

src/engine/strat_engine/backstore/backstore.rs
```diff
@@ -716,6 +713,35 @@ impl Backstore {
...
+            .expect("always exists");
+        let cache_tier_bs_summary: Option<BlockSizeSummary> =
+            self.block_size_summary(BlockDevTier::Cache);
+        if let Err(err) = data_tier_bs_summary.validate() {
+            warn!("Disabling pool changes for this pool: {}", err);
+            ActionAvailability::NoPoolChanges
+        } else if let Some(Err(err)) = cache_tier_bs_summary.map(|ct| ct.validate()) {
+            warn!("Disabling pool changes for this pool: {}", err);
+            ActionAvailability::NoPoolChanges
+        } else {
```
jbaublitz, line 741: Is this more on principal or am I missing a protection that this provides? My understanding is that the cache always gets greedily allocated so it seems that putting this in `NoPoolChanges` mode for the cache may come a little bit too late as the device would have already been extended to use varying sized blocks. If the cache were lazily allocated, I'd understand this a bit more.

jbaublitz (changes requested):
Some changes requested and some questions for clarification.

src/engine/strat_engine/backstore/backstore.rs
mulkieran: I think that this condition will not happen, exactly for the reason you specify. But if it does, then NoPoolChanges would be correct.

src/engine/strat_engine/engine.rs
mulkieran: Perhaps I should reword that? It is totally ok for the physical sizes to differ from the logical sizes, and we would always allow that. How about: "Some of the devices specialized for initializing the pool have different logical sectors sizes or have different physical sector sizes." ?

src/engine/strat_engine/device.rs
mulkieran: No good reason!

src/engine/strat_engine/device.rs
mulkieran: Done.

src/engine/strat_engine/pool.rs
mulkieran: Let me think about that one...

src/engine/strat_engine/backstore/backstore.rs
mulkieran: I've added a note.

jbaublitz (approved):

mulkieran:
Superceded.

PR #3182, opened 2022-11-23
Require clippy for task that uses clippy in toolchain action

mulkieran:


mulkieran:
tested, merging immediately.

Issue #3185, opened 2022-11-24
stratisd does not run in container if the host PID namespace is not shared


leseb:
When the container runs with its own PID namespace stratisd crashes like so:

```
[root@localhost ~]# podman run --user root --privileged -ti -v /run/dbus/system_bus_socket:/run/dbus/system_bus_socket:ro -v /dev:/dev -v /run/udev:/run/udev --ipc=host -v /usr/share/dbus-1/system.d/:/usr/share/dbus-1/system.d/ -v /usr/lib/udev/rules.d/:/usr/lib/udev/rules.d/ stratis /usr/libexec/stratisd
thread 'main' panicked at 'assertion failed: !is_in_root_namespace()?', src/engine/strat_engine/ns.rs:30:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

When applying `--pid=host`, the container runs fine:

```
[root@localhost ~]# podman run --user root --privileged -ti -v /run/dbus/system_bus_socket:/run/dbus/system_bus_socket:ro -v /dev:/dev -v /run/udev:/run/udev --ipc=host -v /usr/share/dbus-1/system.d/:/usr/share/dbus-1/system.d/ -v /usr/lib/udev/rules.d/:/usr/lib/udev/rules.d/ --pid=host stratis /usr/libexec/stratisd --log-level debug
[2022-11-24T15:12:30Z INFO  stratisd::stratis::run] stratis daemon version 3.2.3 started
[2022-11-24T15:12:30Z INFO  stratisd::stratis::run] Using StratEngine
...
...
```

PR #3186, opened 2022-11-24
do not unshare mount namespace if stratisd pid is 1

leseb:
If stratis runs in a container as PID 1, unsharing the mount namespace will not work and we should work the user about that. Also stratisd crashes while asserting that the mount namespace has been unshared.

Related: https://github.com/stratis-storage/stratisd/issues/3185

Signed-off-by: Sébastien Han <seb@redhat.com>

### DIFF
--- a/src/engine/strat_engine/ns.rs
+++ b/src/engine/strat_engine/ns.rs
@@ -24,17 +24,17 @@ pub const NS_TMPFS_LOCATION: &str = "/run/stratisd/ns_mounts";
 
 /// Unshare the mount namespace if the stratisd mount namespace for this thread is in the root namespace.
 // Precondition: If running in a container, PID is not 1 OR the container does share the host PID.
-pub fn unshare_namespace() -> StratisResult<()> {
+pub fn unshare_mount_namespace() -> StratisResult<()> {
     // Only create a new mount namespace if the thread is in the root namespace.
-    if is_in_root_namespace()? {
+    if is_in_root_mount_namespace()? {
         unshare(CloneFlags::CLONE_NEWNS)?;
     }
-    assert!(!is_in_root_namespace()?);
+    assert!(!is_in_root_mount_namespace()?);
     Ok(())
 }
 
 /// Check if the stratisd mount namespace for this thread is in the root namespace.
-pub fn is_in_root_namespace() -> StratisResult<bool> {
+pub fn is_in_root_mount_namespace() -> StratisResult<bool> {
     let pid_one_stat = stat(INIT_MNT_NS_PATH)?;
     let self_stat = stat(format!("/proc/self/task/{}/ns/mnt", gettid()).as_str())?;
     Ok(pid_one_stat.st_ino == self_stat.st_ino && pid_one_stat.st_dev == self_stat.st_dev)
--- a/src/engine/strat_engine/ns.rs
+++ b/src/engine/strat_engine/ns.rs
@@ -22,6 +22,8 @@ const INIT_MNT_NS_PATH: &str = "/proc/1/ns/mnt";
 /// Path to where private namespace mounts are mounted
 pub const NS_TMPFS_LOCATION: &str = "/run/stratisd/ns_mounts";
 
+/// Unshare the mount namespace if the stratisd mount namespace for this thread is in the root namespace.
+// Precondition: If running in a container, PID is not 1 OR the container does share the host PID.
 pub fn unshare_namespace() -> StratisResult<()> {
     // Only create a new mount namespace if the thread is in the root namespace.
     if is_in_root_namespace()? {
--- a/src/stratis/run.rs
+++ b/src/stratis/run.rs
@@ -16,6 +16,8 @@ use tokio::{
     task,
 };
 
+use nix::unistd::getpid;
+
 #[cfg(feature = "dbus_enabled")]
 use crate::dbus_api::DbusAction;
 use crate::{
@@ -48,7 +50,20 @@ async fn signal_thread() -> StratisResult<()> {
 /// Always check for devicemapper context.
 pub fn run(sim: bool) -> StratisResult<()> {
     if !sim {
-        unshare_namespace()?;
+        // If stratisd is running as PID 1, then it is definitely in a container and the
+        // unshare_namespace() command method will fail silently, since the mount namespace will
+        // remain the same after the unshare command. Also, containers give us a native mount
+        // isolation so unsharing the mount namespace is unnecessary. So any filesystem mounts
+        // inside a container won't be visible by the host.
+        //
+        // Note that it is possible for stratisd to be running in a container and not as PID 1, but
+        // this situation is less easily detected. However, stratisd will not fail in the
+        // unshare_namespace() method if that is the case.
+        if getpid().as_raw() != 1 {
+            unshare_namespace()?;
+        } else {
+            debug!("Running as PID 1, not unsharing mount namespace");
+        }
     }
 
     set_up_crypt_logging();
--- a/src/stratis/run.rs
+++ b/src/stratis/run.rs
@@ -22,7 +22,8 @@ use nix::unistd::getpid;
 use crate::dbus_api::DbusAction;
 use crate::{
     engine::{
-        set_up_crypt_logging, unshare_namespace, Engine, SimEngine, StratEngine, UdevEngineEvent,
+        set_up_crypt_logging, unshare_mount_namespace, Engine, SimEngine, StratEngine,
+        UdevEngineEvent,
     },
     stratis::{
         dm::dm_event_thread, errors::StratisResult, ipc_support::setup, stratis::VERSION,
@@ -51,16 +52,16 @@ async fn signal_thread() -> StratisResult<()> {
 pub fn run(sim: bool) -> StratisResult<()> {
     if !sim {
         // If stratisd is running as PID 1, then it is definitely in a container and the
-        // unshare_namespace() command method will fail silently, since the mount namespace will
+        // unshare_mount_namespace() command method will fail silently, since the mount namespace will
         // remain the same after the unshare command. Also, containers give us a native mount
         // isolation so unsharing the mount namespace is unnecessary. So any filesystem mounts
         // inside a container won't be visible by the host.
         //
         // Note that it is possible for stratisd to be running in a container and not as PID 1, but
         // this situation is less easily detected. However, stratisd will not fail in the
-        // unshare_namespace() method if that is the case.
+        // unshare_mount_namespace() method if that is the case.
         if getpid().as_raw() != 1 {
-            unshare_namespace()?;
+            unshare_mount_namespace()?;
         } else {
             debug!("Running as PID 1, not unsharing mount namespace");
         }
--- a/src/engine/mod.rs
+++ b/src/engine/mod.rs
@@ -7,7 +7,7 @@ pub use self::{
     shared::{total_allocated, total_used},
     sim_engine::SimEngine,
     strat_engine::{
-        crypt_metadata_size, get_dm, get_dm_init, set_up_crypt_logging, unshare_namespace,
+        crypt_metadata_size, get_dm, get_dm_init, set_up_crypt_logging, unshare_mount_namespace,
         StaticHeader, StaticHeaderResult, StratEngine, StratKeyActions, ThinPoolSizeParams, BDA,
         CLEVIS_TANG_TRUST_URL,
     },
--- a/src/engine/strat_engine/backstore/blockdevmgr.rs
+++ b/src/engine/strat_engine/backstore/blockdevmgr.rs
@@ -759,7 +759,7 @@ mod tests {
         backstore::devices::{ProcessedPathInfos, UnownedDevices},
         cmd,
         names::KeyDescription,
-        ns::{unshare_namespace, MemoryFilesystem},
+        ns::{unshare_mount_namespace, MemoryFilesystem},
         tests::{crypt, loopbacked, real},
     };
 
@@ -986,7 +986,7 @@ mod tests {
     }
 
     fn test_clevis_initialize(paths: &[&Path]) {
-        unshare_namespace().unwrap();
+        unshare_mount_namespace().unwrap();
         let pool_name = Name::new("pool_name".to_string());
         let _memfs = MemoryFilesystem::new().unwrap();
         let pool_uuid = PoolUuid::new_v4();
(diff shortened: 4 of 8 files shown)


src/engine/strat_engine/ns.rs
```diff
@@ -27,6 +27,8 @@ pub fn unshare_namespace() -> StratisResult<()> {
     if is_in_root_namespace()? {
         unshare(CloneFlags::CLONE_NEWNS)?;
     }
+    // unshare() seems to be working fine even if called on a PID 1 process, so the assertion below is
+    // important to trigger.
```
jbaublitz, line 31: My understanding from our meeting is that this assertion is actually triggered if in a container. Should the below assertion be removed?

jbaublitz (commented):
Just one clarification question on this.

src/engine/strat_engine/ns.rs
leseb: It depends on how we want to tackle this. I think the assertion is important for the non-container case. This assertion fails when the container does not share the host PID and runs as PID 1 because unsharing PID 1 does not seem to work.
It didn't want to change the existing behavior and simply warn the user. If we want to change this I would move `unshare_namespace()?;` under this condition https://github.com/stratis-storage/stratisd/pull/3186/files#diff-3604c519d3539d03673ac6805e179a57af023dbee55fee25322beb1be846798dR64 to avoid the failure entirely.

jbaublitz: I'm open to either. From my point of view, I'd almost rather exit with that error message that warn and then trigger an assertion immediately after. If we can cleanly exit, that would be preferable to me. Thoughts, @mulkieran?

mulkieran: @jbaublitz: I would favor the clean exit, also.

leseb: Thanks both, made the requested changes.

jbaublitz:
@leseb I see you opted to not take the clean exit approach. Was there something that came up that leads you to believe not unsharing if stratisd is PID 1 is better?

I just want outline what I see as our options so that we can all discuss which one would be best.

1. We can require `--pid=host` and exit cleanly if we detect that it hasn't been used.

### Pros
Allows us to leave our mount namespace check unchanged.

### Cons
Causes permission denied in rootless podman. Requires us to run container as root.

2. We can guard unsharing the mount namespace if the PID is detected to be 1. (This is the solution it appears you implemented)

### Pros
Allows us to use rootless podman.

### Cons
Leaves keyfile directory (albeit protected by root read permissions) visible in the container. Special care would be needed to protect the root user from arbitrary access to avoid passphrases being readable.

3. We can unshare the mount namespace and only trigger the assertion if the PID is detected to be 1.

### Pros
May not have any pros. This depends a bit on how containers work. If we unshare the namespace for PID 1 and all future processes started in the container inherit that unshared namespace of PID 1, I see no benefit as any bash shell spawned in the container would be in the same namespace as stratisd, rendering the unshare effectively useless for the purposes of keeping the mount out of view. However, if future processes do not inherit the unshared namespace, this might be an improvement.

### Cons
Same as 2. Perhaps adds an additional layer of indirection, but if a user can get root access, they can easily join the namespace. If the pros above in 3 are invalid, this will be exactly the same cons as 2, and this solution is not viable.

@mheon Tagging you here too for feedback, particularly for the last point.

mheon:
I think that using unshare as security, if it's being done by PID 1 in the container, will not be effective.  Our definitive source of truth for container namespaces is the primary container PID - so if you do an unshare of the mount namespace with that PID, all subsequent `podman exec` calls will join the new namespace, which I suspect undoes much of the security advantage you would otherwise see.

jbaublitz:
Okay! We can disregard option 3. As long as option 2 seems better to you, @leseb, I'll approve it and we can merge this when @mulkieran is ready!

leseb:
@jbaublitz The rootless Podman is not possible due to another issue, it seems that is_in_root_namespace() fails to stat /proc/1/ns/mnt. And using --pid=host won't solve that problem.

Moving away from the rootless Podman problem, I decided not to unshare the mount ns when running as PID 1 because the unshare does not work, so the assertion will always fail. I wrote a simple script to verify this, see:

```
[root@localhost ~]# podman run --privileged -v /run/udev:/run/udev -e RUST_LOG=info  stratis /run/udev/pid
[2022-12-02T09:51:01Z INFO  pid] main pid: 1
[2022-12-02T09:51:01Z INFO  pid] ppid: 0
[2022-12-02T09:51:01Z INFO  pid] tid: 1
[2022-12-02T09:51:01Z INFO  pid] pid_one_stat.st_ino: 4026532760
[2022-12-02T09:51:01Z INFO  pid] self_stat.st_ino: 4026532760
[2022-12-02T09:51:01Z INFO  pid] pid_one_stat.st_dev: 4
[2022-12-02T09:51:01Z INFO  pid] self_stat.st_dev: 4
[2022-12-02T09:51:01Z INFO  pid] unshare
[2022-12-02T09:51:01Z INFO  pid] pid_one_stat.st_ino: 4026532765
[2022-12-02T09:51:01Z INFO  pid] self_stat.st_ino: 4026532765
[2022-12-02T09:51:01Z INFO  pid] pid_one_stat.st_dev: 4
[2022-12-02T09:51:01Z INFO  pid] self_stat.st_dev: 4
thread 'main' panicked at 'assertion failed: !is_in_root_namespace()?', src/main.rs:57:9
```

We can see that `st_ino` and `st_dev` are already identical, even after unsharing. This assumes that the test is correct though (`Ok(pid_one_stat.st_ino == self_stat.st_ino && pid_one_stat.st_dev == self_stat.st_dev)`) :).

Success case:

```
[root@localhost ~]# podman run --privileged -v /run/udev:/run/udev -e RUST_LOG=info --pid=host stratis /run/udev/pid
[2022-12-02T09:51:05Z INFO  pid] main pid: 106440
[2022-12-02T09:51:05Z INFO  pid] ppid: 106438
[2022-12-02T09:51:05Z INFO  pid] tid: 106440
[2022-12-02T09:51:05Z INFO  pid] pid_one_stat.st_ino: 4026531841
[2022-12-02T09:51:05Z INFO  pid] self_stat.st_ino: 4026532760
[2022-12-02T09:51:05Z INFO  pid] pid_one_stat.st_dev: 4
[2022-12-02T09:51:05Z INFO  pid] self_stat.st_dev: 4
[2022-12-02T09:51:05Z INFO  pid] success
```

Here the inode number is different so we are not in the root mount namespace.
The test script I used can be found here https://gist.github.com/leseb/cbc975e07ddb3851a95f0cdd692c50b2.

jbaublitz:
> The rootless Podman is not possible due to another issue, it seems that is_in_root_namespace() fails to stat /proc/1/ns/mnt. And using --pid=host won't solve that problem.

I think it's actually the other way around based on what @mheon told me. That using `--pid=host` is what causes us to fail in rootless podman because `/proc/1/ns/mnt` will not be in a namespace that is a child of the name space used when using rootless podman. That's basically what I was trying to communicate above. If we use option 2 without `--pid=host`, we may be able to use rootless podman. Have you tried that?

leseb:
> > The rootless Podman is not possible due to another issue, it seems that is_in_root_namespace() fails to stat /proc/1/ns/mnt. And using --pid=host won't solve that problem.
> 
> I think it's actually the other way around based on what @mheon told me. That using `--pid=host` is what causes us to fail in rootless podman because `/proc/1/ns/mnt` will not be in a namespace that is a child of the name space used when using rootless podman. That's basically what I was trying to communicate above. If we use option 2 without `--pid=host`, we may be able to use rootless podman. Have you tried that?

With rootless podman I get:

```
[vagrant@localhost ~]$ podman run --privileged -v /run/udev:/run/udev -e RUST_LOG=info  stratis /run/udev/pid
[2022-12-02T14:36:23Z INFO  pid] main pid: 1
[2022-12-02T14:36:23Z INFO  pid] ppid: 0
[2022-12-02T14:36:23Z INFO  pid] tid: 1
[2022-12-02T14:36:23Z INFO  pid] pid_one_stat.st_ino: 4026532758
[2022-12-02T14:36:23Z INFO  pid] self_stat.st_ino: 4026532758
[2022-12-02T14:36:23Z INFO  pid] pid_one_stat.st_dev: 4
[2022-12-02T14:36:23Z INFO  pid] self_stat.st_dev: 4
[2022-12-02T14:36:23Z INFO  pid] unshare
[2022-12-02T14:36:23Z INFO  pid] pid_one_stat.st_ino: 4026532764
[2022-12-02T14:36:23Z INFO  pid] self_stat.st_ino: 4026532764
[2022-12-02T14:36:23Z INFO  pid] pid_one_stat.st_dev: 4
[2022-12-02T14:36:23Z INFO  pid] self_stat.st_dev: 4
thread 'main' panicked at 'assertion failed: !is_in_root_namespace()?', src/main.rs:57:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


[vagrant@localhost ~]$ podman run --privileged -v /run/udev:/run/udev -e RUST_LOG=info --pid=host stratis /run/udev/pid
[2022-12-02T14:36:31Z INFO  pid] main pid: 106935
[2022-12-02T14:36:31Z INFO  pid] ppid: 106933
[2022-12-02T14:36:31Z INFO  pid] tid: 106935
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Nix(EACCES)', src/main.rs:35:25
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

So yes if we don't call unshare() (option 2, as implemented by the current code) if stratisd is PID 1 we should be able to use rootless Podman since we won't `stat /proc/1/ns/mnt`. Does that make sense? Sorry for the confusion!

Unfortunately, this won't be enough to allow rootless entirely. Also, Podman in my experience is just a quick and easy way to execute containers. The end goal is to run on Kubernetes and the container engine will run as root anyways.

Snippet from the rootless failure after skipping the unshare():

```
[vagrant@localhost ~]$ podman run --user root --privileged -ti -v /run/dbus/system_bus_socket:/run/dbus/system_bus_socket:ro -v /dev:/dev -v /run/udev:/run/udev --ipc=host -v /usr/share/dbus-1/system.d/:/usr/share/dbus-1/system.d/ -v /usr/lib/udev/rules.d/:/usr/lib/udev/rules.d/ --device=/dev/vdb stratis /usr/lib/udev/rules.d/stratisd --log-level debug
[2022-12-02T14:46:57Z DEBUG stratisd::stratis::run] Running as PID 1, not unsharing mount namespace
[2022-12-02T14:46:57Z INFO  stratisd::stratis::run] stratis daemon version 3.4.1 started
[2022-12-02T14:46:57Z INFO  stratisd::stratis::run] Using StratEngine
[2022-12-02T14:46:57Z INFO  stratisd::engine::strat_engine::liminal::identify] Beginning initial search for Stratis block devices
[2022-12-02T14:46:57Z ERROR stratisd::engine::strat_engine::backstore::crypt::shared] Device /dev/vdb does not exist or access denied.
[2022-12-02T14:46:57Z WARN  stratisd::engine::strat_engine::backstore::crypt::shared] Failed to acquire a context for device /dev/vdb; failed with error: IO error occurred: Block device required (os error 15)
[2022-12-02T14:46:57Z WARN  stratisd::engine::strat_engine::liminal::identify] udev identified device /dev/vdb as a LUKS device, but could not read LUKS header from the device, disregarding the device: Cryptsetup error: IO error occurred: Block device required (os error 15)
[2022-12-02T14:46:57Z WARN  stratisd::engine::strat_engine::liminal::identify] udev identified device /dev/vdc as a Stratis device but device /dev/vdc could not be opened for reading: Permission denied (os error 13), disregarding the device
[2022-12-02T14:46:57Z WARN  stratisd::engine::strat_engine::liminal::identify] udev identified device /dev/vde as a Stratis device but device /dev/vde could not be opened for reading: Permission denied (os error 13), disregarding the device
[2022-12-02T14:46:57Z WARN  stratisd::engine::strat_engine::liminal::identify] udev identified device /dev/dm-0 as a Stratis device but device /dev/dm-0 could not be opened for reading: Permission denied (os error 13), disregarding the device
[2022-12-02T14:46:57Z ERROR stratisd::stratis::run] The devicemapper thread exited with an error: Failed to initialize DM context; DM error: DM Core error: DM context not initialized due to IO error: Permission denied (os error 13); shutting down stratisd...
Failed to initialize DM context; DM error: DM Core error: DM context not initialized due to IO error: Permission denied (os error 13)
```

jbaublitz:
Okay! @mulkieran I'm okay with this PR if you are!

jbaublitz (approved):

src/engine/strat_engine/ns.rs
```diff
@@ -27,6 +27,8 @@ pub fn unshare_namespace() -> StratisResult<()> {
     if is_in_root_namespace()? {
         unshare(CloneFlags::CLONE_NEWNS)?;
     }
+    // unshare() seems to be working fine even if called on a PID 1 process, so the assertion below is
```
mulkieran, line 30: Plz remove comment here and put a pre-condition in a comment above the method.

Suggested text is: "Precondition: If running in a container, PID is not 1 OR the container does not share the host PID.".

Feel free to revise if the wording is incorrect.

src/stratis/run.rs
```diff
@@ -48,7 +50,22 @@ async fn signal_thread() -> StratisResult<()> {
 /// Always check for devicemapper context.
 pub fn run(sim: bool) -> StratisResult<()> {
     if !sim {
-        unshare_namespace()?;
+        // If we are in a container, stratisd PID will be 1 and 'unshare' will not work. Calling
```
mulkieran, line 53: Plz change this comment to, proposed:

"If stratisd is running as PID 1, then it is definitely in a container and the unshare_namespace() command method will fail, since the pid of the process will remain the same after the unshare command. Note that it is possible for stratisd to be running in a container and not as PID 1, but this
situation is less easily detected. stratisd will fail in the unshare_namespace() method if that is the case."

Please revise the comment if incorrect.

mulkieran (changes requested):
The code looks good. Would you revise the code comments into the more declarative form suggested in the PR review comments?

src/stratis/run.rs
leseb: Revised but kept the original intent. Thank you.

src/engine/strat_engine/ns.rs
leseb: Revised but kept the original intent. Thank you.

leseb:
@jbaublitz @mulkieran thanks for the reviews, glancing at the code again I believe we could gain some clarity but doing the following methods rename:

* unshare_namespace --> unshare_mount_namespace
* is_in_root_namespace --> is_in_root_mount_namespace

This will be closer to what the code is effectively doing, without having to look at the definition of `CLONE_NEWNS`.

If you agree, what would you prefer between these two options:

* create a new PR
* add a separate commit to this PR

Thanks!

jbaublitz:
Yep, works for me!

mulkieran (approved):

PR #3191, opened 2022-12-02
Increase amount of space allocated for metadata

jbaublitz:
Closes https://github.com/stratis-storage/stratisd/issues/3192

@bleve This patch should resolve your metadata issue if you would like to test it. I've already tested both components of it individually, and it appears to resolve the problem.

### DIFF
--- a/src/engine/strat_engine/cmd.rs
+++ b/src/engine/strat_engine/cmd.rs
@@ -36,8 +36,14 @@ use crate::{
     stratis::{StratisError, StratisResult},
 };
 
-// The maximum allowable size of the thinpool metadata device
+/// The maximum allowable size of the thinpool metadata device
 const MAX_META_SIZE: MetaBlocks = MetaBlocks(255 * ((1 << 14) - 64));
+/// This constant is an empirical value. Traditional IO patterns that cause break
+/// do not cause the metadata to exceed 8 times the prediction by thin_metadata_size.
+/// While this works as an upfront estimate for metadata size, we also listen for
+/// the low water mark on the metadata device if needed space exceeds 8 times the
+/// prediction.
+const THIN_META_MULT_FACTOR: u64 = 8;
 
 /// Find the executable with the given name by looking in likely locations.
 /// Return None if no executable was found.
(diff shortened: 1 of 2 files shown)


jbaublitz:
@mulkieran I think we could easily add an automated test to our unit tests soon to ensure we never run out of metadata space if you'd like.

src/engine/strat_engine/cmd.rs
```diff
@@ -456,7 +456,7 @@ pub fn thin_metadata_size(
         .read_to_string(&mut output)?;
     if is_ok {
         Ok(min(
-            Sectors(
+            8u64 * Sectors(
```
mulkieran, line 459: We had better make this a constant and document it as best we can, even though there's not much to say.

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -259,6 +259,17 @@ fn status_to_usage(status: Option<ThinPoolStatus>) -> Option<ThinPoolUsage> {
     })
 }
 
+/// Convert the thin pool status to the metadata low water mark.
```
mulkieran, line 262: This is fine (exactly what we do for ThinPoolUsage), but I'm starting to wonder why we don't just store the whole ThinPoolStatus in the thinpool implementation and extract the data we need from that when we want it. We can either tackle that in this PR, or I'll file an issue and we can work on that later.

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -953,6 +979,7 @@ impl ThinPool {
         pool_uuid: PoolUuid,
         backstore: &mut Backstore,
         new_thin_limit: Option<u64>,
+        is_lowater: bool,
```
mulkieran, line 982: Better document this parameter as well.

mulkieran (changes requested):
A couple of requests and one comment.

From the code, I conclude that stratisd was already receiving the thin metadata low water mark event, so there is literally no work to do on the event handling side?

src/engine/strat_engine/thinpool/thinpool.rs
jbaublitz: Resolved.

src/engine/strat_engine/thinpool/thinpool.rs
jbaublitz: Resolved.

src/engine/strat_engine/cmd.rs
jbaublitz: Resolved.

mulkieran (approved):

PR #3205, opened 2022-12-15
Use assert matches instead of matches crate

mulkieran:
Related https://github.com/stratis-storage/project/issues/221

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,6 +21,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "assert_matches"
+version = "1.5.0"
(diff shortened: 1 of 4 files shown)


jbaublitz (commented):
Are we using the new crate anywhere?

mulkieran:
> Are we using the new crate anywhere?

I did a quick scan of the places where we already use assert_matches, looking for a setup where we could take advantage of the greater expressiveness. The one place that looked hopeful, well it turned out to be rather too simple to need the greater expressiveness.

jbaublitz (approved):

PR #3207, opened 2022-12-16
Do not expect redundancy value in CreatePool parameters

### LINKED ISSUE #2969: Remove redundancy option from PoolCreate method

mulkieran:
Closes #2969

### DIFF
--- a/src/dbus_api/api/manager_3_5/methods.rs
+++ b/src/dbus_api/api/manager_3_5/methods.rs
@@ -1,0 +1,117 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use std::path::Path;
+
+use dbus::{arg::Array, Message};
+use dbus_tree::{MTSync, MethodInfo, MethodResult};
+use futures::executor::block_on;
+
+use crate::{
+    dbus_api::{
+        blockdev::create_dbus_blockdev,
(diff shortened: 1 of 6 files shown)


mulkieran:
All stratis-cli related tests should fail.

mulkieran:
@jbaublitz I'm not convinced that abstracting the shared content of ```create_pool()``` will do any good over the long term, but I'm not really opposed to doing it, either.

src/dbus_api/api/manager_3_5/methods.rs
```diff
@@ -0,0 +1,117 @@
...
+    let base_path = m.path.get_name();
+    let message: &Message = m.msg;
+    let mut iter = message.iter_init();
+
+    let name: &str = get_next_arg(&mut iter, 0)?;
+    let devs: Array<'_, &str, _> = get_next_arg(&mut iter, 2)?;
+    let (key_desc_tuple, clevis_tuple): EncryptionParams = (
+        Some(get_next_arg(&mut iter, 3)?),
+        Some(get_next_arg(&mut iter, 4)?),
+    );
```
jbaublitz, line 37: It looks like the arg count is off by 1.

jbaublitz (changes requested):
Just one issue I'm seeing. I don't think sharing the code is really necessary here either.

src/dbus_api/api/manager_3_5/methods.rs
mulkieran: Fixed, thanks.

jbaublitz (approved):

Issue #3195, opened 2022-12-06
statically compile udev lib binaries 'stratis-str-cmp' and 'stratis-base32-decode'


leseb:
As part of the containerization effort, we need to copy `stratis-str-cmp` and `stratis-base32-decode`, on the host under `/usr/lib/udev/` so that Stratis's udev rule can work adequately and `/dev/stratis` devices can be created.
Statically compiling allows us to not rely on any libraries and avoid potential execution issues, even though the scripts are relatively simple.

Today, `stratis-utils` governs the generation of the util binaries, due to the increased numbers of dependencies it cannot be statically compiled with `RUSTFLAGS='-C target-feature=+crt-static`.

One idea would be to decouple/extract both binaries into individual ones, I believe this is how things used to be.
Happy to discuss whatever approach is the best for the team given the above requirement.

Also, happy to work on the solution under the team's guidance.
Thanks!

jbaublitz:
@leseb One note about static linking in Rust is that without musl, there will always be at least a glibc dynamic link dependency. Would that concern you?

PR #3210, opened 2022-12-19
build udev lib bins statically

leseb:
This patch extracts both stratis-base32-decode and stratis-str-cmp from stratis-util. They are now two distinct binaries.
Also, when built, they are built statically.

```
$ ldd target/x86_64-unknown-linux-gnu/release/stratis-base32-decode
        statically linked
```
Related: https://github.com/stratis-storage/stratisd/issues/3195
Co-authored-by: mulhern <amulhern@redhat.com>
Signed-off-by: Sébastien Han <seb@redhat.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,8 +31,11 @@ endif
 MIN_FEATURES = --no-default-features --features engine,min
 SYSTEMD_FEATURES = --no-default-features --features engine,min,systemd_compat
 EXTRAS_FEATURES =  --no-default-features --features engine,extras,min
+UDEV_FEATURES = --no-default-features --features udev_scripts
 
 DENY = -D warnings -D future-incompatible -D unused -D rust_2018_idioms -D nonstandard_style
+
+STATIC_FLAG = -C target-feature=+crt-static
 
 CLIPPY_DENY = -D clippy::all -D clippy::cargo -A clippy::multiple-crate-versions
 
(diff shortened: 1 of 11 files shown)


leseb:
@jbaublitz @mulkieran this is a draft PR that attempts to address: https://github.com/stratis-storage/stratisd/issues/3195, looking for early feedback. Thank you.

jbaublitz:
Since this statically links the executables, I actually have nothing to request at this time. @mulkieran Do you have any feedback on this?

jbaublitz (changes requested):
Other than CI, that is. :) It looks like the code is failing to build due to a linking error.

leseb:
> Other than CI, that is. :) It looks like the code is failing to build due to a linking error.

I was missing `glibc-static` as a dependency.

leseb:
I see the CI intermittently failing on `python-based-tests` with the error:

```
RuntimeError: Unable to create a pool stratis_FKOZU with devices ['/dev/loop3', '/dev/loop4', '/dev/loop5', '/dev/loop6']: At least one of the devices specified was unsuitable for initialization: Could not obtain udev information for block device /dev/loop5: Block device /dev/loop5 could not be found in the udev database
```

I also saw it there after a recent merge https://github.com/stratis-storage/stratisd/actions/runs/3716119513/jobs/6302081476 so I don't think this is due to my code changes.

@jbaublitz @mulkieran what do you think?

jbaublitz:
Yes, we have intermittent CI failures that should be eventually resolved by the udev sync work. That is definitely not something you have to worry about.

jbaublitz (approved):

Makefile
```diff
@@ -36,6 +36,9 @@ EXTRAS_FEATURES =  --no-default-features --features extras,min
 
 DENY = -D warnings -D future-incompatible -D unused -D rust_2018_idioms -D nonstandard_style
 
+STATIC_FLAG = -C target-feature=+crt-static
+STATIC_TARGET = x86_64-unknown-linux-gnu
```
mulkieran, line 40: Could you explain the purpose of the ```STATIC_TARGET``` constant? The Makefile has a separate ```TARGET``` environment variable which we use to set the target argument as necessary.

mulkieran (changes requested):
I have one request for clarification, see inline comments.

In addition, it seems to me undesirable that dependencies are recompiled with the target-features flag necessary for the static build even though almost none of the dependencies are required for the udev scripts.

I'm going to be discussing with the team whether we should take the step of breaking out the dependencies into individual, more complex, entries as was done for, e.g., our chrono dependency. We could then make most dependencies optional and specifically require them for the individual features specified in the Cargo.toml file. For the udev scripts, and for this PR, you would then introduce a new feature, called perhaps ```udev_scripts``` and require precisely the rather few dependencies that are actually needed for these scripts.

We have been introducing more complex dependencies on an as-needed basis up until this point. But I also have a goal, which is to limit the features of the crates that we require to those that are actually necessary for stratisd. This could have a number of benefits, including reducing compiled code size and limiting the number of reported CVEs.

mulkieran:
@leseb We decided that it would be a good idea to see how splitting out the dependencies into a separate engine feature goes. That work is in progress, but not quite done: https://github.com/stratis-storage/stratisd/pull/3214. Also, we did some experiments with building packages from this PR (the spec file uses some of the Makefile targets) and ran into some problems between the build and install steps. We'll keep you informed about those, also.

mulkieran:
@leseb We've merged the relevant PR, which was #3218 . Please rebase and try the feature-based approach to compiling these two small scripts that I outlined in my review comment.

Makefile
leseb: `STATIC_TARGET` is used so that cargo can build on a different target name. Essentially I thought it was easier to apply a specific target inside the makefile instead of passing a different `TARGET` to the makefile before its execution. If I try to compile without changing the target name I get this (known) cargo error: 

```
error: cannot produce proc-macro for async-trait v0.1.59 as the target x86_64-unknown-linux-gnu` does not support these crate types
make: *** [Makefile:230: build-udev-utils] Error 101`. 
```

Which is solved by forcing the target name.

jbaublitz (approved):

mulkieran (changes requested):
Two final (I think) requests.

mulkieran (approved):

Issue #3105, opened 2022-09-08
Might want to have a third alternative that compiles in neither IPC mechanism


mulkieran:
Currnently, there is no way to compile in neither IPC mechanism. It is likely to be desirable to do this, though, as we make use of the stratisd engine source code to support various applications or tools of which the most obvious to me is stratis-dumpmetadata. At present we compile this application with the other IPC mechanism (i.e., not D-Bus) and that reduces the size of the debug executable by (160 - 116) / 160 /approx 40 / 160 = 1/4.

jbaublitz:
This seems reasonable to me! Do you think this would be a good issue for @bmr-cymru?

mulkieran:
Let's tentatively do that. It is in no way urgent yet...

mulkieran:
This turns out to be more complicated than was originally expected because the IPC mechanism picks up udev events and transmits them to the engine so that the engine can react to them and notify the IPC mechanism of any changes made so that the IPCmechanism, if it is D-Bus can update D-Bus properties and so forth.

So, to run w/out any IPC mechanism currently would prevent the engine's handling of udev events.

A solution to this is to add a third, dummy IPC mechanism, that passes the udev events along to the engine and otherwise does nothing.

PR #3214, opened 2023-01-03
Explode dependencies into individual entries

mulkieran:
Closes https://github.com/stratis-storage/stratisd/issues/3105

### DIFF
--- a/src/stratis/run.rs
+++ b/src/stratis/run.rs
@@ -20,14 +20,18 @@ use nix::unistd::getpid;
 
 #[cfg(feature = "dbus_enabled")]
 use crate::dbus_api::DbusAction;
+
+#[cfg(any(feature = "dbus_enabled", feature = "min"))]
+use crate::stratis::ipc_support::setup;
+
 use crate::{
     engine::{
         set_up_crypt_logging, unshare_mount_namespace, Engine, SimEngine, StratEngine,
         UdevEngineEvent,
     },
     stratis::{
-        dm::dm_event_thread, errors::StratisResult, ipc_support::setup, stratis::VERSION,
-        timer::run_timers, udev_monitor::udev_thread,
+        dm::dm_event_thread, errors::StratisResult, stratis::VERSION, timer::run_timers,
+        udev_monitor::udev_thread,
     },
 };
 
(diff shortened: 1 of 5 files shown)


mulkieran:
Still builds w/ related ci modification in https://github.com/stratis-storage/ci/pull/360

src/stratis/run.rs
```diff
@@ -90,11 +94,15 @@ pub fn run(sim: bool) -> StratisResult<()> {
     runtime.block_on(async move {
         async fn start_threads<E>(engine: Arc<E>, sim: bool) -> StratisResult<()> where E: 'static + Engine {
             let (trigger, should_exit) = channel(1);
+
+            #[allow(unused_variables)]
```
jbaublitz, line 98: should we `#[cfg()]` flag this instead?

jbaublitz (changes requested):
Just one comment, other than that this looks good.

src/stratis/run.rs
mulkieran: Done. (The parser won't allow a flag inside a destructuring let, so it's a bit verbose, but still better, I think.)

mulkieran:
also rebased.

src/stratis/run.rs
```diff
@@ -90,11 +94,18 @@ pub fn run(sim: bool) -> StratisResult<()> {
     runtime.block_on(async move {
         async fn start_threads<E>(engine: Arc<E>, sim: bool) -> StratisResult<()> where E: 'static + Engine {
             let (trigger, should_exit) = channel(1);
+
+            #[cfg(any(feature = "dbus_enabled", feature = "min"))]
             let (udev_sender, udev_receiver) = unbounded_channel::<UdevEngineEvent>();
+            #[cfg(all(not(feature = "dbus_enabled"), not(feature = "min")))]
+            let (udev_sender, _) = unbounded_channel::<UdevEngineEvent>();
```
jbaublitz, line 101: Both of the last two approaches seems to cause problems. The last one would generate a channel that would progressively take up more memory and never be drained and this one will generate a channel disconnect error every time a udev event is received. Neither seems ideal. I would recommend either requiring an IPC mechanism for the time being or having a separate thread to trigger udev event handling in the engine even if there is no IPC mechanism available.

jbaublitz (changes requested):
I think this may require larger changes after looking more closely.

mulkieran:
Obsoleted by #3379

PR #3220, opened 2023-01-09
Fix repetitive logging when pool is in NoPoolChanges state

### LINKED ISSUE #3171: When pool is in NoPoolActions mode warnings are logged every ten seconds
This is due to the fact that the timer thread kicks off to check the pools and filesystems, but that the check() methods refuse and log that refusal every time at the warning level.

This behavior is at least partly correct, since this routine action can't be taken, and that situation is something that ought to be logged at the warning level in the general case, because it, too, endangers the pool.

I think that the problem is that we log it so many times and, in this case, at too high a level, not that we log it at all.

One idea was to avoid calling the check() methods when NoPoolActions mode is set. That means no warnings are shown about the problem, we rely on warnings from the step where the pool is put into NoPoolActions in the first place.

Another idea is to return a this-is-currently-a-disallowed-action error from the check() methods. Then, that error could be treated differently. It could be logged at lower than warning level, either info or, my preference, debug.

jbaublitz:
Closes #3171

### DIFF
--- a/.github/workflows/cargo.yml
+++ b/.github/workflows/cargo.yml
@@ -55,6 +55,6 @@ jobs:
       - name: Run comparisons of version specs with available Fedora packages
         # yamllint disable rule:line-length
         run: |
-          COMPARE_FEDORA_VERSIONS=./ci/dependency_management/compare_fedora_versions FEDORA_RELEASE=rawhide IGNORE_ARGS="--ignore-category low" make -f Makefile_dependencies check-fedora-versions
-          COMPARE_FEDORA_VERSIONS=./ci/dependency_management/compare_fedora_versions FEDORA_RELEASE=f37 IGNORE_ARGS="--ignore-category low" make -f Makefile_dependencies check-fedora-versions
-          COMPARE_FEDORA_VERSIONS=./ci/dependency_management/compare_fedora_versions FEDORA_RELEASE=f36 IGNORE_ARGS="--ignore-category low" make -f Makefile_dependencies check-fedora-versions
+          COMPARE_FEDORA_VERSIONS=./ci/dependency_management/compare_fedora_versions FEDORA_RELEASE=rawhide IGNORE_ARGS="--ignore-category low --ignore-high=stratisd_proc_macros" make -f Makefile_dependencies check-fedora-versions
+          COMPARE_FEDORA_VERSIONS=./ci/dependency_management/compare_fedora_versions FEDORA_RELEASE=f37 IGNORE_ARGS="--ignore-category low --ignore-high=stratisd_proc_macros" make -f Makefile_dependencies check-fedora-versions
+          COMPARE_FEDORA_VERSIONS=./ci/dependency_management/compare_fedora_versions FEDORA_RELEASE=f36 IGNORE_ARGS="--ignore-category low --ignore-high=stratisd_proc_macros" make -f Makefile_dependencies check-fedora-versions
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -140,9 +140,9 @@ jobs:
       - name: Run comparisons of version specs with available Fedora packages
         # yamllint disable rule:line-length
         run: |
-          COMPARE_FEDORA_VERSIONS=./ci/dependency_management/compare_fedora_versions FEDORA_RELEASE=rawhide make -f Makefile_dependencies check-fedora-versions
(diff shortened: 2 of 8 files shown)


src/engine/strat_engine/pool.rs
```diff
@@ -139,17 +139,26 @@ fn check_metadata(metadata: &PoolSave) -> StratisResult<()> {
...
             ActionAvailability::Full
-        }
-    } else {
-        ActionAvailability::Full
-    };
-    max(avail, backstore.action_availability())
+        },
+        backstore.action_availability(),
+    );
+    if state != ActionAvailability::Full {
```
mulkieran, line 155: There is also a log message emitted in ```StratPool::setup()```.  One of these messages might be considered superfluous; you decide.

mulkieran (changes requested):
I believe we will have to bump the stratisd_proc_macros version number to 0.2.0, because the change in the constructor value of the returned error can be viewed as a breaking change for the current version of stratisd.

mulkieran:
> I believe we will have to bump the stratisd_proc_macros version number to 0.2.0, because the change in the constructor value of the returned error can be viewed as a breaking change for the current version of stratisd.

Actually, to ease Fedora packaging, it might be helpful to break this PR into two parts. Thinking...

mulkieran:
> > I believe we will have to bump the stratisd_proc_macros version number to 0.2.0, because the change in the constructor value of the returned error can be viewed as a breaking change for the current version of stratisd.
> 
> Actually, to ease Fedora packaging, it might be helpful to break this PR into two parts. Thinking...

I think my original thought was the best. We can increase the stratisd_proc_macros version to 0.2.0 and require that version in stratisd's Cargo.toml. For Fedora packaging, we'll use a side-tag.

src/engine/strat_engine/pool.rs
jbaublitz: You're right! Resolved.

src/engine/strat_engine/pool.rs
```diff
@@ -269,12 +269,10 @@ impl StratPool {
         let mut needs_save = metadata.thinpool_dev.fs_limit.is_none()
             || metadata.thinpool_dev.feature_args.is_none();
 
-        if action_avail != ActionAvailability::NoPoolChanges {
-            needs_save |= match thinpool.check(uuid, &mut backstore) {
-                Ok(check) => check.0,
-                Err(e) => return Err((e, backstore.into_bdas())),
-            };
-        }
+        needs_save |= match thinpool.check(uuid, &mut backstore) {
```
mulkieran, line 272: For background here of why we decided to put this in an if condition:

1. A pool with inconsistent block sizes was created using a previous version of stratisd.
2. On set up, current version of stratisd detects and puts in NoPoolChanges mode.
3. Then check method is run and, because new version of stratisd has greedy allocation algorithm for thinpool metadata device, the metadata device is extended. This makes needs_save true.
4. Then, because needs_save is true, write_metadata is called, and pool setup returns an error.

The thing we were trying to avoid, then, is having the pool setup encounter what we thought of an an unnecessary error. It was also the case that after the expanson of the thin meta device, the DM devices were inconsistent w/ the pool metadata, and that couldn't be fixed w/out a stratisd restart.

Another way the pool could encounter an error on setup is if the ```started``` status has changed, or needs to be updated. I believe we could have encountered the error in our previous testing using inconsistent block sizes. I may try to formulate another test around that.

My proposal, though, which eliminates the code here entirely is to simply remove the two places where we call ```ThinPool::check``` outside of separate event threads, as it is my belief that both calls were introduced before the timer thread was introduced, and that the introduction of the timer thread resolves the problems that these calls were expected to handle.

That's in https://github.com/stratis-storage/stratisd/pull/3223, which I'm putting up for your review.

mulkieran (changes requested):
Code looks good, but I'ld like to see if you're on board w/ removing the check() calls in ```setup()``` and ```add_blockdevs()``` entirely, see https://github.com/stratis-storage/stratisd/pull/3223 and explanation below.

src/engine/strat_engine/pool.rs
jbaublitz: Resolved.

mulkieran (changes requested):
This looks good.

Please add "--ignore-high=stratisd_proc_macros" to ```IGNORE_ARGS``` variable in GitHub workflows files that have ```compare_fedora_versions``` commands.

PR #3222, opened 2023-01-10
stratisd_proc_macros: use license field in Cargo.toml

bgurney-rh:


mulkieran (approved):

mulkieran:
@jbaublitz I think it's legitimate to consider this package to have a standard MPL-2.0 license, since it does not depend on cryptsetup in any way. We can signal this by using the "license" key, instead of the "license-file" key. If we do that, rust2rpm will be satisfied, and copy the proper license information into the spec file.

jbaublitz:
> @jbaublitz I think it's legitimate to consider this package to have a standard MPL-2.0 license, since it does not depend on cryptsetup in any way. We can signal this by using the "license" key, instead of the "license-file" key. If we do that, rust2rpm will be satisfied, and copy the proper license information into the spec file.

I think there's a misconception here. Our `license-file` directive is no different from specifying the `license` key in the stratisd case because we don't add any additional information to the license indicating that we depend on libcryptsetup and that we become effectively GPL. I think this stems from a misconception when we were discussing this. `license-file` allows you to specify non-standard licenses was my interpretation, but we haven't pointed to a non-standard license. I think we could just as easily use `license` in stratisd as here. I think if we want to make that clear, we have to add it to the README.

jbaublitz (approved):

mulkieran:
> > @jbaublitz I think it's legitimate to consider this package to have a standard MPL-2.0 license, since it does not depend on cryptsetup in any way. We can signal this by using the "license" key, instead of the "license-file" key. If we do that, rust2rpm will be satisfied, and copy the proper license information into the spec file.
> 
> I think there's a misconception here. Our `license-file` directive is no different from specifying the `license` key in the stratisd case because we don't add any additional information to the license indicating that we depend on libcryptsetup and that we become effectively GPL. I think this stems from a misconception when we were discussing this. `license-file` allows you to specify non-standard licenses was my interpretation, but we haven't pointed to a non-standard license. I think we could just as easily use `license` in stratisd as here. I think if we want to make that clear, we have to add it to the README.



> > @jbaublitz I think it's legitimate to consider this package to have a standard MPL-2.0 license, since it does not depend on cryptsetup in any way. We can signal this by using the "license" key, instead of the "license-file" key. If we do that, rust2rpm will be satisfied, and copy the proper license information into the spec file.
> 
> I think there's a misconception here. Our `license-file` directive is no different from specifying the `license` key in the stratisd case because we don't add any additional information to the license indicating that we depend on libcryptsetup and that we become effectively GPL. I think this stems from a misconception when we were discussing this. `license-file` allows you to specify non-standard licenses was my interpretation, but we haven't pointed to a non-standard license. I think we could just as easily use `license` in stratisd as here. I think if we want to make that clear, we have to add it to the README.

@jbaublitz I agree with your point regarding the stratisd license situation. What i'm talking about here is not how users, or we, interpret the ```license-file``` key, but how _cargo_ interprets it (https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields) (also rust2rpm). I'm not yet sure what to do in stratisd that is best for us, cargo, and rust2rpm; but I think we can satisfy cargo and rust2rpm for ```stratisd_proc_macros``` without any further thought, so might as well go ahead.

PR #3223, opened 2023-01-11
Remove two calls to ThinPool::check

mulkieran:


src/engine/strat_engine/serde_structs.rs
```diff
@@ -37,6 +37,7 @@ pub struct PoolSave {
     pub backstore: BackstoreSave,
     pub flex_devs: FlexDevsSave,
     pub thinpool_dev: ThinPoolDevSave,
+    // TODO: This data type should no longer be optional in Stratis 4.0
```
jbaublitz, line 40: Was removing this comment intentional? I think it should still be there.

jbaublitz (changes requested):
Just one comment. Other than that, I think this will be an improvement.

src/engine/strat_engine/serde_structs.rs
mulkieran: I think you are reading it backwards. I added this comment, because I believe we intended to make the started field mandatory in Stratis 4.0.

jbaublitz: Whoops! I'll approve this.

PR #3242, opened 2023-01-26
Log if error when wiping just initialized block devices

mulkieran:


jbaublitz (commented):
Would it be reasonable to return a `NoActionRollback` error variant here containing the original error and the rollback error but not setting an `ActionAvailability` level?

mulkieran:
@jbaublitz Ok, I think this is ready again.

jbaublitz (approved):

mulkieran:
unusual error; bears investigation.

mulkieran:
rebased

PR #3254, opened 2023-02-02
Add valgrind tests to CI

### LINKED ISSUE #3230: Add valgrind task to CI
We know that we can run valgrind on stratisd and have detected at least one memory leak, now fixed. Can we run it in our nightly CI?

jbaublitz:
Closes #3230

### DIFF
--- a/.github/workflows/valgrind.yml
+++ b/.github/workflows/valgrind.yml
@@ -1,0 +1,275 @@
+---
+name: valgrind checks
+
+# yamllint disable-line rule:truthy
+on:
+  schedule:
+    - cron: 15 3 * * *
+
+  workflow_dispatch:
+
+# cancel the in-progress workflow when PR is refreshed.
+# yamllint disable rule:line-length
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.head_ref || github.sha }}
+  cancel-in-progress: true
+
+jobs:
+  checks:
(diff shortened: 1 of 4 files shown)


jbaublitz:
@mulkieran I'm wondering whether we want to further up the timeout values for the purposes of valgrind. The current timeout values in #3255 work outside of valgrind but valgrind seems to slow it down enough that we may need even longer timeouts. What are your thoughts?

mulkieran:
@jbaublitz I don't see any problem w/ longer timeouts.

jbaublitz:
@mulkieran Yeah unfortunately the timeouts don't occur each time I'm going to run it a few more times through rebases before I consider it truly not a problem.

mulkieran (changes requested):
This all looks good except that development rust toolchain version needs to move up to 1.68.0.

At present, the valgrind report does emit some things that look like errors, but none of the tasks fail.

Can the tasks fail? If they can only report but not fail, running them nightly rather than on every PR would probably be enough.

jbaublitz:
They can't fail based on valgrind errors so nightly might be best.

mulkieran:
> They can't fail based on valgrind errors so nightly might be best.

Ok! Let's swap it over to nightly once the PR is ready.

jbaublitz:
I think this is ready.

bmr-cymru (approved):

PR #3274, opened 2023-03-08
Metadata rework

jbaublitz:
Related to stratis-storage/project#553


jbaublitz:
@mulkieran I have a testing question for you. Now that we can no longer create legacy pools in stratisd, how do we want to go about testing the old pool unlock mechanism, etc.? I don't see a good way to do this other that using two separate versions of stratisd, one for legacy pool creation and one for testing the current version.

jbaublitz:
Blocked by #3404

Makefile
```diff
@@ -259,6 +260,13 @@ build-stratis-base32-decode:
 # so we use two distinct targets to build the two binaries
 build-udev-utils: build-stratis-str-cmp build-stratis-base32-decode
 
+## Build the test extras
```
mulkieran, line 263: Please add building with these flags to clippy Makefile target and to .PHONY list also.

mulkieran (changes requested):
Quick preliminary request...

mulkieran:
@jbaublitz Plz rebase again when you get the chance.

mulkieran:
@jbaublitz Plz rebase and resolve conflicts.

jbaublitz:
Okay this should be up to date again

mulkieran (changes requested):
Plz update stratisd-proc-macros lowest version to 0.2.1. And plz see comment above about Makefile, thx: https://github.com/stratis-storage/stratisd/pull/3274#pullrequestreview-1589816447 .

jbaublitz:
TODO:

- [ ] Reserve space for encryption metadata in unencrypted case
- [ ] Reserve space for RAID metadata in the future
- [ ] Reserve space for integrity metadata in the future

martinpitt:
The cockpit test failure is some raid regression in rawhide, unrelated to this PR. I saw it in https://github.com/fedora-selinux/selinux-policy/pull/1962 as well. Debugging it there, unfortunately not very easy to reproduce. Please ignore.

martinpitt:
@jbaublitz We did a quick hack to skip that test, as it's problematic to investigate. (We now do this in a draft PR). You can retry this and #3516 , sorry for the noise!

martinpitt:
Hmm, so `stratis pool create TEST1 {dev_1} {dev_2}` fails with
> stratisd failed to perform the operation that you requested. It returned the following information via the D-Bus: ERROR: Could not allocate sufficient space for thinpool devices. 

dev1/2 are two loop devices of 4 GiB each. They are sparsely allocated, but the tmt machines should have enough (> 20 GB) disk space.

But your own tests panic a lot as well, could that be related?

jbaublitz:
@martinpitt Yes, this is definitely a problem on our end. No need to look into this. I think I incorrectly did accounting for free space at the block device layer and need to appropriately account for the new metadata space we're reserving for dm-integrity.

jbaublitz:
@martinpitt I think it's fixed now.

martinpitt:
Thanks @jbaublitz !

jbaublitz:
@martinpitt It looks like the cockpit test is hanging. Do I just issue `/packit test` to unstick it?

martinpitt:
@jbaublitz I don't actually know (that'd be a question for the packit team), but it seems it actually worked -- some already finished, and the others look like real "in progress"

jbaublitz:
@martinpitt Sorry for the ping before just trying it, it does indeed seem to have worked.

mulkieran:
@jbaublitz Plz rebase when you get the chance, thx.

mulkieran:
@jbaublitz Plz rebase when you get the chance.

mulkieran (changes requested):
Requesting a rebase since there are conflicts

src/engine/strat_engine/serde_structs.rs
```diff
@@ -71,6 +73,36 @@ pub trait Recordable<T: Serialize> {
...
+/// List of optional features for pools.
+#[derive(Debug, Deserialize, Eq, PartialEq, Serialize, Hash, Copy, Clone)]
+pub enum PoolFeatures {
+    Raid,
+    Integrity,
+    Encryption,
+}
+
+/// Features enabled on the given pool to be stored in the metadata.
+#[derive(Debug, Deserialize, Eq, PartialEq, Serialize)]
```
mulkieran, line 85: This is the first use of a ```HashSet``` for our Serde structs. With thin pool features args, which are also really conceptually a set, we went with a ```Vec<String>``` (reason unknown). In our Record implementation for ```ThinPoolDevSave``` we alphabetically sort the feature args, so that a textual diff won't make them appear to have changed when they haven't. It would be helpful to make sure that the PoolFeatures are also sorted in a canonical way, probably alphabetically, when they are written.  For this reason, I think it would be reasonable to use a ```Vec``` instead, since the features end up in arbitrary order in the HashSet and then the way to write them ordered is probably some ```serialize_with``` annotation and that seems like a lot of bits and pieces for limited gain.

mulkieran (changes requested):
I have one comment on the most recent commit only...

martinpitt:
The [cockpit test failued](https://artifacts.dev.testing-farm.io/aa3349cf-e7a6-4873-a205-7c4577a457f6/) with

> warning: There was an error encountered when reading the metadata for the devices found for pool with UUID 21a8d032-ce4f-4c2f-9961-e998e507f74c; timestamp indicates data was written, but no data successfully read

but I'm going to ignore these as long as your own upstream tests fail, ok?

src/engine/strat_engine/serde_structs.rs
jbaublitz: Resolved.

jbaublitz:
> The [cockpit test failued](https://artifacts.dev.testing-farm.io/aa3349cf-e7a6-4873-a205-7c4577a457f6/) with
> 
> > warning: There was an error encountered when reading the metadata for the devices found for pool with UUID 21a8d032-ce4f-4c2f-9961-e998e507f74c; timestamp indicates data was written, but no data successfully read
> 
> but I'm going to ignore these as long as your own upstream tests fail, ok?

Yes this was a bug in code I pushed. It should be resolved now.

src/engine/strat_engine/backstore/backstore/v2.rs
```diff
@@ -0,0 +1,1626 @@
...
+            .collect::<PathBuf>();
+        let enc = match (metadata_enc_enabled, unlock_method, passphrase.as_ref()) {
+            (false, Some(_), _) => return Err((
+                StratisError::Msg(
+                    "Metadata reported that encryption was not enabled but unlock method was provided"
+                        .to_string()
+                ),
+                data_tier.block_mgr.into_bdas(),
+            )),
+            (false, _, Some(_)) => return Err((
```
mulkieran, line 402: I believe that this condition is already covered in ```LiminalDevices::start_pool_new()```.  I think that it would  make sense to move the ```(false, Some(_), _)``` condition into that method also to avoid setting up the cache tier unnecessarily. I think probably the ```(true, None, _)``` condition as well.

mulkieran (changes requested):
One comment, based on testing.

mulkieran (changes requested):
Still working through results; this is all I have so far.

mulkieran (changes requested):
I've identified a somewhat important inconsistency between ```start_pool_legacy``` and ```start_pool_new```. What guarantees that the entry for the pool is found in both cases is that the metadata had to be obtained from the entry to find out the metadata version in order to find out what method to dispatch to. So the reasoning should be the same for both methods.

In any case, I would make ```start_pool_new``` and ```start_pool_legacy``` non-public.

I would probably then decide to use assertions in both places I've marked and extend the expect message to be more explicit.

But turning both locations into error messages is also a legitimate decision.

src/engine/strat_engine/liminal/liminal.rs
```diff
@@ -233,10 +276,229 @@ impl LiminalDevices {
...
+                Ok(o) => o,
+                Err(e) => return Err((e, bdas)),
+            };
+
+            let passphrase = match (
+                metadata.features.contains(PoolFeatures::Encryption),
+                unlock_method,
+                passphrase_fd,
+            ) {
+                (false, _, Some(_)) => {
```
mulkieran, line 361: Could you convert lines 361 to lines 368 (two matches) to the following two matches instead:

```
(false, None, None) => { None },
(false, _, _) => {
    return Err((StratisError::Msg(format!("Metadata reported that encryption was not enabled but some unlock method was provided.")));
}
```
}

src/engine/strat_engine/liminal/liminal.rs
```diff
@@ -157,49 +175,74 @@ impl LiminalDevices {
...
-        let unlocked_devices = match (encryption_info, unlock_method) {
-            (Ok(Some(_)), None) => {
+            })?;
+
+        // Here we take a reference to entries in stopped pools because the call to unlock_pool
+        // below requires the pool being unlocked to still have its entry in stopped_pools.
+        // Removing it here would cause an error.
+        let encryption_info = pool.encryption_info();
+        let unlocked_devices = match (encryption_info, unlock_method, passphrase_fd) {
+            (Ok(Some(_)), None, _) => {
```
mulkieran, line 218: The proposed change is much like the other one suggested. Start the match with:

```
(Ok(None), None, None) => Vec::new(),
(Ok(None), _, _) => {
    return Err(StratisError::Msg(format!(
                    "Pool with UUID {pool_uuid} is not encrypted but some unlock method was provided"
                )));    
}
```

And remove the matches now at line 223, and between 235 and 244. This aligns the code with my other suggestion and also removes a dependency on the ordering of the conditions of unlock method set and passphrase fd present by combining them into one.

src/engine/strat_engine/crypt/handle/v1.rs
```diff
@@ -0,0 +1,1632 @@
...
+                Ok(Some(dev))
+            }
+        }
+        Err(e) => Err(e),
+    }
+}
+
+/// Set up a handle to a crypt device using either Clevis or the keyring to activate
+/// the device.
+pub fn setup_crypt_handle(
```
mulkieran, line 412: Please remove ```pub``` modifier.

src/engine/strat_engine/crypt/handle/v2.rs
```diff
@@ -0,0 +1,1252 @@
...
+                Ok(Some(dev))
+            }
+        }
+        Err(e) => Err(e),
+    }
+}
+
+/// Set up a handle to a crypt device using either Clevis or the keyring to activate
+/// the device.
+pub fn setup_crypt_handle(
```
mulkieran, line 177: Please remove ```pub``` modifier.

mulkieran (changes requested):
Extremely minor matching requests...

src/engine/strat_engine/crypt/shared.rs
```diff
@@ -838,20 +484,47 @@ pub fn activate(
...
             CryptActivate::empty(),
-        ),
-        "Failed to activate device with name {}",
-        name
-    );
+        )
+        .is_err()
+    {
+        if let Some(p) = passphrase {
+            let key_slot = if unlock_method == UnlockMethod::Keyring {
```
mulkieran, line 504: Are you able to re-order this so that you get the keyslot number regardless of whether there is a passphrase or not, but taking into account the unlock method? The reason this seems desirable is the error messages. At present if passphrase is not specified but unlock method is wrong (the examples below assume that the pool was initially encrypted with a key in the kernel keyring and are now stopped) you get the following:

```
stratis pool start
--name spool4 --unlock-method=clevis
Execution failed:
stratisd failed to perform the operation that you requested. It
returned the following information via the D-Bus: ERROR: An attempt to
set up pool with UUID b935bc7e-4e40-4908-85b8-284a99b4492a from the
assembled devices failed; Device
stratis-1-private-b935bc7e4e40490885b8284a99b4492a-crypt was activated
but is reporting that it is inactive.
```

but if you also pass a passphrase:

```
stratis pool start
--name spool4 --unlock-method=clevis --capture-key
Enter passphrase followed by the return key:
Verify passphrase entered:
Execution failed:
stratisd failed to perform the operation that you requested. It
returned the following information via the D-Bus: ERROR: An attempt to
set up pool with UUID 158e01fd-0fed-4a94-a804-06ec263f6632 from the
assembled devices failed; Clevis keyslot not found.
```

Error messages are the same for legacy pools in the same situation.

I argue that the "Clevis keyslot not found" message is friendlier and not quite so low-level.

mulkieran (changes requested):
Just one specific request, applies to both kinds of pools

src/engine/strat_engine/liminal/liminal.rs
jbaublitz: This makes the error message less specific. Can you explain a little bit more why this is desirable?

mulkieran: This caught my attention because for every test, it's being done on legacy pools and new pools at the same time. Both situations, passphrase + method can happen at once, and, due to a different ordering between the code for the new and legacy pools, when both were set, the legacy pool error message picked the opposite thing to report on from the new pool error message. In itself, that is not so important, and might not have been noticed if we were not doing testing that way.

The thing I'm worrying about is future possibility of multiple token slots + the fact that the CLI user is likely to get this message. Given that there's potential for even more variety of ways to specify a way to start an encypted pool in future, all of which will not make sense for an unencrypted pool, I was thinking it would be helpful to just put them all in a single bucket.

src/engine/strat_engine/crypt/shared.rs
```diff
@@ -838,20 +484,44 @@ pub fn activate(
             )));
         }
     }
-    log_on_failure!(
+    if let Some(p) = passphrase {
```
mulkieran, line 487: The particular logical arrangement right here leads to an undesirable consequence:

If a passphrase is given _and_ it is incorrect the pool or device as the case may be will not be opened, _even_ if the correct key is in the keyring (in the case where the pool is encrypted with a key), or the equivalent notion if the pool was encrypted using Clevis.

The problem looks like this. In each case, the pool is encrypted only with the method it is expected to use when being started and in each case, a passphrase is specified but it is incorrect. I think we agreed that the passphrase would act as a fallback, in case the method specified failed without it.

```
$ sudo stratis pool start
--name spool1 --unlock-method keyring --capture-key
Enter passphrase followed by the return key:
Verify passphrase entered:
Execution failed:
stratisd failed to perform the operation that you requested. It
returned the following information via the D-Bus: ERROR: Cryptsetup
error: IO error occurred: Operation not permitted (os error 1).

$ sudo stratis pool start
--name spool2 --unlock-method keyring --capture-key
Enter passphrase followed by the return key:
Verify passphrase entered:
Execution failed:
stratisd failed to perform the operation that you requested. It
returned the following information via the D-Bus: ERROR: An attempt to
set up pool with UUID 6468c4db-fea9-4b75-8450-80dd1f051033 from the
assembled devices failed; Cryptsetup error: IO error occurred:
Operation not permitted (os error 1).

$ sudo stratis pool start
--name spool3 --unlock-method clevis --capture-key
Enter passphrase followed by the return key:
Verify passphrase entered:
Execution failed:
stratisd failed to perform the operation that you requested. It
returned the following information via the D-Bus: ERROR: Cryptsetup
error: IO error occurred: Operation not permitted (os error 1).

$ sudo stratis pool start
--name spool4 --unlock-method clevis --capture-key
Enter passphrase followed by the return key:
Verify passphrase entered:
Execution failed:
stratisd failed to perform the operation that you requested. It
returned the following information via the D-Bus: ERROR: An attempt to
set up pool with UUID 78389678-0e95-411a-b71a-81d2e7c64649 from the
assembled devices failed; Cryptsetup error: IO error occurred:
Operation not permitted (os error 1).
```

mulkieran (changes requested):
This is exactly one problem, but as shown it affects both kinds of pools.

src/engine/strat_engine/crypt/shared.rs
jbaublitz: I'm actually interested in discussing this in stand up. I did that somewhat intentionally only because let's say we rework boot from root work where it always prompts for a passphrase, for example. Is it more or less desirable to enter the wrong passphrase and have it still unlock because of Clevis or to fail because you entered the wrong passphrase?

jbaublitz: Resolved in stand up.

jbaublitz: I think we decided that this is actually the behavior we want.

mulkieran: This is the desired behavior, otherwise the user might be mislead into believing that an incorrect passphrase is actually correct.

mulkieran, line 487: I'm proposing a restructuring like the following:

```
let key_slot = ...
if let Some(p) = passphrase {
  log_on_failure!(device.activate_handle()...
} else {
  device.token_handle()...
}
```
i.e., hoisting the ```key_slot``` assignment out of the enclosing if. The reason for that is to avoid somewhat ugly error messages like:

```$ sudo stratis pool start
--name spool1 --unlock-method=clevis
Execution failed:
stratisd failed to perform the operation that you requested. It
returned the following information via the D-Bus: ERROR: Cryptsetup
error: IO error occurred: No such file or directory (os error 2).
$ sudo stratis pool start
--name spool2 --unlock-method=clevis
Execution failed:
stratisd failed to perform the operation that you requested. It
returned the following information via the D-Bus: ERROR: An attempt to
set up pool with UUID 9b8eb004-c479-478e-9406-13c8f786add4 from the
assembled devices failed; Cryptsetup error: IO error occurred: No such
file or directory (os error 2).
```

which you can get when you have encrypted with a key but you specify unlock method as clevis.

mulkieran (changes requested):
Still doing testing, just one request so far.

src/engine/strat_engine/crypt/shared.rs
jbaublitz: I'm not confident this is the best way to handle it only because keyslots are entirely bound to the passphrase implementation. Would augmenting the error handling for activating by token to detect errno 2 and return a better error message be sufficient? I did some digging into the cryptsetup source code and it looks like this will only be returned on failure to fetch that token/keyslot.

mulkieran: Yes. That would also be fine.

src/engine/strat_engine/crypt/handle/v2.rs
```diff
@@ -0,0 +1,1252 @@
...
+
+    /// Query the device metadata to reconstruct a handle for performing operations
+    /// on an existing encrypted device.
+    ///
+    /// This method will check that the metadata on the given device is
+    /// for the LUKS2 format and that the LUKS2 metadata is formatted
+    /// properly as a Stratis encrypted device. If it is properly
+    /// formatted it will return the device identifiers (pool and device UUIDs).
+    ///
+    /// NOTE: This will not validate that the proper key is in the kernel
```
mulkieran, line 474: I think in this implementation the ```can_unlock``` method no longer exists, so it makes sense to remove mention of it.

mulkieran (changes requested):
Still working...just caught sight of an out of place comment.

src/engine/strat_engine/crypt/shared.rs
jbaublitz: I tested my change and it works as expected.

src/engine/strat_engine/crypt/handle/v2.rs
jbaublitz: Resolved.

src/engine/strat_engine/crypt/shared.rs
```diff
@@ -838,20 +484,55 @@ pub fn activate(
...
-            None,
-            CryptActivate::empty(),
-        ),
-        "Failed to activate device with name {}",
-        name
-    );
+                None
+            };
+        log_on_failure!(
+            device.activate_handle().activate_by_passphrase(
```
mulkieran, line 502: Supposing you get the correct keyslot, but _also_ supply a bad passphrase, the error message is very low-level: ```Cryptsetup
error: IO error occurred: Operation not permitted (os error 1)```. It might be better to map to something a bit more user-friendly here as well.

mulkieran (changes requested):
One more remark on errors when encountering failures when the problem is bad encryption information...

src/jsonrpc/server/pool.rs
```diff
@@ -27,11 +23,10 @@ pub async fn pool_start(
     unlock_method: Option<UnlockMethod>,
     prompt: Option<RawFd>,
 ) -> StratisResult<bool> {
-    if let (Some(fd), Some(kd)) = (prompt, key_get_desc(Arc::clone(&engine), id.clone()).await?) {
-        key_set(engine.clone(), &kd, fd).await?;
-    }
-
-    Ok(engine.start_pool(id, unlock_method).await?.is_changed())
+    Ok(engine
```
mulkieran, line 26: Because of the change here, I think you will also want a change in ```src/bin/stratis-min/stratis-min.rs``` specifically around the error message ``` "--prompt and an unlock_method of clevis are mutually exclusive"```. I think that should no longer be true after this PR, because it is not true on the D-Bus side.

This test code here also illustrates the current behavior: https://github.com/stratis-storage/stratisd/pull/3635 .

mulkieran (changes requested):
Request about the client side of stratis-min

src/engine/strat_engine/crypt/shared.rs
jbaublitz: This error is not specific to this PR and existed previously. I am willing to put up another PR that we can merge separately, but I don't think it belongs here.

PR #3275, opened 2023-03-08
Rename clippy lint derive_hash_xor_eq

bgurney-rh:
Related: stratis-storage/project#587


bgurney-rh:
A simple rename will cause clippy errors in Rust versions before 1.68.0.  We have to find another way.

PR #3277, opened 2023-03-10
Modify clippy allow for derive_hash_xor_eq

bgurney-rh:
Related: stratis-storage/project#587


mulkieran (approved):
I'm accepting this, _but_ we'll remove every annotation visible in this commit as soon as we bump up to 1.68.0.

PR #3285, opened 2023-03-15
Add rollback of Pool setup when starting a pool fails

jbaublitz:
Related to #3274


mulkieran (changes requested):
There was one tiny thing I noticed a while back when glancing at it several days ago. Doing a serious pass now.

src/engine/strat_engine/liminal/identify.rs
```diff
@@ -171,30 +176,81 @@ impl DeviceInfo {
     }
 }
 
-impl From<StratBlockDev> for DeviceInfo {
+impl From<StratBlockDev> for Vec<DeviceInfo> {
     fn from(bd: StratBlockDev) -> Self {
+        fn is_valid(blockdev: &StratBlockDev) -> StratisResult<(bool, bool)> {
```
mulkieran, line 181: I think for this is_valid() method it would be helpful to put in a header comment that documents what it is guarding against by using the blkid probes.

mulkieran (changes requested):
This seems to me to hang together very well. I have a few requests for comments, but that's it.

mulkieran:
Addressing a problem with behavior when some filesystems can be torn down but not others.

jbaublitz:
@mulkieran I think the most recent commit should resolve that.

mulkieran (commented):
I _think_ it would be better if the teardown of a thin device was also idempotent. When we stop pools, removal is handled correctly, which is good.  But we may find another way to get ourselves into the same phantom filesystem problem that we had beforee. I think that tearing down filesystems _could_ use the same approach as teardown for the other devices, and be made idempotent that way.

mulkieran:
> I _think_ it would be better if the teardown of a thin device was also idempotent. When we stop pools, removal is handled correctly, which is good. But we may find another way to get ourselves into the same phantom filesystem problem that we had beforee. I think that tearing down filesystems _could_ use the same approach as teardown for the other devices, and be made idempotent that way.

I would also be willing to defer consideration of this proposal until later. It may not gain us much during any remotely normal operation.

mulkieran (approved):

PR #3288, opened 2023-03-17
Emit correct D-Bus signals when blockdevs are added

### LINKED ISSUE #3267: D-Bus signal not properly sent when devices are added
The root cause is https://github.com/stratis-storage/stratisd/pull/3223, because it removes the pool diffing all together.

mulkieran:
Closes #3267

### DIFF
--- a/src/engine/engine.rs
+++ b/src/engine/engine.rs
@@ -28,7 +28,7 @@ use crate::{
             MappingCreateAction, MappingDeleteAction, Name, PoolDiff, PoolEncryptionInfo,
             PoolIdentifier, PoolUuid, RegenAction, RenameAction, ReportType, SetCreateAction,
             SetDeleteAction, SetUnlockAction, StartAction, StopAction, StoppedPoolsInfo,
-            StratFilesystemDiff, StratPoolDiff, UdevEngineEvent, UnlockMethod,
+            StratFilesystemDiff, UdevEngineEvent, UnlockMethod,
         },
     },
     stratis::StratisResult,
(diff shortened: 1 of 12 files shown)


mulkieran:
dbus-monitor tests may succeed

mulkieran:
Only one test is failing, ```test_pool_add_data_after_cache```. It is not yet clear why.

mulkieran:
There's a bug in the test. It re-adds the same data device as it used to make the pool originally.

mulkieran:
And with what should have been no semantic change the tests are all passing. It may be that the test failure was an anomaly, caused by the final change signal not being caught. If that is the case, though, we may have to move the tests back to nightly, since we don't want that volatility.

mulkieran:
@jbaublitz If the dbus-monitor test has even one transient failure, the final commit should be remove and the test left in nightly. Otherwise, I think this is ready.

src/engine/engine.rs
```diff
@@ -465,6 +465,9 @@ pub trait StateDiff {
     /// the new_state argument as this method should always return the new values
     /// for any properties that are inconsistent.
     fn diff(&self, new_state: &Self) -> Self::Diff;
+
+    /// Return all contained values unchanged.
+    fn unchanged(&self) -> Self::Diff;
```
jbaublitz, line 470: Is this for optimization purposes? To avoid running unnecessary calculations?

jbaublitz (commented):
Just one question.

src/engine/engine.rs
mulkieran: Because diffing cached() with cached() looks a bit silly: https://github.com/stratis-storage/stratisd/pull/3288/commits/7cd97525c0daeeec16919c105908705f4ad3f805#diff-39f0bed1717b84fc076fa34b429732430038740d516da7e9b39cf963aec11318R918 .

jbaublitz: That makes sense!

jbaublitz (approved):

mulkieran:
rebased.

mulkieran:
unrelated flaky test failure.

Issue #3290, opened 2023-03-21
Checking consistent block size on existing pools


mulkieran:
We enforce a rule on consistent block sizes somewhat stringently. Unfortunately, there is a case where our choice of how we compare causes a problem. When a pool is encrypted, we compare the block sizes of the opened devices to the block sizes of the devices that have been presented for adding to a pool. However, when we encrypt, cryptsetup may change the logical block size of the device, generally to increase it to match the physical block size. So, even though the devices in the pool and the devices being presented to be added have exactly the same block sizes, stratisd will reject the devices.

Note that this problem will disappear if crypt device is moved away from the block devices as we intend w/ our design. So, this is sort of a temporary problem, although it still has to be fixed.

The modified approach can be the following:

For the devices to be added, we only know what their block sizes are now, not what their block sizes will be after cryptsetup is applied to them. So, when pool is encrypted, we must read the block sizes on the opened devices and determine if the encryption would be able to make the devices of compatible size. Likely the rule should be that we can expect that encryption could increase the logical block size to match the physical block size of the devices being added, if necessary to make a match. So, we allow block devices that would allow that, but then we enforce that cryptsetup makes the logical block size to match, and if, for some reason it fails, we do a rollback using existing functionality.

Related bz: https://bugzilla.redhat.com/show_bug.cgi?id=2170318

mulkieran:
So, the first step is to read the block sizes from the base devices, and not from the opened cryptsetup devices.

mulkieran:
However, we need to store in the StratisBlockdev struct not just the block sizes of the base device, but also the block  sizes of the opened crypt device if any.

```
StratBlockSizes {
   physical: BlockSizes,
   luks: Option<BlockSizes>,
}
```

when adding devices, we must use the sector_size which has been used in the existing crypt devices to enforce the compatible logical block size for the crypt device.

PR #3291, opened 2023-03-22
Incorporate information about crypt devices into sector size calculations

mulkieran:
Related #3290

### DIFF
--- a/src/engine/strat_engine/backstore/backstore.rs
+++ b/src/engine/strat_engine/backstore/backstore.rs
@@ -238,6 +238,7 @@ impl Backstore {
             devices,
             mda_data_size,
             encryption_info,
+            None,
         )?);
 
         Ok(Backstore {
@@ -262,6 +263,7 @@ impl Backstore {
         pool_name: Name,
         pool_uuid: PoolUuid,
         devices: UnownedDevices,
+        sector_size: Option<u32>,
     ) -> StratisResult<Vec<DevUuid>> {
         match self.cache_tier {
             Some(_) => unreachable!("self.cache.is_none()"),
@@ -280,6 +282,7 @@ impl Backstore {
                         .map(EncryptionInfo::try_from)
                         .transpose()?
                         .as_ref(),
+                    sector_size,
                 )?;
 
                 let cache_tier = CacheTier::new(bdm)?;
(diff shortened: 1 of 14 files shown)


mulkieran:
Need to separate validation for purposes of pool availability from information extraction for purposes of specifying sector size.

mulkieran:
Should we consider the relation of the logical sector size of the encrypted devices on a newly added cache to the logical sector sizes of the encrypted devices in the data tier? I feel that we have to.

mulkieran:
Real error detected in python-based tang tests.

In test_binding_and_adding it looks like the pool is created properly, and the pool is bound properly, but then, when devices are added, things get very weird.

```
[2023-04-05T00:21:07Z DEBUG stratisd::engine::strat_engine::backstore::devices] Verifying device using blkid probe: superblock probe: None, number of partitions: None
[2023-04-05T00:21:07Z DEBUG stratisd::engine::strat_engine::backstore::devices] Verifying device using blkid probe: superblock probe: None, number of partitions: None
[2023-04-05T00:21:09Z ERROR stratisd::engine::strat_engine::backstore::crypt::shared] Device tang does not exist or access denied.
[2023-04-05T00:21:09Z WARN  stratisd::engine::strat_engine::backstore::crypt::initialize] Failed to format device /dev/loop5 with LUKS2 header; failed with error: IO error occurred: Cannot allocate memory (os error 12)
[2023-04-05T00:21:09Z ERROR stratisd::engine::strat_engine::backstore::crypt::shared] Cannot determine device type. Incompatible activation of device?
[2023-04-05T00:21:09Z ERROR stratisd::engine::strat_engine::backstore::crypt::shared] This operation is supported only for LUKS2 device.
```

"Device tang does not exist or access denied." is surprising.
Then the problem "Cannot allocate memory" for failure to format device error is also strange.

I feel like for some reason non-Some ```luks2_params``` parameter is actually being passed to libcryptsetup format() method and something is going wrong with the way the crypt_format() call is invoked.

mulkieran:
Apparent success.

mulkieran:
I don't think there's any point to passing the sector_size as Bytes instead of u32, even in the upper layers. So, e.g., ```Backstore::add_datadevs()``` just takes an ```Option<u32>``` and passes it all the way down to crypt.rs, where it is incorporated into the CryptParamLuks2 struct, if present.

PR #3315, opened 2023-04-14
Run with --verify-devices set

mulkieran:


mulkieran:
This is only a test...

PR #3326, opened 2023-04-19
test

mulkieran:


mulkieran:
Do get this to work, we'll have to move test_harness.py into stratisd test directory and modify it a bit.

PR #3329, opened 2023-04-20
On UnlockPool, start the pool directly instead of relying on catching udev events

### LINKED ISSUE #3324: org.storage.stratis3.Manager.r1 UnlockPool does nothing anymore
Cockpit is still using the "r0" API level of org.storage.stratis3, but since Stratis 3.5.3 the UnlockPool method of the manager doesn't seem to actually unlock the pool.

```
# stratis pool list --stopped
Name    UUID                                   # Devices   Key Description   Clevis
pool0   2e609392-28d8-4f75-9e39-b3197c5ccf44           2   pool0             N/A   

# stratis key set pool0 --capture-key
Enter key data followed by the return key:

# busctl call org.storage.stratis3 /org/storage/stratis3 org.storage.stratis3.Manager.r0 UnlockPool ss 2e609392-28d8-4f75-9e39-b3197c5ccf44 keyring
(bas)qs true 2 "e7244a9d2516457ca8b44e92b570e23d" "26923cafb76c4386a5590e41555972c3" 0 "Ok"

# stratis pool list
Name   Total / Used / Free   Properties   UUID   Alerts

# stratis pool list --stopped
Name    UUID                                   # Devices   Key Description   Clevis
pool0   2e609392-28d8-4f75-9e39-b3197c5ccf44           2   pool0             N/A   

# busctl call org.storage.stratis3 /org/storage/stratis3 org.storage.stratis3.Manager.r2 StartPool 's(bs)' 2e609392-28d8-4f75-9e39-b3197c5ccf44 1 keyring
(b(oaoao))qs true "/org/storage/stratis3/0" 2 "/org/storage/stratis3/1" "/org/storage/stratis3/2" 1 "/org/storage/stratis3/3" 0 "Ok"

# stratis pool list
Name               Total / Used / Free    Properties                                   UUID   Alerts
pool0   7.97 GiB / 1.05 GiB / 6.92 GiB   ~Ca, Cr, Op   2e609392-28d8-4f75-9e39-b3197c5ccf44   WS001
```

As you can see, after calling r0.Unlock, the pool is still listed as stopped.  Calling r2.StartPool works as expected.

I guess Cockpit should catch up to at least "r2", but this looks like a unintended regression to me.

mulkieran:
Supercedes #3328 
Closes #3324

### DIFF
--- a/src/engine/strat_engine/backstore/devices.rs
+++ b/src/engine/strat_engine/backstore/devices.rs
@@ -112,7 +112,7 @@ pub fn get_devno_from_path(path: &Path) -> StratisResult<Device> {
 /// because they will not have the opportunity to be processed.
 pub fn find_stratis_devs_by_uuid(
     pool_uuid: PoolUuid,
-    uuids: Vec<DevUuid>,
+    uuids: &[DevUuid],
 ) -> StratisResult<HashMap<DevUuid, (DevicePath, Device)>> {
     let mut map = HashMap::new();
     if uuids.is_empty() {
--- a/src/engine/types/actions.rs
+++ b/src/engine/types/actions.rs
@@ -221,59 +221,82 @@ impl Display for MappingDeleteAction<Key> {
     }
 }
 
-/// A type for the return type of idempotent unlocking actions.
-pub struct SetUnlockAction<T> {
-    unlocked: Vec<T>,
+/// A type for the return type of idempotent unlocking actions on an
+/// individual pool.
+pub enum SetUnlockAction<T> {
+    /// If the pool was already started.
+    Identity,
+    /// If the unlock action resulted in the pool being started.
+    Started(Vec<T>),
 }
 
 impl<T> SetUnlockAction<T> {
-    /// Create a new return type with newly unlocked resources and resources that
-    /// are still locked.
+    /// Create a new SetUnlockAction where the pool is started.
     pub fn new(unlocked: Vec<T>) -> SetUnlockAction<T> {
-        SetUnlockAction { unlocked }
+        SetUnlockAction::Started(unlocked)
     }
 
-    /// Create a new return type where no newly unlocked resources are reported.
+    /// Create a new SetUnlockAction when no newly unlocked resources are
+    /// reported but a pool is started as a consequence.
     pub fn empty() -> SetUnlockAction<T> {
-        SetUnlockAction {
-            unlocked: Vec::new(),
-        }
+        SetUnlockAction::Started(Vec::new())
+    }
+
+    ///  Create a new SetUnlockAction where the pool was already set up.
+    pub fn identity() -> SetUnlockAction<T> {
+        SetUnlockAction::Identity
     }
 }
 
 impl<T> EngineAction for SetUnlockAction<T> {
     type Return = Vec<T>;
 
+    /// true if a thing was unlocked.
     fn is_changed(&self) -> bool {
-        !self.unlocked.is_empty()
+        match self {
+            SetUnlockAction::Identity => false,
+            SetUnlockAction::Started(unlocked) => !unlocked.is_empty(),
+        }
     }
 
     fn changed(self) -> Option<Vec<T>> {
-        if self.unlocked.is_empty() {
-            None
-        } else {
-            Some(self.unlocked)
+        match self {
+            SetUnlockAction::Identity => None,
+            SetUnlockAction::Started(unlocked) => {
+                if unlocked.is_empty() {
+                    None
+                } else {
+                    Some(unlocked)
+                }
+            }
         }
     }
 }
 
 impl Display for SetUnlockAction<DevUuid> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        if self.unlocked.is_empty() {
-            write!(
-                f,
-                "No new devices were able to be unlocked; no action was taken"
-            )
-        } else {
-            write!(
-                f,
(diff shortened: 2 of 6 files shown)


bgurney-rh (approved):

src/engine/types/actions.rs
```diff
@@ -221,59 +221,82 @@ impl Display for MappingDeleteAction<Key> {
...
-/// A type for the return type of idempotent unlocking actions.
-pub struct SetUnlockAction<T> {
-    unlocked: Vec<T>,
+/// A type for the return type of idempotent unlocking actions on an
+/// individual pool.
+pub enum SetUnlockAction<T> {
+    /// If the pool was already started.
+    Identity,
+    /// If the unlock action resulted in the pool being started.
+    Started(Vec<T>),
```
jbaublitz, line 230: In general, I don't like designing data structures with two states that are essentially equivalent to `None` values. We actually had an issue up about designing the `Vec<_>`-based action structures this way, and I had a problem with it back then for this reason. I feel like having a single state that represents the `None` or empty value is always better for code readability reasons. I think this would be better if it followed the format of the other `Vec`-based action types.

src/engine/strat_engine/backstore/devices.rs
```diff
@@ -112,7 +112,7 @@ pub fn get_devno_from_path(path: &Path) -> StratisResult<Device> {
 /// because they will not have the opportunity to be processed.
 pub fn find_stratis_devs_by_uuid(
     pool_uuid: PoolUuid,
-    uuids: Vec<DevUuid>,
+    uuids: &Vec<DevUuid>,
```
jbaublitz, line 115: Cippy has historically told me to write this as `&[DevUuid]` and not `&Vec<DevUuid>`. I'm not sure why it's not complaining now.

jbaublitz (changes requested):
Just two notes, but the design looks a lot better to me from a standpoint of returning the unlocked UUIDs.

bgurney-rh (approved):

src/engine/strat_engine/backstore/devices.rs
mulkieran: Done.

src/engine/types/actions.rs
mulkieran: I recall the conversation. But here, this seemed best, because 1) the type returned by ```StratEngine::unlock_pool()``` must implement Display for the purposes of logging by the handle_action macro and 2) in the dbus_api ```unlock_pool()``` method that is being modified, it is necessary to distinguish between no devices unlocked because the pool is set up and no devices unlocked when the pool is not set up because it is necessary to know whether to modify the D-Bus properties or not.

There are ways to address this, like have ```StratEngine::unlock_pool``` return a tuple of say a bool and the unmodified SetUnlockAction struct and define Display for the tuple. But that doesn't seem like a better solution to me.

Another thing to do would be to skip the pre-check for whether the pool exists in ```StratEngine::unlock_pool()```.  In that case, though, even if a pool with the UUID exists, stratisd will end up trying to unlock and if succesfull set up the pool, even though it has the same UUID as an existing pool.

What about returning an Err in ```StratPool::unlock_pool()``` if another pool with the same UUID already exists? That is reasonably backward compatible and allows reverting the changes to  ```SetUnlockAction```. Because in previous versions of stratisd on UnlockPool, the udev events would be caught, stratisd would try to set up the pool, and then it would log a warning on failure in the background.

mulkieran: That last suggestion isn't all that good, because it is only relevant in the highly unusual case where devices with the UUID exist in stopped pools _and_ there is a set up pool with the same UUID. Where there already is a set up pool _and_ there are no devices to unlock in the stopped pools, stratisd would previously return Ok with 0 unlocked UUIDs.

jbaublitz: I'm a little bit confused here. My understanding of the code is that an error is always returned if we attempt to start a pool and no devices are unlocked or even if some are unlocked and rolled back and the pool is not set up. I don't really understand why we need to differentiate between those two cases in the success (`Ok`) case mainly because I think the only time we'll get `Ok(_)` and no newly unlocked devices is if the pool is already set up. Can you explain the case you're thinking of in a little more detail where this distinction is necessary?

mulkieran: @jbaublitz We could get ```<empty list>``` in the case where the pool is already started _and_ in the case where previously teardown failed to teardown any crypt devices (somehow) but on calling ```UnlockPool``` the pool is now started. In prior stratisd, in the second case, no attempt would be made to start the pool, because no udev events would occur (I think). In current situation, an attempt will be made to start the pool even if all devices are currently unlocked when ```PoolUnlock``` is called and the pool may be started.

We could modify behavior of start_pool() based on calling context such that it exits early, without actually starting the pool, with Ok if no devices are unlocked and it was invoked in unlock context.

In that case, no attempt would be made to start the pool and ```Ok(<empty list>)``` would always signify pool not started while ```Ok(<non-empty-list>)``` would always signify pool started and ```SetUnlockAction``` could keep its original definition.

jbaublitz: Okay! I think I understand what you're getting at now. Let me review this one more time. I may not ask for any changes and approve it now that I'm a little more clear on what you're getting at. I think you're right about this distinction being important and I can see why that case would not count as the identity portion of things if the pool is still newly started even if no new devices are unlocked.

jbaublitz (approved):

mulkieran:
Ok. Gonna squash the whole thing.

PR #3337, opened 2023-05-04
dracut: remove dependency on plymouth

mberndt123:
I'm adding stratis root fs support on the NixOS distribution, and they asked me to make it work without plymouth, so I've replaced that with systemd-ask-password.

### DIFF
--- a/dracut/90stratis/stratis-rootfs-setup
+++ b/dracut/90stratis/stratis-rootfs-setup
@@ -19,10 +19,9 @@ if $(stratis-min pool is-stopped "$STRATIS_ROOTFS_UUID"); then
 	if $(stratis-min pool is-encrypted "$STRATIS_ROOTFS_UUID"); then
 		ATTEMPTS_REMAINING=3
 		if
-			! while [ $((ATTEMPTS_REMAINING--)) -gt 0 ]
-			do
+			! while [ $((ATTEMPTS_REMAINING--)) -gt 0 ]; do
 				systemd-ask-password --id="stratis:$STRATIS_ROOTFS_UUID" "Enter password for Stratis pool with UUID $STRATIS_ROOTFS_UUID containing root filesystem" |
-				stratis-min pool start --prompt --unlock-method=keyring "$STRATIS_ROOTFS_UUID" && break
+					stratis-min pool start --prompt --unlock-method=keyring "$STRATIS_ROOTFS_UUID" && break
 			done
 		then
 			echo Failed to start pool with UUID $STRATIS_ROOTFS_UUID using a passphrase >&2
--- a/dracut/90stratis/stratis-rootfs-setup
+++ b/dracut/90stratis/stratis-rootfs-setup
@@ -17,10 +17,14 @@ done
 
 if $(stratis-min pool is-stopped "$STRATIS_ROOTFS_UUID"); then
 	if $(stratis-min pool is-encrypted "$STRATIS_ROOTFS_UUID"); then
-		if ! plymouth ask-for-password \
-			--command="stratis-min pool start --prompt --unlock-method=keyring $STRATIS_ROOTFS_UUID" \
-			--prompt="Enter password for Stratis pool with UUID $STRATIS_ROOTFS_UUID containing root filesystem" \
-			--number-of-tries=3; then
+		ATTEMPTS_REMAINING=3
+		if
+			! while [ $((ATTEMPTS_REMAINING--)) -gt 0 ]
+			do
+				systemd-ask-password --id="stratis:$STRATIS_ROOTFS_UUID" "Enter password for Stratis pool with UUID $STRATIS_ROOTFS_UUID containing root filesystem" |
+				stratis-min pool start --prompt --unlock-method=keyring "$STRATIS_ROOTFS_UUID" && break
+			done
+		then
 			echo Failed to start pool with UUID $STRATIS_ROOTFS_UUID using a passphrase >&2
 			exit 1
 		fi
--- a/dracut/90stratis/module-setup.sh
+++ b/dracut/90stratis/module-setup.sh
@@ -12,7 +12,6 @@ check() {
 		xfs_growfs \
 		xfs_db \
 		udevadm \
-		/usr/bin/systemd-ask-password \
 		/usr/sbin/thin_metadata_size \
 		/usr/lib/udev/stratis-str-cmp \
 		/usr/lib/udev/stratis-base32-decode ||
@@ -22,7 +21,7 @@ check() {
 
 # called by dracut
 depends() {
-	echo dm
+	echo dm systemd-ask-password
 	return 0
 }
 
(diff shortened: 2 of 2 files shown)


jbaublitz:
Hey @mberndt123! I remember originally trying this when I was developing this work, but couldn't quite remember what went wrong and why I opted for plymouth. I've set up an environment to test this and on Fedora, the problem I'm seeing is that `systemd-ask-passphrase` returns an error (errno 6, No such device or address) three times and then the boot fails to start. Because this change would break Fedora, I'd say we're definitely open to accepting this change if you can find a way to make it work on Fedora too.

mberndt123:
Hey @jbaublitz,

I'm using Fedora myself and that exact script works fine on my machine. It might be the case that some other dracut module on my system pulls in functionality that is required for systemd-ask-password to function, and that module is not present on your system.
Dracut comes with a `01systemd-ask-password` module, so I've added a dependency on that in `module-setup.sh` (and also removed some plymouth leftovers that shouldn't be required any longer). Can you try it again and see if it works now?

jbaublitz:
That doesn't resolve the issue for me. Just so you're aware, systemd-ask-passphrase defaults to using plymouth for passphrase prompts. Are you doing something different on your system?

mberndt123:
I can't think of anything I'm doing differently. I've generated initramfs images both with and without the plymouth module and either works fine. Without plymouth it just shows a password prompt in the console.

Do you have any suggestions how I might be able to reproduce this?

mberndt123:
I've tried to disable as many dracut modules as reasonably possible…
```
$ lsinitrd initrd 
Image: initrd: 23M
========================================================================
Version: dracut-059-3.fc38

Arguments:  --omit 'plymouth crypt dbus-broker dbus ifcfg network network-manager nss-softokn i18n kernel-modules-extra rootfs-block terminfo usrmount fs-lib memstrack' -v

dracut modules:
systemd
systemd-ask-password
systemd-initrd
dm
kernel-modules
stratis
udev-rules
dracut-systemd
base
shutdown
========================================================================
```
… but the generated initramfs can still boot my system just fine.

Can you provide a virtual machine to reproduce this with?

jbaublitz:
@mberndt123 Please see https://stratis-storage.github.io/stratify/ for an overview of how I'm setting up the VMs that I'm using. After that, I check out your git repo, run `make build-all`, and then `sudo make install`. Can you try that process and see if it's reproducible for you?

mberndt123:
Thanks @jbaublitz. It's a bit late here to try that now (00:27), but I hope I'll get around to it tomorrow.

mberndt123:
Did you use Fedora 37 or 38 for that? I'm on Fedora 38.

jbaublitz:
I'm using Fedora 38 too.

jbaublitz:
Hi @mberndt123! I've done a little bit more debugging and this may just be triggering a bug in our code. Part of the reason I believe this is the case is that when I pull your commits in before a commit that reworked the crypt module rather heavily, booting is successful. Futhermore, after this commit, I'm seeing some really bizarre behavior in the recovery console. Let me see if I can track down exactly what's going on before you invest any more time in this.

jbaublitz:
I've confirmed that the rpassword update is what breaks `systemd-ask-passphrase`! I will get more information on what we can do once I talk with @mulkieran.

mberndt123:
Wow, good news! I was testing this change by simply copying those shell scripts to `/usr/lib/dracut/modules.d/90stratis` and regenerating the initramfs, whereas you did things properly and installed the complete branch, including the broken rpassword. So it makes sense we got different results.

Maybe a solution is to have `systemd-ask-password` store the password in the kernel keyring and have Statis read it from there, then you don't need rpassword at all.

jbaublitz:
I've been working on #3347 and once that's merged, I'll probably ask for some PR revisions around error handling in your PR, but other than that, I think I've got this working!

jbaublitz:
Okay, I've tested #3347 and fixed some bugs. I'm going to leave a review now.

dracut/90stratis/stratis-rootfs-setup
```diff
@@ -17,10 +17,14 @@ done
...
-			--prompt="Enter password for Stratis pool with UUID $STRATIS_ROOTFS_UUID containing root filesystem" \
-			--number-of-tries=3; then
+		ATTEMPTS_REMAINING=3
+		if
+			! while [ $((ATTEMPTS_REMAINING--)) -gt 0 ]
+			do
+				systemd-ask-password --id="stratis:$STRATIS_ROOTFS_UUID" "Enter password for Stratis pool with UUID $STRATIS_ROOTFS_UUID containing root filesystem" |
+				stratis-min pool start --prompt --unlock-method=keyring "$STRATIS_ROOTFS_UUID" && break
+			done
+		then
```
jbaublitz, line 27: In general, I've been unable to find any documentation on what a while loop actually evaluates to in bash. I think this should be restructured to have the while loop first, setting `$RC` to one or zero based on `$?` after the piped command is run, and then the if statement should come after, checking the value of `$RC`.

jbaublitz (changes requested):

dracut/90stratis/stratis-rootfs-setup
mberndt123: How a while loop evaluates is documented in the bash manual:
https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#index-while
That said, I've made the change you've proposed. It made the code a little longer and `$RC` is checked in two different places, but if it helps you make sense of the code better, then why not…

jbaublitz: The documentation is sufficient. If you want to drop that last commit, I've tested it and it works. I'm willing to approve just the first two commits.

jbaublitz:
@mberndt123 Okay, so we have tracked down all of the problems we've seen with this code and one of them appears to be https://github.com/stratis-storage/devicemapper-rs/issues/856. For some reason, your code seems to trigger this bug in our code, so we'll likely want to fix this first, but as stated above, I'll approve this PR after you drop the most recent commit, and then we'll probably wait on fixing that problem with our semaphore handling.

mberndt123:
Wow, that's crazy that a change like this can uncover that sort of bug!

jbaublitz (approved):

mulkieran (approved):

mulkieran:
@mberndt123 I'm afraid the CI doesn't like your fomatting. Plz run ```make fmt-shell``` in the top-level directory and fix up that which needs to be fixed up. Thanks for the PR!

mulkieran:
@jbaublitz Have we merged all the pre-requisiites and is it fine to merge this now?

mberndt123:
Actually while I was reading this script, I thought about whether polling is really the optimal solution here. Instead of calling `stratis-min pool is-stopped "$STRATIS_ROOTFS_UUID"` repeatedly until the pool shows up, we could have a new command like `stratis-min pool wait "$STRATIS_ROOTFS_UUID"` that will block until that pool shows up or a timeout expires. I could implement that if you think it's a good idea, obviously in a PR separate from this one.

Which brings me to another question: where can I get in touch with the Stratis team to discuss this kind of proposal? Because on [stratis-storage.github.io](https://stratis-storage.github.io) I can only see links to GitHub and Twitter. Should I just raise an issue on GitHub to discuss that?

mulkieran:
> Actually while I was reading this script, I thought about whether polling is really the optimal solution here. Instead of calling `stratis-min pool is-stopped "$STRATIS_ROOTFS_UUID"` repeatedly until the pool shows up, we could have a new command like `stratis-min pool wait "$STRATIS_ROOTFS_UUID"` that will block until that pool shows up or a timeout expires. I could implement that if you think it's a good idea, obviously in a PR separate from this one.
> 
> Which brings me to another question: where can I get in touch with the Stratis team to discuss this kind of proposal? Because on [stratis-storage.github.io](https://stratis-storage.github.io) I can only see links to GitHub and Twitter. Should I just raise an issue on GitHub to discuss that?

Raising an issue on GitHub is really the best for us.

mberndt123:
Any updates on this?

jbaublitz:
@mberndt123 We're just waiting on #3367.

mulkieran:
@mberndt123 Plz rebase at your convenience and we should be able to merge. Thanks!

mberndt123:
@mulkieran ok I've rebased it

mulkieran:
thanks!

mberndt123:
@mulkieran @jbaublitz thank you!

PR #3352, opened 2023-05-25
Return an error if there is no unlock method

mulkieran:


jbaublitz:
I just wanted to mention that maybe we should also handle the case (although less likely to be triggered) in `handle.rs` too.

mulkieran:
There seems to be a bit of a bug in that, the pool is still started and fully running. If stratisd is stopped and restarted, the pool should appear in either stopped or started pools, but appears in neither.

src/engine/strat_engine/liminal/device_info.rs
```diff
@@ -669,17 +669,22 @@ impl DeviceSet {
...
+    /// uncertainty, since the error means that encryption information is
+    /// completely missing from one device.
+    pub fn stopped_pool_info(&self) -> StoppedPoolInfo {
+        let info = gather_encryption_info(
             self.internal.len(),
             self.internal.values().map(|info| info.encryption_info()),
         )
-        .ok()
-        .map(|info| StoppedPoolInfo {
+        .unwrap_or({
```
jbaublitz, line 680: I don't think this does the right thing. Errors are only returned when they're a mixture of encrypted and unencrypted devices. I think this aspect of the code should be handled in `engine/shared.rs` in the gather methods. Gather takes a function as a parameter so it can likely be modified to calculate whether it's `PoolEncryption::KeyDesc(MaybeInconsistent::Yes)`, etc.

jbaublitz (changes requested):

src/engine/strat_engine/liminal/device_info.rs
mulkieran: When the two arguments to gather_encryption_info are constructed, the iterator derived from internal.values() returns ```None``` in one place, so that the flattened length is less than the unflattened.

I think that what you're saying is that when the information is being gathered, it is possible to determine what is the encryption info on the devices w/ valid encryption info.

So, if all the devices w/ valid LUKS metadata are KeyDescription only, then you would like to see ```PoolEncryptionInfo::KeyDesc(MaybeInconsistent::Yes)``` rather than both showing inconsistent?

I think we can do that, but my own position is that in the general case where the LUKS info is actually missing because invalid, that information is not certainly known, so I decided to use the most confused value.

If all the devices had valid LUKS2 metadata, and each had a key description, but some key descriptions were different from others, then I would think that it would be appropriate to return
`PoolEncryptionInfo::KeyDesc(MaybeInconsistent::Yes)```,  but it actually make more sent to me to return ```Both``` as I do here, in this situation.

mulkieran:
I don't think that ```gather_encryption_info``` should return an error under any circumstances. At present, ```encryption_info()``` on one device in a ```DeviceSet``` may return None. If other return ```Some```,  then the answer is for the ```PoolEncryptionInfo``` to be maximally inconsistent, as it is here. But, that can be done in in the ```gather_encryption_info``` method. I'll give that a go. How that relates to the related method ```gather_pool_name``` is an interesting question. As ever, I'ld like to do this with a set and some rules. 

If exactly one element of the set and that is ```None```, then the pool encryption info is None.

If two elements of the set, then something is inconsistent.
None + anything => maximally inconsistent, since there should be some encryption info on that other device and we have no idea what it is.

Then feed the rest to the from method for PoolEncryptionInfo, as before.

PR #3362, opened 2023-06-12
Run with prior release

### LINKED ISSUE #3325: Run old versions of testing repo on newer versions of stratisd
For HEAD revision of stratisd, we can try running older versions of testing repo in nightly.

mulkieran:
Closes #3325


mulkieran:
Older revision will still fail, because it checks r6 properties.

bgurney-rh (approved):

PR #3369, opened 2023-06-28
Use shadow-rs to supply build information of Stratis

mulkieran:


mulkieran:
This PR just existed to demonstrate that you could insert information, like the Rust toolchain version, into the stratisd executable, such that you could also use the command-line to ask what the version was. There's a bunch of miscellanous information, like the git hash of the commit, and I'm sure it all works great, but we likely won't pursue using it.

PR #3379, opened 2023-07-10
Add dummy IPC mechanism when neither IPC mechanism is enabled

### LINKED ISSUE #3105: Might want to have a third alternative that compiles in neither IPC mechanism

jbaublitz:
Related to testing for #3274

Closes #3105

### DIFF
--- a/.github/workflows/fedora.yml
+++ b/.github/workflows/fedora.yml
@@ -42,6 +42,9 @@ jobs:
           - task: PROFILEDIR=debug make -f Makefile build-min
             toolchain: 1.70.0  # CURRENT DEVELOPMENT RUST TOOLCHAIN
             components: cargo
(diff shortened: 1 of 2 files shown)


mulkieran (changes requested):
Code is fine, AFAICT. Could you review and fix comments and error messages in src/stratis/ipc_support/dummy.rs? They all relate to JSON IPC mechanism.

jbaublitz:
Okay, this should be ready.

mulkieran (approved):

PR #3380, opened 2023-07-10
Do not send out signal on UserInfo change before r3 

mulkieran:
Related https://github.com/stratis-storage/testing/pull/189


mulkieran:
This should pass all tests, because the D-Bus monitor script should detect that for revisions 0-2 inclusive, UserInfo does not emit a changed signal (according to the introspection data).

mulkieran:
This is trivial unless it fails some test in an interesting way.

Issue #3353, opened 2023-05-30
Paradox when pool is in stopped pools data structure but is fully started


mulkieran:
We can engineer this state by:

1. Corrupting Stratis LUKS metadata so that a device is recognized as a Stratis device but has no valid encryption information to unlock a pool.
2. Restarting stratisd.

stratisd detects that there is no valid encryption information and puts the pool into stopped pools data structure. But, because the pool was not taken down previously, it is fully started and functional. If it were torn down, then it would not be possible to just restart it, due to the invalid LUKS info. It seems like it would not help the user to tear it down, on the other hand, stratisd is reporting the fully functional pool as stopped.

PR #3381, opened 2023-07-11
Bug fix for liminal device code

jbaublitz:


mulkieran (commented):
This looks correct to me, but I'm worried that it might have a bad consequence, due to the issue discussed here: https://github.com/stratis-storage/stratisd/issues/3353 . The consequence would be that there is a pool that is setup but has no mounted filesystems. But the LUKS metadata gets corrupted. The pool cannot be started and all the devices are torn down and now the pool is basically impossible to set up in the normal way and may be entirely unrecoverable. That may very well be better than the alternatives, but I'ld like to introduce the possibility for your consideration.

jbaublitz:
I think there are arguments on both ends. If we do tear down the pool, we could permanently lose data. If we don't, we could leave the devices in an unlocked state unintentionally which could leave sensitive data available to an attacker. I think data loss may be more of a concern here and I'm happy to simply default to putting the partially constructed pool into the partially constructed pool slot to be torn down through `stratis pool stop` later given that we now have the ability to do that through stratisd. What are your thoughts?

mulkieran:
> I think there are arguments on both ends. If we do tear down the pool, we could permanently lose data. If we don't, we could leave the devices in an unlocked state unintentionally which could leave sensitive data available to an attacker. I think data loss may be more of a concern here and I'm happy to simply default to putting the partially constructed pool into the partially constructed pool slot to be torn down through `stratis pool stop` later given that we now have the ability to do that through stratisd. What are your thoughts?

That seems like the best choice that we have available.

jbaublitz:
I think this is actually a real bug. I need to fix the tests.

jbaublitz:
Blocked by #3385

jbaublitz:
Merged in #3389

PR #3382, opened 2023-07-11
Start running long unused test in CI again

mulkieran:


mulkieran:
Maybe it is better to restore the test...we can see.

jbaublitz (approved):

PR #3389, opened 2023-07-20
Tidy up engine handling and D-Bus interface management for partially stopped pools

### LINKED ISSUE #3388: object paths related to partially torn down pool not being removed
LiminalDevices::stop_pool() returns an Err if the stop did not complete. 

StratEngine::stop_pool() passes the error through.

D-Bus stop_pool implementations. both, receive the Err only, return the error message on the D-Bus and do not send a message to update the D-Bus layer.

Confirmed in stratisd 3.5.7 and in https://github.com/stratis-storage/stratisd/pull/3381 .

IMPORTANT note: ```stratis pool list``` has a disagreement with ```busctl tree```. ```busctl tree``` shows the extra object paths after the partial teardown, but ```stratis pool list``` does not show a pool when the pool is partially torn down. After the pool is started, suddenly ```stratis pool list``` has entries for two identical pools. This is explicable by the fact that our GetManagedObjects implementation must call get_data() for the object path to get the UUID and then do a lookup on the UUID in the engine. This fails, so the pool properties result is none, so there is no entry for the object path corrresponding to that pool in the GetManagedObjects result. When the pool is setup, new object paths are added for the pool and its devices, and the pool is now in the pools data structure, so there are now duplicate entries in the stratis-cli result.

mulkieran:
Closes #3388

### DIFF
--- a/src/dbus_api/api/manager_3_2/methods.rs
+++ b/src/dbus_api/api/manager_3_2/methods.rs
@@ -164,32 +164,40 @@ pub fn stop_pool(m: &MethodInfo<'_, MTSync<TData>, TData>) -> MethodResult {
     })
     .unwrap_or(false);
 
-    let msg = match handle_action!(block_on(
+    let action = handle_action!(block_on(
         dbus_context
             .engine
             .stop_pool(PoolIdentifier::Uuid(pool_uuid), false)
-    )) {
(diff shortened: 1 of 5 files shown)


bgurney-rh (approved):

jbaublitz (approved):
I think this looks good.

PR #3413, opened 2023-08-23
stratisd_proc_macros v0.2.1

mulkieran:


mulkieran:
test failure needs investigation...

mulkieran:
Updating Cargo.lock caused the test failure to go away. This is strange to me. What does the build run against if Cargo.lock is not updated?

PR #3427, opened 2023-08-28
Increase serde_derive dependency lower bound to 1.0.185

mulkieran:
Related https://github.com/stratis-storage/project/issues/656


mulkieran:
This update will allow us to drop all our serde_derive related patches.

mulkieran:
Actually, I think this is completely obviously correct, hence trivial.

PR #3437, opened 2023-09-13
V2 of stratis-fstab-setup to order setup before fsck

jbaublitz:
Related to #3436

### DIFF
--- a/systemd/stratis-fstab-setup-v2
+++ b/systemd/stratis-fstab-setup-v2
@@ -1,0 +1,61 @@
+#!/bin/bash
+
+if [ -z "$1" ]; then
+	echo Stratis symlink with pool name and filesystem name in systemd format \
+	\(dev-stratis-\$POOL_NAME-\$FILESYSTEM_NAME\) required as an argument. >&2
+	exit 1
+fi
+
+
+if [ "$1" != dev-stratis* ]; then
+	echo Argument provided was in incorrect format. Stratis symlink with pool name \
+	and filesystem name in systemd format \(dev-stratis-\$POOL_NAME-\$FILESYSTEM_NAME\) \
+	required as an argument. >&2
+	exit 1
+fi
+
+PATH="$1"
+
+IFS='-' read -ra SEGMENTS <<< "$PATH"
+
+if [ "${#SEGMENTS[@]}" != 4 ]; then
+	echo Argument provided was in incorrect format. Stratis symlink with pool name \
+	and filesystem name in systemd format \(dev-stratis-\$POOL_NAME-\$FILESYSTEM_NAME\) \
+	required as an argument. >&2
+	exit 1
+fi
+
+POOL_NAME="${$SEGMENTS[2]}"
+
+i=0
+while ! stratis-min pool is-stopped --name "$POOL_NAME" >/dev/null; do
+	echo Waiting on pool with name $POOL_NAME...
+	sleep 1
+	if [ "$i" = 5 ]; then
+		break
+	fi
+	i=$(($i + 1))
+done
+
+if $(stratis-min pool is-stopped --name "$POOL_NAME"); then
+	if $(stratis-min pool is-bound --name "$POOL_NAME"); then
+		if ! stratis-min pool start --name --unlock-method=clevis "$POOL_NAME"; then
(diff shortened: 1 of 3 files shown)


systemd/stratis-fstab-setup-v2
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+if [ -z "$1" ]; then
+	echo Stratis symlink with pool name and filesystem name in systemd format \
+	\(dev-stratis-\$POOL_NAME-\$FILESYSTEM_NAME\) required as an argument. >&2
+	exit 1
+fi
+
+
+if [ "$1" != dev-stratis* ]; then
```
bmr-cymru, line 10: This expression isn't working for me in testing. I had to change it to:

```bash
if [[ "$1" != dev-stratis* ]]; then
```

systemd/stratis-fstab-setup-v2
```diff
@@ -0,0 +1,61 @@
...
+	echo Argument provided was in incorrect format. Stratis symlink with pool name \
+	and filesystem name in systemd format \(dev-stratis-\$POOL_NAME-\$FILESYSTEM_NAME\) \
+	required as an argument. >&2
+	exit 1
+fi
+
+PATH="$1"
```
bmr-cymru, line 17: Needs to be something like `DEVPATH` to avoid overwriting `PATH` - otherwise we'll get command not found for the binaries the script runs (stratis-min and sleep).

systemd/stratis-fstab-setup-v2
```diff
@@ -0,0 +1,61 @@
...
+IFS='-' read -ra SEGMENTS <<< "$PATH"
+
+if [ "${#SEGMENTS[@]}" != 4 ]; then
+	echo Argument provided was in incorrect format. Stratis symlink with pool name \
+	and filesystem name in systemd format \(dev-stratis-\$POOL_NAME-\$FILESYSTEM_NAME\) \
+	required as an argument. >&2
+	exit 1
+fi
+
+POOL_NAME="${$SEGMENTS[2]}"
```
bmr-cymru, line 28: To evaluate index 2 of the array this wants to be:

```bash
POOL_NAME="${SEGMENTS[2]}"
```

bmr-cymru (changes requested):
I've tested v2 with some minor changes to fix errors I saw running the script. Overall I think the fstab syntax is much better with v2 and I think this is preferable to putting the pool UUID into the options string but I still hit boot failures with the updated stratis-fstab-setup.

I resolved these by adding a `Before=local-fs-pre.target` to the `stratis-fstab-setup@.service` unit.

systemd/stratis-fstab-setup-v2
```diff
@@ -0,0 +1,61 @@
...
+	and filesystem name in systemd format \(dev-stratis-\$POOL_NAME-\$FILESYSTEM_NAME\) \
+	required as an argument. >&2
+	exit 1
+fi
+
+PATH="$1"
+
+IFS='-' read -ra SEGMENTS <<< "$PATH"
+
+if [ "${#SEGMENTS[@]}" != 4 ]; then
```
bmr-cymru, line 21: This won't work if there are embedded '-' characters in the pool or filesystem name (e.g. p1/fs1-snap). Systemd escapes '-' as `\x2d` in unit names but by the time the argument is passed to the stratis-fstab-setup script the escaping has been expanded to the literal '-' character.

jbaublitz:
@bmr-cymru I'm considering closing this out and moving the `Before=pre-local-fs.target` into #3436. Are you okay with that?

bmr-cymru:
@jbaublitz yes that seems reasonable to me. I will re-test with a clean set up and the changes from #3436 today.

Generally I think the changes you proposed here are an improvement in terms of the fstab syntax if we could get the proper escaping and decoding implemented but we can re-visit that in the future - the change to `Before=local-fs-pre.target` seems to be the key change to getting the fsck ordered properly.

jbaublitz:
@bmr-cymru I agree that it'd be an improvement. See #3438 for hopefully a better version of the pool name syntax that doesn't require the path or filesystem name.

PR #3443, opened 2023-09-18
Use updated branch for loopdev patch entry

bgurney-rh:
Related https://github.com/stratis-storage/project/issues/662


mulkieran:
The changes to Cargo.lock are because the new version of bindgen depended on in the loopdev branch requires a newer version of bitllags. So there are now duplicate bitflags dependencies.

mulkieran (approved):

PR #3448, opened 2023-09-21
.packit.yaml: Use cargo-vendor-filterer

bgurney-rh:


bgurney-rh:
Source RPM is 11.87MB (is usually 40 MB when not filtered.)

mulkieran (approved):

PR #3449, opened 2023-09-23
Make wait_for_pools method more robust

mulkieran:


mulkieran:
It didn't fail, so I'm going to merge.

PR #3459, opened 2023-10-02
Remove manual_trigger field from copr build

mulkieran:


mulkieran:
The build kicks off immediately. What happens w/ the test run.

mulkieran:
They run correctly. One failed because of a repeated error that might bear investigating.

PR #3468, opened 2023-10-17
Test new mutex in libcryptsetup-rs

jbaublitz:
To verify that tests pass.


jbaublitz:
@mulkieran Since the Tang tests and the unit tests in per-thread-mutex pass, I'm considering https://github.com/stratis-storage/libcryptsetup-rs/pull/330 safe.

PR #3479, opened 2023-10-24
Remove vendored submodule from libdbus-sys

mulkieran:
Related https://github.com/stratis-storage/project/issues/673


bgurney-rh (approved):
Package successfully builds with centos-stream-9 config.

PR #3483, opened 2023-10-25
Update to revision r7

jbaublitz:


jbaublitz:
Does the spec file need to be updated somewhere for packit to succeed? It looks like it's failing because of a version mismatch.

mulkieran:
> Does the spec file need to be updated somewhere for packit to succeed? It looks like it's failing because of a version mismatch.

https://github.com/stratis-storage/ci/pull/485 should do the trick. Let's ignore the build errors for now for this PR. Likely we should improve the ```create_archive.py``` script so that this failure won't happen next time.

mulkieran (changes requested):
For tests-with-testing-repo failure, set ```--highest-revision-number=7```.

mulkieran (approved):

PR #3492, opened 2023-11-02
github actions: update FEDORA_RELEASE to F39, F38

bgurney-rh:
Related: stratis-storage/project#676


mulkieran (changes requested):
Might as well just add libblkid-rs to the ```--ignore-high``` arguments everywhere check-fedora-versions target is run.

mulkieran (approved):

PR #3494, opened 2023-11-09
Fix alignment problems

jbaublitz:
I think the alignment issue was introduced in #2854. See commit message for more details.

### DIFF
--- a/src/engine/strat_engine/cmd.rs
+++ b/src/engine/strat_engine/cmd.rs
@@ -440,6 +440,8 @@ pub fn clevis_luks_regen(dev_path: &Path, keyslot: c_uint) -> StratisResult<()> 
 
 /// Determine the number of sectors required to house the specified parameters for
 /// the thin pool that determine metadata size.
+///
+/// Precondition: block_size is a power of 2.
 pub fn thin_metadata_size(
     block_size: Sectors,
     pool_size: Sectors,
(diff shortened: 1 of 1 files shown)


bgurney-rh (approved):

jbaublitz:
I want to verify that `MAX_METADATA_SIZE` would be 1 MiB aligned.

jbaublitz:
It is not. I'm going to add a round down function for the `MAX_META_SIZE`.

src/engine/strat_engine/cmd.rs
```diff
@@ -468,16 +468,22 @@ pub fn thin_metadata_size(
         })?
         .read_to_string(&mut output)?;
     if is_ok {
-        Ok(min(
-            THIN_META_MULT_FACTOR
+        let round = block_size - Sectors(1);
```
mulkieran, line 471: Since the computation, specifically the ```& !*round``` only works when ```block_size``` is a power of 2, could you put a statement about that somewhere? I think an inexpensive assertion would actually be rather hard to get, so just a pre-condition in the comments is probably the best.

src/engine/strat_engine/cmd.rs
```diff
@@ -468,16 +468,22 @@ pub fn thin_metadata_size(
...
                         .map_err(|e| StratisError::Msg(e.to_string()))?,
-                ),
-            MAX_META_SIZE.sectors(),
-        ))
+                )
+                + round)
+                & !*round,
+        );
+        assert!(determined_size % block_size == Sectors(0));
+        let max = Sectors(*(MAX_META_SIZE.sectors() - round) & !*round);
```
mulkieran, line 484: I don't understand why here ```round``` is subtracted from the value. I would have though adding it would be what was required.

mulkieran (changes requested):
One small request and a question.

bmr-cymru (approved):

src/engine/strat_engine/cmd.rs
jbaublitz: This is rounding down because if `MAX_META_SIZE` is not aligned to 1MiB, we want the largest multiple of 1 MiB that is below `MAX_META_SIZE`. After all, we don't want to round up to the nearest multiple if that is indeed the absolute max. Increasing it could cause an error.

jbaublitz: Ah, I think I see. You're recommending we just do the bitwise and at the end? I think you're correct about that.

jbaublitz: Resolved.

src/engine/strat_engine/cmd.rs
jbaublitz: Resolved.

jbaublitz:
@mulkieran Do you want this to go in first or the transaction revert commit?

mulkieran:
> @mulkieran Do you want this to go in first or the transaction revert commit?

I think this should go in first.

mulkieran (approved):

PR #3502, opened 2023-11-16
Packit updates

mulkieran:
Related https://github.com/stratis-storage/project/issues/673


mulkieran:
Ok. I was under a misunderstanding. At present, fedora-development = fedora-rawhide so Cockpit only builds for those. If we designate our tests as building for fedora-all, that will cause srpm builds to kick off to support every one of these tests, regardless of what target is set in the copr configuration.

jbaublitz (approved):

PR #3511, opened 2023-11-28
Specify nrext64 option when creating XFS filesystem

### LINKED ISSUE #3504: Avoid creating XFS filesystems with 64 bit extent count
The flag is ```nrext64```. We want to make sure that is set to no, to avoid situation where XFS changes its default configuration. mkfs.xfs 6.1.0 already has that flag. It began to be available in Linux 5.19. In order to get this correct, we would have to scrape the XFS version, to avoid specifying the option to an older version of xfs which rejects the version. 6.5.0 has large extent counts enabled by default (it was released in Oct 2023). https://fossies.org/linux/xfsprogs/doc/CHANGES

mulkieran:
Closes #3504

### DIFF
--- a/src/engine/strat_engine/cmd.rs
+++ b/src/engine/strat_engine/cmd.rs
@@ -233,12 +233,8 @@ fn get_clevis_executable(name: &str) -> StratisResult<PathBuf> {
 }
 
 /// Create a filesystem on devnode. If uuid specified, set the UUID of the
-/// filesystem on creation. If 'noalign', set the noalign option.
-/// NOTE: It used to be desirable to set the noalign option when creating the
-/// MDV, because it was very small; since then the size of the MDV has been
-/// increased, and it is no longer necessary to pass the noalign option when
-/// creating the MDV.
-pub fn create_fs(devnode: &Path, uuid: Option<StratisUuid>, noalign: bool) -> StratisResult<()> {
+/// filesystem on creation.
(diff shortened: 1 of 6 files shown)


mulkieran:
Comment fix, nothing more.

mulkieran:
+ an error string fix, some were just empty strings before...

bgurney-rh (approved):

src/engine/strat_engine/cmd.rs
```diff
@@ -232,13 +232,77 @@ fn get_clevis_executable(name: &str) -> StratisResult<PathBuf> {
...
+    // become more and more true with time. If the version is not high enough,
+    // this will be obvious when the command-line invocation fails.
+    // The check only verifies that the major number is greater than 5, because
+    // before that time, the larger extent counter size was not used by default,
+    // so, for earlier versions, it is unnecessary to turn it off explicitly.
+    // Choose not to use a semantic version package, because there is no Rust
+    // package availabe that necessarily adheres to the xfsprogs versioning
+    // scheme.
+    let use_nrext64_option = match get_mkfs_xfs_version().and_then(|v| {
+        v.split('.')
```
jbaublitz, line 281: We may not want to add this now, but this crate would simplify this code a lot. https://docs.rs/semver/latest/semver/

jbaublitz (approved):

PR #3516, opened 2023-12-12
Add origin field to StratFilesystem

### LINKED ISSUE #3515: Support origin field for filesystem snapshots
If a filesystem is a snapshot, give it an origin field. Store this field in filesystem metadata and expose it on the D-Bus.

mulkieran:
Closes #3515

### DIFF
--- a/src/dbus_api/filesystem/filesystem_3_7/props.rs
+++ b/src/dbus_api/filesystem/filesystem_3_7/props.rs
@@ -1,0 +1,18 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use dbus::arg::IterAppend;
+use dbus_tree::{MTSync, MethodErr, PropInfo};
+
+use crate::dbus_api::{
+    filesystem::shared::{self, get_filesystem_property},
+    types::TData,
+};
+
+pub fn get_fs_origin(
+    i: &mut IterAppend<'_>,
+    p: &PropInfo<'_, MTSync<TData>, TData>,
+) -> Result<(), MethodErr> {
+    get_filesystem_property(i, p, |(_, _, f)| Ok(shared::fs_origin_prop(f)))
+}
--- a/src/engine/sim_engine/pool.rs
+++ b/src/engine/sim_engine/pool.rs
@@ -476,14 +476,15 @@ impl Pool for SimPool {
         &mut self,
         _pool_name: &str,
         fs_uuids: &[FilesystemUuid],
-    ) -> StratisResult<SetDeleteAction<FilesystemUuid>> {
+    ) -> StratisResult<SetDeleteAction<FilesystemUuid, FilesystemUuid>> {
         let mut removed = Vec::new();
         for &uuid in fs_uuids {
             if self.filesystems.remove_by_uuid(uuid).is_some() {
                 removed.push(uuid);
             }
         }
-        Ok(SetDeleteAction::new(removed))
+
+        Ok(SetDeleteAction::new(removed, vec![]))
     }
 
     fn rename_filesystem(
@@ -884,7 +885,7 @@ mod tests {
             .unwrap();
         let fs_uuid = fs_results[0].1;
         assert!(match pool.destroy_filesystems(pool_name, &[fs_uuid]) {
-            Ok(filesystems) => filesystems == SetDeleteAction::new(vec![fs_uuid]),
+            Ok(filesystems) => filesystems == SetDeleteAction::new(vec![fs_uuid], vec![]),
             _ => false,
         });
     }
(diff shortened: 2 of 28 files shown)


src/dbus_api/filesystem/filesystem_3_7/props.rs
```diff
@@ -0,0 +1,18 @@
...
+use crate::dbus_api::{
+    filesystem::shared::{self, get_filesystem_property},
+    types::TData,
+};
+
+pub fn get_fs_origin(
+    i: &mut IterAppend<'_>,
+    p: &PropInfo<'_, MTSync<TData>, TData>,
+) -> Result<(), MethodErr> {
+    get_filesystem_property(i, p, |(_, _, f)| Ok(shared::fs_size_limit_prop(f)))
```
bmr-cymru, line 17: Should this be ``shared::fs_origin_prop(f)``?

When no origin is set for a filesystem does this return ``(false, "")``?

bmr-cymru (changes requested):

mulkieran:
We can not add an origin field without including the abllity to update the origin field in the case that a filesystem is removed and another filesystem has an origin field which points to that filesystem. So we must determine, before this PR is merged, what the rule is supposed to be. There are two possibilities:

- set the origin field to None
- redirect it to point to the same filesystem as the removed filesystem's origin pointed to, if that exists

mulkieran:
We're best off setting the origin field to None. TODO: handle D-Bus signals.

mulkieran:
We may want to make the D-Bus property settable, for debugging purposes.

martinpitt:
@mulkieran same issue as in the other PR -- please ignore testing-farm:fedora-rawhide-x86_64:cockpit here, we will investigate it today.

martinpitt:
@mulkieran We reported the root cause of our test failures to https://github.com/packit/packit-service/issues/2294 and worked around the problem on our side. So from now on the cockpit tests should succeed again. Not sure why they are stuck here, but you may try `/packit test` ? (I'm not allowed to in this project)

martinpitt:
Sorry, not your fault. We reported that lvm failure to https://bugzilla.redhat.com/show_bug.cgi?id=2256433 and will disable the test in rawhide, as it causes an unsalvageable mess. Please ignore.

mulkieran:
rebased

mulkieran:
Will need to add support in stratisd-min and stratis-min for obtaining the value of the origin field.

mulkieran:
rebased.

mulkieran:
rebased

bmr-cymru (approved):
I've tested this with the changes from stratis-cli#1045 and I'm seeing all the behaviour I expect: for regular file systems the origin field is unset and for snapshots I get the UUID of the origin volume.

The only thing I was unsure of reading the patches was whether the value printed for the report should be "None" rather than "Not set" to be consistent with the CLI output for file systems with no origin, but looking at the rest of the report code this is consistent with what we report when `size_limit` is not set so I think this is fine.

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -1576,6 +1577,22 @@ impl ThinPool {
...
+    pub fn unset_fs_origin(&mut self, fs_uuid: FilesystemUuid) -> StratisResult<bool> {
+        let changed = {
+            let (_, fs) = self.get_mut_filesystem_by_uuid(fs_uuid).ok_or_else(|| {
+                StratisError::Msg(format!("No filesystem with UUID {fs_uuid} found"))
+            })?;
+            fs.unset_origin()
+        };
+        let (name, fs) = self
+            .get_filesystem_by_uuid(fs_uuid)
+            .expect("Looked up above.");
```
jbaublitz, line 1590: I don't see any reason for this not to be inside the `if changed ...` block. That would avoid looking it up if nothing has changed.

src/engine/sim_engine/pool.rs
```diff
@@ -476,14 +476,33 @@ impl Pool for SimPool {
...
+            .collect();
+
+        let mut updated_origins = vec![];
+        for sn_uuid in snapshots {
+            if let Some((_, fs)) = self.filesystems.get_mut_by_uuid(sn_uuid) {
+                if fs.unset_origin() {
+                    updated_origins.push(sn_uuid);
+                }
+            }
+        }
```
jbaublitz, line 503: I think this could all be collapsed into a single iterator using `filesystems_mut()` for the iterator and `fold()` at the end to collect the updated origins.

src/engine/types/actions.rs
```diff
@@ -575,38 +575,44 @@ impl Display for DeleteAction<Key> {
...
             None
         } else {
-            Some(self.changed)
+            Some((self.changed, self.updated))
         }
     }
 }
 
-impl Display for SetDeleteAction<FilesystemUuid> {
+impl<U> Display for SetDeleteAction<FilesystemUuid, U> {
```
jbaublitz, line 615: I think erring on the side of specificity is good for these `impl`s so using `FilesystemUuid` might be better. Otherwise, we have a frequent problem of overlapping impls. That would also allow us to add to the log message indicating what UUIDs had their origin updated.

mulkieran:
@jbaublitz I think it is ready now.

mulkieran:
squashed...

PR #3519, opened 2023-12-27
Increase itertools dependency lower bound to 0.12.0

mulkieran:


mulkieran:
"Guest couldn't be provisioned", so I'll ignore failed tests.

PR #3522, opened 2024-01-03
Fix up prediction tests to match new filesystem interface

mulkieran:
Related #3432


mulkieran:
The tests are fixed and passing, will merge after #3523

PR #3528, opened 2024-01-04
Try both Clevis and passphrase in /etc/fstab setup

### LINKED ISSUE #3349: Pools with NBDE can never be unlocked with a passphrase during boot
A encrypted pool that should be unlocked during boot can be bound to a Tang server so that the boot can be unattended. However, if the Tang server is not reachable during boot, the stratis-fstab-setup program will just fail and not allow entering a passphrase at all.

This is different from LUKS. With LUKS, you get the opportunity to enter a passphrase even when a device uses NBDE as well.

jbaublitz:
Closes #3349

### DIFF
--- a/systemd/stratis-fstab-setup
+++ b/systemd/stratis-fstab-setup
@@ -19,23 +19,29 @@ done
 
 if $(stratis-min pool is-stopped "$POOL_UUID"); then
 	if $(stratis-min pool is-bound "$POOL_UUID"); then
-		if ! stratis-min pool start --unlock-method=clevis "$POOL_UUID"; then
+		if stratis-min pool start --unlock-method=clevis "$POOL_UUID"; then
+			exit 0
+		else
 			echo Failed to start pool with UUID $POOL_UUID using Clevis. >&2
-			exit 1
-		fi
-	elif $(stratis-min pool is-encrypted "$POOL_UUID"); then
-		if ! systemd-ask-password \
-			"Enter password for pool with UUID $POOL_UUID" |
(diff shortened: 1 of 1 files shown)


jbaublitz:
@mvollmer I tested this patch and it resolves the linked issue. The one question I had was mainly about whether you had this problem at boot time too. I think we may take a similar approach but I'm not sure.

jbaublitz:
@mulkieran I'm putting this in review after testing and bug fixes.

mvollmer (commented):
I think this works, but the code was a bit hard to follow.

jbaublitz:
@mvollmer I've re-requested your review. I think I've resolved the issues you mentioned.

mulkieran (approved):

jbaublitz:
@mvollmer After investigating a little bit more, I can confirm that the same problem affects the initrd scripts, but it's a little bit harder to solve there largely because we have intentionally separated the Clevis unlock method and the passphrase unlock method into separate dracut modules. This is largely because Anaconda does not support Clevis but does support the keyring unlock method for cryptsetup. These are distributed in separate packages as a result so Anaconda could theoretically support one without the other. I'll have to look into whether there's a way to set a flag based on whether executables are present or not in a dracut module. If we did that, we could potentially merge both dracut modules into a single, dynamic one that would also satisfy the requirement for Anaconda.

mulkieran:
@mvollmer We're ready to merge as soon as you're satisfied w/ the changes.

mvollmer (approved):
Thanks!

PR #3542, opened 2024-01-30
Run udev tests for 40 minutes

mulkieran:


mulkieran:
It's trivial if the tests run to completion.

mulkieran:
The first commit is enough to cause the tests not to time out anymore, but the failure on the generic test comes back.

mulkieran:
Besides using ```stop_pool``` we just need to start the pool when we bring up stratisd again. I _think_ that is a fair test _in most cases_. I think, though, that enhancing stratis-dumpmetadata to overwrite pool metadata would be exactly what we need here to make a more correct test.

mulkieran:
Something has just gone wrong w/ our loop tests on every platform:

```
test: TANG_URL=localhost RUST_LOG=stratisd=debug STRATISD=/usr/libexec/stratisd STRATIS_DUMPMETADATA=/usr/bin/stratis-dumpmetadata PYTHONPATH=./src make -f Makefile tang-tests dump-metadata-tests
```

All seem to be yielding clevis errors with three indicating that cryptsetup is not found. All of this should be quite orthogonal to any changes here.

mulkieran:
Waiting for resolution of #3543

jbaublitz (approved):

PR #3543, opened 2024-02-01
Relax fairness restriciton on AllOrSomeLock

jbaublitz:
Related to #3542

### DIFF
--- a/src/engine/structures/lock.rs
+++ b/src/engine/structures/lock.rs
@@ -292,8 +292,7 @@ where
     /// Returns true if the current request should be put in the wait queue.
     /// * Always returns false if the index for the given request is in the record of woken
     /// tasks.
-    /// * Otherwise, returns true if any of the following conditions are met:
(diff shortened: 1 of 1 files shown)


jbaublitz:
You can disregard these failures. These are a bug that I introduced in this PR.

jbaublitz:
I think they should be fixed now.

jbaublitz:
@mulkieran The tests-with-testing-repo failure is actually interesting mainly because it indicates that our D-Bus signal code might have to be reworked if we accept this change because we will not longer be able to guarantee that the requests are processed in the order that they're received. The testing-farm failure is unexpected and I'm going to investigate further because it looks like the same problem.

jbaublitz:
@mulkieran Looks like we're just waiting on Cockpit tests and I expect those to pass.

mulkieran (changes requested):

mulkieran (approved):

PR #3560, opened 2024-03-04
main.yml: Use new revision of 3.6 release of testing

mulkieran:


mulkieran:
This is just a trivial update to our tests, and that has passed, it looks like everything else is infrastructure issues, so I'll merge.

PR #3568, opened 2024-03-13
Add serializer for Strings

mulkieran:


src/engine/strat_engine/serde_structs.rs
```diff
@@ -12,12 +12,47 @@
...
+const MAXIMUM_STRING_SIZE: usize = 255;
+
+// Safely returns a truncated String. Can not use truncate() as that panics if
+// len is not on a char boundary.
+fn safe_truncate(value: &str, len: usize) -> String {
+    let bytes_value = value.as_bytes();
+    let len = std::cmp::min(len, bytes_value.len());
+    String::from_utf8_lossy(&bytes_value[..len])
+        .trim_matches(std::char::REPLACEMENT_CHARACTER)
+        .to_string()
```
jbaublitz, line 30: Hmm, why not just index the `String`? That would avoid the conversion to bytes and it could be converted directly from a `str` back to a `String`.

mulkieran: The ```str``` index operation is just like ```truncate()```; it'll panic if I try to split a character encoding in the middle. We allow multibyte characters in filesystem names and pool names and it makes sense to allow them in UserInfo property, too.

jbaublitz: Ah, I read up more on strings in Rust and I think that we'll at the very least want to add in a comment to use [this method](https://doc.rust-lang.org/std/string/struct.String.html#method.floor_char_boundary) once it stabilizes.

mulkieran: Comment added!

jbaublitz: I also just found [this method](`https://doc.rust-lang.org/std/primitive.str.html#method.is_char_boundary`) and I think this would simplify the code quite a bit if we used truncate instead, but just checked it beforehand to see if it's on the byte-boundary and move the index back if it isn't. That would avoid the unnecessary allocation that we have here.

mulkieran: @jbaublitz There is a missing link in your comment...

mulkieran: @jbaublitz Maybe you're thinking of ```is_char_boundary```?

mulkieran: @jbaublitz Yes you were! GitHub hid [this link](https://doc.rust-lang.org/std/primitive.str.html#method.is_char_boundary), though, because the URL was between '`' chars. Sneaky.

mulkieran:
@jbaublitz Ok, this is ready again.

src/engine/strat_engine/serde_structs.rs
```diff
@@ -12,12 +12,61 @@
...
+}
+
+// Return a new String, split at the highest index which lies on a char
+// boundary of value which is no greater than len.
+fn safe_split_at(value: &str, len: usize) -> String {
+    value
+        .split_at(our_floor_char_boundary(value, len))
+        .0
+        .to_string()
+}
```
jbaublitz, line 45: One last comment on this. This doesn't avoid the unnecessary allocation which we can do using a slice index and that will return a `&str` which we can then serialize without an intermediary allocation. It shouldn't panic because we've checked the char boundary.

src/engine/strat_engine/serde_structs.rs
```diff
@@ -12,12 +12,61 @@
...
+
+// Serialize a string. Only the first MAXIMUM_STRING_SIZE bytes of the string
+// are serialized.
+fn serialize_string<S>(value: &str, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    String::serialize(&safe_split_at(value, MAXIMUM_STRING_SIZE), serializer)
```
jbaublitz, line 53: Here we should be able to call `str::serialize` with the above change.

jbaublitz (changes requested):
Just one last comment.

mulkieran:
@jbaublitz Ok, ready again.

martinpitt:
The [cockpit f40 failure](https://artifacts.dev.testing-farm.io/19c9e0ce-209e-4ca6-b671-a304519ac7ed) looks unrelated -- something in mdraid/linux takes way too much time to create a raid array. Please ignore or retry.

jbaublitz (approved):

mulkieran:
squashed

PR #3570, opened 2024-03-18
Reject UserInfo values that do not satisfy validate_name

mulkieran:


src/engine/shared.rs
```diff
@@ -169,19 +169,19 @@ pub fn validate_name(name: &str) -> StratisResult<()> {
...
-            "Name contains characters not allowed in udev symlinks: {name}"
+            "Provided string contains characters not allowed in udev symlinks: {name}"
         )));
     }
 
     let name_path = Path::new(name);
     if name_path.components().count() != 1 {
         return Err(StratisError::Msg(format!(
-            "Name is a path with 0 or more than 1 components: {name}"
+            "Provided string is a path with 0 or more than 1 components: {name}"
```
jbaublitz, line 179: I would personally be confused if I got this error message. I'm wondering if it would be better to separate out the case of `.count() == 0` and say something like "Provided string is empty" and then for `> 1 || .is_absolute()` we could say something like "Provided string contains `/`.

jbaublitz (commented):
This isn't explicitly related to this PR, but this PR brought my attention to the error messages here.

src/engine/shared.rs
mulkieran: I agree it is confusing. I've tried to improve things.

jbaublitz (approved):

PR #3574, opened 2024-03-22
Fix a minor bug in pid file handling

mulkieran:
Related: stratis-storage/project#697


bgurney-rh (approved):

mulkieran:
Looks like #3578 is the preferred solution.

PR #3580, opened 2024-03-25
Increase nix dependency lower bound to 0.28.0

mulkieran:
Related https://github.com/stratis-storage/project/issues/699


mulkieran:
@jbaublitz The deprecated allows are for ```flock```, which is deprecated in favor of the ```Flock``` implementation.  That will require a bit more work, so I plan to file an issue and addres that separately.

jbaublitz (approved):

PR #3585, opened 2024-04-01
Increase lower bounds of Stratis-project dependencies

mulkieran:
Related https://github.com/stratis-storage/project/issues/699


mulkieran:
At this point, we should expect build failure on f39 and f38, but not f40 or rawhide.

PR #3588, opened 2024-04-03
Create pipe so file descriptor is not dropped early

mulkieran:
Closes #3586

PR #3591, opened 2024-04-03
Regression test for key set in stratis-min

jbaublitz:
Related to #3588


mulkieran (approved):

jbaublitz:
We should maybe consider doing password verification, but that seems low priority.

mulkieran:

Issue #3299, opened 2023-03-29
We should add a coverage test to stratis-min testing


mulkieran:
It should be easy to get complete coverage information for the stratis-min chunk of code.

mulkieran:
The first step is getting the complete coverage. As ever, we will have to exclude a few things, but llvm-cov tool does not have a convenient way of annotating lines, unlike the standard Python coverage tools. This means that there is more of a benefit to using, e..,g., codecov because we can get changes in the coverage induced by a PR, which is an informative kind of information. See example: https://github.com/rust-netlink/netlink-packet-route/pull/59#issuecomment-1788598827

mulkieran:
Coverage at 64.8%.

PR #3594, opened 2024-04-08
Add stratis-min tests

bgurney-rh:
Related https://github.com/stratis-storage/stratisd/issues/3299


tests/stratis_min.rs
```diff
@@ -214,6 +214,22 @@ fn test_stratis_min_pool_rename() {
     test_with_stratisd_min_sim(stratis_min_pool_rename);
 }
 
+#[test]
+// Test stopping a pool using an invalid name, that will fail with a
+// "Uuid error".
+fn stratis_min_pool_stop_invalid_name() {
```
mulkieran, line 220: As written, the test is almost a parser test because the stratis-min code is interpreting the name as a UUID, trying to read it in, and failing, hence the UUID error. Since it never interacts with stratisd-min at all, you should use just the one part structure, like ```test_stratis_min_report_bad_subreport``` does.

mulkieran (changes requested):

tests/stratis_min.rs
bgurney-rh: OK; I think I see what's going on.  I split the two tests; one stops a started pool, and the other is the parser test.

tests/stratis_min.rs
```diff
@@ -214,6 +214,63 @@ fn test_stratis_min_pool_rename() {
...
+#[test]
+// Test stopping a pool using a valid name.
+fn test_stratis_min_pool_stop_name() {
+    test_with_stratisd_min_sim(stratis_min_pool_stop_name);
+}
+
+#[test]
+// Test stopping a pool using an invalid name, that will fail with a
+// "Uuid error".
+fn test_stratis_min_pool_stop_invalid_name() {
```
mulkieran, line 250: Best to move this one test above the "tests with sim engine" comment. Otherwise, all the tests look good as far as I can tell.

mulkieran (changes requested):

tests/stratis_min.rs
bgurney-rh: OK; resolved.

bgurney-rh:
Removed a couple of tests (test_stratis_min_init_cache_add_data_add_cache and test_stratis_min_create_pool_encrypted_rebind); also removed the ".gitignore" changes.

bgurney-rh:
Squashed into a single commit.

mulkieran (approved):

PR #3609, opened 2024-05-06
Test loopbacked

mulkieran:


mulkieran:
Verify that error also occurs on loopbacked devices.

mulkieran:
Even writing 50 MiB instead of 100 means the test passes. I'm going to try doubling.

Issue #3604, opened 2024-05-01
For stratis-min, failure to scan a UUID when a UUID is expected should really be a parsing error


mulkieran:
There is no way that I can figure out how to do this. clap has conditional parsing for very specific things, like ```required_unless_present```. What I need is a ```value_parser_if``` function but that doesn't exist, and this problem, error code 1 instead of 2, really isn't _that_ important.

PR #3610, opened 2024-05-06
Add tests to cover all unformatted UUID commands

mulkieran:
Related #3604


mulkieran:
I believe that it is also possible not just to test but to make these into parser errors.

bgurney-rh (approved):

PR #3613, opened 2024-05-07
Do verification on the stratisd side for Clevis configs to avoid hanging

jbaublitz:
Related to latchset/clevis#241

### DIFF
--- a/src/engine/strat_engine/backstore/crypt/consts.rs
+++ b/src/engine/strat_engine/backstore/crypt/consts.rs
@@ -36,3 +36,5 @@ pub const DEFAULT_CRYPT_METADATA_SIZE: Bytes = Bytes(16 * IEC::Ki as u128);
 pub const DEFAULT_CRYPT_KEYSLOTS_SIZE: Bytes = Bytes(16352 * IEC::Ki as u128);
 
 pub const CLEVIS_TOKEN_NAME: &str = "clevis\0";
+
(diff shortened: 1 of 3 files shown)


mulkieran (changes requested):
This all makes sense to me. My only concern is that it might be possible to pass a config that causes a stack overflow during the recursive execution of ```all_tang_configs_have_thp```. I think there should be some way to prevent that from happening.

jbaublitz:
That is arguably already possible in other parts of the code given that we take a similar recursive approach when parsing Clevis tokens from disk to provide the reconstructed config to the user over the D-Bus. How would you recommend approaching preventing this in both cases? Adding a parameter to limit the number of recursions we can reach?

mulkieran:
> That is arguably already possible in other parts of the code given that we take a similar recursive approach when parsing Clevis tokens from disk to provide the reconstructed config to the user over the D-Bus. How would you recommend approaching preventing this in both cases? Adding a parameter to limit the number of recursions we can reach?

Until now, we weren't recursively exploring the config, we were essentially passing it directly through to Clevis. But now that we're recursively exploring something that is sent over the D-Bus, I think that the additional caution is warranted. I'm not sure if parsing the Clevis tokens from disk is analogous, but it may well be.

A recursion limit would be fine, I think.

jbaublitz:
@martinpitt Is this a known issue with cockpit? It seems to be affecting all new PRs with stratisd. @mulkieran believes this is not a problem in our master branch.

PR #3617, opened 2024-05-13
Require emulated TPM2 for Stratis tests

mulkieran:


mulkieran:
All our tests failed w/ Artemis, guest couldn't be provisioned error. Cockpit tests failed with some error that is surely unrelated to stratisd changes. So we can be pretty sure that it is the configuration changes, for our tests only, that are breaking something.

mulkieran:
Ok. So this approach is guaranteed not to work. We would probably have to provision in the normal way and then use swtpm to emulate the TPM2 module on the provided virtual machine.

mulkieran:
our tests seem to be stuck running with no information at all.

mulkieran:
I failed to background a step...

mulkieran:
shell syntax  interpretation problems; I think this attempt may work

mulkieran:
rebased

plans/all.fmf
```diff
@@ -19,6 +28,9 @@ prepare:
   - name: Show test system information
     how: shell
     script: free -m; lsblk -i; lscpu; cat /proc/1/sched
+  - name: Show TPM2 information
+    how: shell
+    script: tpm2_getrandom 16 | hexdump
```
jbaublitz, line 33: Should this go in the final version? I thought it was just for testing that the TPM was correctly provisioned during development of the PR.

jbaublitz (changes requested):

plans/all.fmf
mulkieran: I don't mind dropping it.

mulkieran:
@jbaublitz This is ready again.

jbaublitz (approved):

PR #3620, opened 2024-05-15
Use tenacity for retries in Python tests

mulkieran:
Closes https://github.com/stratis-storage/project/issues/626

### DIFF
--- a/.github/workflows/support.yml
+++ b/.github/workflows/support.yml
@@ -46,6 +46,7 @@ jobs:
               python3-psutil
               python3-pyudev
               python3-semantic_version
+              python3-tenacity
             task: PYTHONPATH=./src make -f Makefile lint
             working-directory: ./tests/client-dbus
           - dependencies: black python3-isort
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -24,6 +24,7 @@ jobs:
               python3-psutil
               python3-pyudev
               python3-semantic_version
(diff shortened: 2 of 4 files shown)


mulkieran:
Likely our TMT tests are failing due to a bug in the PR.

mulkieran:
rebased and conflicts resolved, I think.

mulkieran:
Retries as they were constructed were causing TMT tests to timeout. I'm not sure quite why. This PR now only does retries for one thing, starting stratisd, and does it in a much simplified way (which is probably better overall).

mulkieran:
And the tests are all passing now...

mulkieran:
The log entry is a little odd, because we're using the block approach, and not a method.

```
INFO:root:Retrying <unknown> in 0.5 seconds as it raised DBusException: org.freedesktop.DBus.Error.ServiceUnknown: The name is not activatable.
```

The ```<unknown>``` is because no name can be put on the thing that is retried. Also note the lack of line information.

mulkieran:
I do not believe that the tenacity automatic logging is that useful. But we can glean the marginally interesting fact that one retry is always needed under normal operation, but never more, from the test output.

mulkieran:
If this passes, I'll consider it ready.

mulkieran:
I think tenacity is the winner among the retry packages, since it allows retrying arbitrary blocks of code, i.e., you do not have to turn what you want to retry into a function.

Issue #3622, opened 2024-05-22
Support merging a snapshot based on stored instruction in filesystem level metadata


mulkieran:
Related https://github.com/stratis-storage/project/issues/597

Depends on https://github.com/stratis-storage/stratisd/issues/3621 .

We assume just one thing, that filesystem metadata is augmented with a new field, merge, with type ```Option<bool>```.

Thus, the merge must be performed on pool setup, subsequent to the filesystem metadata being written. Note also that this changes how we need to read the filesystem metadata. At present, we can read it serially, setting up each fileystem as we process the metadata. However, once we are implementing merge, we must retain something about the filesystems as a group. Specifically, we must scan all the filesystems to find out which are in a scheduled merge relationship and we will not know if there are any such until we have scanned all, because the metadata is only on the snapshot, not on the origin, and both must be brought together for the merge.

If we think that such a situation could be a problem, it could be allowable to store supplemental information for the origin. If that were done, then, generally, the filesystems not so marked could all be set up immediately, and only those scheduled for merge, either as origin or snapshot, would have their processing delayed until the end.

If a merge fails, the best thing is to keep both filesystems exactly as they were in their prior state and bring the pool up. However, this condition of the pool must be identified on the D-Bus as problem with the pool, "scheduled merge not completed" or something like that.

In the event of a merge, the merged snapshot replaces its origin. Conceptually, it stops existing, and its meaning as the origin of some other snapshot disappears. What this means is that the possible origin values of other filesystems on the pool must be updated to None if they currently point to the merged snapshot and the updated metadata must be written for each of these filesystems. What about other snapshots pointing to the origin in this merge? I would think they should stay in case it turns out to be necessary to role back to a previous snapshot. But what if it turns out that they post-date the snapshot rolled back to? Maybe if that is the case they should lose their origin field.

In the case that somehow, there are two snapshots scheduled to be merged into the same origin, it still seems like the pool should be brought up, neither merge should be begun, and a warning message on the pool should be placed on the D-Bus.

An important implementation question is how to handle reading filesystem metadata. ~~The metadata does not take up much size in memory, so reading it and retaining it seems fine.~~ This is already how the metadata is read. Performing merge operations, manipulating the filesystem metadatas accordingly, and then using those to guide setup of each individual filesystem ought to work.

bmr-cymru:
One additional thought I had in regards to this: how should the filesystem UUID be handled during a merge?

[...]

Issue #3621, opened 2024-05-21
Support issuing an instruction for Stratis filesystem merge and storing the instruction in filesystem metadata


mulkieran:
Related https://github.com/stratis-storage/project/issues/597 

We do not plan to do merges into anything other than a snapshot's origin. Given this the tasks are:

1. Augment the filesystem metadata with a new field, ```merge```, indicating that the filesystem is to be merged. The field makes no sense if there is no origin set.
  * So, we could make the field ```Option<bool>```. If the filesystem has no origin, then the field is ```None```. If the filesystem has an origin, the the field is ```Some(true)``` or ```Some(false)```. If the filesystem has the concept of origin, but no ```merge``` field then it might have a ```None``` value for ```merge``` anyway. Treat ```None``` as ```Some(false)``` in the case where the the origin field is present. Ensure that next time the field is written it is written as ```Some(false)```.

  * The alternative is to make it a non-optional field. But that will not work anyway until 4.0. We could leave a Note to make it non-optional. But otherwise, the first option will always be consistent.

2. Augment the D-Bus API with a new command ```RequestMerge``` on a pool. It can take the pool object path and the UUID or name of the filesystem. Since it can not merge into anything but the origin, that is all the information it needs. It should error on (a) no origin for the filesystem for which the request is made, (b) other filesystem already scheduled to merge into same origin,  ~~(c) a difference between the designated filesystem's size limit and the origin size limit. Why? Why not make it an automatic adjustment when the merge occurs?~~ We will have to work out a way to reconcile a number of the fields in the filesystem metadata. Assuming the request is acceptable, then the filesystem metadata must be updated with the merge field set to ```Some(true)```. NOTE: we have decided that different filesystem size limit is not a reason to reject the merge request. That value will be updated from the snapshot's value.

mulkieran:
Note that we already report the origin of any snapshot on the D-Bus and we maintain this origin when filesystems are deleted, etc. If a merge is scheduled, the snapshot will maintain a to-be-merged status. This should be observable in the filesystem information via a D-Bus property.

mulkieran:
We will first add a debug option to get filesystem metadata

PR #3623, opened 2024-05-23
Support snapshot revert operation

mulkieran:
Closes https://github.com/stratis-storage/stratisd/issues/3621
Closes https://github.com/stratis-storage/stratisd/issues/3622

### DIFF
--- a/src/dbus_api/consts.rs
+++ b/src/dbus_api/consts.rs
@@ -67,6 +67,7 @@ pub const FILESYSTEM_CREATED_PROP: &str = "Created";
 pub const FILESYSTEM_SIZE_PROP: &str = "Size";
 pub const FILESYSTEM_SIZE_LIMIT_PROP: &str = "SizeLimit";
 pub const FILESYSTEM_ORIGIN_PROP: &str = "Origin";
+pub const FILESYSTEM_MERGE_SCHEDULED_PROP: &str = "MergeScheduled";
 
 pub const BLOCKDEV_INTERFACE_NAME_3_0: &str = "org.storage.stratis3.blockdev.r0";
 pub const BLOCKDEV_INTERFACE_NAME_3_1: &str = "org.storage.stratis3.blockdev.r1";
--- a/src/dbus_api/filesystem/filesystem_3_7/api.rs
+++ b/src/dbus_api/filesystem/filesystem_3_7/api.rs
@@ -4,11 +4,27 @@
 
 use dbus_tree::{Access, EmitsChangedSignal, Factory, MTSync, Property};
 
-use crate::dbus_api::{consts, filesystem::filesystem_3_7::props::get_fs_origin, types::TData};
+use crate::dbus_api::{
+    consts,
+    filesystem::filesystem_3_7::props::{
+        get_fs_merge_scheduled, get_fs_origin, set_fs_merge_scheduled,
+    },
+    types::TData,
+};
 
 pub fn origin_property(f: &Factory<MTSync<TData>, TData>) -> Property<MTSync<TData>, TData> {
     f.property::<(bool, String), _>(consts::FILESYSTEM_ORIGIN_PROP, ())
         .access(Access::Read)
         .emits_changed(EmitsChangedSignal::True)
         .on_get(get_fs_origin)
+}
+
+pub fn merge_scheduled_property(
+    f: &Factory<MTSync<TData>, TData>,
+) -> Property<MTSync<TData>, TData> {
+    f.property::<bool, _>(consts::FILESYSTEM_MERGE_SCHEDULED_PROP, ())
+        .access(Access::ReadWrite)
+        .emits_changed(EmitsChangedSignal::True)
+        .on_get(get_fs_merge_scheduled)
+        .on_set(set_fs_merge_scheduled)
 }
--- a/src/dbus_api/filesystem/filesystem_3_7/mod.rs
+++ b/src/dbus_api/filesystem/filesystem_3_7/mod.rs
@@ -5,4 +5,4 @@
 mod api;
 mod props;
 
-pub use api::origin_property;
+pub use api::{merge_scheduled_property, origin_property};
--- a/src/dbus_api/filesystem/filesystem_3_7/props.rs
+++ b/src/dbus_api/filesystem/filesystem_3_7/props.rs
@@ -2,12 +2,16 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-use dbus::arg::IterAppend;
+use dbus::arg::{Iter, IterAppend};
 use dbus_tree::{MTSync, MethodErr, PropInfo};
 
-use crate::dbus_api::{
-    filesystem::shared::{self, get_filesystem_property},
-    types::TData,
+use crate::{
+    dbus_api::{
+        consts,
+        filesystem::shared::{self, get_filesystem_property},
+        types::TData,
+    },
+    engine::PropChangeAction,
 };
 
 pub fn get_fs_origin(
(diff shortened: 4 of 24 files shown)


martinpitt:
Ech, can we please stop breaking rawhide in new an "interesting" ways every *censored* day? I'm still catching up with the fallout from the firewalld/selinux debacle..

```
+ chpasswd
chpasswd: (user root) pam_chauthtok() failed, error:
Authentication token manipulation error
chpasswd: (line 1, user root) password not changed
```

Again, clearly not your fault. I'm still busy with investigating/reporting the other regressions from last night, so if someone else could have a look that'd be nice. However, as #3653 went green, it's worth retrying.

martinpitt:
The timing coincides with https://bodhi.fedoraproject.org/updates/FEDORA-2024-26a844bb1c . This also happens in podman PRs, e.g. https://github.com/containers/podman/pull/23274 or https://github.com/containers/podman/pull/23275 . 

Trivially reproducible in a rawhide VM with updating shadow-utils. With shadow-utils-2:4.15.1-8.fc41.x86_64 (*not* the latest version from that bodhi update, that's -9):

```
# echo root:foobar | chpasswd
chpasswd: (user root) pam_chauthtok() failed, error:
Authentication token manipulation error
chpasswd: (line 1, user root) password not changed
```

That most recent bodhi update reverted the change again:
```
dnf update https://kojipkgs.fedoraproject.org//packages/shadow-utils/4.15.1/9.fc41/x86_64/shadow-utils-4.15.1-9.fc41.x86_64.rpm
```
and with that it works. So please retry.

martinpitt:
I hacked around the shadow-utils b0rkage in https://github.com/cockpit-project/cockpit/pull/20802 . Can you please retry the failed tests (or ignore the rawhide failures if you don't care)?

mulkieran:
rebased. This was just a conflict w/ the filesystem metadata PR, so it was not a lot of work to resolve the conflicts.

mulkieran:
I think that likely we'll need to expose the message method in devicemapper and that that will be the most efficient and easiest thing.

mulkieran:
Now running into ENOSPC on both ubuntu and fedora.

mulkieran:
On both Fedora and Ubuntu, loop_test_rename is just hanging...

mulkieran:
Huzzah! At least the thinpool checks are succeeding (by reducing the size of the filesystems).

mulkieran:
Next step is to set the XFS filesystem UUID.

mulkieran:
rebased. Need to fix up engine level unit tests.

mulkieran:
See if engine tests fail

mulkieran:
Now all existing tests are succeeding again. What happens if we add the revert test back in again?

mulkieran:
The initial pool is not created in the test. We will give up on the revert Rust unit test, and test it externally, instead.

mulkieran:
More boring type errors to fix

bmr-cymru:
Setting/clearing the `MergeScheduled` property and applying the merge the next time the pool is set up is all working well with the current branch. I've added support to the `snapm` Stratis plugin and tested that we get the expected behaviour from a revert operation.

I know  we discussed this previously but now that I've had a chance to test it out some more I'm unsure about the cleaning up of the origin links after the revert in commit b081b6563f9809033a2bccc8284f20b0908df28f:

```
If a snapshot was taken after the one reverted to, remove its origin
field.

If a snapshot was reverted, any snapshot that points at it should have
its origin field removed.
```

1. I think for the first case it may make more sense to leave the origin set as it is: even though the snapshot was taken at a later point in time it is still a snapshot of the now-reverted origin. Since revert on thin is just a case of switching one `thin_id` address space for another the sequence of snapshot creation doesn't really matter.

This would be better for `snapm`'s uses since we use `Origin` to identify snapshots volumes, so the current behaviour causes us to lose track of snapshot set members when an earlier snapshot set is reverted:

```
# Create two snapshot sets:
root@localhost:~/src/git/snapm# snapm snapset create stratis1 /opt /srv
SnapsetName:      stratis1
MountPoints:      /opt, /srv
NrSnapshots:      2
Time:             2024-09-05 17:43:24
UUID:             c9ccc991-c6b3-539f-9273-21e5fc1cb1ce
Status:           Active
Autoactivate:     yes
Bootable:         no
root@localhost:~/src/git/snapm# snapm snapset create stratis2 /opt /srv
SnapsetName:      stratis2
MountPoints:      /opt, /srv
NrSnapshots:      2
Time:             2024-09-05 17:43:29
UUID:             c4c11f79-e15d-51b8-a4a5-951471699a4d
Status:           Active
Autoactivate:     yes
Bootable:         no
root@localhost:~/src/git/snapm# snapm snapset list
SnapsetName  Time                 NrSnapshots Status  MountPoints             
stratis1     2024-09-05 17:43:24            2 Active  /opt, /srv              
stratis2     2024-09-05 17:43:29            2 Active  /opt, /srv              

# Revert the first snapshot set:
root@localhost:~/src/git/snapm# snapm snapset revert stratis1
WARNING - Snaphot set stratis1 is in use: reboot required to complete revert
root@localhost:~/src/git/snapm# umount /opt /srv
root@localhost:~/src/git/snapm# stratis pool stop --name p1
root@localhost:~/src/git/snapm# stratis pool start --name p1

# The second snapshot set is now "orphaned":
root@localhost:~/src/git/snapm# snapm snapset list
SnapsetName  Time                 NrSnapshots Status  MountPoints             
root@localhost:~/src/git/snapm# stratis filesystem
Pool   Filesystem                             Total / Used / Free / Limit        Device                                                 UUID                                
p1     fs1                                    1 GiB / 73 MiB / 951 MiB / None    /dev/stratis/p1/fs1                                    590756f3-1eaf-47db-8dbb-224b3c5c038c
p1     fs1-snapset_stratis2_1725554609_-opt   1 GiB / 73 MiB / 951 MiB / None    /dev/stratis/p1/fs1-snapset_stratis2_1725554609_-opt   94e61615-b7d7-4ac6-8b1e-f8260908b539
p1     fs2                                    2 GiB / 73 MiB / 1.93 GiB / None   /dev/stratis/p1/fs2                                    9db8871e-ff2e-479b-9575-85f215553a78
p1     fs2-snapset_stratis2_1725554609_-srv   2 GiB / 73 MiB / 1.93 GiB / None   /dev/stratis/p1/fs2-snapset_stratis2_1725554609_-srv   6a1fb22d-aec8-47b6-af34-3ba3518376b8
```

This leaves the user unable to manage the snapshots via `snapm` and prevents the snapshot being reverted into the origin. If the origin link was maintained the user would be able to revert or delete the snapshot set via `snapm` as required.

Looking at LVM2 for comparison it's a bit inconsistent - for COW snapshots the origin of other snapshots is preserved after the merge operation but for thin it behaves similarly to the current Stratis implementation: the origin field of _all_ other snapshots is cleared and they become "normal" thin LVs. This seems to have caused some confusion for users in the past and there was a proposal a while ago to add a new `--keep-origin` option to maintain the link.

2. In the second case I think it would be better to set the origin of any snapshots-of-snapshots to the reverted origin UUID. This again allows us to still detect them as snapshots in `snapm` and would allow them to be reverted into the origin (which seems like a valid operation since the content is a snapshot of the data that exists on that device after the revert).

LVM2 does do this for thin snapshots-of-snapshots: the origin is updated following the merge to point to the merged origin volume:

```
root@localhost:~/src/git/stratis-cli# lvs
  LV                VG     Attr       LSize  Pool  Origin      Data%  Meta%  Move Log Cpy%Sync Convert
  pool0             fedora twi-aotz--  2.00g                   6.31   12.21                           
  test0             fedora Vwi-aotz--  1.00g pool0             6.32                                   
  test0-snap1       fedora Vwi---tz-k  1.00g pool0 test0                                              
  test0-snap1-snap1 fedora Vwi---tz-k  1.00g pool0 test0-snap1                                        

root@localhost:~/src/git/snapm# lvconvert --merge fedora/test0-snap1
  Volume fedora/test0-snap1 replaced origin fedora/test0

root@localhost:~/src/git/stratis-cli# lvs
  LV                VG     Attr       LSize  Pool  Origin Data%  Meta%  Move Log Cpy%Sync Convert
  pool0             fedora twi-aotz--  2.00g              6.31   12.21                           
  test0             fedora Vwi-a-tz--  1.00g pool0        6.32                                   
  test0-snap1-snap1 fedora Vwi---tz-k  1.00g pool0 test0                                         
```

mulkieran:
(1) is done

mulkieran:
For (2), the only issue is whether, if we do it, we ought to make ```destroy_filesystems``` patch up the graph, also. In both cases, we can have origin0, snap1, and snap1_a, where snap1_a is a snap of snap1. 

For the revert case, snap1 overwrites origin1, and the proposal is that snap1_a's origin should be changed to origin0, instead of being set to None, as it is now.

For the destroy case, if snap1 is destroyed, should snap1_a's origin be changed to origin0? It is currently set to None.

mulkieran:
(2) is done

mulkieran:
Another change is implemented. If reverting snap1_a into snap_1, the origin field is set to origin0, not to None as formerly.

mulkieran:
I think we should make the  filesystem_destroy operation patch up the origin as well, to make it congruent with revert behavior. But that's more of a pain, because of the D-Bus.

bmr-cymru:
> I think we should make the filesystem_destroy operation patch up the origin as well, to make it congruent with revert behavior. But that's more of a pain, because of the D-Bus.

I think that makes sense. I'll try to get these changes tested in my snapshot manager VMs.

PR #3637, opened 2024-06-25
Metadata rework

jbaublitz:
Follow up on #3274


mulkieran:
@jbaublitz Plz rebase and resolve conflicts.

plans/all.fmf
```diff
@@ -1,38 +1,99 @@
...
-    script: free -m; lsblk -i; lscpu; cat /proc/1/sched
-  - name: Record mkfs.xfs version
-    how: shell
-    script: mkfs.xfs -V
-discover:
-  how: fmf
-execute:
-  how: tmt
-  exit-first: false
+/python:
```
mulkieran, line 8: You can use inheritance here, and should use it to make it clear what is different between the two setups. It turns out that the inheritance works fine, so something like this will do: https://github.com/stratis-storage/stratisd/pull/3640/commits/690e2066c9480dce8aaca120e6b2af38b4dfe891

mulkieran (changes requested):
Two requests...still working.

plans/all.fmf
jbaublitz: Ah, I think I missed the part on inheritance and the factored out step was just being replaced by the specified step.

src/dbus_api/api/prop_conv.rs
```diff
@@ -66,7 +66,7 @@ pub fn locked_pools_to_prop(pools: &LockedPoolsInfo) -> StoppedOrLockedPools {
 }
 
 /// Convert a stopped pool data structure to a property format.
-pub fn stopped_pools_to_prop(pools: &StoppedPoolsInfo) -> StoppedOrLockedPools {
+pub fn stopped_pools_to_prop(pools: &StoppedPoolsInfo, metadata: bool) -> StoppedOrLockedPools {
```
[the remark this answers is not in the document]
jbaublitz: Resolved.

plans/all.fmf
jbaublitz: Resolved.

mulkieran:
Please go ahead and extend the TMT-based udev and bind tests for v2 pools. https://github.com/stratis-storage/stratisd/pull/3643 is an indication of how to do that. The actual test changes are not at all considerable, feel free to choose something else if you want.

mulkieran (changes requested):
In pool-level metadata, "features" is double-nested. Here is an excerpt from the metadata print out.

```
  "features": {
    "features": [
      "Encryption"
    ]
  },
```

jbaublitz:
> In pool-level metadata, "features" is double-nested. Here is an excerpt from the metadata print out.
> 
> ```
>   "features": {
>     "features": [
>       "Encryption"
>     ]
>   },
> ```

This is not a bug, but rather the result of the default behavior for serde derive. We chose to wrap the features in a new-type struct which is resulting in the double nesting. To change the behavior you're seeing if it's important to you, I think we could either replace the new-type struct with a Vector in the metadata or we could rename one of the keys.

mulkieran:
> > In pool-level metadata, "features" is double-nested. Here is an excerpt from the metadata print out.
> > ```
> >   "features": {
> >     "features": [
> >       "Encryption"
> >     ]
> >   },
> > ```
> 
> This is not a bug, but rather the result of the default behavior for serde derive. We chose to wrap the features in a new-type struct which is resulting in the double nesting. To change the behavior you're seeing if it's important to you, I think we could either replace the new-type struct with a Vector in the metadata or we could rename one of the keys.

It is certainly a consequence of the default behavior for ```serde_derive``` and it works perfectly correctly within stratisd. The reason I noticed it at all was while writing tests external to stratisd that verify invariants that should hold for the metadata. The question is, is it the metadata structure we want to support or would we prefer to avoid that apparently redundant key?

jbaublitz:
> > > In pool-level metadata, "features" is double-nested. Here is an excerpt from the metadata print out.
> > > ```
> > >   "features": {
> > >     "features": [
> > >       "Encryption"
> > >     ]
> > >   },
> > > ```
> > 
> > 
> > This is not a bug, but rather the result of the default behavior for serde derive. We chose to wrap the features in a new-type struct which is resulting in the double nesting. To change the behavior you're seeing if it's important to you, I think we could either replace the new-type struct with a Vector in the metadata or we could rename one of the keys.
> 
> It is certainly a consequence of the default behavior for `serde_derive` and it works perfectly correctly within stratisd. The reason I noticed it at all was while writing tests external to stratisd that verify invariants that should hold for the metadata. The question is, is it the metadata structure we want to support or would we prefer to avoid that apparently redundant key?

I've removed the redundant key.

jbaublitz:
I'm going to try rerunning the tests here. I've been unable to reproduce the failures I'm seeing locally in a VM with the same code.

jbaublitz:
I think this might be a timing issue which is why I can't reproduce it locally. Last time rawhide succeeded.

tests/client-dbus/tests/udev/test_udev.py
```diff
@@ -484,17 +492,30 @@ def test_duplicate_pool_name(
...
+                        )
+                    )
                     for dev in sublist
                 ],
             )
 
-            wait_for_udev(CRYPTO_LUKS_FS_TYPE, self._lb_mgr.device_files(luks_tokens))
+            wait_for_udev(
+                CRYPTO_LUKS_FS_TYPE if _LEGACY_POOL is not None else STRATIS_FS_TYPE,
+                self._lb_mgr.device_files(luks_tokens),
```
mulkieran, line 517: Change this line to:
```
self._lb_mgr.device_files(luks_tokens if _LEGACY_POOL is not None else non_luks_tokens),
```

and the failure will be resolved.

Also, in the case where ```_LEGACY_POOL``` is  ```None```, the expression will be identical to the one on line 519. But that should be acceptable.

jbaublitz: My last change seemed to fix the problem! Resolving.

mulkieran (changes requested):
Typo located...

src/engine/strat_engine/backstore/blockdev/v2.rs
```diff
@@ -0,0 +1,467 @@
...
+        if Some(s) == self.new_size
+            || (self.new_size.is_none() && s == self.bda.dev_size().sectors())
+        {
+            Ok(None)
+        } else {
+            Ok(Some(s))
+        }
+    }
+
+    fn grow(&mut self) -> StratisResult<bool> {
```
mulkieran, line 291: I _believe_ that the implementation of this method should also include an adjustment to the ```integrity_meta_allocs``` value since the amount of metadata space required increases with the size of the device.

mulkieran (changes requested):
One more...

mulkieran (approved):
Approved, modulo the requirement to remove temporary test redirects.

Issue #3634, opened 2024-06-20
D-Bus notification for org.storage.stratis3.filesystem.r*.Used is missing


mvollmer:
When slowing filling a filesystem with data, the value of org.storage.stratis3.filesystem.r*.Used is updated periodically (as expected), but there are no change notifications for that.

There are notifications when org.storage.stratis3.pool.r*.TotalPhysicalUsed every 10 seconds, and I would expect to get notifications for filesystem.Used at the same time.

stratisd-3.6.7-1.fc40.x86_64

Here is the output of D-Bus monitor. You can see the notifications for pool.TotalPhysicalUsed in there, bot nothing for filesystem.Used.

[stratis-used.log](https://github.com/user-attachments/files/15912658/stratis-used.log)

mulkieran:
We've reproduced this behavior.

mulkieran:
We have reproduced, using a test that makes use of the monitor-dbus test script:

1. When a filesystem is added to the pool, the pool TotalPhysicalUsed property is updated.
2. When we write 1 MiB to a file on the filesystem, using dd, and doing it iteratively, with 4096B block size + only 128 KiB at a time, we see that no signal for the filesystem Used property is sent out for the changes.
3. After we have stopped writing to the filesystem, and we wait a while for the D-Bus to settle the updated value does appear on the D-Bus, but no signal is sent. The dbus monitor script reports the discrepancy.

Our test that automatically uses the dbus monitor script did not report any discrepancy. However, it writes the data all at once, instead of in 128KiB iterations. Since it did not report an error we can conclude that for the test either:
1. The Used property value is not updated.
2. The signal is sent out.
I think that (1) is likely the correct explanation.

We have an _idea_ of why this is happening, which is that the ```Used``` property signal gets sent out generally only when the ```Size``` property signal gets sent out as well, and, of course, that property is updated less frequently. But that hypothesis does not explain the observed behavior of the Used property getting the correct value.

mulkieran:
I believe the fix can be localized to the ```check_fs``` and ```handle_extension``` methods.

PR #3639, opened 2024-06-26
Issue stratisd 3634

mulkieran:
Related #3634


mulkieran:
This approach is acceptable only if it is ok to spawn a thread for every filesystem.

mulkieran:
With simple test:
1. Create a pool
2. Create a filesystem
3. Write 1 MiB to the filesystem

The filesystem Used signal is not emitted on HEAD (although pool TotalPhysicalUse signal is); it is emitted with this PR.

mulkieran:
@jbaublitz This draft PR contains all the essentials and is somewhat tested. Plz let me know what you think.

jbaublitz (commented):
I looked at this for a while and the only change I wanted to potentially discuss is that I'm a bit unclear on what the benefit is of running a diff in the case of a filesystem that isn't mounted. Is that just for the sake of code simplicity? I see how it's necessary to run the diff code even when the extend size is zero, but I'm not entirely sure why we're doing that in the case where the filesystem isn't mounted.

mulkieran:
> I looked at this for a while and the only change I wanted to potentially discuss is that I'm a bit unclear on what the benefit is of running a diff in the case of a filesystem that isn't mounted. Is that just for the sake of code simplicity? I see how it's necessary to run the diff code even when the extend size is zero, but I'm not entirely sure why we're doing that in the case where the filesystem isn't mounted.

I agree that's too much. I think to address that I should change ```should_extend()``` into ```should_update()``` and have it return an enum result with two values, Extend and Update.

mulkieran:
ubuntu test failures look like infrastructure

jbaublitz:
> > I looked at this for a while and the only change I wanted to potentially discuss is that I'm a bit unclear on what the benefit is of running a diff in the case of a filesystem that isn't mounted. Is that just for the sake of code simplicity? I see how it's necessary to run the diff code even when the extend size is zero, but I'm not entirely sure why we're doing that in the case where the filesystem isn't mounted.
> 
> I agree that's too much. I think to address that I should change `should_extend()` into `should_update()` and have it return an enum result with two values, Extend and Update.

Why not just still used `.filter_map()` but only return None in the case that the filesystem isn't mounted? I feel like that still would fix the bug while also maybe being a little bit conceptually simpler than adding an enum.

mulkieran:
> > > I looked at this for a while and the only change I wanted to potentially discuss is that I'm a bit unclear on what the benefit is of running a diff in the case of a filesystem that isn't mounted. Is that just for the sake of code simplicity? I see how it's necessary to run the diff code even when the extend size is zero, but I'm not entirely sure why we're doing that in the case where the filesystem isn't mounted.
> > 
> > 
> > I agree that's too much. I think to address that I should change `should_extend()` into `should_update()` and have it return an enum result with two values, Extend and Update.
> 
> Why not just still used `.filter_map()` but only return None in the case that the filesystem isn't mounted? I feel like that still would fix the bug while also maybe being a little bit conceptually simpler than adding an enum.

We need two bits of info to know whether the filesystem should just be visited or whether it should also be extended. I'm trying out calling extend_size from should_visit only if the filesystem should be extended and returning 0 if the filesystem should not be extended.

mulkieran:
@jbaublitz Ok. This is ready again.

jbaublitz (approved):
After our discussion, this looks good to me.

PR #3648, opened 2024-07-10
Return true from extend_thin_meta_device less

mulkieran:


src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -1117,7 +1117,10 @@ impl ThinPool {
                 meta_growth,
             );
 
-            (ext.is_ok(), ext)
+            match ext {
+                Ok(Sectors(0)) | Err(_) => (false, ext),
```
jbaublitz, line 1121: Isn't this unreachable?

jbaublitz (changes requested):

src/engine/strat_engine/thinpool/thinpool.rs
mulkieran: @jbaublitz You might be misreading the diff.

jbaublitz: Ah I see, so this is an optimization to not save the metadata if it's not extended at all? I am failing to see a case where this will ever be reached. Unreachable might be the wrong term, because it could be reached, but `meta_growth` is guaranteed to be larger than 0. I don't see a case where this will be reached with the current implementation. I was originally thinking that `ext` could be 0 but I don't see a case where `meta_growth` would be non-zero but `ext` would be zero.

jbaublitz: Let me dig into do_extend and see if maybe it's rounding down. That could potentially cause the behavior you're seeing.

mulkieran: I understand your argument. I think we can leave this in pending for now and work on simplifying it later.

jbaublitz:
Is this still relevant or can it be closed?

mulkieran:
> Is this still relevant or can it be closed?

It may be. I need to take another look.

mulkieran:
To do this better we should just save the thinpool's metadata at the start of the check method and see if it has change when the check method is about to return, thereby eliminating complexity.

PR #3651, opened 2024-07-16
Online re-encrypt

### LINKED ISSUE #3597: Add support for online encrypt, decrypt, and reencrypt in stratisd metadata version 2
We've had some discussions with the cryptsetup maintainers, and it seems like we should be able to support online operations for encrypt, decrypt and reencrypt in version 2 of the metadata. Only reencrypt is not feasible for version 1 of the metadata because decrypt and encrypt would require shifting the data to either make room for the crypt metadata in the encrypt case or to bring the Stratis metadata to the beginning of the device in the case of decrypt. As a result, reencrypt may be supported in version 1, but all three will be support in version 2.

jbaublitz:
Closes #3597 
Blocked by #3637 
Blocked by stratis-storage/libcryptsetup-rs#347


jbaublitz:
It appears as though this will fail until we release the updated libcryptsetup-rs on crates.io.

martinpitt:
See https://bodhi.fedoraproject.org/updates/FEDORA-2024-aa10dc94fe . The automated tests are a sea of red.

martinpitt:
OTOH, *this particular* PR failed the RPM build on a missing Rust crate, so that's different.

jbaublitz:
I think we need to wait for a Fedora 41 compose for now. It seems like that's why CI is failing.

mulkieran:
I would expect all Packit builds to succeed at this point.

mulkieran:
@jbaublitz Please rebase when you get the chance

src/dbus_api/pool/pool_3_8/api.rs
```diff
@@ -102,3 +102,50 @@ pub fn clevis_infos_property(f: &Factory<MTSync<TData>, TData>) -> Property<MTSy
...
+    f.method("ReencryptPool", (), reencrypt_pool)
+        // b: true if successful
+        //
+        // Rust representation: bool
+        .out_arg(("results", "b"))
+        .out_arg(("return_code", "q"))
+        .out_arg(("return_string", "s"))
+}
+
+pub fn decrypt_pool_method(f: &Factory<MTSync<TData>, TData>) -> Method<MTSync<TData>, TData> {
```
mulkieran, line 143: Can we consider calling this UnencryptPool rather than DecryptPool? decrypt generally at least connotes a slightly different thing.

mulkieran (changes requested):
Requesting a nomenclature change as it might be clearer this way.

src/dbus_api/pool/pool_3_8/api.rs
jbaublitz: Do you mind if we solicit feedback on this? I do not know what will be clearer to the average user in this case. Even from a quick search of opinions across the internet, some people associate the verb "unencrypt" only with the state of being "unencrypted" (never having been encrypted) and not with any usage of the verb. Personally I don't think decrypt necessarily denotes something different as much as you can temporarily or persistently decrypt data. Temporarily decrypting seems to be the process of unlocking a device where persistently decrypting data seems to indicate changing the device to store unencrypted data.

src/dbus_api/pool/pool_3_8/api.rs
```diff
@@ -102,3 +102,50 @@ pub fn clevis_infos_property(f: &Factory<MTSync<TData>, TData>) -> Property<MTSy
...
+pub fn encrypt_pool_method(f: &Factory<MTSync<TData>, TData>) -> Method<MTSync<TData>, TData> {
+    f.method("EncryptPool", (), encrypt_pool)
+        // Optional key descriptions of key in the kernel keyring
+        // a: array of zero or more elements
+        // b: true if a token slot is specified
+        // i: token slot
+        // s: key description
+        //
+        // Rust representation: Vec<((bool, u32), String)>
+        .in_arg(("key_desc", "a((bu)s)"))
```
mulkieran, line 115: Probably should go with plural here, ```key_descs``` and same below as ```clevis_infos```.

mulkieran (changes requested):
Another minor name request...

src/dbus_api/pool/pool_3_8/api.rs
jbaublitz: Resolved.

src/dbus_api/pool/pool_3_8/api.rs
mulkieran: We decided in standup that this nomenclature is Ok.

mulkieran (changes requested):
I think that ```reencrypt``` is timing out on a 50G device pool.

Steps are:

1. Make a new encrypted pool with a 50G device encrypted with a key in the kernel keyring.
2. > stratis pool encryption reencrypt --name=pn
3. Timeout occurs. Also, there was an interesting WARN message: ```[2025-04-03T19:44:37Z WARN  stratisd::engine::strat_engine::crypt::shared] Failed to read hotzone area starting at 34359738368```. I don't yet know what a hotzone area is, but I expect it is where the data transformation is actually occurring.
4. Timeout occurs: ```Execution failed:
stratis attempted communication with the daemon, stratisd, over the D-Bus, but stratisd did not respond in the allowed time. ```
5. Oddly, this problem also seems to occur in stratis-cli CI. https://github.com/stratis-storage/stratis-cli/actions/runs/14249887889/job/39939916279?pr=1139 . The reason that is really odd is that stratis-cli is running against the sim engine in these tests.

jbaublitz:
@mulkieran I think there may be a few issues here. One was a bug in the D-Bus interface that I've resolved with the most recent update to this PR. That may explain the timeout given that the end result in reencryption was a panic in both the sim and strat engines. I am going to try to reserve a test machine with more storage to try out the updated PR with more physical storage to see if the other problem you raised still exists.

jbaublitz:
After #3805 is resolved, we will have to add additional signal handling for the new free token slot property in this PR.

jbaublitz:
Before I proceed @fsuba @mulkieran are we okay with the timestamp having a granularity of seconds? I feel like that should be sufficient if it's sufficient for other timestamps in our metadata.

mulkieran:

src/engine/strat_engine/crypt/handle/v2.rs
```diff
@@ -729,6 +719,196 @@ impl CryptHandle {
...
+        encryption_info: &InputEncryptionInfo,
+    ) -> StratisResult<Self> {
+        let tmp_header = format!("/tmp/temp-header-{pool_uuid}");
+        {
+            let mut file = OpenOptions::new()
+                .create(true)
+                .truncate(true)
+                .write(true)
+                .open(&tmp_header)?;
+            file.write_all(&[0; 4096])?;
```
mulkieran: Ideally this 4096 would be a named constant somewhere.

src/engine/strat_engine/thinpool/filesystem.rs
```diff
@@ -453,6 +454,13 @@ impl StratFilesystem {
     pub fn thin_id(&self) -> ThinDevId {
         self.thin_dev.id()
     }
+
+    /// Get the sector size reported by libblkid for this filesystem.
+    pub fn block_size(&self) -> StratisResult<u64> {
```
mulkieran: Better to change the name so that it is a logical sector size and to return in Bytes if it is not too inconvenient.

src/engine/strat_engine/crypt/handle/v2.rs
```diff
@@ -729,6 +719,196 @@ impl CryptHandle {
...
+        }
+
+        let mut device = CryptInit::init(Path::new(&tmp_header))?;
+        let data_offset = Bytes::from(16 * IEC::Mi).sectors();
+        device.set_data_offset(*data_offset)?;
+
+        let sectors = thinpool
```
mulkieran: From here to line 756 seems like very much an ```unwrap_or_else``` situation, and is possibly better in the v2 thinpool under some title like ```get_default_logical_sector_size```.

src/engine/strat_engine/crypt/handle/v2.rs
```diff
@@ -729,6 +719,196 @@ impl CryptHandle {
...
+            Some(("aes", "xts-plain")),
+            CryptParamsReencrypt {
+                mode: CryptReencryptModeInfo::Encrypt,
+                direction: CryptReencryptDirectionInfo::Forward,
+                resilience: "checksum".to_string(),
+                hash: "sha256".to_string(),
+                data_shift: 0,
+                max_hotzone_size: 0,
+                device_size: 0,
+                luks2: CryptParamsLuks2 {
```
mulkieran, line 780: Seems like here you can just reuse ```params``` struct defined at line 757.

src/engine/strat_engine/crypt/handle/v2.rs
```diff
@@ -729,6 +719,196 @@ impl CryptHandle {
...
+            key,
+            new_slot,
+        )
+    }
+
+    /// Encrypt an unencrypted pool.
+    pub fn decrypt(self, pool_uuid: PoolUuid) -> StratisResult<()> {
+        let activation_name = format_crypt_backstore_name(&pool_uuid);
+        let sector_size = {
+            let mut probe = BlkidProbe::new_from_filename(self.luks2_device_path())?;
```
mulkieran: If there is a simple way to abstract this chunk, it should definitely be done. It's been recurring in a bunch of places.

src/engine/strat_engine/crypt/shared.rs
```diff
@@ -992,6 +996,205 @@ pub fn clevis_decrypt(
...
+}
+
+pub fn handle_setup_reencrypt(
+    luks2_path: &Path,
+    encryption_info: &EncryptionInfo,
+) -> StratisResult<(u32, SizedKeyMemory, u32)> {
+    let mut device = acquire_crypt_device(luks2_path)?;
+
+    let mut keys = get_all_passphrases(&mut device, encryption_info)?;
+    let (single_ts, single_key) = keys.pop().expect("Must have at least one unlock mechanism");
```
mulkieran: ```split_off``` might be better here, so that you can give a name to the rest of the passphrases.

src/engine/strat_engine/crypt/shared.rs
```diff
@@ -992,6 +996,205 @@ pub fn clevis_decrypt(
...
+                    ))
+                }
+            },
+        }
+    }
+
+    Ok(passphrases)
+}
+
+pub fn handle_setup_reencrypt(
```
mulkieran, line 1098: Here it would be good to have a header comment explaining what this does. Each of the passphrases has to be updated in some way w/ respect to the volume key, but it is not clear exactly how.

src/engine/strat_engine/crypt/handle/v1.rs
```diff
@@ -1003,6 +1003,27 @@ impl CryptHandle {
         Ok(())
     }
 
+    /// Encrypt an unencrypted pool.
```
mulkieran: This header comment and the one below need a little revising, since they are both re-encrypt and also the object being worked on is a single crypt device.

src/engine/sim_engine/pool.rs
```diff
@@ -925,6 +929,30 @@ impl Pool for SimPool {
...
+        &mut self,
+        _: &Name,
+        _: PoolUuid,
+        enc: &InputEncryptionInfo,
+    ) -> StratisResult<CreateAction<EncryptedDevice>> {
+        self.encryption_info = convert_encryption_info(Some(enc), None)?;
+        Ok(CreateAction::Created(EncryptedDevice))
+    }
+
+    fn reencrypt_pool(&mut self, _: &Name) -> StratisResult<ReencryptedDevice> {
```
mulkieran: Please change sim engine's reencrypt so that it returns an error if pool is not encrypted. Better for stratis-cli testing.

src/engine/strat_engine/crypt/handle/v2.rs
```diff
@@ -729,6 +719,196 @@ impl CryptHandle {
         Ok(())
     }
 
+    /// Encrypt an unencrypted pool.
+    pub fn encrypt(
+        pool_uuid: PoolUuid,
+        thinpool: &mut ThinPool<v2::Backstore>,
```
mulkieran: Does not need to be mut.

jbaublitz:
I've run the tests a few times and it seems like the error is gone for reencryption but there is a new error that I will have to try to debug for encryption in the testing infrastructure.

src/engine/strat_engine/crypt/handle/v2.rs
jbaublitz: This was already fixed in the updates to enable read lock acquisition during encryption/reencryption/decryption.

src/engine/strat_engine/crypt/handle/v2.rs
jbaublitz: This will no longer make sense with the most recent commits for read lock support.

src/engine/strat_engine/thinpool/filesystem.rs
```diff
@@ -18,6 +18,7 @@ use devicemapper::{
     Bytes, DevId, DmDevice, DmName, DmOptions, DmUuid, Sectors, ThinDev, ThinDevId, ThinPoolDev,
     ThinStatus,
 };
+use libblkid_rs::BlkidProbe;
```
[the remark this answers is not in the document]
jbaublitz: @mulkieran I believe this is just plain incorrect. I'm going to mark as resolved.

Cargo.toml
```diff
@@ -79,7 +79,7 @@ version = "1.2.3"
 optional = true
 
 [dependencies.chrono]
-version = "0.4.20"
+version = "0.4.31"
```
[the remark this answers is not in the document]
jbaublitz: @mulkieran I believe this is not a very useful comment. I'm going to mark as resolved.

src/engine/strat_engine/crypt/handle/v2.rs
jbaublitz: @mulkieran Are you referring to the logical block size chunk?

src/engine/strat_engine/crypt/handle/v2.rs
jbaublitz: `unwrap_or_else` will not work here because in the else part, there are methods that can return early. This generally does not work well with `unwrap_or_else` unless the entire expression is supposed to return an error.

I've added the min logical sector size into the general thinpool implementation. I usually try to make the methods as widely as available as possible unless there is a specific constraint requiring special handling in v1 or v2.

src/engine/strat_engine/backstore/backstore/v1.rs
```diff
@@ -798,6 +798,47 @@
...
+        for (bd, (slot, key, new_slot)) in self
+            .blockdevs()
+            .into_iter()
+            .map(|(_, _, bd)| bd)
+            .zip(key_info)
+        {
+            bd.do_reencrypt(slot, key, new_slot)?;
+        }
+        Ok(())
+    }
```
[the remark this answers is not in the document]
jbaublitz: I think I can disregard this.

src/engine/strat_engine/backstore/backstore/v1.rs
```diff
@@ -798,6 +798,47 @@ impl Backstore {
...
+            return Err(StratisError::Msg(
+                "Requested pool does not appear to be encrypted".to_string(),
+            ));
+        };
+
+        operation_loop(
+            self.blockdevs_mut().into_iter().map(|(_, _, bd)| bd),
+            |blockdev| blockdev.setup_reencrypt(),
+        )
+    }
```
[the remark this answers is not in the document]
jbaublitz: @mulkieran I think this is legitimate but I wanted to get your feedback on the doc comment largely because I don't think it's correct. I think the order returned by this method is deterministic.

src/engine/strat_engine/crypt/handle/v2.rs
```diff
@@ -729,6 +719,222 @@ impl CryptHandle {
...
+        encryption_info: &InputEncryptionInfo,
+    ) -> StratisResult<(u32, (u32, SizedKeyMemory))> {
+        let tmp_header = format!("/tmp/temp-header-{pool_uuid}");
+        {
+            let mut file = OpenOptions::new()
+                .create(true)
+                .truncate(true)
+                .write(true)
+                .open(&tmp_header)?;
+            file.write_all(&[0; 4096])?;
```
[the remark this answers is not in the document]
jbaublitz: @mulkieran I'm torn on this one. I'm going to look into a bit more tomorrow.

src/dbus_api/pool/pool_3_9/methods.rs
```diff
@@ -0,0 +1,243 @@
...
+    );
+
+    let mut guard = get_mut_pool!(dbus_context.engine; pool_uuid; default_return; return_message);
+    let (name, _, _) = guard.as_mut_tuple();
+
+    let result = guard.start_reencrypt_pool();
+    let guard = guard.downgrade();
+    let result = result.and_then(|key_info| guard.do_reencrypt_pool(key_info));
+    let mut guard = block_on(dbus_context.engine.upgrade_pool(guard));
+    let result = result.and_then(|_| guard.finish_reencrypt_pool(&name));
```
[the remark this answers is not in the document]
jbaublitz: I don't think this is valid. The two workflows are different and require different handling.

src/engine/sim_engine/pool.rs
```diff
@@ -925,6 +931,68 @@ impl Pool for SimPool {
         }
     }
 
+    fn encrypt_pool_idem_check(&self) -> StratisResult<CreateAction<EncryptedDevice>> {
```
mulkieran, line 934: Would it be possible to put in sleeps of, say, 3 minutes, for the parts of these encryption operations that we would expect in the strat engine to be long-running?

mulkieran (changes requested):
If it is possible to do the request here, the sooner the better, as it will help w/ testing. @jbaublitz For this request, do not bother unless you really want to. I found a way to manage the testing via sort-of mocking the call in stratis-cli tests, so I'm not sure its worth the trouble any more to sleep bits of the sim engine.

jbaublitz:
Okay, I've addressed all PR revisions from CodeRabbit in this PR. I think it can be closed whenever you like @mulkieran.

jbaublitz:
@mulkieran Can we close this?

mulkieran:
We don't think we need this anymore...

PR #3652, opened 2024-07-18
Revision 8 bump

jbaublitz:
Blocked by #3637


mulkieran:
@jbaublitz Could you refresh this in whatever way is necessary?

jbaublitz:
@mulkieran Feel free to review again, but I think I moved everything around appropriately after looking at what was released in 3.7.0.

mulkieran:
> @mulkieran Feel free to review again, but I think I moved everything around appropriately after looking at what was released in 3.7.0.

Ok. I'll handle the stratis-cli and testing repo failures.

mulkieran (changes requested):
The test errors ``` TypeError: Fewer items found in D-Bus signature than in Python arguments``` are because the r8 StartPool method does not take a file descriptor for the passphrase but this change was introduced to support the metadata rework. I think that the thing to do is to port that change from the r7 revision to the r8 revision and remove it from the r7 revision. I think doing it in a separate commit will help keep that step clear.

Correction: I had this backward. I believe that it is the testing repo that needs to be updated and that the StartPool failures are what should be expected.

src/dbus_api/pool/mod.rs
```diff
@@ -275,6 +275,38 @@ pub fn create_dbus_pool<'a>(
...
+                .add_m(pool_3_0::unbind_clevis_method(&f))
+                .add_m(pool_3_5::init_cache_method(&f))
+                .add_m(pool_3_0::add_cachedevs_method(&f))
+                .add_m(pool_3_0::bind_keyring_method(&f))
+                .add_m(pool_3_0::unbind_keyring_method(&f))
+                .add_m(pool_3_0::rebind_keyring_method(&f))
+                .add_m(pool_3_0::rebind_clevis_method(&f))
+                .add_m(pool_3_0::rename_method(&f))
+                .add_m(pool_3_3::grow_physical_device_method(&f))
+                .add_m(pool_3_7::get_metadata_method(&f))
```
mulkieran, line 295: There's one method here that is missing, ```get_fs_metadata_method```. Added in r7.

mulkieran (changes requested):

mulkieran (approved):

mulkieran:
In the dbus monitor tests I'm finding that, for r7 only, the updated value of the D-Bus StoppedPools property has the following extra chunks, while the final value does not. Most likely this means, that, for r7 only, a signal was put on the D-Bus for the enriched, r8, value of StoppedPools. At the end of the check, though, when the GetManagedObjects() call is made, the StoppedPools value associated with r7 is the correct one, without the metadata_version, etc. fields.

```
                    dbus.String("metadata_version"): dbus.Struct(
                        (dbus.Boolean(True), dbus.UInt64(2)),
                        signature=None,
                        variant_level=1,
                    ),
                    dbus.String("features"): dbus.Struct(
                        (
                            dbus.Boolean(True),
                            dbus.Dictionary({}, signature=dbus.Signature("sb")),
                        ),
                        signature=None,
                        variant_level=1,
                    ),

```

src/dbus_api/pool/mod.rs
```diff
@@ -407,6 +440,22 @@ pub fn get_pool_properties(
             consts::POOL_OVERPROV_PROP => shared::pool_overprov_enabled(pool),
             consts::POOL_NO_ALLOCABLE_SPACE_PROP => shared::pool_no_alloc_space(pool),
             consts::POOL_METADATA_VERSION_PROP => shared::pool_metadata_version(pool)
+        },
```
mulkieran, line 443: I believe that you should remove POOL_METADATA_VERSION_PROP from the 3_7 interface.

mulkieran (changes requested):

src/dbus_api/pool/mod.rs
jbaublitz: Okay, done.

mulkieran (approved):

mulkieran:
merging this, stratis-cli, and testing PRs, and then all should be back to normal.

PR #3656, opened 2024-07-24
Add D-Bus method to get filesystem metadata

mulkieran:
Related https://github.com/stratis-storage/stratisd/issues/3621

### DIFF
--- a/src/engine/sim_engine/pool.rs
+++ b/src/engine/sim_engine/pool.rs
@@ -758,6 +758,29 @@ impl Pool for SimPool {
     fn metadata_version(&self) -> StratSigblockVersion {
         StratSigblockVersion::V2
     }
+
+    fn current_fs_metadata(&self, fs_name: Option<&str>) -> StratisResult<String> {
+        serde_json::to_string(
+            &self
+                .filesystems
+                .iter()
+                .filter_map(|(name, uuid, fs)| {
+                    if fs_name.map(|n| *n == **name).unwrap_or(true) {
+                        Some((uuid, fs.record(name, *uuid)))
+                    } else {
+                        None
+                    }
+                })
+                .collect::<HashMap<_, _>>(),
+        )
+        .map_err(|e| e.into())
+    }
+
+    fn last_fs_metadata(&self, fs_name: Option<&str>) -> StratisResult<String> {
+        // The sim pool doesn't write data, so the last fs metadata and the
+        // current fs metadata are, by definition, the same.
+        self.current_fs_metadata(fs_name)
+    }
 }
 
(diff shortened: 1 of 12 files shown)


src/engine/sim_engine/pool.rs
```diff
@@ -758,6 +758,26 @@ impl Pool for SimPool {
...
+            .filter_map(|(name, uuid, fs)| {
+                if fs_name.map(|n| *n == **name).unwrap_or(true) {
+                    Some(serde_json::to_string(&fs.record(name, *uuid)).map_err(|e| e.into()))
+                } else {
+                    None
+                }
+            })
+            .collect::<StratisResult<Vec<_>>>()
+            .map(|v| v.join("\n"))
+    }
```
jbaublitz, line 774: If I'm understanding this correctly, this means if fs_name is `None` it will return a bunch of JSON blobs separated by newlines for all filesystems present in the pool. Wouldn't it be better to put this all into a JSON object mapping UUIDs to metadata? I'm unaware of any JSON parser that can handle multiple separate blobs of JSON at once.

src/engine/strat_engine/thinpool/thinpool.rs
```diff
@@ -728,6 +728,37 @@ impl<B> ThinPool<B> {
...
+            .iter()
+            .filter_map(|fssave| {
+                if fs_name.map(|n| *n == fssave.name).unwrap_or(true) {
+                    Some(serde_json::to_string(fssave).map_err(|e| e.into()))
+                } else {
+                    None
+                }
+            })
+            .collect::<StratisResult<Vec<_>>>()
+            .map(|v| v.join("\n"))
```
jbaublitz, line 760: Same comment here as in the SimPool.

jbaublitz (changes requested):
There's one aspect of data structure that I think we need to rethink, but other than that this looks good.

src/engine/sim_engine/pool.rs
mulkieran: @jbaublitz That's exactly what it does. I was of two minds about this, but then I found that ```jq``` is fine with distinct blobs of JSON and I grew more satisfied with it.
```
$ cat junk.txt | jq
{}
{
  "key": 2
}
```

But your proposed solution is likely better.

jbaublitz: I don't know of any JSON parser in a programming language like Python that will do the same unfortunately.

jbaublitz (approved):
This looks better. Thanks!

mulkieran:
infrastructure problem on tests, ignoring

PR #3665, opened 2024-08-05
Issue stratisd 3663

mulkieran:


mulkieran:
It's not clear that we have a compelling reason to even keep the legacy pool prediction correct. I feel like we should drop it entirely in this PR.

mulkieran:
The tests succeed if the prediction script drops the code that subtracts the crypt metadata allocs...

mulkieran:
In theory, all tests will pass now.

mulkieran:
The one test failure is almost certainly anomalous.

PR #3671, opened 2024-08-12
Round up to the nearest eight sectors

mulkieran:
Related #3274


mulkieran:
@jbaublitz Can you confirm that rounding up to the nearest 8 sectors is the intention?

PR #3672, opened 2024-08-13
Allow logging in stratis-predict-usage utility

mulkieran:


mulkieran:
I think this should be entirely non-controversial. The size of the executable increases from 1.4 to 1.6 MiB, and the behavior does not change at all.

PR #3674, opened 2024-08-15
Enforce set invariant in destroy_filesystems

mulkieran:
Related https://github.com/stratis-storage/stratisd/issues/3621


mulkieran:
Probably trivial enough...

PR #3675, opened 2024-08-20
Assert snap name

mulkieran:


mulkieran:
@jbaublitz So, this method is invoked in just two places: https://github.com/stratis-storage/stratisd/blob/master/src/engine/strat_engine/pool/v1.rs#L1107 and https://github.com/stratis-storage/stratisd/blob/master/src/engine/strat_engine/pool/v2.rs#L1013 . In both methods, there is a check for whether the snapshot name is already present and if that name is present, then there's a return. So, as stratisd is now, this method can not be invoked with a snapshot name that is already in use. However, if the method is invoked directly with an in-use name, then execution proceeds and just here https://github.com/stratis-storage/stratisd/blob/master/src/engine/strat_engine/thinpool/thinpool.rs#L583 the newly created snapshot is inserted and the previously existing filesystem of the same name is removed from the table. With the code as it is, an assertion should always succeed. But, if there ever becomes some other way to invoke the thinpool ```snapshot_filesystem``` method then it might fail. Because the thing that ensures the precondition one stack frame away from the assertion of the pre-condition, I'm willing to change the assertion into an error return instead.

jbaublitz (approved):
I personally believe that this is not riskier than many of our other asserts, except for the side effect of clobbering the filesystem. That exception may be enough to warrant additional handling, but personally I am okay with the assert. Approving. If you decide to change it to an error, feel free to request my review again.

mulkieran:
Ok. I'll stick w/ the assert.

Issue #3650, opened 2024-07-16
Update to syn 2 in stratisd-proc-macros


jbaublitz:

jbaublitz:
@mulkieran Would you be able to do this? I'm not going to be able to get around to this for a long time because of other priorities. I was mostly making a note in case we had missed this.

mulkieran:
@jbaublitz I deferred addressing this because I'm unfamiliar with syn and then just forgot about it, so I'm not sure.

PR #3682, opened 2024-09-04
Increase syn dependency lower bound to 2.0.0

mulkieran:
Closes https://github.com/stratis-storage/stratisd/issues/3650

### DIFF
--- a/stratisd_proc_macros/src/lib.rs
+++ b/stratisd_proc_macros/src/lib.rs
@@ -8,16 +8,16 @@ use proc_macro::TokenStream;
 use proc_macro2::{Group, Span, TokenStream as TokenStream2, TokenTree};
 use quote::quote;
 use syn::{
-    parse, parse_str, punctuated::Punctuated, token::Comma, Attribute, Block, FnArg, Ident,
-    ImplItem, ImplItemMethod, Item, Lit, Meta, MetaList, NestedMeta, Pat, PatIdent, PatType, Path,
-    PathSegment, Receiver, ReturnType, Stmt, Token, Type, TypePath,
+    parse, punctuated::Punctuated, token::Comma, Attribute, Block, FnArg, Ident, ImplItem,
+    ImplItemFn, Item, LitStr, Pat, PatIdent, PatType, Path, PathSegment, Receiver, ReturnType,
+    Stmt, Token, Type, TypePath,
 };
 
 /// Add guard for mutating actions when the pool is in maintenance mode.
 ///
 /// This method adds a statement that returns an error if the pool is set
 /// to limit available actions.
-fn add_method_guards(method: &mut ImplItemMethod, level: Ident) {
+fn add_method_guards(method: &mut ImplItemFn, level: Ident) {
     let stmt = if let ReturnType::Type(_, ty) = &method.sig.output {
         if let Type::Path(TypePath {
             path: Path { segments, .. },
@@ -136,7 +136,7 @@ fn process_token_stream(token: TokenTree) -> TokenTree {
 
 /// Take the body of the method and wrap it in an inner method, replacing the
 /// body with a check for an error that limits available actions.
-fn wrap_method(f: &mut ImplItemMethod) {
+fn wrap_method(f: &mut ImplItemFn) {
     let wrapped_ident = Ident::new(
         format!("{}_wrapped", f.sig.ident.clone()).as_str(),
         Span::call_site(),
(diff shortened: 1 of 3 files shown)


mulkieran:
stratisd_proc_macros compiles...but panics.

mulkieran:
Now panicking in a better place...

mulkieran:
compiles

mulkieran:
If it passes, it should be good.

stratisd_proc_macros/src/lib.rs
```diff
@@ -193,59 +193,45 @@ fn wrap_method(f: &mut ImplItemMethod) {
...
-                    } else {
-                        panic!("pool_mutating_action attribute must be in form #[pool_mutating_action(\"REJECTION LEVEL\")]");
-                    }
+    if let Some((index, return_value)) = attrs
+        .iter()
+        .enumerate()
+        .filter(|(_, a)| a.path().is_ident("pool_mutating_action"))
+        .map(|(i, a)| {
+            if let Meta::List(MetaList { ref tokens, .. }) = a.meta {
+                if let TokenTree::Literal(litstr) = tokens.clone().into_iter().next().expect("stratisd code always supplies string literal argument to pool_mutating_actions") {
```
jbaublitz, line 202: I think this can be further destructured using [this enum](https://docs.rs/syn/latest/syn/enum.Lit.html). It will allow you to avoid converting to a String and then trimming the quotation marks, instead just getting the String value.

stratisd_proc_macros/src/lib.rs
```diff
@@ -193,59 +193,45 @@ fn wrap_method(f: &mut ImplItemMethod) {
...
+                        parse_str::<Ident>(litstr.to_string().trim_matches('"')).unwrap_or_else(|_| { panic!("{} is expected to be a valid identifier", litstr.to_string().trim_matches('"'))})
+                    )
+                } else {
+                    panic!("pool_mutating_action attribute must be in form #[pool_mutating_action(\"REJECTION LEVEL\")]")
                 }
+            } else {
+                panic!("pool_mutating_action attribute must be in form #[pool_mutating_action(\"REJECTION LEVEL\")]")
             }
+        })
+        .last() {
```
jbaublitz, line 214: What is the rationale for using last? Is it in order to preserve the same behavior as before where the last value would be the last value found in the loop? If so, I might request that we change this to ensure there is exactly one and use that value or return None.

jbaublitz (changes requested):

stratisd_proc_macros/src/lib.rs
mulkieran: That is the rationale. I was trying to avoid avoidable behavior changes. But I'm all for changing the behavior, so I'll do that while I'm here.

stratisd_proc_macros/src/lib.rs
mulkieran: litstr type is  TokenTree::Literal, which is a struct with private fields and a rather awkward set of methods in syn2. I think that syn2  probably expects that the normal way to encode that argument value is without the '"' in the first place, as a plain Ident, but I don't want to hit every spot where the ```pool_mutating_action``` annotation is used to change that, so I left it alone.

jbaublitz: Ah, I see. I think changing this to use an ident would be better after looking at the types. Would you be willing to do that? I think it would generally be better from the standpoint of upgrading to the new API. I'll also look into the recommended way of doing this because I have some other crates I maintain that need updates.

stratisd_proc_macros/src/lib.rs
mulkieran: I went with panic if encountered twice, I think that's more appropriate.

stratisd_proc_macros/src/lib.rs
mulkieran: I would, but I'ld like to postpone that until we start the next release. I'll file an issue.

mulkieran:
@jbaublitz Ok, it is ready again.

stratisd_proc_macros/src/lib.rs
jbaublitz: Okay sounds good!

jbaublitz (approved):

jbaublitz (changes requested):
I'm retracting my approval just to suggest one more possible avenue. Does [parse_nested_meta](https://docs.rs/syn/latest/syn/struct.Attribute.html#method.parse_nested_meta) do what we want here? Maybe not. I'll look into it more when I get back, but it looks promising.

mulkieran:
@jbaublitz The last commit tightens it up a bit. It can be tightened up even further by specifying parse_args with Ident type, but that requires changing all the uses every where to actually be Idents, not strings.

jbaublitz:
Would you mind holding this until I get back? I want to investigate further.

mulkieran:
@jbaublitz Yeah, that's fine.

mulkieran:
> @mulkieran I found a better way to do it:
> 
> ```
>             if let Ok(level) = a.parse_args::<LitStr>() {
>                 let err_str = level.value();
>                 (i, level.parse::<Ident>().expect(format!("Could not parse value {err_str} for rejection level as an Ident").as_str()))
>             } else {
>                 panic!("pool_mutating_action attribute must be in form #[pool_mutating_action(\"REJECTION LEVEL\")]")
>             }
> ```
> 
> This requires no changes to the way we currently provide the rejection level.

Yeah. That's a better way to parse. But the expect is still rejected, so I'll stick with panic.

jbaublitz:
What's rejecting it?

mulkieran:
> What's rejecting it?

Compiler doesn't like the ```as_str()``` invocation.

jbaublitz (approved):

mulkieran:
squashed.

PR #3686, opened 2024-09-13
Initial test that does nothing

mulkieran:
Related https://github.com/stratis-storage/project/issues/728


mulkieran:
Test ran until it ran out of TMT time, 20 min. Not surprising, since all it does is start and stop stratisd.

PR #3688, opened 2024-09-17
Avoid AttributeError if calling stop before start

mulkieran:


mulkieran:
test failures are due to Packit having some mild infrastructure problem.

PR #3700, opened 2024-10-14
Accept "adv" key as well as "thp" key

### LINKED ISSUE #3699: stratisd CreatePool method does not know about Clevis tang "adv" properties, only "thp".
We are seeing this test failure with Fedora 40 and stratisd 3.7.1-1.fc40 from udpates-testing:

https://cockpit-logs.us-east-1.linodeobjects.com/pull-0-a30106a4-20241012-013105-fedora-40-updates-testing/log.html#281-2

Creating a Stratis pool with a tang configuration fails with this error:

```
Either thumbprints for all Tang servers or a directive to trust all Tang servers is required
```

I believe this was introduced with #3613.

Cockpit uses the `adv` property to deliver the thumbprint to clevis instead of the `thp` one, but stratisd demands `thp` to be used.

The `adv` property is [documented](https://man.archlinux.org/man/clevis-encrypt-tang.1.en) by clevis as a proper alternative to `thp`, so I think stratisd should also recognize it.

mulkieran:
Closes #3699

### DIFF
--- a/src/engine/strat_engine/backstore/backstore/v1.rs
+++ b/src/engine/strat_engine/backstore/backstore/v1.rs
@@ -25,10 +25,7 @@ use crate::{
                 devices::UnownedDevices,
                 shared::BlockSizeSummary,
             },
-            crypt::{
-                back_up_luks_header, handle::v1::CryptHandle, interpret_clevis_config,
-                restore_luks_header,
-            },
+            crypt::{back_up_luks_header, handle::v1::CryptHandle, restore_luks_header},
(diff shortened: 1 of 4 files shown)


mulkieran:
@jbaublitz This is a port of #3701 .  It has to distribute some bits of that PR into v2 as well, and that's probably the only thing worth checking for correctness.

mulkieran:
rebased

jbaublitz (approved):

mulkieran:
rebased. We expect all tests to succeed.

PR #3708, opened 2024-10-17
Generalize traversing a Clevis config

mulkieran:


mulkieran:
There is some way to do the recursive call non-greedily...

mulkieran:
The challenge problem is to turn the ```sss``` ident into a ```FnMut``` and get it to compile again.

src/engine/strat_engine/crypt/shared.rs
```diff
@@ -204,14 +204,40 @@ pub fn clevis_info_from_metadata(
...
+///    )
+/// }
+fn traverse_clevis_config<T>(
     pin: &str,
     clevis_config: &Value,
     recursion_limit: u64,
-) -> StratisResult<bool> {
+    tang_func: &dyn Fn(&Map<String, Value>) -> StratisResult<T>,
+    tpm2_func: &dyn Fn(&Map<String, Value>) -> StratisResult<T>,
+    sss: &mut (&T, &mut dyn FnMut(T, T) -> StratisResult<T>),
```
jbaublitz, line 236: Can you explain why you made these references? I don't see that pattern used in most closure usage in the Rust code I've seen. My understanding is that `Fn()` itself already represents a reference.

src/engine/strat_engine/crypt/shared.rs
```diff
@@ -236,14 +255,16 @@ fn all_tang_configs_have_url_trust_info(
...
             if let Some(obj) = obj.get("pins").and_then(|val| val.as_object()) {
-                obj.iter().try_fold(true, |b, (pin, config)| {
-                    Ok(
-                        b && all_tang_configs_have_url_trust_info(
-                            pin,
-                            config,
-                            recursion_limit - 1,
-                        )?,
-                    )
+                obj.iter().try_fold(sss.0.clone(), |acc, (pin, config)| {
```
jbaublitz, line 258: Can you explain why you opted for a reference here? My brief analysis of the ownership here leads me to believe that a move and an owned data type would be okay here.

jbaublitz (commented):
I'm not entirely clear on some of the design choices here.

src/engine/strat_engine/crypt/shared.rs
mulkieran: I can change sss.0 so that it is not a reference. I've added a commit for that. But you can't move sss into the closure, because it is used after the closure returns. So, AFAICT, if you add a move, you have to be able to clone sss.

jbaublitz: I think this gets back to my other comment: why do all of these need to be references? I feel like ownership may remove the need for cloning.

mulkieran: ```sss``` is recursively passed and then used on the return value. I don't believe that it can be moved into the closure for that reason.

jbaublitz (commented):
Just to give a little bit of context, my main concern here is that we don't yet have a reason to generalize this function. Because of the complexity this adds to the traversal, I generally would want to see a case where this generalization would be needed to avoid code duplication. If this is the only code using the abstraction, I hesitate to introduce the complexity just yet as I believe this version is harder to read upfront.

After playing with the code a little bit I see why you have written it the way you have. I don't see a much better way after trying quite a few different variations.

Do you have any concrete plans to use this abstraction or are you doing it out of principle? If there's no current foreseen usage of this, I would vote that we hold off on this until we bump into a situation where it's necessary.

PR #3712, opened 2024-10-29
Support snapshot revert

### LINKED ISSUE #3621: Support issuing an instruction for Stratis filesystem merge and storing the instruction in filesystem metadata

### LINKED ISSUE #3622: Support merging a snapshot based on stored instruction in filesystem level metadata

mulkieran:
Closes #3621
Closes #3622

### DIFF
--- a/src/dbus_api/consts.rs
+++ b/src/dbus_api/consts.rs
@@ -67,6 +67,7 @@ pub const FILESYSTEM_CREATED_PROP: &str = "Created";
 pub const FILESYSTEM_SIZE_PROP: &str = "Size";
 pub const FILESYSTEM_SIZE_LIMIT_PROP: &str = "SizeLimit";
 pub const FILESYSTEM_ORIGIN_PROP: &str = "Origin";
+pub const FILESYSTEM_MERGE_SCHEDULED_PROP: &str = "MergeScheduled";
 
 pub const BLOCKDEV_INTERFACE_NAME_3_0: &str = "org.storage.stratis3.blockdev.r0";
 pub const BLOCKDEV_INTERFACE_NAME_3_1: &str = "org.storage.stratis3.blockdev.r1";
--- a/src/dbus_api/filesystem/filesystem_3_7/api.rs
+++ b/src/dbus_api/filesystem/filesystem_3_7/api.rs
@@ -4,11 +4,27 @@
 
 use dbus_tree::{Access, EmitsChangedSignal, Factory, MTSync, Property};
 
-use crate::dbus_api::{consts, filesystem::filesystem_3_7::props::get_fs_origin, types::TData};
+use crate::dbus_api::{
+    consts,
+    filesystem::filesystem_3_7::props::{
+        get_fs_merge_scheduled, get_fs_origin, set_fs_merge_scheduled,
+    },
+    types::TData,
+};
 
 pub fn origin_property(f: &Factory<MTSync<TData>, TData>) -> Property<MTSync<TData>, TData> {
     f.property::<(bool, String), _>(consts::FILESYSTEM_ORIGIN_PROP, ())
         .access(Access::Read)
         .emits_changed(EmitsChangedSignal::True)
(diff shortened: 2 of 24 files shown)


mulkieran:
Need to fix up tests so that they call StartPool correctly...

mulkieran:
Experienced timeouts on valgrind tests...

PR #3716, opened 2024-11-12
Modifications to integrity metadata reservation

jbaublitz:
Related to #3274 
Related to #3637


jbaublitz:
I'm going to add a few tests here for the parameters, but after that, this should be ready.

src/bin/utils/cmds.rs
```diff
@@ -84,6 +84,20 @@ pool is encrypted, setting this option has no effect on the prediction."),
...
+                         .long("integrity-tag-size")
+                         .num_args(1)
+                         .help("Size of the integrity checksums to be stored in the integrity metadata. The checksum size depends on the algorithm used for checksums. Units are bytes.")
+                         .next_line_help(true)
+                     )
+                     .arg(
+                         Arg::new("integrity_journal_size")
+                         .long("integrity-journal-size")
+                         .num_args(1)
+                         .help("Size of the integrity journal. Default is 128 MiB. Units are 4KiB blocks.")
```
mulkieran, line 99: Better just keep everything as bytes. I told Vojtech we would just do that, so we might as well stick with it.

mulkieran (changes requested):

src/dbus_api/api/manager_3_8/methods.rs
```diff
@@ -130,3 +145,103 @@ pub fn start_pool(m: &MethodInfo<'_, MTSync<TData>, TData>) -> MethodResult {
...
+            Ok(j) => Some((pin, j)),
+            Err(e) => {
+                let (rc, rs) = engine_to_dbus_err_tuple(&StratisError::Serde(e));
+                return Ok(vec![return_message.append3(default_return, rc, rs)]);
+            }
+        },
+        None => None,
+    };
+
+    let journal_size =
```
mulkieran, line 190: Better just to make units of journal size bytes everywhere.

mulkieran (changes requested):

mulkieran:

src/engine/strat_engine/engine.rs
```diff
@@ -494,9 +494,20 @@ impl Engine for StratEngine {
         name: &str,
         blockdev_paths: &[&Path],
         encryption_info: Option<&EncryptionInfo>,
+        journal_size: Option<Bytes>,
+        tag_size: Option<Bytes>,
     ) -> StratisResult<CreateAction<PoolUuid>> {
         validate_name(name)?;
         let name = Name::new(name.to_owned());
+        let rounded_journal_size = match journal_size {
```
mulkieran, line 502: I think we will have to validate the tag size as well, matching our expectations to the dm-integrity expectations. I don't think there is any thing but trouble at this time that could result from allowing odd-valued tag sizes.

mulkieran (changes requested):
One request so far...

src/engine/strat_engine/engine.rs
jbaublitz: I don't think this is as simple as checking for odd numbers. I'm not opposed to doing some validating, but just checking odd numbers feels incomplete. All of the values I've seen for tag size are powers of two. I think the thing to remember with tag size, however, is that that is only used to calculate the tag area which will be rounded to 4k regardless of whether the number is odd or not so I think we should either leave it as is or only support valid values of tags that integrity supports.

mulkieran: I think we should only support the values that dm-integrity itself supports. We can relax that restriction at some time later if there ever appears to be any value to doing so, which seems very unlikely. It is the case that our design is to do external metadata, so integrity support in stratisd will only use this value to calculate the size of the external metadata device. But, at some time, we hope that we will be able to support integrity on NPo2 devices. If there is code that is shared for the implementation of these, then tag sizes that integrity does not support will be more of a problem. My other concern is that, at present, the D-Bus spec restricts the tag size to a byte. That means max is 128. Is that enough, or should the size be doubled?

mulkieran: What I meant is: if we combine the powers of 2 restriction and the 8 digits restriction, the max possible value is 128. I also have not seen a larger tag size than 128 bytes. Is that fact enough to go on?

mulkieran: ok!

mulkieran:
@jbaublitz Speaking of tag size and also journal size, it is possible to make both 0. This still leave the, relatively small, superblock pre-allocated. It will legitimate for a user to specify 0 for both if they are using NPo2 devices and want to turn on integrity later and it is legitimate now if they never want to turn on integrity at all. So I believe that it is entirely acceptable to allow zero for both. I'll just add some more about that in the stratis-cli docs, unless you disagree.

jbaublitz:
> @jbaublitz Speaking of tag size and also journal size, it is possible to make both 0. This still leave the, relatively small, superblock pre-allocated. It will legitimate for a user to specify 0 for both if they are using NPo2 devices and want to turn on integrity later and it is legitimate now if they never want to turn on integrity at all. So I believe that it is entirely acceptable to allow zero for both. I'll just add some more about that in the stratis-cli docs, unless you disagree.

Do you mean we should leave things the way they are? If you believe this is best in order to support other features like NPo2 devices, I'm willing to leave it as it's a relatively small overhead for the device.

mulkieran:
> > @jbaublitz Speaking of tag size and also journal size, it is possible to make both 0. This still leave the, relatively small, superblock pre-allocated. It will legitimate for a user to specify 0 for both if they are using NPo2 devices and want to turn on integrity later and it is legitimate now if they never want to turn on integrity at all. So I believe that it is entirely acceptable to allow zero for both. I'll just add some more about that in the stratis-cli docs, unless you disagree.
> 
> Do you mean we should leave things the way they are? If you believe this is best in order to support other features like NPo2 devices, I'm willing to leave it as it's a relatively small overhead for the device.

I think that we should continue to allow 0 for both values.

Issue #3598, opened 2024-04-25
Feature request for multiple encrypt unlock mechanisms


jbaublitz:
A while back, @mvollmer requested the ability to add multiple keyslot bindings to encrypted pools. For example, instead of using the `sss` pin in Clevis, a user could simply add one keyslot that uses TPM and one keyslot that uses Tang. This seems like a feature that other users may be interested in.

Because we currently only support a static token slot in older versions of stratisd (token slot 1 for the keyring implementation and token slot 2 for Clevis), we will likely have to do some special handling in the new API for tokens slots 1 and 2. However, token slot 0 is no longer used in version 2 of the metadata and could be used for either.  In version 1 of the metadata, this would still be reserved for the Stratis token. All token slots higher than 2 would be available for either implementation.

jbaublitz:
A few initial thoughts in terms of design:
1. The unbind method could be collapsed into a single method (it would take a token slot as a parameter, remove the keyslot associated with the token, and then remove the token). Rebind and bind would still have to remain separate methods.
2. There could be an option to find the first free token slot if the user does not care which slot the token ends up in.
3. Special care will have to be taken to ensure that there is always 1 keyslot available to unlock the device.

mvollmer:
> A while back, @mvollmer requested the ability to add multiple keyslot bindings to encrypted pools.

"Request" is a strong word. :-) There are conceptual problems with using the "sss" pin, AFAICS: it can not be modified incrementally, and has to be constructed from scratch for every change. This requires all secrets that should go into its configuration. With LUKS slots, you only need one when adding a new slot, and none when removing a slot.

Also, I think there is no passphrase pin for tang. I can only find "null", "sss", "tang", "tpm2", and "tpm2plus". Thus clevis seems to be designed for non-interactive unlocking only.

(Please check all this if you think it might not be true. It has been a while and I might be wrong.)

mvollmer:
Another argument is of course: "This is how LUKS works, why would Stratis be different."

jbaublitz:
@mvollmer Apologies for the misinterpretation. :) I maybe thought you were more set on this as a feature than you were.

We're actually having some interesting conversations right now because of another point you raised around usability that I mostly agree with: having to remember the passphrase and the key description is not super usable.

Right now we're having some discussions about adding in the ability to specify multiple unlock mechanism, but this conflicts somewhat with our efforts to obscure the key description a bit.

[...]

PR #3720, opened 2024-11-20
Add infrastructure for supporting multiple token slots per unlock method

jbaublitz:
Related to #3598


jbaublitz:
@mulkieran As we discussed, I'll let you handle the tests that involve stratis-cli and the cert tests.

jbaublitz:
@mulkieran I'm going to say that udev tests passing is good enough to call https://github.com/stratis-storage/libcryptsetup-rs/pull/379 ready for merge. We should likely put out a new release for the locking updates and this method when you get a chance.

src/dbus_api/api/manager_3_8/api.rs
```diff
@@ -38,21 +38,23 @@ pub fn create_pool_method(f: &Factory<MTSync<TData>, TData>) -> Method<MTSync<TD
...
+        // a: array of zero or more elements
+        // b: true if a token slot is specified
+        // i: token slot
         // s: pin name
         // s: JSON config for Clevis use
         //
-        // Rust representation: (bool, (String, String))
-        .in_arg(("clevis_info", "(b(ss))"))
+        // Rust representation: Vec<((bool, u32), String, String)>
+        .in_arg(("clevis_info", "a((bi)ss)"))
```
mulkieran, line 57: ^bi^bu   I think.

src/dbus_api/api/manager_3_8/api.rs
```diff
@@ -38,21 +38,23 @@ pub fn create_pool_method(f: &Factory<MTSync<TData>, TData>) -> Method<MTSync<TD
...
+        // i: token slot
         // s: key description
         //
-        // Rust representation: (bool, String)
-        .in_arg(("key_desc", "(bs)"))
-        // Optional Clevis information for binding on initialization.
-        // b: true if the pool should be encrypted and able to be unlocked
-        // using Clevis.
+        // Rust representation: Vec<((bool, u32), String)>
+        .in_arg(("key_desc", "a((bi)s)"))
```
mulkieran, line 48: ^bi^bu I think.

mulkieran (changes requested):
Two small request...

jbaublitz:
The usage of `i` was pretty widespread but I think I've tracked down and changed all of the places I used it.

src/dbus_api/pool/pool_3_8/methods.rs
```diff
@@ -0,0 +1,399 @@
...
+        }
+    };
+    Ok(vec![msg])
+}
+
+pub fn bind_keyring(m: &MethodInfo<'_, MTSync<TData>, TData>) -> MethodResult {
+    let message: &Message = m.msg;
+    let mut iter = message.iter_init();
+    let key_desc_str: String = get_next_arg(&mut iter, 0)?;
+    let token_slot_tuple: (bool, u32) = get_next_arg(&mut iter, 1)?;
```
mulkieran, line 149: Expectations for token_slot_tuple differ from published api signature which is ```(b(bu))```

mulkieran (changes requested):
Found one disagreement between published D-Bus API and dbus method. I may be able to find one more...

src/dbus_api/pool/pool_3_8/methods.rs
jbaublitz: The implementation uses the correct data type. I've fixed the signature.

src/dbus_api/pool/mod.rs
```diff
@@ -444,7 +445,7 @@ pub fn get_pool_properties(
             consts::POOL_UUID_PROP => uuid_to_string!(pool_uuid),
             consts::POOL_ENCRYPTED_PROP => shared::pool_enc_prop(pool),
             consts::POOL_AVAIL_ACTIONS_PROP => shared::pool_avail_actions_prop(pool),
-            consts::POOL_KEY_DESC_PROP => shared::pool_key_desc_prop(pool),
+            consts::POOL_KEY_DESC_PROP => shared::pool_key_descs_prop(pool),
```
mulkieran, line 448: Right here you need to change ```POOL_KEY_DESC_PROP``` to ```POOL_KEY_DESCS_PROP``` and I think similar things on line 449 also.

mulkieran (changes requested):
Problem with GetManagedObjects result...

src/dbus_api/api/manager_3_8/api.rs
```diff
@@ -38,21 +38,23 @@ pub fn create_pool_method(f: &Factory<MTSync<TData>, TData>) -> Method<MTSync<TD
...
+        // a: array of zero or more elements
+        // b: true if a token slot is specified
+        // i: token slot
         // s: pin name
         // s: JSON config for Clevis use
         //
-        // Rust representation: (bool, (String, String))
-        .in_arg(("clevis_info", "(b(ss))"))
+        // Rust representation: Vec<((bool, u32), String, String)>
+        .in_arg(("clevis_info", "a((bu)ss)"))
```
mulkieran, line 57: ^```(bu)ss```^```(bu)(ss)``` OR you can modify EncryptionInfos struct in dbus_api/api/manager_3_8/methods.rs to match this spec as it is.

This is an example of pesky bug we've run into before, where the D-Bus API silently drops badly formed data if it is marshalled into an array due to a dbus-rs bug.

mulkieran (changes requested):
One more D-Bus bug found...still working.

jbaublitz:
@mulkieran This should be all set now!

src/bin/stratis-min/stratis-min.rs
```diff
@@ -42,18 +41,8 @@ fn parse_args() -> Command {
...
-                            .num_args(1)
-                            .value_parser(PossibleValuesParser::new(UnlockMethod::VARIANTS)),
-                    )
-                    .arg(
-                        Arg::new("prompt")
-                            .long("prompt")
-                            .num_args(0)
-                            .requires("unlock_method"),
-                    ),
+                    .arg(Arg::new("token_slot").long("token_slot").num_args(1))
```
mulkieran, line 44: For ```token_slot``` arguments should be able to use ```.value_parser(clap::value_parser!(u32))``` in the parser, i.e., here, and then drop the extra handling when you process the parsed arg.

src/dbus_api/pool/mod.rs
```diff
@@ -444,8 +445,8 @@ pub fn get_pool_properties(
             consts::POOL_UUID_PROP => uuid_to_string!(pool_uuid),
             consts::POOL_ENCRYPTED_PROP => shared::pool_enc_prop(pool),
             consts::POOL_AVAIL_ACTIONS_PROP => shared::pool_avail_actions_prop(pool),
-            consts::POOL_KEY_DESC_PROP => shared::pool_key_desc_prop(pool),
-            consts::POOL_CLEVIS_INFO_PROP => shared::pool_clevis_info_prop(pool),
+            consts::POOL_KEY_DESCS_PROP => shared::pool_key_descs_prop(pool),
+            consts::POOL_CLEVIS_INFOS_PROP => shared::pool_clevis_info_prop(pool),
```
mulkieran, line 449: ^```pool_clevis_info_prop```^```pool_clevis_infos_prop```

mulkieran (changes requested):
One more D-Bus problem and a request regarding stratis-min parser.

jbaublitz:
@mulkieran I just fixed a bug where creating a pool with Clevis only failed. Please update any testing to use the most recent version with the bug fix.

src/engine/sim_engine/pool.rs
```diff
@@ -364,27 +377,57 @@ impl Pool for SimPool {
...
+                            return Err(StratisError::Msg(format!(
+                                "Key slot {t} is already in use by a Clevis token"
+                            )));
+                        }
+                    }
+                } else {
+                    t
+                }
+            }
+            OptionalTokenSlotInput::None => encryption_info.free_token_slot(),
```
mulkieran, line 405: Right here, it no token specified, a new token slot is found. To be consistent, since this is keyring, should check against existing encryption info and return Identity if already bound with this key.

mulkieran (changes requested):
Just the one thing we discussed in standup so far...

src/engine/sim_engine/pool.rs
```diff
@@ -420,27 +463,33 @@ impl Pool for SimPool {
...
         })
     }
 
     fn rebind_keyring(
         &mut self,
+        token_slot: Option<u32>,
         new_key_desc: &KeyDescription,
     ) -> StratisResult<RenameAction<Key>> {
-        let encryption_info = match pool_enc_to_enc!(self.encryption_info()) {
+        let token_slot = token_slot.ok_or_else(|| {
```
mulkieran, line 488: I believe that you can leave this check out and the sim engine will align better w/ what the strat engine does. As far as I can tell, the strat engine, if given no token slot, will grab the key description associated with the top token slot and do an idempotency check against that. This seems consistent w/ our other decisions.

mulkieran (changes requested):
Still working...

src/engine/sim_engine/pool.rs
```diff
@@ -449,25 +498,35 @@ impl Pool for SimPool {
...
+            )))
+        }
     }
 
     // The sim engine does not store token info so this method will always return
     // RenameAction::Identity.
-    fn rebind_clevis(&mut self) -> StratisResult<RegenAction> {
-        let encryption_info = match pool_enc_to_enc!(self.encryption_info()) {
+    fn rebind_clevis(&mut self, token_slot: Option<u32>) -> StratisResult<RegenAction> {
+        let token_slot = token_slot.ok_or_else(|| {
```
mulkieran, line 525: I believe that it makes sense to leave out this check here also.

mulkieran (changes requested):
This is the last one I found in testing...

src/engine/sim_engine/pool.rs
jbaublitz: After looking into this, I remembered this was intentional and now recall why I made this decision. This would require reworking the API to pass the key store into this method. I don't think we should rework the API just for the sim_engine, but I'm open to other suggestions.

mulkieran (changes requested):
I'm noting one more discrepancy of the sim with strat engine. The strat engine behavior seems correct to me:
* In the strat engine, it is forbidden to unbind the last remaining binding. In the sim engine, you can remove the last remaining binding, which is a little awkward.

(If, when unbinding, you specify a non-existant token slot, stratid will return Ok(false) and stratis-cli will report that no change occurred. This is true for sim and strat engine and that seems correct to me.)

This is with keyring, not tested with Clevis yet.

jbaublitz:
@mulkieran This should be ready to go again.

jbaublitz:
@mulkieran Just realized I forgot to update the introspection data. That is now resolved.

src/engine/strat_engine/crypt/shared.rs
```diff
@@ -100,9 +101,10 @@ fn key_desc_to_passphrase(key_description: &KeyDescription) -> StratisResult<Siz
 // in memory (this is automatically done when formatting a LUKS2 device).
 pub fn add_keyring_keyslot(
     device: &mut CryptDevice,
+    token_slot: Option<u32>,
```
mulkieran, line 104: Need to update Postcondition to match argument.

Also, I believe that the first error message that is returned, "Failed to initialize keyslot with existing Clevis key" becomes slightly stale with this PR. It should be something more like "Failed to initialize keyslot with existing key", as it is no longer guaranteed to be a Clevis key that you tried to use.

mulkieran (changes requested):
One minor request...

src/engine/strat_engine/crypt/shared.rs
jbaublitz: This should be resolved.

jbaublitz:
@mulkieran I've updated the commit with the unit tests with a list of everything tested. Let me know if you want any changes.

mulkieran:
Clevis test checks out, this is the error encountered by the CLI:

```
# date; stratis pool bind tang
--trust-url spool1 http://127.0.0.1/
Fri Jan 24 13:52:33 EST 2025
Execution failed:
stratisd failed to perform the operation that you requested. It
returned the following information via the D-Bus: ERROR: Command
failed: cmd: "/usr/bin/clevis" "luks" "bind" "-y" "-d" "/dev/dm-1"
"tang" "{\"url\":\"http://127.0.0.1/\"}", exit reason: 1 stdout:
stderr:
Unable to find a free slot in /dev/dm-1
Error adding new binding to /dev/dm-1
```

mulkieran:
key error:

```
# date; stratis pool bind keyring
spool1 testkey33
Fri Jan 24 15:47:24 EST 2025
Execution failed:
stratisd failed to perform the operation that you requested. It
returned the following information via the D-Bus: ERROR: Cryptsetup
error: IO error occurred: Invalid argument (os error 22).
```

This error is more nicely formatted than the Clevis error, but less informative:

The logs are actually more helpful:

```
[2025-01-24T20:47:24Z ERROR
stratisd::engine::strat_engine::crypt::shared] All key slots full.
[2025-01-24T20:47:24Z WARN
stratisd::engine::strat_engine::crypt::shared] Failed to initialize
new keyslot with existing key; failed with error: IO error occurred:
Invalid argument (os error 22)
```

The first line must be from libcryptsetup itself, the second is from stratisd.

We have a general rule that we never do in the CLI what can be done in the daemon. But likely we should give the user a more informative error in both cases. stratisd could check the existing encryption info and return that it looks like the limit for this device has been reached. It could do this for both Clevis and key actions. I think that would probably be better than anything the CLI could do.

mulkieran:
@jbaublitz Plz rebase.

PR #3740, opened 2025-01-07
Fix bug in sim engine

jbaublitz:
This commit fixes a bug where an error starting a pool in the sim engine will cause the pool to disappear instead of being readded to stopped pools.


mulkieran:
Every TMT test has passed; GitHub is just unaware....

PR #3746, opened 2025-01-22
Increase itertools dependency lower bound to 0.14.0

mulkieran:


mulkieran:
Should succeed on rawhide, fail everywhere else.

PR #3747, opened 2025-01-22
Increase libcryptsetup-rs dependency lower bound to 0.12.0

mulkieran:
Related https://github.com/stratis-storage/project/issues/748


mulkieran:
rawhide is now building fine. The other Fedora releases will have to wait another day anyhow, so we can push this one as soon as it is ready.

PR #3751, opened 2025-01-31
Show only officially supported commands in help text

mulkieran:


mulkieran:
The new help looks like this:

```
$ ./target/release/stratisd-tools --help
Top-level command used to invoke one of a set of stratisd tools

Usage: stratisd-tools <EXECUTABLE>

Arguments:
  <EXECUTABLE>  

Options:
  -h, --help     Print help
  -V, --version  Print version

Executables:
* stratis-dumpmetadata
```

Whereas the current help looks like this:

```
$ ./target/release/stratisd-tools --help
Top-level command used to invoke one of a set of stratisd tools

Usage: stratisd-tools <EXECUTABLE>

Arguments:
  <EXECUTABLE>  [possible values: stratis-checkmetadata, stratis-dumpmetadata, stratis-printmetadata]

Options:
  -h, --help     Print help
  -V, --version  Print version
```

This change allows us to hide the tools that we are not documenting or releasing yet. Also, in the future, we are likely to release more of these tools, just as thinp has done and the list if it was just a line after ```<EXECUTABLE>``` would get long an jumbly. This way, it is possible to have a nicely formatted list at the the end.

The executable can still  be invoked in exactly the same way.

bgurney-rh (approved):

PR #3752, opened 2025-01-31
Make stratis-legacy-pool a command of stratisd-tools

mulkieran:


bgurney-rh (approved):

Cargo.toml
```diff
@@ -295,11 +291,11 @@ engine = [
 ]
 default = ["dbus_enabled", "engine"]
 dbus_enabled = ["dep:dbus", "dep:dbus-tree"]
-extras = ["dep:pretty-hex"]
+extras = ["dep:pretty-hex", "test_extras"]
 min = ["dep:termios"]
 systemd_compat = ["dep:bindgen"]
 udev_scripts = ["dep:data-encoding"]
-test_extras = ["engine"]
+test_extras = []
```
jbaublitz, line 298: Are we leaving this for packaging?

jbaublitz (changes requested):
Could you also update the git hooks accordingly?

Cargo.toml
mulkieran: I can also remove it. I think there is no problem with that.

mulkieran: The reasoning behind leaving it in wasn't that strong.

mulkieran: The one good reason for leaving this in was not having to update the spec file. So, yes, it was for packaging convenience. If we change it, we change the spec file. But that is no big deal, really, except that it could break the Packit/TMT structure briefly. But, we can work around that.

mulkieran: Actually, I don't even think it was for packaging. I'll see about removing it.

mulkieran:
Packages worked. Tests would have failed anyway, because of necessity to respond yes.

mulkieran:
@jbaublitz This is ready again.

jbaublitz (approved):

PR #3769, opened 2025-02-24
Fix bug in backstore setup with cache

### LINKED ISSUE #3768: Assertion can be triggered when a pool with a cache device and encryption fails to be set up
The two problems that need to occur for this to happen on pool start are:
* No unlock mechanism can be available
* There is a cache

This will not take down stratisd but it does cause some state problems internally and gives a difficult to understand error message.

jbaublitz:
Closes #3768

### DIFF
--- a/tests/client-dbus/tests/udev/test_start.py
+++ b/tests/client-dbus/tests/udev/test_start.py
@@ -1,0 +1,118 @@
+# Copyright 2021 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Test starting additional stratisd processes.
+"""
+
+# isort: STDLIB
+import os
+
+# isort: LOCAL
+from stratisd_client_dbus import Manager, Pool, get_object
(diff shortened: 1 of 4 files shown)


jbaublitz:
I want to check on V1 too, but will need to do that a little bit later.

jbaublitz:
I've confirmed that this seems to be the result of encryption management in the backstore which does not effect V1. V1 should be able to remain unchanged.

mulkieran (changes requested):
Looks correct, but I think we should take the opportunity to add an additional test. I think it would be best at the pool level, so probably it should be part of our Python infrastructure.

mulkieran (changes requested):
t would be nice if when you added a test it would just be automatically run but that is not the case at this time. I'll see if it is possible to consolidate these tests, so that actually does happen.

mulkieran:
@jbaublitz On second thoughts, probably just better to add a new target in tests/client-dbus/Makefile and then in tests-fmf/python.fmf.

jbaublitz:
I'm doing some investigation into unittest and I'm finding that the discover verb may actually do what we want. I'm going to do some further testing and I'll report back.

jbaublitz:
I went with your suggestion as it's quicker. This should run the tests now.

tests/client-dbus/tests/udev/test_start.py
```diff
@@ -0,0 +1,118 @@
...
+
+class TestFailedStart(UdevTest):
+    """
+    Test creating an encrypted pool and stopping it, removing the key from the keyring,
+    and then start it.
+
+    This is a regression test for a bug.
+    """
+
+    def failed_start_regression_test(self):
```
mulkieran, line 37: Change the name from ```failed_start_regression_test``` to ```test_failed_start...```. unittest's standard naming convention expects every actual test to have a ```test_``` prefix.

mulkieran (changes requested):

mulkieran (approved):

PR #3775, opened 2025-02-26
Fix jsonrpc module to work with metadata V2

### LINKED ISSUE #3774: JSON RPC has gone out of sync with V2 of the metadata
The verbs exposed in stratis-min do not necessarily reflect how we want perform unlocking in early boot with V2 of the metadata.

The `is-bound` verb should be removed, the `is-encrypted` verb should be updated to handle V2, and the scripts in early boot should be updated to use this new CLI.

jbaublitz:
Closes #3774


jbaublitz:
@drckeefe Feel free to take a look at whether this fixes your issue once the build completes.

bmr-cymru (approved):

mulkieran (commented):
This looks correct to me, but I haven't tested it yet. Also, if it is really the case that if ```metadata_version``` is ```V1``` then ```info``` is ```Some``` and ```features``` is ```None```, and vice-versa if ```metadata_version``` is ```V2```, then we might be able to enforce that in the StoppedPoolInfo struct in some robust way.

jbaublitz:
> This looks correct to me, but I haven't tested it yet. Also, if it is really the case that if `metadata_version` is `V1` then `info` is `Some` and `features` is `None`, and vice-versa if `metadata_version` is `V2`, then we might be able to enforce that in the StoppedPoolInfo struct in some robust way.

I think this could potentially be done through an enum, but I'm not sure that would get us much, as we already duplicate that information in the metadata version. Another way to potentially enforce this is to create a new enum with V1 and V2 variants that contain the data for either features or info to avoid duplicating information that we'll have to check in two places.

mulkieran:
Superceded by #3779

PR #3777, opened 2025-02-27
Remove allows of clippy::needless_collect

mulkieran:


jbaublitz:
@mulkieran Did this get fixed in clippy?

mulkieran:
> @mulkieran Did this get fixed in clippy?

Yes. It appears that a lot of people reported false positives on this one, including for this particular case.

jbaublitz:
Okay, great! Glad they did something about it.

mulkieran:
@jbaublitz At the time the mutability constraint required the intermediate collect the Rust version we were using was 1.58. At present even our lowest supported Rust version is 1.74.0. So I think the mutability constraint must have been relaxed in the interim. Since I'm proposing actually changing code I'm putting this one up for your review.

jbaublitz (approved):

PR #3595, opened 2024-04-11
Add password verification for stratis-min

jbaublitz:
Closes #3593

PR #3783, opened 2025-03-05
Use password verification when setting passphrase

### LINKED ISSUE #3593: Do password verification on user-entered password value in stratis-min
We do this in stratis-cli, because it is a good idea and we would like stratis-min to match.

mulkieran:
Closes #3593 
Related https://github.com/stratis-storage/stratis-cli/pull/1129
Supercedes https://github.com/stratis-storage/stratisd/pull/3595

### DIFF
--- a/src/jsonrpc/client/key.rs
+++ b/src/jsonrpc/client/key.rs
@@ -20,7 +20,7 @@ pub fn key_set(key_desc: KeyDescription, keyfile_path: Option<&str>) -> StratisR
             do_request!(KeySet, key_desc; file.as_raw_fd())
         }
         None => {
-            let password = prompt_password()?
+            let password = prompt_password(true)?
                 .ok_or_else(|| StratisError::Msg("Password provided was empty".to_string()))?;
 
             let (read_end, write_end) = pipe()?;
--- a/src/jsonrpc/client/pool.rs
+++ b/src/jsonrpc/client/pool.rs
@@ -34,7 +34,7 @@ pub fn pool_start(
     prompt: bool,
 ) -> StratisResult<()> {
     if prompt {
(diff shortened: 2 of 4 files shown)


mulkieran:
Sample interaction:

```
# stratis-min key set my-key --capture-key
Enter passphrase followed by return: 
Verify passphrase: 
# stratis-min key list
Key Description   
my-key
# stratis-min pool create pn /dev/bn --key-descs my-key:0
# stratis-min pool stop --name pn
# stratis-min key unset my-key
# stratis-min pool start dff64558-a6a3-4f54-b5f7-0ec785ae091d --prompt --token-slot any
Enter passphrase followed by return:
```

jbaublitz (changes requested):
The reason I paused this work and did not continue is that it requires changes to the fstab script, the dracut code, and also testing for all of those things to make sure we haven't broken anything. If you want to prioritize all of that right now, we can do it, but we need that to be in this PR too.

mulkieran:
> The reason I paused this work and did not continue is that it requires changes to the fstab script, the dracut code, and also testing for all of those things to make sure we haven't broken anything. If you want to prioritize all of that right now, we can do it, but we need that to be in this PR too.

As far as I understand, it doesn't require any changes to the fstab script or other things you mention. It would only require such changes if it were changing the behavior of the stratis-min _pool start_ call, by requiring the passphrase twice in that circumstance. But the point of this PR is not to do that, but only to require password verification on the _key create_ call, where it is customary to do password verification, and which is not, as far as I understand, made use of in any of our fstab or dracut scripts.

jbaublitz (approved):
Okay, I see what you mean. Approved.

PR #3784, opened 2025-03-05
Trim password before checking for emptiness

mulkieran:


mulkieran:
I think that cockpit failure is unlikely to be our fault. The change here is that stratis-min will send even an empty passphrase on the IPC mechanism to try to unlock a pool that required to be unlocked in order to mount a filesystem specified in fstab. It did that in most circumstances before, but now it should do it in every circumstance.

src/jsonrpc/client/pool.rs
```diff
@@ -34,8 +34,7 @@ pub fn pool_start(
     prompt: bool,
 ) -> StratisResult<()> {
     if prompt {
-        let password = prompt_password(false)?
-            .ok_or_else(|| StratisError::Msg("Password provided was empty".to_string()))?;
+        let password = prompt_password(false)?;
```
jbaublitz, line 37: Is it intentional allowing an empty password on pool start?

jbaublitz (commented):
This seems to be the same behavior as stratis-cli which I think it probably for the best, but I'm realizing with the changes you made here that a passphrase of all spaces would be considered the empty passphrase. I'm not sure that's a bad thing, but I think this could raise a problem for a passphrase that starts with a space when passed in via the D-Bus vs. either of our CLIs. As I recall, all of the password work happens on the CLI side, so if a user were to pass in a passphrase with a space at the beginning, I'm pretty sure that would behave differently in the CLI and the daemon.

mulkieran:
> This seems to be the same behavior as stratis-cli which I think it probably for the best, but I'm realizing with the changes you made here that a passphrase of all spaces would be considered the empty passphrase. I'm not sure that's a bad thing, but I think this could raise a problem for a passphrase that starts with a space when passed in via the D-Bus vs. either of our CLIs. As I recall, all of the password work happens on the CLI side, so if a user were to pass in a passphrase with a space at the beginning, I'm pretty sure that would behave differently in the CLI and the daemon.

I had to check, but it looks like stratis-cli strips all the whitespace already. So this change would bring stratis-min more into alignment with stratis-cli.

src/jsonrpc/client/pool.rs
mulkieran: Yes. However it is intended to address a very strange behavior that was reported to me but that I have not been able to observe myself. I may withdraw it.

jbaublitz:
> > This seems to be the same behavior as stratis-cli which I think it probably for the best, but I'm realizing with the changes you made here that a passphrase of all spaces would be considered the empty passphrase. I'm not sure that's a bad thing, but I think this could raise a problem for a passphrase that starts with a space when passed in via the D-Bus vs. either of our CLIs. As I recall, all of the password work happens on the CLI side, so if a user were to pass in a passphrase with a space at the beginning, I'm pretty sure that would behave differently in the CLI and the daemon.
> 
> I had to check, but it looks like stratis-cli strips all the whitespace already. So this change would bring stratis-min more into alignment with stratis-cli.

I agree that it would bring it more into alignment with stratis-cli, but I'm not sure we're doing the right thing in stratis-cli because of the case I highlighted above. The end result is that the same passphrase entered into stratis-cli vs. the D-Bus API will work in one and not in the other. That seems like it will be confusing for users.

mulkieran:
> > > This seems to be the same behavior as stratis-cli which I think it probably for the best, but I'm realizing with the changes you made here that a passphrase of all spaces would be considered the empty passphrase. I'm not sure that's a bad thing, but I think this could raise a problem for a passphrase that starts with a space when passed in via the D-Bus vs. either of our CLIs. As I recall, all of the password work happens on the CLI side, so if a user were to pass in a passphrase with a space at the beginning, I'm pretty sure that would behave differently in the CLI and the daemon.
> > 
> > 
> > I had to check, but it looks like stratis-cli strips all the whitespace already. So this change would bring stratis-min more into alignment with stratis-cli.
> 
> I agree that it would bring it more into alignment with stratis-cli, but I'm not sure we're doing the right thing in stratis-cli because of the case I highlighted above. The end result is that the same passphrase entered into stratis-cli vs. the D-Bus API will work in one and not in the other. That seems like it will be confusing for users.


> > > This seems to be the same behavior as stratis-cli which I think it probably for the best, but I'm realizing with the changes you made here that a passphrase of all spaces would be considered the empty passphrase. I'm not sure that's a bad thing, but I think this could raise a problem for a passphrase that starts with a space when passed in via the D-Bus vs. either of our CLIs. As I recall, all of the password work happens on the CLI side, so if a user were to pass in a passphrase with a space at the beginning, I'm pretty sure that would behave differently in the CLI and the daemon.
> > 
> > 
> > I had to check, but it looks like stratis-cli strips all the whitespace already. So this change would bring stratis-min more into alignment with stratis-cli.
> 
> I agree that it would bring it more into alignment with stratis-cli, but I'm not sure we're doing the right thing in stratis-cli because of the case I highlighted above. The end result is that the same passphrase entered into stratis-cli vs. the D-Bus API will work in one and not in the other. That seems like it will be confusing for users.

The motivation for passing the empty string when starting a pool is gone, so I removed that code. If you look at the code, you'll see that my code only makes a very minor change in the direction of correctness. Before, all passphrases gathered by systemd-ask-passphrase were considered non-empty, because they included the newline character. Now, if a passphrase is the empty string + a newline + (optional) whitespace, it is considered an empty passphrase and ```Ok(None)``` is returned.

But regardless of the change in this PR, whitespace is _always_ stripped from the passphrase, at both ends.

It turns out that there was a complicating factor that might have obscured the small bug that this PR fixes. When systemd-ask-passphrase gets its passphrase indirectly through plymouth, if the user just presses return, systemd-ask-passphrase receives an error code and does not process the passphrase at all. The effect is different if systemd-ask-passphrase is used directly in my tty. In that case, it receives and processes the empty string.

With regard to whether it is a good idea to strip out whitespace as well as newline, I think that is a different issue. We have been stripping out whitespace in both stratis-min and stratis-cli for a long time. We could change that easily, by just rstripping a newline in each instead. If that's what you would prefer, it's probably just another one line change.

I can see about adding that.

mulkieran:
@jbaublitz Ok. I switched to just stripping the newline.

jbaublitz (approved):

PR #3786, opened 2025-03-05
Reduce crypt log level error to warn

mulkieran:


jbaublitz (commented):
Could you provide some additional info on this?

Maybe it would be helpful to explain my rationale for this. There were a good number of times when I was debugging the cryptsetup code where I wanted to see what the underlying cryptsetup error was. The reason I put it as error is it could be helpful to have the underlying cause jump out as an error if there is an underlying cryptsetup problem. Furthermore, cryptsetup is very, very sparing in what they mark as error based on my review of the source code, so I'm not sure we actually want this if we want the few times that the error is designated to jump out at us.

jbaublitz:
Would you be willing to compromise on `warn!`? I think it's worth highlighting separately from regular information that we display at the `info!` level which includes a lot of successful operations.

mulkieran:
> Would you be willing to compromise on `warn!`? I think it's worth highlighting separately from regular information that we display at the `info!` level which includes a lot of successful operations.

Ok. I'll go w/ warn!.

mulkieran:
@jbaublitz this is ready again.

jbaublitz (approved):

PR #3789, opened 2025-03-10
Fix Makefile so that auditable checks can be run

mulkieran:


mulkieran:
f40 failures are infrastructure failures, we can ignore.

PR #3791, opened 2025-03-11
Also send signal if token slot is None

mulkieran:


jbaublitz:
I need some time to convince myself that we don't need to check the `Legacy` variant in the create actions as well. That would be the logical equivalent of `None` in rebind so I just need to take a look. I also just want to verify `unbind`.

jbaublitz (changes requested):
I think `bind` is safe. It checks whether a binding already exists for the given binding method and if it does, it sets the boolean for older interfaces to `false`. This seems correct.

I think that `unbind` has the same problem as rebind as it does a similar lowest token slot calculation, omitting the `None` variant which in r8 will default to the lowest token slot.

bgurney-rh (approved):

mulkieran:
TMT tests are hung, and I don't think they'll ever complete, so we will  have to ignore.

Issue #3348, opened 2023-05-22
stratis-fstab-setup@.service might use the network but does not wait for it


mvollmer:
For a pool with NBDE, stratis-fstab-setup will run clevis during boot, as it should. For clevis to have a chance to work, the network needs to be up enough for the the tang server to be reachable, but the stratis-fstab-setup@.service units don't have any dependency on network-online.target or similar.

Putting "_netdev" into the fstab entry doesn't help either: it delays the actual mounting of the filesystem, but it doesn't delay starting of stratis-fstab-setup@.service, which is started as early as allowed by its own dependencies.

I think stratis-fstab-setup (the script) should probably just sit in a loop and retry clevis on every change to the network status, whenever there might be a chance for it to work. (But it's not clear when to give up, hmm.)

 Just waiting for network-online.target and then running clevis once might be enough for most cases, but please see https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/

jbaublitz:
Just for reference, we use network-online.target for the root fs. CoreOS wanted to write their own handling of Stratis in the root filesystem as a result of this. We could potentially add an exponential backoff retry mechanism instead for both as suggested by CoreOS, but a stopgap that would probably be workable sooner would be to add a requirement for `network-online.target`. What would be your preference?

mvollmer:
> Just for reference, we use network-online.target for the root fs.

I don't immediately see how this is possible... Doesn't network-online.target indirectly depend on -.mount?  Or does this all happen inside the initrd?

> What would be your preference?

Something like this in stratis-fstab-setup, maybe:

```
if $(stratis-min pool is-stopped "$POOL_UUID"); then
	if $(stratis-min pool is-bound "$POOL_UUID"); then
		while ! systemctl is-active network-online.target; do
		  	echo Waiting for network
			sleep 1
		done
		if ! stratis pool start --unlock-method=clevis --uuid "$POOL_UUID"; then
			echo Failed to start pool with UUID $POOL_UUID using Csslevis. >&2
			exit 1
		fi
       else ...
```

(Note that by the time network-online.target has been reached, stratis-min has stopped working...)

jbaublitz:
> > Just for reference, we use network-online.target for the root fs.
> 
> I don't immediately see how this is possible... Doesn't network-online.target indirectly depend on -.mount? Or does this all happen inside the initrd?

[...]

PR #3811, opened 2025-04-02
Support key description and Clevis in StoppedPools and metadata

### LINKED ISSUE #3805: StoppedPools property in r8 doesn't tell what kind of unlocking method should be tried for V2 pools
The StoppedPools property has "clevis_info" and "key_description" fields for V1 pools, but not for V2 pools. These fields were useful to know what unlock methods to try when starting the pool. For example, if there was no key description, Cockpit would never ask for a passphrase when starting a V1 pool.

With V2 pools, I am not sure how to achieve the same thing. How can we know that it is pointless to try with a passphrase?

I.e., if you have a pool that has only a clevis "tang" slot, but no passphrase slot, we would like to call StartPool once without a file descriptor (and without a token slot). When this fails, we would like to know whether we should ask the user for a passphrase or not.

Conversely, if there is no "clevis" slot, we would like to not even try to call StartPool without a file descriptor. Doing it anyway is mostly harmless UI-wise, but still.

Also, how can a API client use the "token slot" parameter of StartPool when there is no way to know which token slots are used how when a pool is stopped?

Related:

 - How would we create V1 pools when testing with Stratis 3.8?  Working with V1 pools is a very important feature of Stratis 3.8, and we will likely have to have code paths in Cockpit that are specific to managing V1 pools. We would like to test that in our CI. It would be great if there would be a tool or dev-mode or something in Stratis 3.8 that can create legacy V1 pools. Is there? (Haven't really looked yet...)

 - Is there a way to know how many token slots are still free? We would like to prevent the user from attempting to add a new slot when we can know upfront that it will fail because no slot is free anymore.

And thanks a lot for the new metadata format! I really dig it! :-)

jbaublitz:
Closes #3805 
Related https://github.com/stratis-storage/project/issues/762


jbaublitz:
@mulkieran I'm going to start testing this now but you are also welcome to start inital review if you'd like.

src/engine/types/keys.rs
```diff
@@ -407,6 +409,13 @@ impl EncryptionInfo {
             .next()
             .cloned()
     }
+
+    /// Returns number of free token slots in the encryption information.
+    #[allow(clippy::cast_possible_truncation)]
```
mulkieran, line 414: Can you move this allow so it is on top of the line that has the possible truncation, not the whole method?

mulkieran (changes requested):
Still working...minor request.

src/engine/types/keys.rs
jbaublitz: I tried doing that and when I did, Rust claimed that for some reason that invocation was unstable.

jbaublitz:
@mvollmer Are the test failures here indicating that something broke and the machine is failing to start or is this an infrastructure error? https://artifacts.dev.testing-farm.io/3608c965-1d5a-4f8e-9ce4-554a8f15b275/

jbaublitz:
@mulkieran I think I will have to do some additional work on this after looking into #3348 more.

mvollmer:
> @mvollmer Are the test failures here indicating that something broke and the machine is failing to start or is this an infrastructure error? https://artifacts.dev.testing-farm.io/3608c965-1d5a-4f8e-9ce4-554a8f15b275/

This is a known bug in the current rawhide kernel: https://bugzilla.redhat.com/show_bug.cgi?id=2362273

jbaublitz:
I was able to resolve the problems in #3348 without making changes to this PR. This should be ready for review.

src/bin/stratis-min/stratis-min.rs
```diff
@@ -103,6 +103,9 @@ fn parse_args() -> Command {
                 Command::new("has-passphrase")
                     .arg(Arg::new("name").long("name").num_args(0))
                     .arg(Arg::new("id").required(true)),
+                Command::new("is-bound")
```
mulkieran, line 106: Please add a test for the new command in tests/stratis_min.rs, probably similar to ```stratis_min_pool_is_encrypted``` test.

mulkieran (changes requested):
Just one request as of now, still working...

src/engine/types/keys.rs
```diff
@@ -271,6 +271,8 @@ pub struct EncryptionInfo {
 }
 
 impl EncryptionInfo {
+    pub const MAX_TOKEN_SLOTS: u8 = 15;
```
mulkieran, line 274: In the commit that introduces ```MAX_TOKEN_SLOTS``` could you use it in the two error messages that hard-code 15 already? They are both in crypt/handle/v2.rs.

mulkieran (changes requested):
Finished, two more requests.

src/jsonrpc/server/pool.rs
```diff
@@ -345,22 +345,69 @@ pub async fn pool_has_passphrase(
...
+            .ok_or_else(|| StratisError::Msg(format!("Could not find pool with name {n}")))?,
+    };
+    if let Some(info) = stopped
+        .stopped
+        .get(pool_uuid)
+        .or_else(|| stopped.partially_constructed.get(pool_uuid))
+    {
+        Ok(match (info.info.as_ref(), info.features.as_ref()) {
+            (Some(_), Some(_)) => unreachable!(),
+            (Some(ei), _) => ei.clevis_info()?.is_some(),
```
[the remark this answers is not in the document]
jbaublitz: Let me look into this. I already fixed this bug, and I'm not sure how it got reintroduced. I specifically fixed it when I was testing, and I think the only reason the tests worked was because I fixed it in this PR.

src/bin/stratis-min/stratis-min.rs
jbaublitz: I'm going to hold off on this until you've gotten back to me about what you want written differently in that test.

src/jsonrpc/server/pool.rs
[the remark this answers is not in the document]
jbaublitz: Can you explain a little bit more what you have in mind for the test? Are you saying create an encrypted and unencrypted pool and then test that I get the correct output from both pools? Something else?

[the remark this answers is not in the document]
jbaublitz: I think I've resolved this then.

jbaublitz:
@mulkieran I believe I've addressed all of the problems.

mulkieran (approved):

PR #3816, opened 2025-04-23
Add PoolSave method to strip cache info from metadata

mulkieran:
Related https://github.com/stratis-storage/project/issues/767


mulkieran:
test error can be ignored this time.

mulkieran:
Proposed instructions for use:

```
> systemctl stop stratisd
``` 

Then, follow instructions above to install stratisd from this PR: https://github.com/stratis-storage/stratisd/pull/3816#issuecomment-2824880679

```
> systemctl start stratisd
```

This should be enough to bring the pool up completely and write updated metadata that does not specify the cache items. To verify that the metadata has been updated:

```
> stratis pool debug get-metadata --name pn --pretty
```

Make sure the "cache_tier" key is missing from the result. If it isn't, the operation has failed.

Next

```
> systemctl stop stratisd
```

Now, resintall stratisd 3.8.0. Probably best is just:

```
> dnf update stratisd
```

since the COPR rpm is, properly speaking, a pre-release.

At this point, you should be running as normal and you can add the cache device back to the pool.

mulkieran:
Superceded by #3864

Issue #3809, opened 2025-04-02
stratisd_proc_macros can't handle unnamed arguments


jbaublitz:
The syntax for a function parameter `_: MyType` causes the proc macro to fail. This is a valid syntax that we want to accept.

PR #3826, opened 2025-05-01
Rework systemd dependencies

### LINKED ISSUE #3348: stratis-fstab-setup@.service might use the network but does not wait for it

jbaublitz:
Closes #3348 

@mvollmer Please take a look at the last commit here. This is based on #3811 but the last commit will go in separately. I believe that this reordering of dependencies in addition to the addition of _netdev will successfully set up a Clevis device.


jbaublitz:
@drckeefe Just so you're aware, this PR will potentially change some of our documentation slightly. The reason for this is that currently Clevis in /etc/fstab is unreliable and may not be able to be set up. This PR requires the addition of `_netdev` to the /etc/fstab entry and then it should work for both Clevis and keyring implementations.

mvollmer:
> I believe that this reordering of dependencies in addition to the addition of _netdev will successfully set up a Clevis device.

I believe so, too. But note that then _all_ entries in fstab for stratis filesystems need _netdev, even the ones for unencrypted pools or encrypted ones without tang. (I think, haven't tested.)

If people get the new dependencies from this PR but do not add "_netdev" to their fstab entries, I would expect the next boot to break. Have you tried this? I might be wrong.

mvollmer:
> I would expect the next boot to break.

With an error from systemd about cyclic dependencies or something like that.

jbaublitz:
@mvollmer I already discussed this with @drckeefe and we decided we're going to try to add a new unit file that specifically requires the _netdev parameter in /etc/fstab and waits for the network. Otherwise they will be identical.

mvollmer (changes requested):
As @jbaublitz has said, let's not add the new dependencies to stratis-fstab-setup, but to a new unit.

jbaublitz:
@mvollmer I believe the most recently update to this PR resolves your concerns and it works for both Clevis and keyring implementations with a change to /etc/fstab based on my testing. Any last requests? 

@mulkieran I think this is ready for your review.

mulkieran (changes requested):
Please grab the first commit https://github.com/stratis-storage/stratisd/pull/3826/commits and put it in a separate PR if feasible. It is the solution to this issue, I believe: https://github.com/stratis-storage/stratisd/issues/3809

mulkieran (approved):

PR #3841, opened 2025-05-15
Load volume key into keyring on activation for resize operation

### LINKED ISSUE #3810: Encrypted pools are sometimes not started during boot
A encrypted V2 pool with a passphrase is not always started during boot. This happens when our CI test reboots too early after adding a cache device to the pool. The pool seems to be in some kind of intermediate state that stratisd-min and stratisd try to fix, but something seems to go wrong with cryptsetup.

When we put a one minute delay before the reboot, the pool is reliably started during boot.

[Journal of a failed test run](https://gist.githubusercontent.com/mvollmer/b39d5fbe7c5a84d9720ebcebf33ab9f1/raw/eebdf15592e54140a774307645c8bdbe4494610d/TestStorageStratisReboot-testEncrypted-fedora-42-127.0.0.2-2301-FAIL.log)

[Journal of a passed run](
https://gist.githubusercontent.com/mvollmer/b39d5fbe7c5a84d9720ebcebf33ab9f1/raw/eebdf15592e54140a774307645c8bdbe4494610d/TestStorageStratisReboot-testEncrypted-fedora-42-127.0.0.2-2201-PASS.log)

In the failed run, I think stratisd-min is able to unlock the pool as expected:

```
stratisd-min[668]: [2025-03-26T17:02:40Z INFO  stratisd::engine::strat_engine::liminal::liminal] Pool with name "pool0" and UUID "32605337-e367-409d-a3a9-e0092d1013ad" set up
```

But then it fails to extend the pool:

```
Mar 26 17:02:40 fedora-42-127-0-0-2-2201 stratisd-min[668]: [2025-03-26T17:02:40Z ERROR stratisd::engine::strat_engine::thinpool::thinpool] Attempted to extend a thinpool meta sub-device belonging to pool with uuid 32605337-e367-409d-a3a9-e0092d1013ad but failed with error: Crypt(IOError(Os { code: 1, kind: PermissionDenied, message: "Operation not permitted" }))
Mar 26 17:02:40 fedora-42-127-0-0-2-2201 stratisd-min[668]: [2025-03-26T17:02:40Z WARN  stratisd::engine::strat_engine::thinpool::thinpool] Device extension failed: Cryptsetup error: IO error occurred: Operation not permitted (os error 1)
```

And later the switch-over to the full-fat stratisd somehow fails as well:

```
Mar 26 17:02:42 fedora-42-127-0-0-2-2201 stratisd[929]: [2025-03-26T17:02:42Z INFO  stratisd::engine::strat_engine::liminal::liminal] Attempt to set up pool failed, but it may be possible to set up the pool later, if the situation changes: An attempt to set up pool with UUID 32605337-e367-409d-a3a9-e0092d1013ad from the assembled devices failed; DM error: DM error: Invalid: Specified new table "LinearDevTargetTable { table: [TargetLine { start: Sectors(0), length: Sectors(7987192), params: Linear(LinearTargetParams { device: Device { major: 8, minor: 0 }, start_offset: Sectors(8192) }) }, TargetLine { start: Sectors(7987192), length: Sectors(2056), params: Linear(LinearTargetParams { device: Device { major: 8, minor: 32 }, start_offset: Sectors(8192) }) }] }" does not match kernel table "LinearDevTargetTable { table: [TargetLine { start: Sectors(0), length: Sectors(7987192), params: Linear(LinearTargetParams { device: Device { major: 8, minor: 0 }, start_offset: Sectors(8192) }) }, TargetLine { start: Sectors(7987192), length: Sectors(10248), params: Linear(LinearTargetParams { device: Device { major: 8, minor: 32 }, start_offset: Sectors(8192) }) }] }"
```

This is not blocking us in any way, but maybe it helps you find a bug in Stratis.

jbaublitz:
Closes #3810

### DIFF
--- a/src/engine/engine.rs
+++ b/src/engine/engine.rs
@@ -422,6 +422,14 @@ pub trait Pool: Debug + Send + Sync {
 
     /// Get number of free token slots for pool.
     fn free_token_slots(&self) -> Option<u8>;
+
+    /// Indicates whether the volume key for this pool is loaded into the keyring.
+    fn volume_key_is_loaded(&self, uuid: PoolUuid) -> StratisResult<bool>;
+
+    /// Load volume key for this pool into the keyring.
+    ///
+    /// Returns true if the key was newly loaded and false if the key was already loaded.
+    fn load_volume_key(&mut self, uuid: PoolUuid) -> StratisResult<bool>;
 }
 
 pub type HandleEvents<P> = (
--- a/src/engine/strat_engine/backstore/backstore/v2.rs
+++ b/src/engine/strat_engine/backstore/backstore/v2.rs
@@ -25,7 +25,7 @@ use crate::{
             },
             crypt::{handle::v2::CryptHandle, manual_wipe, DEFAULT_CRYPT_DATA_OFFSET_V2},
             dm::{get_dm, list_of_backstore_devices, remove_optional_devices, DEVICEMAPPER_PATH},
-            keys::search_key_process,
+            keys::{search_key_process, unset_key_process},
             metadata::{MDADataSize, BDA},
             names::{format_backstore_ids, CacheRole},
             serde_structs::{BackstoreSave, CapSave, PoolFeatures, PoolSave, Recordable},
@@ -871,7 +871,8 @@ impl Backstore {
         if let Some(ref mut cache_tier) = self.cache_tier {
             cache_tier.destroy()?;
         }
-        self.data_tier.destroy()
+        self.data_tier.destroy()?;
+        unset_key_process(&VolumeKeyKeyDescription::new(pool_uuid)).map(|_| ())
     }
 
     /// Teardown the DM devices in the backstore.
@@ -881,7 +882,8 @@ impl Backstore {
         if let Some(ref mut cache_tier) = self.cache_tier {
             cache_tier.block_mgr.teardown()?;
         }
-        self.data_tier.block_mgr.teardown()
+        self.data_tier.block_mgr.teardown()?;
+        unset_key_process(&VolumeKeyKeyDescription::new(pool_uuid)).map(|_| ())
     }
 
     /// Consume the backstore and convert it into a set of BDAs representing
--- a/src/engine/strat_engine/backstore/backstore/v2.rs
+++ b/src/engine/strat_engine/backstore/backstore/v2.rs
@@ -707,7 +707,7 @@ impl Backstore {
                 let table = self.data_tier.segments.map_to_dm();
                 cache.set_origin_table(get_dm(), table)?;
                 cache.resume(get_dm())?;
-                handle.resize(None)?;
+                handle.resize(pool_uuid, None)?;
                 false
             }
             (Some(cache), None, None) => {
@@ -730,7 +730,7 @@ impl Backstore {
                 )];
                 placeholder.set_table(get_dm(), table)?;
                 placeholder.resume(get_dm())?;
-                handle.resize(None)?;
+                handle.resize(pool_uuid, None)?;
                 false
             }
             (None, Some((cap, linear)), None) => {
--- a/src/engine/strat_engine/backstore/backstore/v2.rs
+++ b/src/engine/strat_engine/backstore/backstore/v2.rs
@@ -25,6 +25,7 @@ use crate::{
             },
             crypt::{handle::v2::CryptHandle, manual_wipe, DEFAULT_CRYPT_DATA_OFFSET_V2},
             dm::{get_dm, list_of_backstore_devices, remove_optional_devices, DEVICEMAPPER_PATH},
+            keys::search_key_process,
             metadata::{MDADataSize, BDA},
             names::{format_backstore_ids, CacheRole},
             serde_structs::{BackstoreSave, CapSave, PoolFeatures, PoolSave, Recordable},
@@ -35,7 +36,7 @@ use crate::{
         types::{
             ActionAvailability, BlockDevTier, DevUuid, EncryptionInfo, InputEncryptionInfo,
             KeyDescription, OptionalTokenSlotInput, PoolUuid, SizedKeyMemory, TokenUnlockMethod,
-            UnlockMechanism, ValidatedIntegritySpec,
+            UnlockMechanism, ValidatedIntegritySpec, VolumeKeyKeyDescription,
         },
     },
     stratis::{StratisError, StratisResult},
@@ -1351,6 +1352,29 @@ impl Backstore {
             ActionAvailability::NoPoolChanges
         } else {
             ActionAvailability::Full
+        }
+    }
+
+    /// Check whether a volume key is in the kernel keyring for a crypt device in the backstore.
+    pub fn volume_key_is_loaded(uuid: PoolUuid) -> StratisResult<bool> {
+        Ok(search_key_process(&VolumeKeyKeyDescription::new(uuid))?.is_some())
+    }
+
+    /// Load volume key into the kernel keyring for a crypt device in the backstore.
+    pub fn load_volume_key(uuid: PoolUuid) -> StratisResult<bool> {
+        let crypt_physical_path = &once(DEVICEMAPPER_PATH)
+            .chain(once(
+                format_backstore_ids(uuid, CacheRole::Cache)
+                    .0
+                    .to_string()
+                    .as_str(),
+            ))
+            .collect::<PathBuf>();
+        if Self::volume_key_is_loaded(uuid)? {
+            Ok(false)
+        } else {
+            CryptHandle::load_vk_to_keyring(crypt_physical_path, uuid)?;
+            Ok(true)
         }
     }
 }
--- a/src/engine/strat_engine/crypt/shared.rs
+++ b/src/engine/strat_engine/crypt/shared.rs
@@ -475,6 +475,32 @@ fn device_is_active(device: Option<&mut CryptDevice>, device_name: &DmName) -> S
     }
 }
 
+/// Load the volume key for a V2 pool into the process keyring.
+pub fn load_vk_to_keyring(
+    device: &mut CryptDevice,
+    passphrase: Option<&SizedKeyMemory>,
+    uuid: PoolUuid,
+) -> StratisResult<()> {
+    let vk_kd = VolumeKeyKeyDescription::new(uuid);
+    device.activate_handle().set_keyring_to_link(
+        &vk_kd.to_system_string(),
+        None,
+        None,
+        Some("@p"),
+    )?;
+    if let Some(p) = passphrase {
+        device.activate_handle().activate_by_passphrase(
+            None,
+            None,
+            p.as_ref(),
+            CryptActivate::KEYRING_KEY,
+        )?;
+    } else {
+        activate_by_token(device, None, None, CryptActivate::KEYRING_KEY)?;
+    }
+    Ok(())
+}
+
 /// Activate encrypted Stratis device.
 ///
 /// Precondition: uuid.is_none() if pool metadata version == 1, uuid.is_some() if pool metadata
--- a/src/engine/strat_engine/crypt/shared.rs
+++ b/src/engine/strat_engine/crypt/shared.rs
@@ -4,7 +4,7 @@
 
 use std::{
     fs::OpenOptions,
-    io::{Seek, SeekFrom, Write},
+    io::{self, Seek, SeekFrom, Write},
     mem::forget,
     path::{Path, PathBuf},
     slice::from_raw_parts_mut,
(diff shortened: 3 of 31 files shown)


jbaublitz:
One thing we probably need to decide what to do with is that the volume keys show up in the key list output. We do not want this as they can't be used with the unlock method that we use with the other key descriptions. We can either split out the type entirely or filter based on the format of the volume key names.

jbaublitz:
Putting this back in progress, there's some behavior I need to fix.

mvollmer (commented):
I don't understand these changes enough to comment in detail, I am afraid.

I imagine that "stratis pool start ..." currently stores the received passphrase in the keyring, sets up the pool, and then unsets the key.  With this change, would it leave the key in the keyring?

jbaublitz:
> Does this change mean that also Bind, Rebind, and Unbind can use this key?

@mvollmer No that is not what it will do, we will likely decide to filter this key out of listing operations largely because there are different cryptsetup APIs for passphrases, tokens, and volume keys. It makes no sense to list these keys when they will be largely useless for anything but background processes that require authentication. I will tag you once I've worked out the problems with this PR. I obviously haven't resolved some bugs that still are lingering in the code and I think I've determined what the problem is. I will likely have a fix later today.

jbaublitz:
@mulkieran I discovered that the keyring refreshes every time we call get_persistent_keyring. It should be relatively easy to do that to keep the keyring active. This has some notable drawbacks of ceasing to work when stratisd is down, but if we can add a background task to keep the keyring active and also idempotently ensure that the volume key is loaded into the keyring on stratisd service start or pool start, this should be sufficient to have it work as expected as long as the daemon is up.

bmr-cymru:
24.04 (Noble Numbat) should have 2.7.0 (I was looking for updates for `podman` and `mkosi`, but the latter aren't there yet).

https://launchpad.net/ubuntu/+source/cryptsetup

jbaublitz:
@mvollmer This should be ready for your review. This PR should fix the problem you were having with the tests.

jbaublitz:
A few notes on this PR:
* CI is currently failing due to a network error with the Ubuntu repositories. It's hard to tell when this will resolve.
* This PR adds a property to report whether the volume key is loaded in the keyring. We do not intend to send a change signal for this particular property as it seems somewhat useless here.

src/engine/engine.rs
```diff
@@ -422,6 +422,12 @@ pub trait Pool: Debug + Send + Sync {
 
     /// Get number of free token slots for pool.
     fn free_token_slots(&self) -> Option<u8>;
+
+    /// Indicates whether the volume key for this pool into the keyring.
```
mulkieran: There's a bit of a typo in this comment.

src/engine/strat_engine/backstore/backstore/v2.rs
```diff
@@ -880,7 +882,8 @@ impl Backstore {
         if let Some(ref mut cache_tier) = self.cache_tier {
             cache_tier.block_mgr.teardown()?;
         }
-        self.data_tier.block_mgr.teardown()
+        self.data_tier.block_mgr.teardown()?;
+        unset_key_persistent(Either::Right(&VolumeKeyKeyDescription::new(pool_uuid))).map(|_| ())
```
mulkieran: I do not understand why unsetting the key here is the correct thing to do, as I believe the teardown() method is used in the context of stopping a pool.

jbaublitz: The volume key should only be available in the keyring as long as the pool is active so this is intended.

src/engine/engine.rs

src/stratis/keys.rs
```diff
@@ -0,0 +1,69 @@
...
+                            warn!(
+                                "Failed to load volume key into keyring for pool with UUID {uuid}: {e}"
+                            );
+                        }
+                        Err(e) => {
+                            warn!("Failed to join thread: {e}");
+                        }
+                    }
+                } else {
+                    warn!("Pool with UUID {uuid} not found");
```
mulkieran: I _think_ this should just be an info, since the pool might have been, e.g., stopped after the relevant set of UUIDs was calculated and before it was reached on the refreshing step.

src/stratis/keys.rs
```diff
@@ -0,0 +1,69 @@
...
+use std::sync::Arc;
+
+use tokio::{sync::mpsc::UnboundedReceiver, task::spawn_blocking};
+
+use crate::{
+    engine::{Engine, KeyDescription, PoolIdentifier},
+    stratis::{StratisError, StratisResult},
+};
+
+pub async fn load_vks(
```
mulkieran, line 16: Please add some header docs here.

src/engine/engine.rs
```diff
@@ -422,6 +422,12 @@ pub trait Pool: Debug + Send + Sync {
 
     /// Get number of free token slots for pool.
     fn free_token_slots(&self) -> Option<u8>;
+
+    /// Indicates whether the volume key for this pool is loaded into the keyring.
+    fn volume_key_is_loaded(&self, uuid: PoolUuid) -> StratisResult<bool>;
+
+    /// Load volume key for this pool into the keyring.
```
mulkieran, line 429: Please document the meaning of return value. I believe it means change or not, but might as well say that explicitly.

src/engine/strat_engine/backstore/backstore/v2.rs
mulkieran: Ok!

src/stratis/timer.rs
```diff
@@ -67,18 +73,107 @@ async fn check_pool_and_fs(
...
+            Err(e) => {
+                warn!("Failed to get timeout for keyring refresh: {e}");
+                continue;
+            }
+        };
+        if let Err(e) = process_checks().await {
+            warn!("Failed to refresh keyring: {e}");
+        }
+        trace!("Timed keyring refresh finished");
+        let timeout_adjusted = timeout.saturating_sub(3600);
```
mulkieran: I think we should avoid our timeout actually going to 0, otherwise the task will always be wanting to run. If the timeout is so short that it is better not to refresh at all, we can log a warning. But we should only do it once, on the notify on the ```persistent_keyring_expiry``` file.

src/stratis/timer.rs
```diff
@@ -67,18 +73,107 @@ async fn check_pool_and_fs(
...
+                }
+            }
+        }
+    }
+}
+
+/// Periodically refresh the keyring to ensure the keys are never garbage collected.
+async fn refresh_keyring() {
+    async fn process_timeout() -> StratisResult<u64> {
+        let mut file = File::open("/proc/sys/kernel/keys/persistent_keyring_expiry").await?;
```
mulkieran: I think you want to use ```KEY_EXPIRY_PATH``` here as well as below.

mulkieran (changes requested):
Still working...

tests/client-dbus/src/stratisd_client_dbus/_introspect.py
```diff
@@ -266,6 +266,9 @@
     <property name="Uuid" type="s" access="read">
       <annotation name="org.freedesktop.DBus.Property.EmitsChangedSignal" value="const" />
     </property>
+    <property name="VolumeKeyLoaded" type="v" access="read">
```
mulkieran, line 269: I need to know precisely what combination of this property being false and the values of other properties should cause stratis-cli to post a warning.

mulkieran (commented):
still working...we should consider whether we should implement a warning code in pool list view in another patch release in stratis-cli.

src/stratis/timer.rs
jbaublitz: While I'm okay with the idea of not letting the timeout go to 0, I would like to avoid any case where the keyring can expire. If we allow this, pools will fail to extend in certain cases and I think we should try to avoid that at all costs. I think it would be better to have a minimum timeout of 1 second for everything below one hour and a minimum of half a second if the expiry is 1 second.

tests/client-dbus/src/stratisd_client_dbus/_introspect.py
jbaublitz: I used a variant here specifically because if it's a String, it's an error message. If it's a boolean it indicates whether the volume key is loaded or not. That should be relatively easy to handle in Python. I think that is a lot less verbose than a pattern with a lot of booleans and Variant is designed for cases exactly like this.

src/stratis/timer.rs
jbaublitz: The other option I wanted to propose is that we use the process keyring instead. This means that we will never preserve any keys across restarts of stratisd but we will not have to refresh. Is that preferable?

mulkieran: We should discuss in standup. Any use to using the process keyring generally, but attaching to the persistent keyring when stratisd shuts down?

tests/client-dbus/src/stratisd_client_dbus/_introspect.py
mulkieran: My current plan for stratis-cli was not to show this value directly, but, if it indicated something that the user should consider taking action on, show a warnng in the pool listing. But in most cases, if the value is False, there isn't any need to show a warning. In all cases where the pool is V1, for example. But there are also cases where the pool is V2 where a value of false is no problem, like when the pool is unencrypted. What I'm trying to do is narrow down precisely the situation, on an encrypted V2 pool, where a value of false is a problem to be reported.

jbaublitz: I think that sounds correct!

src/stratis/timer.rs
jbaublitz: I doubt it only because we will have no control over whether the keyring expires between when stratisd shuts down and starts back up.

mulkieran (commented):
I have two questions regarding the "VolumeKeyLoaded" property.

1. Is there still a reason, now that the volume key is in the process keyring, for stratis-cli to still detect and make use of this property in its pool detail listing? Note that the original idea of how to use this property was to add a warning in the case that the user needed to take some action, rather than simply report the value of the property.
2. Is there still a reason, now that the volume key is in the process keyring, for stratisd to report the D-Bus property?

I think the answer is yes to both questions. But, in that case, we still need a precise understanding of under what conditions the warning should be generated.

tests/client-dbus/src/stratisd_client_dbus/_introspect.py
mulkieran: Ok! I'm going with, if the pool is encrypted and the volume key is not loaded and it is V2, show a warning.

src/engine/strat_engine/keys.rs
```diff
@@ -53,14 +82,30 @@ pub fn get_persistent_keyring() -> StratisResult<KeySerial> {
...
+
 /// Search for the given key description in the persistent root keyring.
 /// Returns the key ID or nothing if it was not found in the keyring.
-fn search_key(
-    keyring_id: KeySerial,
-    key_desc: &KeyDescription,
-) -> StratisResult<Option<KeySerial>> {
-    let key_desc_cstring = CString::new(key_desc.to_system_string())
-        .map_err(|_| StratisError::Msg("Invalid key description provided".to_string()))?;
+fn search_key(keyring_id: KeySerial, key_desc: String) -> StratisResult<Option<KeySerial>> {
```
mulkieran, line 103: The header comments are about the persistent keyring and it's really the keyring with the specified ```keyring_id```. Might as well fix them here.

mulkieran (changes requested):
One tiny comment request so far...

src/engine/strat_engine/names.rs
```diff
@@ -52,6 +52,12 @@ impl KeyDescription {
     }
 }
 
+impl VolumeKeyKeyDescription {
+    pub fn to_system_string(&self) -> String {
```
mulkieran, line 56: I think you might be better off creating a type for the result value of ```to_system_string``` as there are more functions than formerly that require the specially formatted system string as an argument.

mulkieran (changes requested):
One design suggestion...

src/engine/strat_engine/names.rs
jbaublitz: I'm not really sure what this adds. The end result needs to be a `String` for the syscalls so adding in an intermediary conversion step feels like it would just make the conversion less ergonomic. From the standpoint of validation, the two methods that would use this signature don't seem to share enough of the format of the key description to provide value in validation. Can you explain a bit more why you see this as a benefit?

src/engine/strat_engine/crypt/shared.rs
```diff
@@ -473,14 +475,67 @@ fn device_is_active(device: Option<&mut CryptDevice>, device_name: &DmName) -> S
     }
 }
 
+/// Load the volume key for a V2 pool into the kernel keyring.
+pub fn load_vk_to_keyring(
```
mulkieran, line 479: I think in header comments the keyring should be the "process" keyring.

src/engine/strat_engine/names.rs
mulkieran: More places than formerly where it is necessary to match-up the system_string() in storing and retirieving. I'm worried that in one place for reading or writing, the conversion will be skipped, and that fact won't be detected immediately.

jbaublitz: Working on implementing this, I believe there are many more places where we don't have any place for a type signature that use `.to_system_string()`. The end result of this is that we have more places where we call `.to_system_string().into_string()` which looks a bit redundant to me. Because we're often using this specifically in contexts that expect a String and less in function signatures that take a String, this appears to be more disruptive than helpful.

jbaublitz: Furthermore I believe that because we're not passing a raw String to any of these APIs in the layer above any of these calls that we're essentially guaranteed to call `.to_system_string()` to even get a String to use for the APIs that take Strings.

PR #3855, opened 2025-06-06
Use a list for the trigger key

mulkieran:


mulkieran:
multiple values for trigger is not actually supposed to work, closing.

PR #3856, opened 2025-06-09
Only create placeholder device when encryption is enabled

### LINKED ISSUE #3840: Only create cache device placeholder when encryption is enabled
We got a suggestion to remove the placeholder cache device in cases where the pool is not encrypted to avoid overhead. Per our discussions with @oniko, we believe that once the pool is encrypted, we should not be changing the device underneath it and will have to have the placeholder device for safety purposes, but I believe that we can safely remove it until we encrypt the pool.

jbaublitz:
Closes #3840

### DIFF
--- a/src/engine/strat_engine/backstore/backstore/v2.rs
+++ b/src/engine/strat_engine/backstore/backstore/v2.rs
@@ -143,7 +143,7 @@ pub struct Backstore {
     data_tier: DataTier<StratBlockDev>,
     /// A linear DM device.
     origin: Option<LinearDev>,
-    /// A placeholder device to be converted to cache or a cache device.
+    /// A cache device.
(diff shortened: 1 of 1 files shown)


jbaublitz:
After we just discussed this, this will likely be demoted to 3.9.0 or 3.9.1 depending on how hard it is to evaluate the need for migration code.

jbaublitz:
Blocked by #3800

jbaublitz:
@mulkieran I believe this may be rendered obsolete by the performance testing. I think the only thing we may need to test is slower drives. If there's a large performance degradation in V2, this may be worth pursuing.

jbaublitz:
I confirmed with @spetrovi that performance is consistent across slower and faster drives. I'm going to close the linked issue.

PR #3858, opened 2025-06-12
Try alternative branches for tests that use non-stratisd repos

mulkieran:
Related https://github.com/stratis-storage/project/issues/755


mulkieran:
Even though both branch checkouts failed, the condition for both checkout succeeding passed. This is very strange.

mulkieran:
This one is working...

mulkieran:
ref value is expanding to 3858/merge which is not what is wanted.

mulkieran:
Packit test failures can be ignored...

mulkieran:
> Approved. Are you planning to just do this for stratis-cli for the time being? I feel like libcryptsetup-rs/libblkid-rs/etc. will be harder as they're Rust packages and it involves changing the Cargo.toml in Github Actions. I'm not even sure what it would take in packit.

This is for stratis-cli and for testing repos. For our Rust dependencies...it will have to be a Rust solution. I'm not sure yet what we could do with Cargo.toml to make that work. It's possible that ```cargo add``` will help.

Issue #3796, opened 2025-03-20
Handle failed cache devices


jbaublitz:
Currently, our plan is to have RAID cover data devices only. Cache is not included and this leads to a problem. With RAID, you would theoretically be able to remove failed devices, but that does not seem like enough justification to add RAID to a cache, which will otherwise not benefit a lot. We need to find another way to handle failed devices in the cache so that users don't get into a position where they cannot start their pool.

There are two separate cases we need to handle: a started and stopped pool. In a started pool, it's relatively easy to add remove cache functionality. That is straightforward. In the stopped pool case, there are a few considerations. If a cache fails to be set up, we could theoretically just remove the cache for the user so that the pool can still be set up and they can re-add the cache. The other option, if we're willing to target only V2 pools for this functionality, would be to provide a way to remove the cache from stopped pools. As the Stratis metadata is available in V2 in all cases, we would theoretically be able to target a specific metadata operation to remove the cache on a stopped pool. This, however, would not be able to be done in V1.

@drckeefe @mulkieran I'm open to your feedback here.

PR #3864, opened 2025-06-17
Allow starting pool without cache

jbaublitz:
Related to #3796 

I will separate out removing the cache from a started pool into a separate PR.

### DIFF
--- a/src/engine/strat_engine/liminal/liminal.rs
+++ b/src/engine/strat_engine/liminal/liminal.rs
@@ -56,6 +56,12 @@ use crate::{
     stratis::{StratisError, StratisResult},
 };
 
+type MetadataResult = StratisResult<(
+    DateTime<Utc>,
+    PoolSave,
+    Option<Vec<Either<PathBuf, PathBuf>>>,
+)>;
+
 /// Devices which stratisd has discovered but which have not been assembled
 /// into pools.
 #[derive(Debug, Default, Eq, PartialEq)]
@@ -191,10 +197,9 @@ impl LiminalDevices {
             luks_info: StratisResult<Option<PoolEncryptionInfo>>,
             infos: &HashMap<DevUuid, Box<LStratisDevInfo>>,
             bdas: HashMap<DevUuid, BDA>,
-            meta_res: StratisResult<(DateTime<Utc>, PoolSave)>,
-            remove_cache: bool,
+            meta_res: MetadataResult,
         ) -> BDARecordResult<(Name, AnyPool)> {
-            let (timestamp, metadata) = match meta_res {
+            let (timestamp, metadata, paths_to_wipe) = match meta_res {
                 Ok(o) => o,
                 Err(e) => return Err((e, bdas)),
             };
@@ -207,7 +212,7 @@ impl LiminalDevices {
                 bdas,
                 timestamp,
                 metadata,
-                remove_cache,
+                paths_to_wipe,
             )
         }
 
@@ -298,7 +303,7 @@ impl LiminalDevices {
         let res = load_stratis_metadata(pool_uuid, &mut infos, remove_cache);
         let (infos, bdas) = split_stratis_infos(infos);
 
-        match start_pool_failure(pools, pool_uuid, luks_info, &infos, bdas, res, remove_cache) {
+        match start_pool_failure(pools, pool_uuid, luks_info, &infos, bdas, res) {
             Ok((name, pool)) => {
                 self.uuid_lookup = self
                     .uuid_lookup
@@ -340,21 +345,27 @@ impl LiminalDevices {
         passphrase_fd: Option<RawFd>,
         remove_cache: bool,
     ) -> StratisResult<(Name, PoolUuid, AnyPool, Vec<DevUuid>)> {
-        #[allow(clippy::too_many_arguments)]
         fn start_pool_failure(
             pools: &Table<PoolUuid, AnyPool>,
             pool_uuid: PoolUuid,
             infos: &HashMap<DevUuid, Box<LStratisDevInfo>>,
             bdas: HashMap<DevUuid, BDA>,
-            meta_res: StratisResult<(DateTime<Utc>, PoolSave)>,
+            meta_res: MetadataResult,
             token_slot: TokenUnlockMethod,
             passphrase_fd: Option<RawFd>,
-            remove_cache: bool,
         ) -> BDARecordResult<(Name, AnyPool)> {
-            let (timestamp, metadata) = match meta_res {
+            let (timestamp, metadata, paths_to_wipe) = match meta_res {
                 Ok(o) => o,
                 Err(e) => return Err((e, bdas)),
             };
+            let processed_paths_to_wipe = paths_to_wipe.map(|v| {
+                v.into_iter()
+                    .map(|e| {
+                        e.left()
+                            .expect("Must be an unencrypted path in V2 of the metadata")
+                    })
+                    .collect::<Vec<_>>()
+            });
 
             let passphrase = match (
                 metadata.features.contains(&PoolFeatures::Encryption),
@@ -397,9 +408,9 @@ impl LiminalDevices {
                 bdas,
                 timestamp,
                 metadata,
+                processed_paths_to_wipe,
                 token_slot,
                 passphrase,
-                remove_cache,
             )
         }
 
@@ -427,7 +438,6 @@ impl LiminalDevices {
             res,
             token_slot,
             passphrase_fd,
-            remove_cache,
         ) {
             Ok((name, pool)) => {
                 self.uuid_lookup = self
@@ -868,7 +878,7 @@ impl LiminalDevices {
             if let Some(true) | None = metadata.started {
                 match metadata_version {
                     StratSigblockVersion::V1 => setup_pool_legacy(
-                        pools, pool_uuid, luks_info, infos, bdas, timestamp, metadata, false,
+                        pools, pool_uuid, luks_info, infos, bdas, timestamp, metadata, None,
                     )
                     .map(Either::Left),
                     StratSigblockVersion::V2 => setup_pool(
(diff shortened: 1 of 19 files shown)


src/engine/strat_engine/pool/v2.rs
```diff
@@ -265,7 +267,8 @@ impl StratPool {
         };
 
         // TODO: Remove in stratisd 4.0
-        let mut needs_save = metadata.thinpool_dev.fs_limit.is_none()
+        let mut needs_save = remove_cache
```
mulkieran: Put the ```remove_cache``` part of the expression elsewhere since it will need to stay in stratisd 4.0 (or move the comment).

mulkieran: Note that this writes out the metadata even if the pool has no cache. I'm fine with that...just mentioning it.

src/engine/strat_engine/pool/v1.rs
```diff
@@ -297,7 +298,8 @@ impl StratPool {
         };
 
         // TODO: Remove in stratisd 4.0
-        let mut needs_save = metadata.thinpool_dev.fs_limit.is_none()
+        let mut needs_save = remove_cache
```
mulkieran: Same comment as in v2 pools...

src/engine/strat_engine/liminal/liminal.rs
```diff
@@ -1221,13 +1264,20 @@ fn load_stratis_metadata(
...
                 StratisError::Msg(format!(
                     "No metadata found on devices associated with pool UUID {pool_uuid}"
                 ))
-            }),
+            })?;
+            if remove_cache {
+                let dev_uuids = remove_cache_from_metadata(&mut metadata);
+                update_infos(infos, dev_uuids);
+            }
+            Ok((timestamp, metadata))
```
mulkieran: Should the timestamp be updated here to the current time?

mulkieran (changes requested):
Still working...one question and a few minor requests.

src/engine/strat_engine/liminal/liminal.rs
mulkieran: > Should the timestamp be updated here to the current time?

I think the answer is no, because the timestamp is calculated when writing it out...yeah, verified, this timestamp represents the ```last_update_time```. Maybe it would be better to change ```timestamp``` to ```last_update_time``` for clarity here.

jbaublitz: I think this would be a bad idea, only because we use timestamp throughout our entire codebase when reading the metadata. If you want to put up a separate PR changing all of the instances of timestamp to last_updated_time, we can but I think that should be separate from this PR.

jbaublitz:
I've added a commit to wipe the cache devices on cache removal. I'm going to keep an eye on the tests and fix anything that pops up then do some manual testing.

mulkieran:
Ignore the testing failures, Packit is unfortunately delayed.

tests/client-dbus/tests/udev/test_start.py
```diff
@@ -113,6 +114,58 @@ def test_failed_start_regression(self):
...
+                    "id": "testpool",
+                    "id_type": "name",
+                    "unlock_method": (False, (False, 0)),
+                    "key_fd": (False, 0),
+                    "remove_cache": True,
+                },
+            )
+            self.assertEqual(rc, 0)
+
+            assert not bool(Pool.Properties.HasCache.Get(get_object(pool_object_path)))
```
mulkieran, line 172: Could you augment this test and the one above by initializing a cache using the previously used cache device?

src/engine/strat_engine/pool/v2.rs
```diff
@@ -227,14 +228,17 @@ impl StratPool {
...
     ///   key description.
     ///   * no StratBlockDev in cachdevs has a key description
+    #[allow(clippy::too_many_arguments)]
     pub fn setup(
         uuid: PoolUuid,
         datadevs: Vec<StratBlockDev>,
         cachedevs: Vec<StratBlockDev>,
         timestamp: DateTime<Utc>,
         metadata: &PoolSave,
+        paths_to_wipe: Vec<PathBuf>,
```
mulkieran: There is a slight fuzziness about the arguments here. There are two, but the way they are being used now, they could be one, say: ```cache_paths_to_wipe: Option<Vec<PathBuf>>```. A value of ```Some``` is equivalent to ```remove_cache``` is true. Is it useful to merge the two arguments in that way? If not, arg name should still be explicit, something like ```cache_paths_to_wipe``` since the logs definitely indicate cache.

src/engine/strat_engine/pool/v2.rs
```diff
@@ -289,6 +294,18 @@ impl StratPool {
                     return Err((err, pool.backstore.into_bdas()));
                 }
             }
+
+            for path in paths_to_wipe {
```
mulkieran: This is actually related to my point above. Why would we skip wiping the devices to wipe just because ```needs_save``` is false? I think the idea is that ```paths_to_wipe``` will always be empty if ```needs_save``` is true, but that is an implicit invariant that is a bit hard to figure out on the fly. And if is true, then it doesn't matter if this block is within or outside the conditional.

tests/client-dbus/tests/udev/test_start.py
jbaublitz: Sure!

src/engine/strat_engine/pool/v2.rs
mulkieran: This and the comment below apply to v1 pools also.

mulkieran: Note that in this scheme it is not possible to remove the ```Option```, because it is possible that no device paths for cache devices have been found. In that case, though, it is still necessary to remove the cache metadata, if present.

src/engine/strat_engine/liminal/liminal.rs
```diff
@@ -183,22 +183,32 @@ impl LiminalDevices {
...
+        remove_cache: bool,
     ) -> StratisResult<(Name, PoolUuid, AnyPool, Vec<DevUuid>)> {
         fn start_pool_failure(
             pools: &Table<PoolUuid, AnyPool>,
             pool_uuid: PoolUuid,
             luks_info: StratisResult<Option<PoolEncryptionInfo>>,
             infos: &HashMap<DevUuid, Box<LStratisDevInfo>>,
             bdas: HashMap<DevUuid, BDA>,
-            meta_res: StratisResult<(DateTime<Utc>, PoolSave)>,
+            meta_res: StratisResult<(DateTime<Utc>, PoolSave, Vec<PathBuf>)>,
```
mulkieran: Here, also, it is not quite clear whether the ```remove_cache``` argument is redundant.

src/engine/strat_engine/liminal/liminal.rs
```diff
@@ -856,8 +888,10 @@ impl LiminalDevices {
                         bdas,
                         timestamp,
                         metadata,
+                        paths_to_wipe,
```
mulkieran: I think ```paths_to_wipe``` will always be ```vec![]```, so the different treatment in the v1 and v2 pools is odd.

mulkieran (changes requested):
This is a critique. I don't think I've found a bug, but I've found an inconsistency where in some places ```paths_to_wipe``` parameter is treated as if it is independent of the ```remove_cache``` parameter, and other places it is treated as if the values are coupled (which they are).

src/engine/strat_engine/liminal/liminal.rs
```diff
@@ -329,17 +337,19 @@ impl LiminalDevices {
         pool_uuid: PoolUuid,
         token_slot: TokenUnlockMethod,
         passphrase_fd: Option<RawFd>,
+        remove_cache: bool,
     ) -> StratisResult<(Name, PoolUuid, AnyPool, Vec<DevUuid>)> {
+        #[allow(clippy::too_many_arguments)]
```
mulkieran: Guessing you can drop this allow.

src/engine/strat_engine/pool/v1.rs
```diff
@@ -261,12 +262,14 @@ impl StratPool {
     ///   * key_description.is_none() -> no StratBlockDev in datadevs has a
     ///   key description.
     ///   * no StratBlockDev in cachdevs has a key description
+    #[allow(clippy::too_many_arguments)]
```
mulkieran: Guessing you can drop the allow here, also.

src/engine/strat_engine/liminal/liminal.rs
```diff
@@ -1221,13 +1280,23 @@ fn load_stratis_metadata(
...
+    match get_metadata(&stratis_infos_ref(infos)) {
+        Ok(opt) => {
+            let (timestamp, mut metadata) = opt
             .ok_or_else(|| {
                 StratisError::Msg(format!(
                     "No metadata found on devices associated with pool UUID {pool_uuid}"
                 ))
-            }),
+            })?;
+            let dev_uuids_and_paths = if remove_cache {
```
mulkieran: Minor stylistic point: Our lowest supported Rust is 1.77.0 these days, and that means that we can use ```bool.then()``` here! https://doc.rust-lang.org/std/primitive.bool.html#method.then .

src/engine/strat_engine/liminal/liminal.rs
```diff
@@ -1204,12 +1231,44 @@ impl Into<Value> for &LiminalDevices {
...
+                    Some(info) => (d.uuid, Some(info.dev_info.devnode.clone())),
+                    None => (d.uuid, None),
+                })
+                .collect::<Vec<_>>()
+        })
+        .unwrap_or_default()
+}
+
+/// Remove any cache devices detected from the infos parameter.
+fn update_infos(infos: &mut HashMap<DevUuid, Box<LStratisInfo>>, dev_uuids: Vec<DevUuid>) {
```
mulkieran: It might not be necessary or useful to make this a separate method.

mulkieran (changes requested):
Need to do a little more testing, but I have no more code comments at this time.

PR #3862, opened 2025-06-17
.packit.yaml: Remove unnecessary job duplication

mulkieran:

PR #3865, opened 2025-06-20
Add identifiers to COPR builds

mulkieran:


mulkieran:
We do not need to use these identifiers anymore, since #3862 was merged.

PR #3890, opened 2025-08-01
Use TMT configuration to set up TPM2

mulkieran:


mulkieran:
Probably time to ask TMT why this does not work.

PR #3907, opened 2025-08-29
Update real version of libdbus-sys dependency to 0.2.6 

mulkieran:
Related https://github.com/stratis-storage/project/issues/809

## Summary by CodeRabbit

* **Chores**
  * Removed a crate-specific exclusion from packaging metadata so dependencies are treated uniformly.
  * Standardized vendor filtering to ensure consistent handling of all dependencies.
  * Improves build consistency and reduces edge cases in dependency preparation.
  * No user-facing feature changes; only build/configuration behavior adjusted.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -813,9 +813,9 @@ dependencies = [
 
 [[package]]
 name = "libdbus-sys"
-version = "0.2.5"
+version = "0.2.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


mulkieran:
We can wait until libdbus-sys clears in Fedora

mulkieran:
This PR is probably not that helpful. Test binaries are now no longer packaged with libdbus-sys in the new version, but the entire source of libdbus is still packaged in the vendor subdirectory. That could perhaps create a licensing problem, so we might as well continue to exclude it from the vendor tar file. And if we continue to do that, then we don't get that benefit from the version bump, and so there's no compelling reason to do this at all.

PR #3908, opened 2025-08-31
Follow the recommended dracut module ordering for out-of-tree modules.

jozzsi:
Move module ordering from 90 to 50

For dracut release v108 or later the recommended ordering for out out of tree modules is 50-59 range. The following is a section from dracut documentation:

> Not using the 50-59 range for out of tree dracut modules will likely
> lead to unintended errors in the initramfs generation process as your
> dracut module will either run too early or too late in the generation process.
> You have been warned.

## Summary by CodeRabbit

- Refactor
  - Moved Stratis initramfs integration to an earlier module stage for more consistent boot behavior and alignment with related components, including Clevis support.
- Chores
  - Updated installation and cleanup routines to target the new module location and removed references to deprecated paths, ensuring consistent setup and teardown of Stratis and Clevis assets.


jbaublitz:
Hi @jozzsi, I'm willing to take a look at this but I'm the one who originally did the ordering for the file and when I tried to put it in that range, I remember bumping into boot failures. As I recall, the ordering was actually very important in our case. I'm not sure if it has to do with the fact that we're a storage stack. I can test again, but I think this will require some work just to validate that this change doesn't cause a regression. I will try to take a look at it within the next 3 weeks.

mulkieran:
dracut 107 is what is currently in Fedora: https://src.fedoraproject.org/rpms/dracut .

jbaublitz:
@jozzsi Usually I'm the one doing the testing. I may have to plan this for the next sprint unless I can get it added to this sprint. It will be a non-trivial amount of work just because we still do not have Anaconda support. I will try to get this scheduled.

PR #3928, opened 2025-10-23
Increase CURRENT DEVELOPMENT ENVIRONMENT to Fedora 42

mulkieran:
Related https://github.com/stratis-storage/project/issues/805


mulkieran:
We have a weird hang on check with udev and checks with tang and the rawhide tests have begun failing on mkfs command. I suspect that the problem with the rawhide tests is caused by a an xfs update and it is completely unrelated to this PR. Not so sure about the checks w/ udev/tang, though.

mulkieran:
All testing farm tests have now passed!

PR #3947, opened 2025-12-19
Additional zbus interfaces

mulkieran:


mulkieran:
No apparent errors on key setting with the alternative fix, so we can go with that.

mulkieran:
rustfmt failure is just an infra issue.

PR #3950, opened 2026-01-02
tests: If exit code from D-Bus method non-zero, report message

mulkieran:
Related https://github.com/stratis-storage/project/issues/836


mulkieran:
Everything in COPR is stuck in pending, merging.

PR #3955, opened 2026-01-15
For cache_removal tests use new device to re-init cache

mulkieran:
Closes https://github.com/stratis-storage/project/issues/836


mulkieran:
Doesn't break the test, and probably fixes it, merging...

PR #3960, opened 2026-01-28
Makefile: separate install target into two subtargets

mulkieran:


mulkieran:
Just an infra failure w/ Packit.

PR #3977, opened 2026-02-25
Rewrite stratis-decode-dm in Rust

mulkieran:
Closes https://github.com/stratis-storage/stratisd/issues/3976

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -184,6 +184,8 @@ jobs:
         run: make test-stratis-min-root
       - name: Run stratis-tools tests
         run: make test-stratisd-tools
+      - name: Run stratis-utils tests
+        run: make test-stratis-utils
 
   shell-checks:
     runs-on: ubuntu-24.04
--- a/Makefile
+++ b/Makefile
@@ -433,6 +433,10 @@ test-stratis-min:
 test-stratisd-tools:
 	RUSTFLAGS="${RUSTFLAGS}" RUST_BACKTRACE=1 RUST_TEST_THREADS=1 cargo test --no-default-features --features "engine,extras" test_stratisd_tools
 
+## Test stratis-utils
+test-stratis-utils:
+	RUSTFLAGS="${RUSTFLAGS}" RUST_BACKTRACE=1 RUST_TEST_THREADS=1 cargo test --no-default-features --features "dbus_enabled,engine" test_stratis_utils
+
 ## Run yamllint on workflow files
 yamllint:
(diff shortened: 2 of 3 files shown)


mulkieran:
there was a bug in the testing repo test, variable shadowed. should be fixed now.

src/bin/utils/cmds.rs
```diff
@@ -200,6 +201,76 @@ impl<'a> UtilCommand<'a> for StratisPredictUsage {
...
+    fn cmd() -> Command {
+        Command::new("stratis-decode-dm")
+            .about("Utility that maps from Stratis filesystem devicemapper path to a Stratis value")
+            .arg(
+                Arg::new("path")
+                    .help("The absolute path of the devicemapper device ('/dev/mapper/<devicemapper-name>')")
+                    .required(true)
+                    .value_parser(
+                        |s: &str| -> Result<PathBuf, String> {
+                            let p = PathBuf::from(s);
```
jbaublitz, line 223: Does it make more sense to parse it as a `PathBuf` and then just check whether the path is absolute in the value_parser closure?

src/bin/utils/decode_dm.rs
```diff
@@ -0,0 +1,198 @@
...
+    fdo::{Error, ManagedObjects},
+    names::OwnedInterfaceName,
+    zvariant::OwnedValue,
+};
+
+const REVISION_NUMBER: &str = env!("CARGO_PKG_VERSION_MINOR");
+const BUS_NAME: &str = "org.storage.stratis3";
+const TOP_OBJECT: &str = "/org/storage/stratis3";
+
+static REVISION: LazyLock<String> = LazyLock::new(|| format!("r{REVISION_NUMBER}"));
```
jbaublitz: I think we can remove this altogether by simply changing the format String in `POOL_INTERFACE` and `FILESYSTEM_INTERFACE`.

src/bin/utils/cmds.rs
mulkieran: It kind of does! But I wanted to make the behavior exactly equivalent to the Python script, and that returns a parser error when path is not absolute. I regret that original decision, but I'ld rather stick with it now.

src/bin/utils/decode_dm.rs
mulkieran: @coderabbitai, how would I implement the reviewer's suggestion?

mulkieran: Ok! That was also copied from the Python script directly, but it's also fine this way, done.

PR #3979, opened 2026-02-27
Make device size ioctls cross-platform

jbaublitz:
Resolves https://issues.redhat.com/browse/RHEL-148313

Tested on PPC


jbaublitz:
Oh definitely. They must have bumped it recently because when I checked on Friday, it was still at 11 according to the Fedora packages website.

mulkieran:
Accidentally closed.

PR #4001, opened 2026-04-13
Edition 2024

mulkieran:
Related https://github.com/stratis-storage/project/issues/843

## Summary by CodeRabbit

* **Compatibility**
  * Improved compatibility with Rust 2024 edition rules and newer compiler lint checks.
  * Updated generated system integration bindings for safer modern Rust handling.

* **Reliability**
  * Preserved existing storage, encryption, D-Bus, and event-processing behavior while addressing compiler and lint issues.

* **Maintenance**
  * Simplified internal pattern matching and data handling without changing user-visible functionality or public interfaces.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -299,6 +299,17 @@ future_incompatible = { level = "deny", priority = 1 }
 unused = { level = "deny", priority = 2}
 rust_2018_idioms = { level = "deny", priority = 3 }
 nonstandard_style = { level = "deny", priority = 4 }
+boxed-slice-into-iter = { level = "deny", priority = 5}
+dependency-on-unit-never-type-fallback = { level = "deny", priority = 5}
+deprecated-safe-2024 = { level = "deny", priority = 5 }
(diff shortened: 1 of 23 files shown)


mulkieran:
[if_let_rescope.txt](https://github.com/user-attachments/files/30359699/if_let_rescope.txt)

These are everything flagged by ```if_let_rescope``` clippy lint. It is clear that we never imagined that the scope of the pattern matched in the if-let would leak into the else, so we didn't write the code as if it would. In edition 2024 it will not leak into the else branch which is what we we always assumed and would prefer.So there is no action to take for this lint.

mulkieran:
There is no action to be taken for edition-2024-expr-fragment-specifier lint, either. The expansion of the macros yields identical code in edition 2024 and edition 2021.

mulkieran:
The final task, before this is ready to go, is to examine the individual tail-expr-drop-order lints to determine if any changes in drop order might affect the behavior of the program.

mulkieran:
@coderabbitai review

PR #4015, opened 2026-04-28
Replace linters and formatters with ruff

mulkieran:
Related https://github.com/stratis-storage/project/issues/64


mulkieran:
@coderabbitai review

tests/client-dbus/tests/misc/test_predict.py
```diff
@@ -58,7 +58,7 @@ def _call_predict_usage_pool(
         if command.returncode != 0:
             raise RuntimeError(
                 f"Invocation of {_STRATIS_PREDICT_USAGE} returned an error: "
-                f"{command.returncode,}, {errs}"
+                f"{command.returncode}, {errs}"
```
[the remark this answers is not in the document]
mulkieran: Thanks @coderabbitai. You've got a point, but I've been entirely consistent there, so I'm going to stick with it (for now).

.github/workflows/main.yml
```diff
@@ -53,7 +52,6 @@ jobs:
           libblkid-devel
           make
           ncurses
-          python3-isort
```
jbaublitz, line 56: With the removal of black and isort, why aren't we replacing it with ruff?

jbaublitz (commented):
Just one question on this.

.github/workflows/main.yml
mulkieran: This one is run exclusively on the source code. We used to need those two dependencies, because of the stratis-decode-dm Python script. But now that stratis-decode-dm is Rust, we can drop these. We should have dropped them with the stratis-decode-dm change, but somehow I missed it.

jbaublitz: Ah that makes sense!

jbaublitz (approved):

PR #4019, opened 2026-05-09
stratisd.service: Remove DefaultDependencies=no and After=multi-user.target

### LINKED ISSUE #3968: stratisd.service: After=multi-user.target creates ordering cycle for dependent services
Hello all! First of all, thanks for the great work on Stratis. I just recently started using it, and it's amazing.

I have a Podman volume plugin backed by Stratis. It works well for containers created after the host has initialized, but for containers using Quadlet (which are essentially systemd units), I have a dependency ordering issue that prevents units with stratis volumes from initializing at boot.

I took some time to analyze the current unit file and test a potential fix for any service that depends on `stratisd.service`.

## Context

For context, `stratisd.service` currently uses:

```ini
DefaultDependencies=no
After=multi-user.target
...
[Install]
WantedBy=multi-user.target
```

This creates an ordering cycle for any service in `multi-user.target` that depends on stratisd. For example, my volume plugin service has:

```ini
After=stratisd.service
Wants=stratisd.service

[Install]
WantedBy=multi-user.target
```

Produces the cycle:
```
podman-volume-stratis.service -> (After) -> stratisd.service -> (After) -> multi-user.target -> podman-volume-stratis.service
```

The `After=multi-user.target` was introduced in #3826 (changing from `After=local-fs.target`) as part of reworking fstab/Clevis dependencies. Since early boot device assembly is now handled by `stratisd-min-postinitrd`, the full `stratisd` daemon doesn't need special boot ordering — it just needs D-Bus, which is available after `basic.target`.

## Proposed fix 

I would remove `DefaultDependencies=no` and `After=multi-user.target` from `stratisd.service.in`. The standard default dependencies already provide `After=sysinit.target basic.target`, which ensures D-Bus availability for the `Type=dbus` service. This allows downstream services to use `After=stratisd.service` without creating cycles. I've tested this locally, and it's working -- at least for my use cases.

I'm willing to submit a PR if you agree with the fix.

kriansa:
## Summary

Removes `DefaultDependencies=no` and `After=multi-user.target` from `stratisd.service.in`.

These two directives created an ordering cycle for any `multi-user.target` service that depends on `stratisd.service`:

```
dependent.service → (After) stratisd.service → (After) multi-user.target → (Wants) dependent.service
```

With default dependencies restored, systemd adds `After=basic.target` automatically, which is sufficient since `stratisd` is a `Type=dbus` service and D-Bus is available after `basic.target`.

## Why this is safe

The Clevis/fstab boot path introduced in #3826 is handled entirely by `stratisd-min-postinitrd` and the `stratis-fstab-setup` units — they are independent of `stratisd.service`. Change 3 in #3826 (moving from `After=local-fs.target` to `After=multi-user.target`) was not needed for the Clevis fix; changes 1 and 2 of that PR (the new `stratis-fstab-setup-with-network@.service` and removing `Before=local-fs-pre.target` from `stratisd-min-postinitrd.service`) are what fixed Clevis/Tang support.

A QEMU-based reproduction at https://github.com/kriansa/stratisd-boot-cycle-repro verifies:
1. The ordering cycle exists with the current unit file
2. Removing the two lines resolves the cycle
3. A simulated Clevis/network fstab service (matching `stratis-fstab-setup-with-network@` ordering) completes successfully after the fix

## Test plan

- [ ] Boot with `stratisd.service` enabled and a downstream service with `After=stratisd.service` in `multi-user.target` — confirm no ordering cycle
- [ ] Confirm `stratisd` starts correctly (D-Bus available via `basic.target`)
- [ ] Test Clevis/Tang fstab path is unaffected

Closes #3968

### DIFF
--- a/systemd/stratisd.service.in
+++ b/systemd/stratisd.service.in
@@ -1,8 +1,6 @@
 [Unit]
 Description=Stratis daemon
 Documentation=man:stratisd(8)
-DefaultDependencies=no
-After=multi-user.target
 
 [Service]
 BusName=org.storage.stratis3


kriansa:
@mulkieran I'm unable to confirm whether the failing test is actually an issue or a transient CI failure. Could you retrigger it just in case?

mulkieran:
> @mulkieran I'm unable to confirm whether the failing test is actually an issue or a transient CI failure. Could you retrigger it just in case?

@kriansa Ignore that rawhide test failure, it's due to the rawhide environment, specifically a util-linux change which will be fixed with the next release of util-linux.

Would you mind rebasing and doing a force-push to bring your PR up-to-date? Thanks!

kriansa:
@mulkieran sure! I did it using the GH interface. Is that OK?

mulkieran:
@coderabbitai review

jbaublitz (commented):
@mulkieran I have just one additional concern. @kriansa You may not know this, but as soon as stratisd starts, it issues a kill signal to stratisd-min. This is not very graceful but it is largely at the suggestion of systemd maintainers who told us that we would not be able to encode that logic in the systemd files. I finally recalled why I added the after and it's because under certain conditions, stratisd would start too early and kill stratisd-min partway through the filesystem setup. This resulted in an incomplete boot. Is there any safe lower bound target we can add to prevent stratisd from starting too soon and killing stratisd-min? There is perhaps a safety net here as I believe D-Bus cannot start until all of the filesystems in /etc/fstab have been mounted so it's possible that that protection alone is enough to guarantee that this won't happen.

mulkieran:
> @mulkieran I have just one additional concern. @kriansa You may not know this, but as soon as stratisd starts, it issues a kill signal to stratisd-min. This is not very graceful but it is largely at the suggestion of systemd maintainers who told us that we would not be able to encode that logic in the systemd files. I finally recalled why I added the after and it's because under certain conditions, stratisd would start too early and kill stratisd-min partway through the filesystem setup. This resulted in an incomplete boot. Is there any safe lower bound target we can add to prevent stratisd from starting too soon and killing stratisd-min? There is perhaps a safety net here as I believe D-Bus cannot start until all of the filesystems in /etc/fstab have been mounted so it's possible that that protection alone is enough to guarantee that this won't happen.

@jbaublitz We should see if removing DefaultDependencies=no, as this PR does, gives us the requirement we need.

mulkieran:
> > @mulkieran I have just one additional concern. @kriansa You may not know this, but as soon as stratisd starts, it issues a kill signal to stratisd-min. This is not very graceful but it is largely at the suggestion of systemd maintainers who told us that we would not be able to encode that logic in the systemd files. I finally recalled why I added the after and it's because under certain conditions, stratisd would start too early and kill stratisd-min partway through the filesystem setup. This resulted in an incomplete boot. Is there any safe lower bound target we can add to prevent stratisd from starting too soon and killing stratisd-min? There is perhaps a safety net here as I believe D-Bus cannot start until all of the filesystems in /etc/fstab have been mounted so it's possible that that protection alone is enough to guarantee that this won't happen.
> 
> @jbaublitz We should see if removing DefaultDependencies=no, as this PR does, gives us the requirement we need.

@jbaublitz I've done further investigation. I believe that removing DefaultDependencies=no adds some After targets that may ensure that stratisd is not started so early that stratisd-min can not complete its job. But I also believe that it adds some dependencies that mean that systemd kills stratisd too early on shutdown, such that it would not be able to support the root filesystem. I think it may be possible to keep DefaultDependencies=no but add individual dependencies that will do the necessary thing. I'm looking into that.

jbaublitz:
@mulkieran Can you list the targets/dependencies that you think will specifically affect shutdown?

mulkieran:
> @mulkieran Can you list the targets/dependencies that you think will specifically affect shutdown?

@jbaublitz 

Remove DefaultDependencies=no (same as setting DefaultDependencies to yes) with stratisd and the following occurs:

1. What affects shutdown:
Conflicts=shutdown.target is added
Before=shutdown.target is added

I believe that, with stratisd config as it is now, systemd will wait until very near the end before finally shutting stratisd down. With this PR, it will start the stratisd shutdown process earlier.

2. What affect startup:
Requires=sysinit.target is added
After=sysinit.target is added
After=basic.target is added

I believe that the presence of these additional requirements _may_ prevent stratisd from starting too early and killing stratisd-min before filesystems are properly set up.

I'm not sure that it would not be possible, with these additional startup dependencies, to allow systemd, rather than stratisd, to handle shutting down stratisd-min-postinitrd service by introducing a Conflicts entry in some system config file, but _that_ is likely a discussion for another time.

jbaublitz:
@mulkieran Encoding that logic in a systemd unit file is not possible or at least was not the last time I talked with the maintainers. I explained the exact logic we needed encoded into the systemd unit file instead of the workaround we ended up with. Their ultimate conclusion was that this logic is too advanced to be encoded using the existing feature set. While that may have changed in the last few years, I doubt it. I have not heard of any updates to the dependency/ordering system in recent times.

jbaublitz:
@mulkieran One additional thing: https://www.freedesktop.org/software/systemd/man/latest/bootup.html. I refer to this document when I'm wondering about boot order. Reviewing this page, I remember the reason I chose multi-user.target. That is the first point in the graph where remote-fs.target is guaranteed to have finished. Because we rely on this for Clevis, I'm not entirely convinced that we want stratisd starting before that point.

mulkieran:
> @mulkieran One additional thing: https://www.freedesktop.org/software/systemd/man/latest/bootup.html. I refer to this document when I'm wondering about boot order. Reviewing this page, I remember the reason I chose multi-user.target. That is the first point in the graph where remote-fs.target is guaranteed to have finished. Because we rely on this for Clevis, I'm not entirely convinced that we want stratisd starting before that point.

I feel like if you had (included in DefaultDependencies, hence previously excluded):
Requires=sysinit.target is added
After=sysinit.target is added
After=basic.target is added

and

After=remote-fs.target

The Clevis availability might be guaranteed, modulo bugs.

jbaublitz:
@mulkieran It's possible that that could solve the problem.

jbaublitz:
@kriansa If we merge this now, we want to make a few changes on our end regarding adding in just a few additional ordering constraints. Would you be able to test those before we merge them to make sure that we're not re-breaking you?

mulkieran (approved):

PR #4034, opened 2026-05-26
Clean up for registration and deregistration code

### LINKED ISSUE #4028: Announcement of filesystems of a just started pool is missing the "Pool" property with 3.9
Here is what Cockpit sees when a pool is started with a filesystem:

```
 UPDATE {"/org/storage/stratis3/8":{"org.storage.stratis3.filesystem.r8":{"Created":"2026-05-19T12:00:33Z","Name":"fsys1","Devnode":"/dev/stratis/pool0/fsys1","Size":"1099511627776","Uuid":"cf54c44803f64ad8ace16ba600b6fdff","MergeScheduled":false,"SizeLimit":[false,""],"Used":[true,"572522496"],"Origin":[false,"00000000000000000000000000000000"]}}}
```

Note that there is no value for the "Pool" property in that initial notification. Thus, Cockpit can't associate that filesystem with a pool.

jbaublitz:
Closes #4028 

I manually tested with dbus-monitor and did a little bit of additional cleanup to make registration and deregistration consistent.

## Summary by CodeRabbit

* **Refactor**
  * Pool creation/start now uses a single registration flow that returns pool, filesystem, and device D-Bus object paths for responses.
  * Pool stopping and pool destruction delegate teardown to a centralized unregister operation that accepts filesystem/device UUID lists.
  * Simplified component lifecycle handling by removing per-filesystem/per-device unregister loops.

* **Bug Fixes**
  * Improved robustness: failures to unregister individual filesystem/block device objects now produce warnings instead of aborting.
  * On stop errors, retained/unpresent components are selectively cleaned up to match current engine state.

### DIFF
--- a/src/dbus/manager/manager_3_2/methods.rs
+++ b/src/dbus/manager/manager_3_2/methods.rs
@@ -15,7 +15,6 @@ use zbus::{
 
 use crate::{
     dbus::{
-        blockdev::unregister_blockdev,
         consts::OK_STRING,
         filesystem::unregister_filesystem,
         manager::Manager,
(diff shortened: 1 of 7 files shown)


mulkieran:
@jbaublitz Looks like the format test failure is real, could you fix? The clippy failure is real too.

mulkieran:
@coderabbitai review

mulkieran:
@coderabbitai review

src/dbus/manager/manager_3_2/methods.rs
```diff
@@ -152,50 +147,8 @@ pub async fn stop_pool_method(
...
-                    }
-                }
-            }
-        }
-        _ => {}
-    }
-
     if let Ok(StopAction::Stopped(_) | StopAction::Partial(_)) = action {
-        if let Err(e) = unregister_pool(connection, manager, &pool).await {
+        if let Err(e) = unregister_pool(connection, manager, &pool, &fs_uuids, &dev_uuids).await {
```
mulkieran: The ```fs_uuids_rem``` values in the removed code address a particular scenario where a pool is partially stopped; the pool begins tearing down the filesystem devices but then encounters a situation where the filesystem is still mounted, and stops. So the pool is partially torn down, but filesystems are still remaining and the pool should not be unregistered. I don't think it can remove,  but I also think that combining the action taken on Stopped and Partial results is probably mistaken and should also be fixed. The same objection holds for the r6 implementation.

mulkieran (changes requested):
I'm not done review yet, but I wanted to note one significant concern in the comment.

mulkieran:
@coderabbitai review

src/dbus/manager/manager_3_2/methods.rs
```diff
@@ -207,6 +165,48 @@ pub async fn stop_pool_method(
         if stopped.map(|s| s.info.is_some()).unwrap_or(false) {
             send_locked_pools_signals(connection).await;
         }
+    } else if action.is_err() {
+        match engine.get_pool(PoolIdentifier::Uuid(pool_uuid)).await {
+            Some(g) => {
```
mulkieran, line 171: I think you can remove all manipulation of blockdev object paths from this code block and just retain the manipulation of filesystem object paths, since this block now runs in a restricted scope where only filesystems can have been removed as the pool is still up and running, just possibly with a few filesystems stripped.

src/dbus/manager/manager_3_6/methods.rs
```diff
@@ -123,6 +85,48 @@ pub async fn stop_pool_method(
         if stopped {
             send_locked_pools_signals(connection).await;
         }
+    } else if action.is_err() {
+        match engine.get_pool(id.clone()).await {
```
mulkieran: Same as above: https://github.com/stratis-storage/stratisd/pull/4034/changes/2fea59b7b185f4c388a55375987d35b468496390#r3436813894 .

mulkieran (changes requested):

mulkieran:
unused import error is surely real...

jbaublitz:
I ran it through the git hooks and I believe this is ready.

mulkieran (changes requested):
mananger_3_2 and manager_3_6 implementation of stop_pool() method are the same wrt. the CleanedUp result, but should be different.

The issue is that in manager_3_2 implementation of stop_pool_method(), the CleanedUp result is unreachable because engine.stop_pool() is invoked with the value false, because partially constructed pools are not implemented in that interface. This method isn't consistent, treating this case as unreachable in one place, but then as reachable in another.

Contrariwise, in manager_3_6 implementation of stop_pool() the CleanedUp result can be returned, so where it treats that variant as unreachable is incorrect.

mulkieran (approved):

PR #4037, opened 2026-06-04
Prioritize master repo

mulkieran:


mulkieran:
It looks like test failure on 43 was due to infra problem, rpm was constructed but install from rpm somehow failed.

mulkieran:
Seems transitory.

PR #4061, opened 2026-07-30
Take partial steps toward upgrading stratisd to edition 2024

mulkieran:
Related https://github.com/stratis-storage/project/issues/843

Continue to allow if-let-rescope lint and edition-2024-expr-fragment-specifier lint. Neither of these lints expose any worrisome patterns. However, if we do not allow them and continue to deny or warn on rust-2024-compatibility lints we will be flooded with annoying messages.

Continue to allow tail-expr-drop-order lint. This lint is the difficult one and probably every instance where a suspicious pattern is reported will have to be manually checked, because drop order of temporaries changes in edition 2024 and that can change the semantics.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -300,6 +300,13 @@ unused = { level = "deny", priority = 2}
 rust_2018_idioms = { level = "deny", priority = 3 }
 nonstandard_style = { level = "deny", priority = 4 }
 
+rust-2024-compatibility = { level = "deny", priority = 5}
+edition-2024-expr-fragment-specifier = { level = "allow", priority = 6}
+if-let-rescope = { level = "allow", priority = 6}
+rust-2024-incompatible-pat = { level = "allow", priority = 6}
+tail-expr-drop-order = { level = "allow", priority = 6}
+unsafe-op-in-unsafe-fn = { level = "allow", priority = 6}
+
 [lints.clippy]
 all = { level = "deny" }
 cargo = { level = "deny", priority = 1 }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -303,7 +303,6 @@ nonstandard_style = { level = "deny", priority = 4 }
 rust-2024-compatibility = { level = "deny", priority = 5}
 edition-2024-expr-fragment-specifier = { level = "allow", priority = 6}
 if-let-rescope = { level = "allow", priority = 6}
-rust-2024-incompatible-pat = { level = "allow", priority = 6}
 tail-expr-drop-order = { level = "allow", priority = 6}
 
 [lints.clippy]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -305,7 +305,6 @@ edition-2024-expr-fragment-specifier = { level = "allow", priority = 6}
 if-let-rescope = { level = "allow", priority = 6}
 rust-2024-incompatible-pat = { level = "allow", priority = 6}
 tail-expr-drop-order = { level = "allow", priority = 6}
-unsafe-op-in-unsafe-fn = { level = "allow", priority = 6}
 
 [lints.clippy]
 all = { level = "deny" }
--- a/src/engine/strat_engine/crypt/shared.rs
+++ b/src/engine/strat_engine/crypt/shared.rs
@@ -384,7 +384,7 @@ fn sss_dispatch(json: &Value, recursion_limit: u64) -> StratisResult<Value> {
 
     let mut pin_map = Map::new();
     for jwe in jwes {
-        if let Value::String(ref s) = jwe {
+        if let Value::String(s) = jwe {
             // NOTE: Workaround for the on-disk format for Shamir secret sharing
             // as written by clevis. The base64 encoded string delimits the end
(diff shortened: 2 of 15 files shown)


mulkieran:
@coderabbitai review

Cargo.toml
```diff
@@ -300,6 +300,11 @@ unused = { level = "deny", priority = 2}
 rust_2018_idioms = { level = "deny", priority = 3 }
 nonstandard_style = { level = "deny", priority = 4 }
 
+rust-2024-compatibility = { level = "deny", priority = 5}
+edition-2024-expr-fragment-specifier = { level = "allow", priority = 6}
+if-let-rescope = { level = "allow", priority = 6}
+tail-expr-drop-order = { level = "allow", priority = 6}
```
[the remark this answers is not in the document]
mulkieran: @coderabbitai We're deferring dealing with tail-expr-drop-order lint until later. it flags many instances and I think we'll have to inspect them _all_.

src/engine/strat_engine/crypt/shared.rs
```diff
@@ -900,12 +900,12 @@ unsafe extern "C" fn open(
...
         Ok(pass) => {
-            let malloc_pass = libc::malloc(pass.as_ref().len());
+            let malloc_pass = unsafe { libc::malloc(pass.as_ref().len()) };
             let pass_slice =
                 unsafe { from_raw_parts_mut::<u8>(malloc_pass.cast::<u8>(), pass.as_ref().len()) };
             pass_slice.copy_from_slice(pass.as_ref());
-            *buffer = malloc_pass.cast::<libc::c_char>();
-            *buffer_len = pass.as_ref().len();
+            unsafe { *buffer = malloc_pass.cast::<libc::c_char>() };
+            unsafe { *buffer_len = pass.as_ref().len() };
```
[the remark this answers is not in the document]
mulkieran: @coderabbitai Your point is very well taken, and since we explicitly mark the call as unsafe here, we should deal with the fact that it is unsafe. But inside a call back and all this is gonna get ugly.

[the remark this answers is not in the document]
mulkieran: @coderabbitai Ok, thank you for the comment. We are doing the allocate-at-least-1-byte trick since this is a callback for a C library, and is is safest to do conventional C tricks in that context, I think.

mulkieran:
@coderabbitai review

src/engine/strat_engine/crypt/shared.rs
```diff
@@ -900,12 +901,29 @@ unsafe extern "C" fn open(
...
+            let pass_len = pass.as_ref().len();
+            let alloc_len = max(pass_len, 1);
+            let malloc_pass = unsafe { libc::malloc(alloc_len) };
+            if malloc_pass.is_null() {
+                let err = StratisError::Msg(format!(
+                    "malloc() returned null when allocating {} bytes for passphrase",
+                    alloc_len
+                ));
+                warn!("{}", err);
+                let guard = CLEVIS_ERROR
```
jbaublitz: I would argue that this is overloading the meaning of this variable. This is specifically to propagate Clevis errors. This error is not specifically a Clevis error and I believe we should return the appropriate errno instead.

src/engine/strat_engine/crypt/shared.rs
```diff
@@ -900,12 +901,29 @@ unsafe extern "C" fn open(
     #[allow(clippy::to_string_in_format_args)]
     match res {
         Ok(pass) => {
-            let malloc_pass = libc::malloc(pass.as_ref().len());
+            let pass_len = pass.as_ref().len();
+            let alloc_len = max(pass_len, 1);
```
jbaublitz, line 905: I understand that this is defensive coding, but is there a case you've found where a passphrase can be empty? I don't believe we allow that in our code.

jbaublitz (changes requested):
Just a few comments on the encryption bits.

src/engine/strat_engine/crypt/shared.rs
mulkieran: If open_safe() returned an error on 0 length passphrase or if SizedKeyMemory struct did some non-zero length guarantee it would be fine to skip the dance for 0 length passphrase in this code. Otherwise, it appears that the empty passphrase could be stowed in that Clevis info that open_safe() reads. It seems safer to do the conventional C thing in this code, rather than worry about guaranteeing non-zero length passphrase in other code.

src/engine/strat_engine/crypt/shared.rs
mulkieran: @coderabbitai Show me how to implement the linked suggestion: https://github.com/stratis-storage/stratisd/pull/4061#discussion_r3707275878 .

mulkieran: Thank you @coderabbitai.

jbaublitz (approved):

mulkieran:
Something is wrong with testing farm infra, I'm gonna merge.

PR #4066, opened 2026-08-13
Use atomic file writing pattern for systemd generators

mulkieran:


mulkieran:
I think cockpit error is caused by rawhide repo being transitorily funky and will go away.

PR #4067, opened 2026-08-14
Remove symlink source before symlinking

mulkieran:
Related #3968

### DIFF
--- a/src/bin/utils/generators/stratis_setup_generator.rs
+++ b/src/bin/utils/generators/stratis_setup_generator.rs
@@ -4,7 +4,7 @@
 
 use std::{
     error::Error,
-    fs::create_dir_all,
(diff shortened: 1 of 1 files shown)


mulkieran:
@coderabbitai review

jbaublitz (changes requested):
I believe making this idempotent might be a better approach. I don't see a benefit of removing and re-adding the symlink.

mulkieran:
> I believe making this idempotent might be a better approach. I don't see a benefit of removing and re-adding the symlink.

@coderabbitai Tell me how to implement the suggestion.

mulkieran:
I just want a package that isn't corrupted so that I can install it. The perils of working on rawhide, I guess.

jbaublitz (approved):

PR #4074, opened 2026-08-20
Increase getrandom dependency actual version to 0.4.3

mulkieran:
Related https://github.com/stratis-storage/project/issues/889


mulkieran:
The F45 install failure is irrelevant and unexplained. It is unreasonable to imagine that it is due to a version bump, so ignoring.

PR #4075, opened 2026-08-20
Increase recommended Rust toolchain to 1.98.0

mulkieran:
Related https://github.com/stratis-storage/project/issues/888


mulkieran:
@coderabbitai review

mulkieran:
> 🔇 Additional comments (4)
> 
> ℹ️ Review info

@coderabbitai I believe that the suspicious_runtime_symbol_definition lint was added earlier than 1.87.0. Our CI compiles a bunch of things with the MSRV and they do seem to be compiling. Also I'm worried that the thing flagged by that lint are actually bad. They were auto-generated by bindgen, and maybe it is auto-generating incorrect code.

mulkieran:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 🧩 Analysis chain
> 
> `@mulkieran` You are correct to question this allowance. My earlier claim about the lint version was incorrect.
> 
> `suspicious_runtime_symbol_definitions` is a rustc lint introduced in Rust 1.98.0. It is not a Clippy lint. The allowance is therefore required for the new recommended toolchain, not for the MSRV build.
> 
> `src/systemd/bindings.rs` only includes bindgen output from `OUT_DIR`. The allowance applies only to that generated-bindings module. It does not suppress the lint for handwritten code.
> 
> The lint can flag an imported C declaration such as `strlen` when bindgen maps C `char` differently from the Rust runtime signature. This can be ABI-compatible because the pointer representation is unchanged. It does not indicate that bindgen generated a replacement implementation of a runtime symbol.
> 
> I could not inspect the exact generated declaration because `build.rs` creates it in `OUT_DIR`, which is not present in the source tree. However, the structure matches the known Rust 1.98 bindgen false-positive case. The allowance is appropriate here.
> ✏️ Learnings added
> 
> You are interacting with an AI system.

@coderabbitai You were correct, suspicious_runtime_symbols_definitions was introduced after Rust 1.87.0. However, you are the bearer of bad news: since you are correct that proves that the jobs that are supposed to compile with the MSRV are compiling with a newer version instead, i.e., there is a bug in our CI and that needs to be fixed.

mulkieran:
@coderabbitai review

mulkieran:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 🧩 Analysis chain
> 
> `@mulkieran` The CI configuration does select the MSRV correctly.
> 
> `Cargo.toml` defines the MSRV as `1.87.0`. In `.github/workflows/main.yml`, the MSRV jobs read that value and pass it to `actions-rust-lang/setup-rust-toolchain@v1` as the explicit `toolchain`.
> 
> The MSRV build succeeds because `src/systemd/bindings.rs` already contains:
> 
> ```rust
> #![allow(unknown_lints)]
> ```
> 
> This attribute precedes:
> 
> ```rust
> #![allow(suspicious_runtime_symbol_definitions)]
> ```
> 
> Rust 1.87.0 does not know `suspicious_runtime_symbol_definitions`. It would normally emit an `unknown_lints` warning, and this project denies warnings. The existing `allow(unknown_lints)` suppresses that warning for the generated bindings module.
> 
> Therefore, the passing MSRV jobs do not indicate that they use Rust 1.98.0. The new allowance remains compatible with Rust 1.87.0.
> 
> You are interacting with an AI system.

@coderabbitai I decided just to allow the functions stratisd uses, of which there are exactly two which use fundamental types. So none of the clippy or compiler lints that previously affected the generated code are triggered and the generated code is only about 8 lines worth. Seems like an improvement. What do you think?

PR #4090, opened 2026-09-10
Add remove_cache API to remove caching from a pool

### LINKED ISSUE #3796: Handle failed cache devices

jbaublitz:
Follow up on #4057 
Closes #3796


src/engine/strat_engine/backstore/backstore/v2.rs
```diff
@@ -1378,6 +1379,74 @@ impl Backstore {
...
+        // Resume with the new linear table.
+        dm.device_suspend(&DevId::Name(&dm_name), DmOptions::private())?;
+
+        // Tear down the now-orphaned cache-sub and meta-sub devices.
+        let (cache_sub_name, _) = format_backstore_ids(pool_uuid, CacheRole::CacheSub);
+        let (meta_sub_name, _) = format_backstore_ids(pool_uuid, CacheRole::MetaSub);
+        remove_optional_devices(vec![cache_sub_name, meta_sub_name])?;
+
+        // Wrap the CacheRole::Cache device as a LinearDev placeholder.
+        let (dm_name, dm_uuid) = format_backstore_ids(pool_uuid, CacheRole::Cache);
```
mulkieran: It's a bit weird that this same dm_name, dm_uuid needs to be gotten on line 1417, also. It seems like it would make more sense to show that the same name is being used throughout this chunk of code.

mulkieran (changes requested):
I have the one stylistic request.

src/engine/strat_engine/backstore/backstore/v2.rs
jbaublitz: Resolved.

mulkieran (approved):

